PR #32, opened 2022-03-13
Overlay Images Node

theflyingzamboni:
-Added Overlay Images node and BoundedIntegerInput
-Corrected bug in Stack Images node when images have only one channel

### DIFF
--- a/backend/nodes/image_nodes.py
+++ b/backend/nodes/image_nodes.py
@@ -251,6 +251,85 @@ class ImResizeToResolutionNode(NodeBase):
         return result
 
 
+@NodeFactory.register("Image (Utility)", "Overlay Images")
+class ImOverlay(NodeBase):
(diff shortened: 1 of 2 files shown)


backend/nodes/image_nodes.py
```diff
@@ -251,6 +251,85 @@ def run(
...
+    """OpenCV transparency overlay node"""
+
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
+        self.description = "Overlay transparent images on base image."
+        self.inputs = [
+            ImageInput("Base"),
+            ImageInput("Overlay A"),
+            BoundedIntegerInput("Opacity A (%)", minimum=1, maximum=99),
```
joeyballentine, line 265: I'm kinda wondering if these would be better as sliders instead?

PR #54, opened 2022-04-27
Added nodes for normal maps

RunDevelopment:
I added 2 new nodes for processing normal maps:

1. Normalize: Given the R and G channels of a normal map, this will produce a normalized RGB normal map.
2. Addition: Given 2 normal maps (RG), this will add two normal maps together and produce a normalized RGB normal map. This node also has strength (weights) options, so it can also be used to produce a weighted average of 2 normal maps.


joeyballentine:
This looks really useful, I'll take a closer look later. Thanks!

Issue #52, opened 2022-04-26
Add a setting to disable hardware acceleration for the GUI 


RunDevelopment:
On Win10, the Nvidia 1080 Ti driver has a bug that causes all Electron applications to become blurry. 

![image](https://user-images.githubusercontent.com/20878432/165321332-d0dd3359-6a21-4606-bae4-75cf3700e03c.png)

The bug has been present for 4 years now, and I lost all hope of it ever getting fixed. Luckily, there is a simple workaround. Disabling hardware acceleration for the Electron app fixes this issue. This workaround works for Slack, VS Code, and Discord, and it works for chaiNNer as well. 

Please add an option in the settings for this, so I don't have to start chaiNNer from the command line with `--disable-gpu` anymore.

RunDevelopment:
Thanks!

PR #60, opened 2022-04-27
Fuzzy search query

RunDevelopment:
This implements a fuzzy search. Basically, as long as a sequence of character (with any number of characters in between these characters) matches the search string, the node matches. Or in other words, the search `foo` will be interpreted as the regex `f.*o.*o`.

I also refactored the "super terrible" code around namespaces and filtering matches. Categories with no nodes will no longer be displayed.

Exampe: Matches for `pylo`

image
(Sorry for the blurry screenshot, see #52)

### DIFF
--- a/src/components/NodeSelectorPanel.jsx
+++ b/src/components/NodeSelectorPanel.jsx
@@ -5,9 +5,7 @@ import {
   InputGroup, InputLeftElement, Tab, TabList, TabPanel,
   TabPanels, Tabs, Text, Tooltip, useColorModeValue, useDisclosure, Wrap, WrapItem,
 } from '@chakra-ui/react';
-import {
-  memo, useContext, useEffect, useState,
-} from 'react';
+import { memo, useContext, useState } from 'react';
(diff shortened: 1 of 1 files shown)


RunDevelopment:
```
Error: Error: Failed to load plugin 'jsx-a11y' declared in '.eslintrc.js » eslint-config-airbnb » /github/workspace/node_modules/eslint-config-airbnb/rules/react-a11y.js': Cannot find module 'eslint-plugin-jsx-a11y'
```

Pretty sure that's not my fault.

joeyballentine:
> ```
> Error: Error: Failed to load plugin 'jsx-a11y' declared in '.eslintrc.js » eslint-config-airbnb » /github/workspace/node_modules/eslint-config-airbnb/rules/react-a11y.js': Cannot find module 'eslint-plugin-jsx-a11y'
> ```
> 
> Pretty sure that's not my fault.

It's not, I think that workflow is just buggy. I'll try out a different one at some point

PR #63, opened 2022-04-28
Fix ESLint CI

RunDevelopment:


RunDevelopment:
Ok, so npm@8.1 installs the dependencies juuuuust right, for ESLint. That explains why it works for me locally.

RunDevelopment:
Okay, I finally understand it. The problem is that `eslint-config-airbnb` declares a bunch of its deps as `peerDependencies` (e.g. `eslint-plugin-jsx-a11y` and `eslint-plugin-import`). The problem with that is that different npm version handle peer deps differently. npm 1, 2, 7, and 8 install all peer deps by default, but all other versions do not. Guess what npm version our CI uses: v6.

The funny thing is: npm v6 is supposed to print warnings "WARN: peer dep whatsitcalled not installed", but it doesn't do that because it errors out because of node-gyp before it gets the chance to.

To fix this, I just installed npm v7.

PR #85, opened 2022-04-29
Pylint

RunDevelopment:


RunDevelopment:
Failed as expected. Good.

RunDevelopment:
There are still a bunch of warning left btw. We'll have to resolve those over time. With "we" I mean someone that is not me. I just don't know enough about Python to fix those warnings.

requirements.txt
```diff
@@ -1,7 +1,10 @@
...
 numpy
+opencv-python
 Pillow
-ncnn_vulkan
-sanic_cors
+pyinstaller
+pylint
 sanic
+sanic_cors
+websockets==9.1
```
joeyballentine, line 9: Websockets isn't necessary

joeyballentine: Sorry I think I misled you with the old one

joeyballentine: Tbh, not a problem.

PR #93, opened 2022-04-30
Ported more code to TS

RunDevelopment:
The main feature here is `safeIpc.ts`. I made safe types for `ipcMain` and `ipcRenderer` that know about all channels in chaiNNer. TS can now verify the names, argument types, and return types of channels. Using the safe types is super simple, just import `ipcMain` and `ipcRenderer` from `safeIpc.ts` instead of `electron`. (Also, this has zero runtime overhead because `safeIpc.ts` literally just re-exports those two objects from `electron` but with different types.)

Adding new channels will now require adding them to the `Channels` interface, but that is a small price to pay for the guaranteeing the correct usage of channels.


src/components/SettingsModal.tsx
```diff
@@ -455,7 +455,6 @@ export const SettingsButton = memo(() => {
       <SettingsModal
         isOpen={isSettingsOpen}
         onClose={onSettingsClose}
-        onOpen={onSettingsOpen}
```
joeyballentine, line 458: Why did you get rid of this?

src/setupIntegratedPython.ts
```diff
@@ -42,7 +49,7 @@ export const extractPython = async (directory, pythonPath, onProgress) => {
   if (['linux', 'darwin'].includes(os.platform())) {
     try {
       log.info('Granting perms for integrated python...');
-      await fs.chmod(pythonPath, 0o7777, (error) => {
+      fs.chmod(pythonPath, 0o7777, (error) => {
```
joeyballentine, line 52: Did I not import the promise version of this?

RunDevelopment: Nope.

src/components/SettingsModal.tsx
RunDevelopment: Because `SettingsModal` has no `onOpen` property. Look at the props at the start of the file.

src/components/SettingsModal.tsx
```diff
@@ -102,7 +107,6 @@ const SettingsModal = ({ isOpen, onClose }) => {
           <Switch
             defaultChecked={colorMode === 'dark'}
             size="lg"
-            value={colorMode === 'dark'}
```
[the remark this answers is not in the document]
RunDevelopment: Because `value` is useless. That [props controls when value this HTML Form element will submit](https://chakra-ui.com/docs/components/form/switch#props). I only noticed because its type is `string | number`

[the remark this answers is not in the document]
joeyballentine: Did you check that it actually didn't affect functionality? I thought that was needed to make the inputs controlled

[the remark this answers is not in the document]
joeyballentine: Maybe chakra does something special for that though, in which case yeah let's get rid of it

src/components/SettingsModal.tsx
joeyballentine: Oh haha. Guess I just missed that when I refactored it into its own file

src/setupIntegratedPython.ts
joeyballentine: Whoops. In that case we should probably switch it for the promise version right?

src/components/SettingsModal.tsx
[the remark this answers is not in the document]
RunDevelopment: Yeah, I did. Nothing changed. Everything works just like before.

src/setupIntegratedPython.ts
RunDevelopment: Sure. One sec

PR #96, opened 2022-05-01
Adjust description and default of Average Color Fix

theflyingzamboni:


backend/nodes/image_nodes.py
```diff
@@ -1415,14 +1415,16 @@ class AverageColorFixNode(NodeBase):
     def __init__(self):
         """Constructor"""
         super().__init__()
-        self.description = """Correct for upscaling model color shift by matching average color of
-            Input Image to that of a smaller Reference Image. If Reference Image
-            contains compression artifacts, the node can downscale it to avoid
-            artifacts in the output."""
+        self.description = """Correct for upscaling model color shift by matching
+         average color of Input Image to that of a smaller Reference Image. 
+         Using significant downscaling increases generalization of averaging effect
```
RunDevelopment, line 1420: Could you please remove the trailing space here? 

Some IDEs (e.g. mine) are configured to trim trailing spaces by default, so the next one to make any changes to this file will remove those spaces. This kind of stuff is super annoying because it clutters the diff.

```suggestion
         Using significant downscaling increases generalization of averaging effect
```

PR #97, opened 2022-05-01
Migrated GlobalNodeState to TS

RunDevelopment:
Migrating GlobalNodeState was quite difficult. This file kinda touches everything that happens inside the application, so it was quite difficult to derive types for it. I pretty such that some types (Especially `NodeData` and `EdgeData`) as still wrong, but they are correct enough for now. I expect pretty much all type definitions I added in this PR to change as I port more and more components.

I'm also not too proud about the seemingly random functions and types I added. Especially the types and functions about edge handles, save files, and node schemas all belong into their own files/classes. I just threw them in there for now, because I know that they will still change.

Honestly, what we really need here are better abstraction boundaries. Especially for save files. Save file stuff is kinda all over the place rn, from what I see.
Well, refactoring complex logic into neat little units in kinda what TS excels at, so hopefully this will be quite painless once enough is ported over to TS.

Also, @joeyballentine, regarding `null`, `undefined`, and default values for parameters. It's kinda all over the place whether optional parameters/properties use `null` or `undefined` (or both) as their nil value. This isn't necessarily an issue for object properties (it's just kinda weird), but it's a problem for function parameters. Optional parameters will only default to their default value if no value or `undefined` is passed, `null` doesn't work. I've already seen a few function calls that pass `null` thinking that the parameter will default to something sensible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,6 +59,7 @@
         "@types/os-utils": "^0.0.1",
         "@types/react-dom": "^18.0.3",
         "@types/semver": "^7.3.9",
+        "@types/uuid": "^8.3.4",
         "@typescript-eslint/eslint-plugin": "^5.21.0",
         "@typescript-eslint/parser": "^5.21.0",
         "@vercel/webpack-asset-relocator-loader": "^1.7.0",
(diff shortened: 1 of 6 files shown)


joeyballentine:
> Optional parameters will only default to their default value if no value or undefined is passed, null doesn't work

I was not aware of this. Did you see any spots this would cause bugs?

RunDevelopment:
> I was not aware of this. Did you see any spots this would cause bugs?

I've seen [this example](https://github.com/joeyballentine/chaiNNer/blob/9c22b55d17cad2fa9f1c57f2a567db2c39f10cf3/src/components/ReactFlowBox.jsx#L196) today. Probably still works because `createNode` has the same `nodeType === 'iterator'` check internally.

joeyballentine (approved):

PR #98, opened 2022-05-01
Add Simple Video Frame Iterator

joeyballentine:


backend/nodes/image_iterator_nodes.py
```diff
@@ -150,3 +154,175 @@ def walk_error_handler(exception_instance):
                     }
                 )
         return ""
+
+
+@NodeFactory.register("Image", VIDEO_ITERATOR_DEFAULT_INPUT_NODE_NAME)
```
RunDevelopment, line 159: Should that really be under the "Image" category?

joeyballentine: Since it deals with each frame as an individual image, yeah. At least, that's how I think about it. It's like an alternative to exporting frames and using the Image File Iterator.

Eventually, I'll have video specific nodes for doing things like interpolation and upscaling with temporally aware models, which will definitely be in a video category. I'll re-assess at that point whether this node belongs there or not.

RunDevelopment: > video category

Well, we'll have lots of refactoring ahead of us then. Right now, the category + node name is used to identify node types, so changing the category of a node is breaking change.

Honestly, having a single ID for node types would be nice anyway.

joeyballentine: > so changing the category of a node is breaking change.

That's why I have migrations.js

joeyballentine: I've made plenty of these breaking changes before. Always just have to convert people's chains when they load them

PR #99, opened 2022-05-01
Added Color Transfer node

theflyingzamboni:


backend/nodes/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
RunDevelopment, line 1: This isn't how you typically include a license in open source. It's not what source is licensed here, so the license effectively applies to everything in this directory.

It seems like [this](https://stackoverflow.com/questions/1523427/what-is-the-common-header-format-of-python-files) is common practice in python, so we should probably follow it. Just move the licensed function into its own file.

joeyballentine: Yeah I definitely don't want to accidentally change the license of some of my files

backend/nodes/image_nodes.py
```diff
@@ -1508,130 +1509,106 @@ def run(self, img: np.ndarray, method: int | str) -> np.ndarray:
...
+                ],
+            ),
+            DropDownInput(
+                "generic",
+                "Reciprocal Scaling Factor",
+                [
+                    {"option": "Yes", "value": 1},
+                    {"option": "No", "value": 0},
+                ],
+            ),
```
RunDevelopment, line 1571: Do we not have a checkbox input?

backend/nodes/image_nodes.py
```diff
@@ -1508,130 +1509,106 @@ def run(self, img: np.ndarray, method: int | str) -> np.ndarray:
...
-            blurred = self.fragment_blur(img, 5, i, 1 << i)
-            # blurred tends to be a bit too transparent
-            self.with_self_as_background(blurred)
-            with_background(result, blurred)
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
+        self.description = "Transfers colors from reference image"
+        self.inputs = [
+            ImageInput("Image"),
```
RunDevelopment, line 1545: Isn't "Image" the default name?

backend/nodes/image_nodes.py
joeyballentine: Nope, haven't needed it yet

backend/nodes/image_nodes.py
joeyballentine: It is, I'm pretty sure

PR #102, opened 2022-05-01
Migrated more code to TS

RunDevelopment:
This mainly ports the DependencyManager. I also refactored a little, so please check that I didn't mess up.

### DIFF
--- a/src/components/DependencyManager.tsx
+++ b/src/components/DependencyManager.tsx
@@ -37,23 +37,45 @@ import {
   VStack,
 } from '@chakra-ui/react';
 import { exec, spawn } from 'child_process';
-import { ipcRenderer } from 'electron';
-import { memo, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
+import { memo, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import semver from 'semver';
+import { ipcRenderer } from '../helpers/safeIpc';
 import { SettingsContext } from '../helpers/contexts/SettingsContext';
-import getAvailableDeps from '../helpers/dependencies';
+import getAvailableDeps, { Dependency } from '../helpers/dependencies';
 import pipInstallWithProgress from '../helpers/pipInstallWithProgress';
+import { PythonKeys } from '../common-types';
 
-const checkSemver = (v1, v2) => {
+const checkSemver = (v1: string, v2: string) => {
   try {
-    return !semver.gt(semver.coerce(v1).version, semver.coerce(v2).version);
+    return !semver.gt(semver.coerce(v1)!.version, semver.coerce(v2)!.version);
   } catch (error) {
     console.log(error);
     return true;
   }
 };
 
-const DependencyManager = ({ isOpen, onClose, onPipListUpdate = () => {} }) => {
+type ParsedPipList = Record<string, string>;
+
+const parsePipOutput = (output: string): ParsedPipList => {
+  const tempPipList = output.split('\n').map((pkg) => pkg.replace(/\s+/g, ' ').split(' '));
+  const pipObj: ParsedPipList = {};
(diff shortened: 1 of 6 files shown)


src/components/DependencyManager.tsx
```diff
@@ -167,73 +172,59 @@ const DependencyManager = ({ isOpen, onClose, onPipListUpdate = () => {} }) => {
     })();
   }, [depChanged]);
 
-  const runPipCommand = (args) => {
+  const runPipCommand = (args: string[]) => {
     setShellOutput('');
     setIsRunningShell(true);
-    const command = spawn(pythonKeys.python, ['-m', 'pip', ...args, '--disable-pip-version-check']);
+    // TODO: hope and pray pythonKeys is non-null
```
joeyballentine, line 178: It should be. Chainner would error before starting up if it was

src/components/DependencyManager.tsx
```diff
@@ -167,73 +172,59 @@ const DependencyManager = ({ isOpen, onClose, onPipListUpdate = () => {} }) => {
...
+      // TODO: hope and pray pythonKeys is non-null
+      pythonKeys!.python,
       dep,
-      (percentage) => {
-        setProgress(percentage);
-      },
+      setProgress,
       (data) => {
-        output += String(data);
+        output += data;
```
joeyballentine, line 218: So idk if you tested this, and maybe I'm thinking of something else, but I'm pretty sure this is necessary or it just outputs a string of hex data

src/components/DependencyManager.tsx
```diff
@@ -326,9 +315,10 @@ const DependencyManager = ({ isOpen, onClose, onPipListUpdate = () => {} }) => {
                               isLoading={isRunningShell}
                               leftIcon={<DownloadIcon />}
                               size="sm"
-                              onClick={async () => {
+                              onClick={() => {
                                 setDepChanged(true);
-                                await updatePackage(dep);
+                                // eslint-disable-next-line @typescript-eslint/no-floating-promises
+                                installPackage(dep, true);
```
joeyballentine, line 321: Do we know for sure this functions the same not being awaited?

joeyballentine:
FYI, this fails on startup

src/components/DependencyManager.tsx
joeyballentine: So actually, it turns out it IS null on the first render before it makes the request to get the keys.

src/components/DependencyManager.tsx
joeyballentine: I was wrong, it appears to work perfectly fine. I think I'm thinking of something in the file download/pip install file

src/components/DependencyManager.tsx
joeyballentine: Verified that it does

joeyballentine:
I changed two `!`s to `?` due to them being null on the first render. I think that just means the file needs to be refactored but we can worry about that later.

RunDevelopment:
> I changed two `!`s to `?` due to them being null on the first render.

That's good. I typically use `!` because I just don't know enough about the code application as a whole yet, so I just assume that everything works.

Also, state being null, undefined, or some other invalid value because it's waiting for some async thing is a super common thing all throughout the application. We might want to try and come up with a general solution for this.

RunDevelopment:
> FYI, this fails on startup

Ah, the problem was that I replaced empty objects with `null`. My bad.

src/components/DependencyManager.tsx
RunDevelopment: TS wouldn't allow the `+=` if `data` wasn't a string (or number). In this case, `pipInstallWithProgress` guarantees that `data` is a string.

Also, I think there are a few more unnecessary conversions. I think TS lint has a rule to detect those.

PR #109, opened 2022-05-02
Remove ESLint from pre-commit hook

RunDevelopment:
With typescript now fully integrated into ESLint, ESLint will do a full project type check. This is quite slow, and the other type-related rules aren't cheap either.

The effect is that I have to wait like ~20sec per commit, which is REALLY annoying. So I just removed ESLint from he pre-commit hook. ESLint is still checked by our CI, so with a PR based-workflow that shouldn't be an issue.


joeyballentine:
Agreed that it is really annoying

PR #114, opened 2022-05-03
Migrated all remaining components to TS

RunDevelopment:


src/common-types.ts
```diff
@@ -12,14 +12,19 @@ export interface IteratorSize extends Size {
...
+}
 export type Input = {
   type: string;
   label: string;
   optional?: boolean;
-  def?: InputValue;
-  default?: InputValue;
-  options?: { value: InputValue }[];
+  def?: InputSchemaValue;
+  default?: InputSchemaValue;
```
joeyballentine, line 26: This makes me think we should probably standardize whether it's default or def

src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -576,21 +575,25 @@ export const GlobalProvider = ({
   );
 
   const useInputData = useCallback(
-    (id: string, index: number) => {
+    // eslint-disable-next-line prefer-arrow-functions/prefer-arrow-functions, func-names
+    function <T extends InputValue>(
```
joeyballentine, line 579: Just curious why you did it like this

src/common-types.ts
RunDevelopment: Oh, yes.

src/helpers/contexts/GlobalNodeState.tsx
RunDevelopment: From the perspective of `useInputData`. the input data is a `InputValue` (string or number (at least for now)), but from the perspective of e.g. `FileInput`, the input data has to be a string. Using generics, we resolve this perspective problem.

RunDevelopment: Ah, about that. As it turns, TSX will interpret `<T>() => {}` as the HTML tag `<T>` followed by an arrow function. That's why I had to use `function`.

joeyballentine: Makes sense

PR #115, opened 2022-05-03
Migrated main files to TS

RunDevelopment:
Regarding `main.ts`: I made a few changes while migrating:

- I removed `ipcMain.handle('get-live-sys-info', ...)`. The channel was unused.
- I fixed the comparison function for automatic updates.
- `role: 'Help'` should have used `'help'`.
- `app.on('uncaughtException', ...)` should have been `process.on('uncaughtException', ...)`

### DIFF
--- a/src/pages/main.tsx
+++ b/src/pages/main.tsx
@@ -15,11 +15,12 @@ import {
 } from '@chakra-ui/react';
 import { Split } from '@geoffcox/react-splitter';
 import { useWindowSize } from '@react-hook/window-size';
-import { app, ipcRenderer } from 'electron';
+import { app } from 'electron';
 import log from 'electron-log';
 import { memo, useEffect, useRef, useState } from 'react';
-import { ReactFlowProvider } from 'react-flow-renderer';
-import useFetch from 'use-http';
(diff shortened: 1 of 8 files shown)


src/pages/main.tsx
```diff
@@ -30,26 +31,33 @@ import NodeSelector from '../components/NodeSelectorPanel';
...
 
   // Queries
   const [backendReady, setBackendReady] = useState(false);
 
-  const options = { cachePolicy: 'no-cache', retries: 10 };
-  const { loading, error, data, response } = useFetch(`http://localhost:${port}/nodes`, options, [
-    port,
-  ]);
+  const { loading, error, data, response } = useFetch<BackendNodesResponse>(
+    `http://localhost:${port}/nodes`,
```
joeyballentine, line 57: Didn't you find a different module for making requests? Shouldn't we just use that everywhere?

joeyballentine:
The typescript check fails

src/pages/main.tsx
RunDevelopment: Yes, but also no. The backend class I made returns promises, and awaiting those in components isn't trivial. It's not hard either, but the main goal of this PR was to port those files over.

PR #116, opened 2022-05-03
Faster image preview

RunDevelopment:
I changed the backend implementation of the image preview to downscale images that are too large. This makes the Load Image preview **a lot** faster and doesn't cause lag for large images anymore. It also saves quite a bit of RAM since the frontend only has a small PNG in memory.

The only downside is that the preview will be slightly blurry if you zoom in. And I mean really zoom in. You'll only notice that the image is smaller than the original once you hit max zoom level. People with DPI scaling will notice this sooner of course.

### DIFF
--- a/backend/nodes/image_nodes.py
+++ b/backend/nodes/image_nodes.py
@@ -69,11 +69,21 @@ class ImReadNode(NodeBase):
 
         import base64
 
+        # resize the image, so the preview loads faster and doesn't lag the UI
+        # 512 was chosen as the target because a 512x512 RGBA 8bit PNG is at most 1MB in size
+        target_size = 512
+        max_size = target_size * 1.2
+        if w > max_size or h > max_size:
+            f = max(w / target_size, h / target_size)
+            img = cv2.resize(
+                img, (int(w / f), int(h / f)), interpolation=cv2.INTER_AREA
+            )
+
         _, encoded_img = cv2.imencode(".png", (img * 255).astype("uint8"))
         base64_img = base64.b64encode(encoded_img).decode("utf8")
 
         return {
-            "image": base64_img,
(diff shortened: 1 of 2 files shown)


joeyballentine:
Great idea, but yeah I have no concerns with this being blurry, its only a preview

backend/nodes/image_nodes.py
```diff
@@ -69,11 +69,21 @@ def get_extra_data(self) -> Dict:
 
         import base64
 
+        # resize the image, so the preview loads faster and doesn't lag the UI
+        # 512 was chosen as the target because a 512x512 RGBA 8bit PNG is at most 1MB in size
+        target_size = 512
+        max_size = target_size * 1.2
```
joeyballentine, line 75: What exactly is the point of this * 1.2?

RunDevelopment: If the image is `target_size + 1` (or some other small integer), then the downscaled image will be a complete mess, I think. Maybe it's not a huge issue, but I wanted to be sure.

Also, are the interpolation methods that don't have this problem?

joeyballentine: Area is the best, at least in opencv. In what way does the result become a mess?

RunDevelopment: It will become blurry. Here's an image with a width of 513 without that check:

image

and with the check:

image

It's not *very* blurry, but it's definitely better with the check.

RunDevelopment: That being said, most of the image I upscale are have powers of two as their size, so I personally wouldn't have a problem with the check being removed.

joeyballentine: Oh I understand now, you only wanna resize it if it's more than just a bit larger than the target size. I see.

PR #117, opened 2022-05-04
Fixed bugs in Threshold nodes

theflyingzamboni:
-Fixed dtypes for threshold nodes
-Removed unused options from Threshold node
-Corrected Adaptive Method options for Threshold (Adaptive) node


backend/nodes/image_nodes.py
```diff
@@ -596,7 +580,7 @@ def run(
             int(c),
         )
 
-        return result
+        return result.astype("float32") / 255
```
joeyballentine, line 583: Wouldn't this mean it would return as uint8?

PR #119, opened 2022-05-04
Added useAsyncEffect

RunDevelopment:
`useAsyncEffect` effect is the async equivalent of `useEffect`. The function guarantees that (if used correctly) no race conditions will occur.

Right now, all the async function all over the React code likely cause many subtle race conditions. `useAsyncEffect` is my attempt at eliminating those. However, it not nearly enough and I will refactor a few other components in the next few days as well.

### DIFF
--- a/src/helpers/hooks/useAsyncEffect.ts
+++ b/src/helpers/hooks/useAsyncEffect.ts
@@ -1,0 +1,137 @@
+// eslint-disable-next-line max-classes-per-file
+import { useEffect } from 'react';
+import log from 'electron-log';
+
+/**
+ * An object that can tell whether the current operation has been canceled.
+ */
+export interface CancellationToken {
+  /**
+   * Whether the operation has been canceled.
+   *
+   * Once an operation has been canceled, it can **never** be un-canceled again.
+   */
+  readonly isCanceled: boolean;
+  /**
+   * A simply utility function that will throw a {@link CancellationError} if the operation has
+   * been canceled.
+   *
+   * @throws {CancellationError}
+   */
+  readonly checkCanceled: () => void;
+  /**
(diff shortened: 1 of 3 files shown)


src/helpers/hooks/useAsyncEffect.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * This is an async replacement for `useEffect`.
+ *
+ * The supplier must not cause effects, except as a callback in `token.effect`.
+ * The `successEffect`, `catchEffect`, and `finallyEffect` functions are allowed to cause effects
+ * and must be executing synchronously.
+ *
+ * THis whole operation is roughly equivalent to:
```
joeyballentine, line 84: THis -> This

joeyballentine:
This is really interesting. I never considered that async functions in useEffects could cause race condition

src/helpers/hooks/useAsyncEffect.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+  };
+
+  readonly causeEffect = (fn: () => void) => {
+    this.checkCanceled();
+    fn();
+  };
+
+  /**
+   * A simply utility function that will set `isCanceled` to `true`.
```
joeyballentine, line 63:
```suggestion
   * A simple utility function that will set `isCanceled` to `true`.
```

src/helpers/hooks/useAsyncEffect.ts
joeyballentine, line 84:
```suggestion
 * This whole operation is roughly equivalent to:
```

joeyballentine (changes requested):

PR #127, opened 2022-05-04
Save file refactor

RunDevelopment:
I extracted the logic for saving and loading save files out into its own file. The `SaveFile` "class" is used as namespace inspired by `JSON`.


joeyballentine:
While you're at it, would this be a good time to implement that thing you mentioned to replace base64 encoding it? Adding a md5 hash and using JSON for the rest? With your base64 check you do, it would be pretty easy to determine if a save file is old or not.

If you don't wanna do that though, I can do it myself later.

RunDevelopment:
I'd rather have the refactor be one PR and the new feature another.

joeyballentine:
Fair

PR #130, opened 2022-05-04
Added safe IPC interface for browser windows

RunDevelopment:


src/helpers/safeIpc.ts
```diff
@@ -122,5 +125,21 @@ interface SafeIpcRenderer extends Electron.IpcRenderer {
   sendToHost<C extends keyof Channels>(channel: C, ...args: ChannelArgs<C>): void;
 }
 
+interface WebContentsWithSaveIcp extends WebContents {
```
joeyballentine, line 128: Is this supposed to be "safe" instead of "save"?

src/helpers/safeIpc.ts
```diff
@@ -122,5 +125,21 @@ interface SafeIpcRenderer extends Electron.IpcRenderer {
...
+  send<C extends keyof Channels>(channel: C, ...args: ChannelArgs<C>): void;
+  sendSync<C extends keyof Channels>(channel: C, ...args: ChannelArgs<C>): ChannelReturn<C>;
+  sendTo<C extends keyof Channels>(
+    webContentsId: number,
+    channel: C,
+    ...args: ChannelArgs<C>
+  ): void;
+  sendToHost<C extends keyof Channels>(channel: C, ...args: ChannelArgs<C>): void;
+}
+export interface BrowserWindowWithSaveIpc extends BrowserWindow {
```
joeyballentine, line 140: Same here

src/helpers/safeIpc.ts
RunDevelopment: Ahhhh, the typos spellcheck can't detect.

PR #143, opened 2022-05-05
Cleaned up around schemata

RunDevelopment:
More than anything else, I just cleaned up around schemata. Changes:

- `NodeSchema` now have a `category` property.
- `/nodes` from the backend now returns a `NodeSchema[]`.
- Added the `SchemaMap` class.
- No more `availableNodes`. All of these instances have been replaced with `schemata: SchemaMap` variables/parameters.
- `createNode` no longer takes a `defaultNodes` argument. The function will just get `defaultNodes` from the schema if needed.


src/components/NodeSelectorPanel.tsx
```diff
@@ -153,21 +171,10 @@ const NodeSelector = ({ data, height }: NodeSelectorProps) => {
...
-                                                  matchesSearchQuery(`${category} ${n.name}`) ||
-                                                  matchesSearchQuery(`${n.subcategory} ${n.name}`)
-                                          );
-
-                                    // don't show categories without nodes
-                                    if (matchingNodes.length === 0) return null;
-
-                                    const namespaceMap = getNamespaces(matchingNodes);
+                                {[...byCategory(matchingNodes)].map(([category, categoryNodes]) => {
+                                    const namespaceMap = getNamespaces(categoryNodes);
```
joeyballentine, line 178: This isn't anything to do with your changes, I just realized I left in the word namespaces here. This is a relic from before I added the subcategories, and used to have nodes grouped by the first part of their names, back when they were named like "IMAGE::LOAD". It should probably be renamed.

RunDevelopment: Alright

joeyballentine (approved):

PR #144, opened 2022-05-05
Fixed bug with optional inputs causing input array misalignment

theflyingzamboni:
-Nodes without default will default to undefined now. This means that indices will never be missing when the inputData object is converted to an Array.
-Removed code from Save Image and Overlay nodes that corrected for parameter misalignment.


src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -131,6 +131,8 @@ const getInputDefaults = (
                 defaultData[i] = input.default;
             } else if (input.options) {
                 defaultData[i] = input.options[0].value;
+            } else {
+                defaultData[i] = undefined;
```
RunDevelopment, line 135: You need to change the type of  the variable:

```diff
-const defaultData: Record<number, InputValue> = {};
+const defaultData: Record<number, InputValue | undefined> = {};
```

PR #146, opened 2022-05-06
Make `backend-ready` a send channel

RunDevelopment:


src/main.ts
```diff
@@ -700,26 +700,23 @@ const doSplashScreenChecks = async () =>
...
                 splash.destroy();
                 mainWindow.show();
-                ipcMain.removeHandler('backend-ready');
             });
         });
 
         mainWindow.webContents.once('dom-ready', () => {
-            ipcMain.removeHandler('backend-ready');
-            ipcMain.handle('backend-ready', async () => {
+            ipcMain.once('backend-ready', async () => {
```
joeyballentine, line 714: Does this not complain now about it already being handled by the other "once"? 

The reason I did this btw is there's a race condition between the app being ready to display and the backend being ready.

Though... Maybe this isn't actually needed? If the frontend has loaded to the point it can send this message, that means that the app is ready, therefore the upper handler isn't needed at all since the backend will only ever be ready after the web contents are ready to show. I'm not sure though

RunDevelopment: You essentially implemented a "on the last of these two" listener, I know that. The current implementation is a bit hacky because it leaves a listener registered and only works if `dom-ready` and `backend-ready` are sent at most once.

I'll make a little function for this that fixes these issues.

joeyballentine: Right. I'm just thinking now that it might not be necessary. It depends how electron works with defining when "dom-ready". Like does it start running the react code before it sends this event? Or will the code that checks if the backend is open only ever actually run after "dom-ready"?

If it's the latter, then we only need to check for the backend ready event

RunDevelopment: https://stackoverflow.com/questions/12637725/what-is-the-dom-ready-event#:~:text=DOM%20ready%20means%20that%20all,and%20any%20other%20linked%20resources).

RunDevelopment: Why is `dom-ready` necessary in the first place. We know that React has rendered at least once before `backend-ready` is fired, because it is fired after an async operation.

joeyballentine: That's exactly what I'm trying to say, I don't think it's necessary.

I'm pretty sure I only did this when I thought it was the cause of the app freezing on opening sometimes

RunDevelopment: Then I'll just remove the `dom-ready` stuff. If it causes freezes again, we'll know how to fix them.

joeyballentine: I don't even think this fixed it. It wasn't consistently reproducible so I really didn't know if anything I was doing was changing anything. I do think it's fixed now though, for some reason. Maybe an electron update, idk

RunDevelopment: Maybe. I read some bug reports about `dom-ready` firing twice.

PR #147, opened 2022-05-06
Moved `convertToUsableFormat` into Header.tsx

RunDevelopment:
This function only used in `Header` and has little to with managing the global node state.

### DIFF
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -18,3 +18,14 @@ export const deepCopy = <T>(value: T): T => JSON.parse(JSON.stringify(value)) as
 export const noop = () => {};
 
 export const copyNode = (node: Readonly<Node<NodeData>>): Node<Mutable<NodeData>> => deepCopy(node);
+
+export interface ParsedHandle {
+    id: string;
+    index: number;
+}
+export const parseHandle = (handle: string): ParsedHandle => {
+    return {
+        id: handle.substring(0, 36), // uuid
+        index: Number(handle.substring(37)),
+    };
+};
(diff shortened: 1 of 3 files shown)


src/helpers/util.ts
```diff
@@ -18,3 +18,14 @@ export const deepCopy = <T>(value: T): T => JSON.parse(JSON.stringify(value)) as
...
 export const copyNode = (node: Readonly<Node<NodeData>>): Node<Mutable<NodeData>> => deepCopy(node);
+
+export interface ParsedHandle {
+    id: string;
+    index: number;
+}
+export const parseHandle = (handle: string): ParsedHandle => {
+    return {
+        id: handle.substring(0, 36), // uuid
+        index: Number(handle.substring(37)),
```
joeyballentine, line 29: I didn't realize you had done this before, I highly suggest you change this back to getting the last element of it splitting by '-', or this will break if a node ever has more than 10 handles

RunDevelopment: No, it won't. If a node has 123456 handles, then this id will be `<36 char UUID>-123456`. This code works perfectly fine for this.

Man, we should really add tests. I'd rather have a test for this than me just saying "trust me bro."

joeyballentine: Oh, I guess I just misunderstood how substring works. I was thinking it would just get the 37th character, not the 37th to the end

joeyballentine: I blame JS not having the nice python syntax for this kind of thing

joeyballentine: And also, I agree we should add tests. It's just a matter of if the effort to do them is worth not putting that effort into other stuff. And also having motivation to write tests 😅

RunDevelopment: You just gotta make it easy to write tests. That's where snapshots come in. With snapshots, testing is as simple as: this functions, these inputs, outputs look fine.

PR #149, opened 2022-05-07
Condense node list styling

joeyballentine:
image

Condenses the list of nodes to help show more at a time. Also, I think it just looks better.


.prettierrc.json
```diff
@@ -2,5 +2,6 @@
 	"printWidth": 100,
 	"singleQuote": true,
 	"singleAttributePerLine": true,
-	"tabWidth": 4
+	"tabWidth": 4,
+	"trailingComma": "es5"
```
RunDevelopment, line 6: Why?

joeyballentine: when I was saving, it kept removing trailing commas, then giving me a warning telling me that it wanted the trailing commas

PR #154, opened 2022-05-07
Added abstraction layer for interpolation methods

### LINKED ISSUE #153: Save files with Resize nodes break when PIL is installed
When PIL gets installed, old save files with Resize nodes break.

This is because the interpolation method input uses the literal interpolation method values of either PIL or OpenCV. The those values depend on what you have installed.

I'm currently working on a fix for this. Since the numerical values of PIL and OpenCV are incompatible with each other, I will choose the PIL values since most of our users likely have PIL installed.

RunDevelopment:
fixes #153

### DIFF
--- a/backend/nodes/image_nodes.py
+++ b/backend/nodes/image_nodes.py
@@ -22,20 +22,13 @@ from .properties.inputs import *
 from .properties.outputs import *
 from .utils.color_transfer import color_transfer
 from .utils.fill_alpha import *
+from .utils.image_resize import resize
 from .utils.image_utils import (
     get_opencv_formats,
     get_pil_formats,
     normalize,
     normalize_normals,
 )
-
-try:
-    from PIL import Image
-
-    pil = Image
-except ImportError:
-    logger.error("No PIL found, defaulting to cv2 for resizing")
-    pil = Image = None
 
 
 @NodeFactory.register("Image", "Load Image")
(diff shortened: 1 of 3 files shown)


RunDevelopment:
While this is a fix, it also breaks save files of users that do not have PIL installed. The interpolation method of resize nodes in those save files will change. Only NN interpolation stays the same.

joeyballentine:
> While this is a fix, it also breaks save files of users that do not have PIL installed. The interpolation method of resize nodes in those save files will change. Only NN interpolation stays the same.

If by this you just mean the order is switched, that's fine. We can add a migration for that.

RunDevelopment:
> We can add a migration for that.

We cannot. The order only changes if PIL isn't installed. If PIL is installed, everything stays the same. Since we don't know whether PIL is installed from the same file alone, we can't fix this.

joeyballentine:
> > We can add a migration for that.
> 
> We cannot. The order only changes if PIL isn't installed. If PIL is installed, everything stays the same. Since we don't know whether PIL is installed from the same file alone, we can't fix this.

I'm confused, it looks like you coded it so it's the same order no matter what, and it's the same order as was there before. What's the problem?

RunDevelopment:
Yes, right now (in this PR), the numerical values of the interpolation methods selected by our users are the fixed (see the new `InterpolationMethod` class).

The problem is that this wasn't the case before. E.g. Cubic interpolation has the numerical value of 3 for PIL and 2 for OpenCV. This is a huge problem because those numerical values are saved directly. 

Let's say we have two users: A and B. User A has PIL installed, but user B has not. If those two users create a Resize node and both select Cubic as the interpolation and then save the chain, then these users will create different save files. The save file of user A will save 3 for cubic interpolation and the user B will save 2.

This is the problem. Both user created the same chain, but depending on whether PIL is installed or not, that chain will be interpreted differently. I.e. if user A opens the save of user B, it will not be cubic interpolation. Likewise, if user B opens the save file of user A, it will not be cubic interpolation.

joeyballentine:
Oh I see now. I was looking at the order of the methods themselves rather than the numbers they use. To be honest I don't think this is a huge deal so long as it's mentioned in the changelog.

PR #156, opened 2022-05-07
Changed Add Caption to PIL for unicode characters

theflyingzamboni:
-change image_resize to pil_utils to hold any functions that use PIL


backend/nodes/utils/pil_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    """Add caption with PIL or fall back to cv2"""
+
+    if pil is not None:
+        img = cv2.copyMakeBorder(
+            img, 0, 42, 0, 0, cv2.BORDER_CONSTANT, value=(0, 0, 0, 255)
+        )
+
+        pimg = Image.fromarray((img * 255).astype("uint8"))
+
+        font_path = os.path.join(os.getcwd(), "src/public/fonts/Roboto-Light.ttf")
```
RunDevelopment, line 72: Does work when we `npm run make` the application?

backend/nodes/utils/pil_utils.py
```diff
@@ -0,0 +1,110 @@
...
+def resize(
+    img: np.ndarray, out_dims: Tuple[int, int], interpolation: int
+) -> np.ndarray:
+    """Perform PIL resize or fall back to cv2"""
+
+    interpolation = INTERPOLATION_METHODS_MAP[interpolation]
+
+    # Try PIL first, otherwise fall back to cv2
+    if pil is not None:
+        pimg = pil.fromarray((img * 255).astype("uint8"))
```
RunDevelopment, line 55: Also, I just noticed that this is a bug. This assumes that the `img` is normalized. Same for where you copied this line.

Could you please fix this by normalizing `img` at the start of the method?

theflyingzamboni: Did I not include normalize? *checks* Apparently not. I had. No idea what happened to it.

RunDevelopment: The Resize nodes currently noramlize the img before passing it to this `resize`. I forgot the change that.

PR #158, opened 2022-05-07
Added missing key in NodeSelector

RunDevelopment:
React was throwing a little hissy fit over a missing `key`.


joeyballentine:
interesting, i didn't get an error, i assumed because the top level component of that component had a key

PR #161, opened 2022-05-08
Node registering refactor

joeyballentine:
This PR refactors the internal node registering system to use unique identifiers for nodes rather than registering them based on category and name. Now, the category and name can be freely changed without having to write migrations.

Along with this, nodes have now been recategorized thanks to @theflyingzamboni

image


joeyballentine:
Note: migrations still have to be written for old saves to work. However, at least this part is finished.

src/common-types.ts
```diff
@@ -45,13 +45,15 @@ export interface NodeSchema {
...
     readonly outputs: Output[];
     readonly defaultNodes?: DefaultNode[];
+    readonly identifier: string;
 }
 
 export interface DefaultNode {
     // Default nodes aren't currently used
     __SPECIAL: never;
     category: string;
     name: string;
```
RunDevelopment, line 55: Default nodes don't have a category and name anymore.
```suggestion
```

src/common-types.ts
```diff
@@ -79,9 +82,8 @@ export interface PythonKeys {
 }
 
 export interface UsableData {
-    category: string;
-    node: string;
     id: string;
+    identifier: string;
```
RunDevelopment, line 86: We should probably call `identifier` `typeId` or `schemaId`. This will be VERY confusing otherwise.

joeyballentine: SchemaId is good. I couldn't think of a good name to call it. I'll fix that

PR #165, opened 2022-05-08
Stale data fix

RunDevelopment:
It's what we talked about in discord.

I also removed `outlineInvalidNodes` and `unOutlineInvalidNodes` because they weren't used.


RunDevelopment:
I didn't fix `createNode` and `duplicateNode` because those are a little more complex. I'll make a separate PR for those 2 after this one.

PR #168, opened 2022-05-08
Refactor data that gets saved to contain only required information

joeyballentine:


src/common-types.ts
```diff
@@ -57,11 +57,7 @@ export interface DefaultNode {
 export interface NodeData {
     readonly id: string;
     readonly parentNode?: string;
-    readonly category: string;
-    readonly subcategory: string;
-    readonly icon: string;
     readonly schemaId: string;
-    readonly type: string;
     readonly isLocked?: boolean;
```
RunDevelopment, line 61: Shouldn't `isLocked` be saved too?

src/components/node/IteratorHelperNode.tsx
```diff
@@ -78,15 +78,15 @@ const IteratorHelperNode = ({ data, selected }: IteratorHelperNodeProps) => {
                     icon={icon}
                     parentNode={parentNode}
                     selected={selected}
-                    type={type}
+                    type={name}
```
RunDevelopment, line 81: Could you please rename the `type` prop?

src/components/node/IteratorHelperNode.tsx
```diff
@@ -78,15 +78,15 @@ const IteratorHelperNode = ({ data, selected }: IteratorHelperNodeProps) => {
...
                 />
                 <NodeBody
                     accentColor={accentColor}
                     category={category}
                     id={id}
                     inputs={inputs}
                     isLocked={isLocked}
-                    nodeType={type}
+                    nodeType={name}
```
RunDevelopment, line 89: And this one also.

src/components/node/IteratorNode.tsx
```diff
@@ -85,7 +83,7 @@ const IteratorNode = memo(({ data, selected }: IteratorNodeProps) => {
                         icon={icon}
                         percentComplete={percentComplete}
                         selected={selected}
-                        type={type}
+                        type={name}
```
RunDevelopment, line 86: And this one.

src/components/node/IteratorNode.tsx
```diff
@@ -108,7 +106,7 @@ const IteratorNode = memo(({ data, selected }: IteratorNodeProps) => {
                         id={id}
                         inputs={inputs}
                         isLocked={isLocked}
-                        nodeType={type}
+                        nodeType={name}
```
RunDevelopment, line 109: And this one.

src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -166,8 +166,24 @@ export const GlobalProvider = ({
...
+                id: n.data.id,
+                ...(n.data.iteratorSize ? { iteratorSize: n.data.iteratorSize } : {}),
+            },
+            id: n.id,
+            position: n.position,
+            type: n.type,
+            selected: n.selected,
+            ...(n.height ? { height: n.height } : {}),
+            ...(n.width ? { width: n.width } : {}),
+            ...(n.zIndex ? { zIndex: n.zIndex } : {}),
```
RunDevelopment, line 182: Why? `n.height` is `number | undefined`, right? This means that you can just say `height: n.height`. If height is undefined, JSON.stringify will remove the property: E.g.

```js
> JSON.stringify({ foo: undefined })
'{}'
```

RunDevelopment:
In general, I don't think changing the `SaveData` interface was such a good idea. Now everybody wanting to create or consume save data has to know how to convert nodes and edges arrays. 

I think it would be better if picking which data to actually save was an implementation detail of `SaveFile.stringify` and `SaveFile.parse`.

src/components/node/IteratorHelperNode.tsx
joeyballentine: Yeah, that's a good idea

joeyballentine:
> In general, I don't think changing the `SaveData` interface was such a good idea. Now everybody wanting to create or consume save data has to know how to convert nodes and edges arrays.
> 
> I think it would be better if picking which data to actually save was an implementation detail of `SaveFile.stringify` and `SaveFile.parse`.

This is a good idea. I'll change that

PR #173, opened 2022-05-09
Overlay fix again

theflyingzamboni:
-Stopped listening to people on stackoverflow and correctly implemented alpha compositing overlay
-Found out RunDevelopment had already done that elsewhere and it was slightly faster
-Used that instead
-Also fixed the grayscale metadata bug
-The saga endeth
-Or...?

### DIFF
--- a/backend/nodes/image_nodes.py
+++ b/backend/nodes/image_nodes.py
@@ -48,7 +48,7 @@ class ImReadNode(NodeBase):
         img = self.result[0]
 
         if img.ndim == 2:
-            h, w, c = img.shape[:2], 1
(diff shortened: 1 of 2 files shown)


backend/nodes/image_nodes.py
```diff
@@ -48,7 +48,7 @@ def get_extra_data(self) -> Dict:
         img = self.result[0]
 
         if img.ndim == 2:
-            h, w, c = img.shape[:2], 1
```
RunDevelopment, line 51: Did we use this pattern is other places too?

theflyingzamboni: No

PR #179, opened 2022-05-10
Split global state

RunDevelopment:
This PR does a bunch of things around `GlobalContext`:

- Split `GlobalContext` into `GlobalContext` (mostly never-changing stuff) and `GlobalVolitileContext` (stuff that changes all the time (nodes, edges))
- Changed some APIs and returned context values to make them more memoizable.
- Use [`use-context-selector`](https://github.com/dai-shi/use-context-selector) for contexts to avoid re-rendering.
- Changed how input elements get their data to avoid re-rendering.
- More efficient way to store nodes and edges in session storage to avoid re-rendering.

### DIFF
--- a/src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
+++ b/src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
@@ -1,7 +1,7 @@
 import { Box, Center, Tooltip } from '@chakra-ui/react';
 import { DragEvent, memo } from 'react';
 import ReactMarkdown from 'react-markdown';
-import { useContext } from 'use-context-selector';
+import { useContext, useContextSelector } from 'use-context-selector';
 import { NodeSchema } from '../../common-types';
 import { GlobalChainContext, GlobalContext } from '../../helpers/contexts/GlobalNodeState';
 import RepresentativeNode from '../node/RepresentativeNode';
@@ -19,7 +19,8 @@ interface RepresentativeNodeWrapperProps {
 }
 
 const RepresentativeNodeWrapper = ({ node }: RepresentativeNodeWrapperProps) => {
-    const { createNode, reactFlowInstance } = useContext(GlobalChainContext);
+    const createNode = useContextSelector(GlobalChainContext, (c) => c.createNode);
+    const reactFlowInstance = useContextSelector(GlobalChainContext, (c) => c.reactFlowInstance);
     const { reactFlowWrapper, setHoveredNode } = useContext(GlobalContext);
 
     return (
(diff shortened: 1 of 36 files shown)


src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
```diff
@@ -18,10 +19,9 @@ interface RepresentativeNodeWrapperProps {
 }
 
 const RepresentativeNodeWrapper = ({ node }: RepresentativeNodeWrapperProps) => {
-    const { createNode, reactFlowInstance, reactFlowWrapper, useHoveredNode } =
-        useContext(GlobalContext);
-
-    const [, setHoveredNode] = useHoveredNode;
+    const createNode = useContextSelector(GlobalChainContext, (c) => c.createNode);
+    const reactFlowInstance = useContextSelector(GlobalChainContext, (c) => c.reactFlowInstance);
```
joeyballentine, line 23: Didn't we get rid of reactFlowInstance? Is this still needed?

src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -178,28 +173,23 @@ export const GlobalProvider = ({
     schemata,
     reactFlowWrapper,
 }: React.PropsWithChildren<GlobalProviderProps>) => {
-    const { useSnapToGrid } = useContext(SettingsContext);
+    const useSnapToGrid = useContextSelector(SettingsContext, (c) => c.useSnapToGrid);
 
-    const [nodes, setNodes, onNodesChange] = useNodesState<NodeData>([]);
-    const [edges, setEdges, onEdgesChange] = useEdgesState<EdgeData>([]);
```
joeyballentine, line 184: Does everything still work without the onNodes/EdgesChange functions?

src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
joeyballentine: After doing a code search I've determined that reactFlowInstance is not needed and can be removed.

src/helpers/contexts/GlobalNodeState.tsx
RunDevelopment: They weren't used at all. That's why I removed them.

src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
RunDevelopment: There are two instances where we use `reactFlowInstance.project()`. How would those be replaced?

joeyballentine: That can be replaced with the new (in v10) useReactFlow hook which contains the project method https://reactflow.dev/docs/api/hooks/use-react-flow/ https://reactflow.dev/docs/api/react-flow-instance/

src/helpers/contexts/GlobalNodeState.tsx
joeyballentine: Oh, I guess the ones in ReactFlowBox do all the work. Makes sense.

src/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
RunDevelopment: Good to know.

PR #181, opened 2022-05-10
Use useReactFlow more

RunDevelopment:
As we discussed in Discord, useReactFlow can be used in a lot more places.

### DIFF
--- a/src/components/inputs/InputContainer.tsx
+++ b/src/components/inputs/InputContainer.tsx
@@ -1,7 +1,7 @@
 import { Box, HStack, Text, useColorModeValue } from '@chakra-ui/react';
 import React, { memo, useContext } from 'react';
 import { Handle, Position } from 'react-flow-renderer';
-import { GlobalContext } from '../../helpers/contexts/GlobalNodeState';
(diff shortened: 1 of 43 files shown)


src/components/inputs/InputContainer.tsx
```diff
@@ -13,10 +13,7 @@ interface InputContainerProps {
 
 const InputContainer = memo(
     ({ children, hasHandle, id, index, label }: React.PropsWithChildren<InputContainerProps>) => {
-        const isValidConnection = useContextSelector(
-            GlobalVolatileContext,
-            (c) => c.isValidConnection
-        );
+        const { isValidConnection } = useContext(GlobalContext);
```
joeyballentine, line 16: How come you didn't need the context selector anymore?

joeyballentine: I'm assuming it's because it's memoized better now?

RunDevelopment: Yes, isValidConnection only depends on `schemata, getNode, getNodes, getEdges` now. All of these should be pretty constant.

PR #185, opened 2022-05-11
Input validation middleware

joeyballentine:
Basically, I converted the inputs to classes and made them validate and transform the passed in values via an "enforce" method. This means that we can be 100% sure any image passed in is 0-1 and fp32, a number is an integer and not a string, a dropdowninput has a valid value, etc


theflyingzamboni:
Unused normalize imports: 
-image adj, image chan, image dim, image filter, image util

int cast:
-image chan l238 (and associated method typing) (I think the constants should return int already, but they weren't before)


file_inputs: 
-change output type hinting from Dict to classes, remove Dict typing import

generic_inputs: 
-change output type hinting from Dict to classes, remove Dict typing import

process: 
-L84, was this TODO completed? Are the inputs that don't require validation being left until later?

backend/nodes/ncnn_nodes.py
```diff
@@ -288,13 +287,14 @@ def __init__(self):
 
     def perform_interp(self, bin_a: np.ndarray, bin_b: np.ndarray, amount: int):
         try:
-            amount_a = amount / 100
-            amount_b = 1 - amount_a
+            amount_a = amount / 100.0
```
RunDevelopment, line 290: Does this even matter?

```py
>>> 4 / 100
0.04
>>> 4 / 100.0
0.04
```

backend/nodes/properties/inputs/numeric_inputs.py
```diff
@@ -0,0 +1,144 @@
...
+            "min": self.minimum,
+            "max": self.maximum,
+            "def": self.default,
+            "step": self.step,
+            "hasHandle": True,
+            "optional": self.optional,
+        }
+
+    def enforce(self, value):
+        return max(float(self.minimum), float(value))
```
RunDevelopment, line 37: Could you enforce the maximum (if present) too?

Other numeric inputs also seem to have this problem. Could you fix it for all of them?

backend/nodes/properties/inputs/numeric_inputs.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class OddIntegerInput(NumberInput):
+    """Input for an odd integer number"""
+
+    def __init__(self, label: str, default: int = 1, minimum: int = 1):
+        super().__init__(label, default=default, minimum=minimum, maximum=None, step=2)
+
+    def enforce(self, value):
+        odd = int(value) - (int(value) % 2)
```
RunDevelopment, line 76: That's interesting. So you round *down* to the nearest odd number. The front end makes a similar correction, but it rounds *up* to the nearest odd number.

RunDevelopment (changes requested):
Looks good, just a few nits.

backend/nodes/ncnn_nodes.py
joeyballentine: Nah, just didn't revert it after actually fixing it

backend/nodes/properties/inputs/numeric_inputs.py
joeyballentine: Good point since maximum can be overridden. Maybe I should default those to -Infinity and Infinity though instead of None? I'll have to test how that works with min/max

backend/nodes/properties/inputs/numeric_inputs.py
RunDevelopment: Let's use the front end version.

I also just notice that this implementation doesn't actually work.
```py
>>> def make_odd(n):
...     return n - (n % 2)
...
>>> make_odd(3)
2
>>> make_odd(4)
4
>>> make_odd(5)
4
>>> make_odd(6)
6
```

joeyballentine: Oh oops, that should be `n - (1 - (n% 2))`

RunDevelopment (approved):

PR #190, opened 2022-05-11
Added clear button to search bar

### LINKED ISSUE #180: Add search bar clear button
Add a small x to the search bar when it has some text to reset the text. Like this:

image

RunDevelopment:
Fixes #180.

No search query:

image

With search query:

image

### DIFF
--- a/src/global.css
+++ b/src/global.css
@@ -91,3 +91,14 @@ body {
 [data-theme=dark] .react-flow__controls-interactive {
     border-radius: 0px 0px 8px 8px;
 }
+
+.clear-search-bar-button {
+    color: var(--chakra-colors-gray-300);
+    cursor: pointer;
+}
+.clear-search-bar-button:hover {
+    color: var(--chakra-colors-gray-000);
+}
+.clear-search-bar-button > * {
+    font-size: 66%;
+}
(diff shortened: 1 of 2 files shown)


src/global.css
```diff
@@ -91,3 +91,14 @@ body {
...
+.clear-search-bar-button {
+    color: var(--chakra-colors-gray-300);
+    cursor: pointer;
+}
+.clear-search-bar-button:hover {
+    color: var(--chakra-colors-gray-000);
+}
+.clear-search-bar-button > * {
+    font-size: 66%;
+}
```
joeyballentine, line 104: How does this look in light theme?

joeyballentine: And why regular css instead of chakra props?

joeyballentine: Also the X should probably be darker in light mode, its kinda hard to see right now

RunDevelopment: In that case, we should probably adjust the chakra variables because the magnifying glass isn't any better.

joeyballentine: The chakra variables you used are just colors. They don't change between dark and light. That's the reason I use useColorModeValue everywhere. If you really wanna do it in regular css, do it like how Zamboni did the controls UI where it selects light and dark theme and use the different color variables. But it's probably easier just to use the style props

RunDevelopment: Done.

image

PR #197, opened 2022-05-11
Fixed VRAM

RunDevelopment:
This fixes the VRAM display on my laptop.

The first problem was that my laptop has an Nvidia GPU and an integrated Intel HD Graphics. This means that the VRAM usage of my NV is 0 most of the time. However, the VRAM usage assumed that 0 = no GPU. So now null = no GPU.

The second problem was that Nvidia SMI would random add new lines into the output. Examples:

```
NVIDIA GeForce GTX 950M
, 4096, 0, 4055, 0, 0

NVIDIA GeForce GTX 950M
, 4096, 0, 4055, 0
, 0
``` 

This messed up the parsing, which caused either NaN or Infinite VRAM usage to be displayed. So I fixed the parsing and added a check, so only finite numbers get passed to the front end.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -459,7 +459,6 @@ const checkNvidiaSmi = async () => {
 
                 vramChecker.stdout.on('data', (data) => {
                     const [, vramTotal, vramUsed] = String(data).split(/\s*,\s*/, 4);
-
                     const usage = (Number(vramUsed) / Number(vramTotal)) * 100;
                     if (Number.isFinite(usage)) {
                         lastVRam = usage;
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -458,9 +458,12 @@ const checkNvidiaSmi = async () => {
                 );
 
                 vramChecker.stdout.on('data', (data) => {
-                    const [, vramTotal, vramUsed] = String(data).split('\n')[0].split(', ');
+                    const [, vramTotal, vramUsed] = String(data).split(/\s*,\s*/, 4);
```
joeyballentine, line 461: Does split not require the regex multiline flag to handle multiline strings?

RunDevelopment: The multiline flag (m) only changes the meaning of `^`and `$`.

joeyballentine: Good to know. Tbh, never used that flag before, but just wanted to check.

PR #212, opened 2022-05-13
Changed duplicate to prevent stale data

RunDevelopment:


RunDevelopment:
With this PR, the `nodes` and `edges` arrays are only used as dependencies (e.g. `useEffect(..., [nodes])`) to detect changes. If we replace them with a simple counter, we could remove the arrays and still detect changes.

PR #213, opened 2022-05-13
Clearer interpolation slider

### LINKED ISSUE #184: Add interpolation slider labels
Label ends of interpolation sliders for PyTorch and NCNN as 'Model A' and 'Model B'. Make sure functions match that ordering.

RunDevelopment:
fixes #184.

---

This PR adds a *potenial* security risk. To evaluate the note string, I have to execute JS code given by the backend. Right now, the slider will just blindly execute this code. If we ever allow third-party nodes, we'll have to rethink this approach.


joeyballentine:
Hopefully by then we can have a better way of interpolating 2+ models anyway and we worn even have to use a normal slider for it. Like id love to have it be any amout of inputs where each input is the model connection + slider

PR #223, opened 2022-05-14
Added support for drag'n'dropping files

RunDevelopment:
I simplified the internal data transfer we do for drag'n'dropping nodes from the node selector panel and added support for dragging in .chn and image files.

The whole thing is implemented via a list of processors that all get a chance to process the dropped data. This makes it easy to add support for more files types and other stuff in the future.

Right now, dropping a single image file and dropping a single .chn file are supported. Multiple files are not supported.

### DIFF
--- a/src/helpers/dataTransfer.ts
+++ b/src/helpers/dataTransfer.ts
@@ -1,0 +1,120 @@
+import { XYPosition } from 'react-flow-renderer';
+import log from 'electron-log';
+import { extname } from 'path';
+import { NodeProto } from './contexts/GlobalNodeState';
+import { SchemaMap } from './SchemaMap';
+import { FileOpenResult, ipcRenderer } from './safeIpc';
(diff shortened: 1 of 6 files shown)


src/helpers/dataTransfer.ts
```diff
@@ -0,0 +1,120 @@
...
+        const saveData = await SaveFile.read(path);
+        return { kind: 'Success', path, saveData };
+    } catch (error) {
+        log.error(error);
+        return { kind: 'Error', path, error: String(error) };
+    }
+};
+
+/**
+ * Returns `false` if the data couldn't not be processed by this processor.
```
joeyballentine, line 36:
```suggestion
 * Returns `false` if the data could not be processed by this processor.
```

RunDevelopment: I hate that I sometimes just leave out words while writing. Thank you for spotting this.

PR #231, opened 2022-05-15
Resize, interpolation, and sliders

theflyingzamboni:
-Changed scale factor for Resize (Factor) and Average Color Fix to percentages.
-Changed SliderInput so that custom end labels can be added. Default is (min, max). None will hide the end labels.
-Reversed weight values for pytorch and ncnn model interpolation so that sliding left increases weight of A and right increases weight of B. Labeled ends A and B accordingly.
-Added migrations for scale factor and interpolation value changes.

### DIFF
--- a/backend/nodes/image_filter_nodes.py
+++ b/backend/nodes/image_filter_nodes.py
@@ -126,7 +126,11 @@ class AverageColorFixNode(NodeBase):
             ImageInput("Image"),
             ImageInput("Reference Image"),
             BoundedNumberInput(
-                "Reference Image Scale Factor", default=0.125, step=0.125
+                "Reference Image Scale Factor (%)",
+                minimum=0.1,
+                maximum=None,
+                default=12.5,
+                step=12.5
             ),
         ]
         self.outputs = [ImageOutput()]
@@ -140,12 +144,12 @@ class AverageColorFixNode(NodeBase):
     ) -> np.ndarray:
         """Fixes the average color of the input image"""
 
-        if scale_factor != 1.0:
+        if scale_factor != 100.0:
             ref_img = cv2.resize(
                 ref_img,
                 None,
-                fx=scale_factor,
-                fy=scale_factor,
+                fx=scale_factor / 100,
+                fy=scale_factor / 100,
                 interpolation=cv2.INTER_AREA,
             )
 
--- a/backend/nodes/image_filter_nodes.py
+++ b/backend/nodes/image_filter_nodes.py
@@ -130,7 +130,7 @@ class AverageColorFixNode(NodeBase):
                 minimum=0.1,
                 maximum=None,
                 default=12.5,
-                step=12.5
+                step=12.5,
             ),
         ]
         self.outputs = [ImageOutput()]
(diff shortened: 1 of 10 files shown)


backend/nodes/image_filter_nodes.py
```diff
@@ -126,7 +126,11 @@ def __init__(self):
             ImageInput("Image"),
             ImageInput("Reference Image"),
             BoundedNumberInput(
-                "Reference Image Scale Factor", default=0.125, step=0.125
+                "Reference Image Scale Factor (%)",
+                minimum=0.1,
```
RunDevelopment, line 130: Is there a reason for this minimum?

Same for `ImResizeByFactorNode`.

backend/nodes/properties/inputs/numeric_inputs.py
```diff
@@ -163,9 +162,23 @@ def __init__(
...
+        return {
+            "type": self.input_type,
+            "label": self.label,
+            "min": self.minimum,
+            "max": self.maximum,
+            "noteExpression": self.note_expression,
+            "ends": self.ends,
+            "def": self.default,
+            "step": self.step,
+            "hasHandle": True,
```
RunDevelopment, line 179: Well, that an assumption. Is that really true right now? Also, will it also be true in the future?

src/components/inputs/SliderInput.tsx
```diff
@@ -21,7 +21,7 @@ interface SliderInputProps extends InputProps {
     max: number;
     def?: number;
     accentColor: string;
-    hideEnds?: boolean;
+    ends?: Array<string | number> | null;
```
RunDevelopment, line 24: Use a tuple type.

Tuples aren't a runtime type/class in JS (yet), but TS has a tuple type.

```suggestion
    ends?: [string | number, string | number] | null;
```

RunDevelopment (changes requested):

RunDevelopment:
Ah, too late, I guess.

backend/nodes/image_filter_nodes.py
theflyingzamboni: Because 0 is an invalid scale number, but you could technically use a sub-1% scale (not sure why you'd want to, but you could).

joeyballentine: Didn't catch this originally, but yeah should probably make it like 0.00001 or something

backend/nodes/properties/inputs/numeric_inputs.py
joeyballentine: it should be customizable yeah, but also it doesn't really matter since at the moment number types don't have any outputs

src/components/inputs/SliderInput.tsx
joeyballentine: Definitely need to wait to merge stuff until you've taken a look haha

backend/nodes/properties/inputs/numeric_inputs.py
theflyingzamboni: > 

What is the question here?

PR #241, opened 2022-05-16
Prevent chaiNNer from going to sleep while processing

### LINKED ISSUE #239: Add prevent-app-suspension to prevent windows from going to sleep while chaiNNer is processing
This is something in electron. See docs here: https://www.electronjs.org/docs/latest/api/power-save-blocker#powersaveblockerstarttype

joeyballentine:
closes #239

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -254,8 +254,8 @@ const registerEventHandlers = () => {
     });
     ipcMain.on('stop-sleep-blocker', () => {
         if (blockerId !== undefined) {
-           powerSaveBlocker.stop(blockerId);
-           blockerId = undefined;
+            powerSaveBlocker.stop(blockerId);
+            blockerId = undefined;
         }
     });
 };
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -244,6 +245,11 @@ const registerEventHandlers = () => {
     });
 
     ipcMain.handle('get-app-version', () => app.getVersion());
+
+    ipcMain.on('start-sleep-blocker', () => {
+        const id = powerSaveBlocker.start('prevent-app-suspension');
+        ipcMain.on('stop-sleep-blocker', () => powerSaveBlocker.stop(id));
+    });
```
RunDevelopment, line 252: This creates a memory leak since the `stop-sleep-blocker` listeners never get removed.

Also, calling(?) `stop-sleep-blocker` when the blocker isn't currently active should be a noop.

```suggestion
    let blockerId: number | undefined;
    ipcMain.on('start-sleep-blocker', () => {
        if (blockerId === undefined) {
            blockerId = powerSaveBlocker.start('prevent-app-suspension');
        }
    });
    ipcMain.on('stop-sleep-blocker', () => {
        if (blockerId !== undefined) {
           powerSaveBlocker.stop(blockerId);
           blockerId = undefined;
        }
    });
```

RunDevelopment (approved):

PR #242, opened 2022-05-16
Fix error popup when backend intentionally killed

### LINKED ISSUE #236: [ BUG ] Installing NCNN gave error output
Installing NCNN gave an error output about restarting chainner.  I did so but NCNN still wasn't installed.  I tried again and again but still it wasn't installed.

Finally, I just let it sit and NCNN and the other dependencies were marked as installed.  If chainner is installing, it'd be nice to have some indication and not get an error.

joeyballentine:
Closes #236

### DIFF
--- a/src/components/DependencyManager.tsx
+++ b/src/components/DependencyManager.tsx
@@ -43,6 +43,7 @@ import semver from 'semver';
 import { useContext } from 'use-context-selector';
 import util from 'util';
 import { PythonKeys } from '../common-types';
+import { GlobalContext } from '../helpers/contexts/GlobalNodeState';
 import { SettingsContext } from '../helpers/contexts/SettingsContext';
 import getAvailableDeps, { Dependency } from '../helpers/dependencies';
 import { useAsyncEffect } from '../helpers/hooks/useAsyncEffect';
@@ -82,6 +83,7 @@ const DependencyManager = ({
     onClose,
     onPipListUpdate = () => {},
 }: DependencyManagerProps) => {
+    const { setIsBackendKilled } = useContext(GlobalContext);
     const { useIsSystemPython } = useContext(SettingsContext);
 
     const [isSystemPython] = useIsSystemPython;
@@ -180,6 +182,7 @@ const DependencyManager = ({
 
     useAsyncEffect(async () => {
         if (depChanged) {
+            setIsBackendKilled(true);
             await ipcRenderer.invoke('kill-backend');
         }
     }, [depChanged]);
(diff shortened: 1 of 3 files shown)


src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -76,6 +76,8 @@ interface Global {
     setIteratorPercent: (id: string, percent: number) => void;
     setHoveredNode: SetState<string | null | undefined>;
     setZoom: SetState<number>;
+    isBackendKilled: boolean;
```
RunDevelopment, line 79: This has nothing to do with current global *node* state.

Why not move this into the execution context?

joeyballentine: Good point

src/helpers/contexts/GlobalNodeState.tsx
```diff
@@ -20,13 +20,13 @@ import {
     Size,
 } from '../../common-types';
 import { useAsyncEffect } from '../hooks/useAsyncEffect';
+import { ChangeCounter, useChangeCounter, wrapChanges } from '../hooks/useChangeCounter';
```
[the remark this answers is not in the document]
joeyballentine: I have the organize imports setting on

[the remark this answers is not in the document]
RunDevelopment: This isn't a VS Code, FWIK. Maybe we could use ESLint to define import sorting instead? That way, it's consistent for all of us.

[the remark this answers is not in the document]
joeyballentine: It's a vscode setting. I've added the eslint rule for it so now it should be consistent. Haven't pushed it yet though

src/components/DependencyManager.tsx
```diff
@@ -602,11 +605,13 @@ export const DependencyManagerButton = memo(() => {
                     />
                 </VStack>
             </Tooltip>
-            <DependencyManager
-                isOpen={isOpen}
-                onClose={onClose}
-                onPipListUpdate={setPipList}
-            />
+            <ExecutionProvider>
```
RunDevelopment, line 608: I'm very sure this doesn't work. This execution context is not the same context as the `useContext(ExecutionContext)` you get `setIsBackendKilled` from. How you tested that this works?

joeyballentine: It doesn't work, I pinged you on discord about it

joeyballentine: @RunDevelopment it is now sorta working properly, but it still popped up with the warning, only it happened once instead of multiple times.

PR #248, opened 2022-05-16
Add sort-imports eslint rule

joeyballentine:


src/helpers/migrations.js
```diff
@@ -1,6 +1,5 @@
-/* eslint-disable import/prefer-default-export */
-import log from 'electron-log';
 import { isEdge, isNode } from 'react-flow-renderer';
```
RunDevelopment, line 1: Removing the eslint-diable here will cause an error.

joeyballentine: Whoops, I thought I undid that

RunDevelopment:
That because `sort-imports` is super conservatives. The order of imports in JS technically matters because imports can have side effects, so changing the order of 2 imports could change the behavior of the program.

So `sort-imports` to the approach of: I'll tell you that the imports have to be ordered in a certain way, but I won't do it for you because that might break things. How convenient.

Just configure the rule to only sort members (e.g. `import {b,a} ...` -> `import {a,b} ...`) and allow everything else otherwise. That's what I did [here](https://github.com/RunDevelopment/refa/blob/deaafbb9d8f8e5d3e2128747b2f24f17e8a8455e/.eslintrc.js#L71).

PR #254, opened 2022-05-16
Changed overlay minimum val to 0

theflyingzamboni:
Overlay minimum should be 0 in case someone wants to pass through the base image without deleting the node.


joeyballentine:
Can you make the other changes @RunDevelopment suggested in that PR?

RunDevelopment:
@joeyballentine Which PR?

RunDevelopment:
Oh, if you meant [this](https://github.com/joeyballentine/chaiNNer/pull/231#discussion_r873206667), I've already done this in #232.

PR #255, opened 2022-05-17
Allow NCNN FP16

joeyballentine:
This PR allows NCNN users to change the CPU/fp16 settings instead of those settings being disabled. The CPU setting doesn't do anything for NCNN yet though.

FP16 on the other hand is now the sole contributor to what gets an NCNN bin saved as fp16 or not. This is to hopefully fix black output on old AMD cards, though the fix is unconfirmed at the moment.

### DIFF
--- a/backend/nodes/ncnn_nodes.py
+++ b/backend/nodes/ncnn_nodes.py
@@ -120,17 +120,12 @@ class NcnnSaveNode(NodeBase):
         full_param_path = os.path.join(directory, full_param)
 
         logger.info(f"Writing NCNN model to paths: {full_bin_path} {full_param_path}")
-        bin_is_fp16 = bin_data.dtype == "float16"
-        bin_file_data = struct.pack(
-            "<I",
-            (FLAG_FLOAT_16 if bin_is_fp16 or os.environ["isFp16"] else FLAG_FLOAT_32),
-        ) + bin_data.astype(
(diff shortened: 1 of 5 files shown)


backend/nodes/ncnn_nodes.py
```diff
@@ -189,18 +189,12 @@ def run(self, net_tuple: tuple, img: np.ndarray) -> np.ndarray:
...
-
         with tempfile.TemporaryDirectory(prefix="chaiNNer-") as tempdir:
             bin_file_data = struct.pack(
                 "<I",
-                (
-                    FLAG_FLOAT_16
-                    if bin_is_fp16 or os.environ["isFp16"]
-                    else FLAG_FLOAT_32
-                ),
+                (FLAG_FLOAT_16 if os.environ["isFp16"] else FLAG_FLOAT_32),
```
RunDevelopment, line 195: `os.environ` only contains strings, right? If so, then this won't work.

Also, could you please extract the `os.environ["isFp16"]` condition out into its own variable? I don't like that it's duplicated here and 2 lines below.

```py
>>> print(1 if False else 0)
0
>>> print(1 if "False" else 0)
1
```

RunDevelopment: Looking at all the `os.environ["isFp16"]` in other places, maybe it would make sense to have 2 functions: 1one to read/parse the env into boolean, and one to set the env using boolean values.

PR #258, opened 2022-05-17
Add unsaved changes warning on quit

### LINKED ISSUE #219: Show "discarding unsaved changes?" message when closing
Just like with *File > New*, show a little dialog that asks the user whether they really want to close without saving.

joeyballentine:
Closes #219 (sorta)

I don't think this is possible to do with chakra menus, as there isn't a way I can think of to block the main process and wait for a response from the renderer. Using the system dialog was pretty easy.

This only half closes the issue since I didn't do the warning on file open.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -927,6 +927,7 @@ const createWindow = async () => {
             const choice = dialog.showMessageBoxSync(mainWindow, {
                 type: 'question',
                 buttons: ['Yes', 'No'],
+                defaultId: 1,
                 title: 'Discard unsaved changes?',
                 message:
                     'The current chain has some unsaved changes. Do you really want to quit without saving?',
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -917,6 +917,24 @@ const createWindow = async () => {
         mainWindow.webContents.openDevTools();
     }
 
+    let hasUnsavedChanges = false;
+    ipcMain.on('update-has-unsaved-changes', (_, value) => {
+        hasUnsavedChanges = value;
+    });
+
+    mainWindow.on('close', (event) => {
+        if (hasUnsavedChanges) {
+            const choice = dialog.showMessageBoxSync(mainWindow, {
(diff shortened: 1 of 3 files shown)


src/main/main.ts
```diff
@@ -917,6 +917,24 @@ const createWindow = async () => {
...
+    mainWindow.on('close', (event) => {
+        if (hasUnsavedChanges) {
+            const choice = dialog.showMessageBoxSync(mainWindow, {
+                type: 'question',
+                buttons: ['Yes', 'No'],
+                title: 'Discard unsaved changes?',
+                message:
+                    'The current chain has some unsaved changes. Do you really want to quit without saving?',
+            });
+            if (choice > 0) event.preventDefault();
```
RunDevelopment, line 934: `choice` is just an index of the `buttons` array, so please be explicit.
```suggestion
            if (choice === 1) event.preventDefault();
```

src/main/main.ts
```diff
@@ -917,6 +917,24 @@ const createWindow = async () => {
...
+    let hasUnsavedChanges = false;
+    ipcMain.on('update-has-unsaved-changes', (_, value) => {
+        hasUnsavedChanges = value;
+    });
+
+    mainWindow.on('close', (event) => {
+        if (hasUnsavedChanges) {
+            const choice = dialog.showMessageBoxSync(mainWindow, {
+                type: 'question',
+                buttons: ['Yes', 'No'],
```
RunDevelopment, line 929: By default, the first button (Yes) will be selected (focused). However, we probably want our users to save those changes.
```suggestion
                buttons: ['Yes', 'No'],
                defaultId: 1,
```

RunDevelopment (changes requested):

src/main/main.ts
joeyballentine: That's what I get for copying and pasting from SO lol

RunDevelopment (approved):

PR #259, opened 2022-05-17
Added import/order rule

RunDevelopment:
There's one caveat: Auto fixing doesn't work for TSX files. So `npm run lint-fix` doesn't fix this rule in TSX files, but interestingly the VSCode ESLint plugin does. Somewhat annoying but bearable.


joeyballentine:
Try adding it to rules as well and see if that makes it auto fixable

RunDevelopment:
It didn't make it any more fixable, but it did find stuff in `migration.js`.

PR #281, opened 2022-05-18
Fixed connection line z index

RunDevelopment:
I wasn't possible to just point npm towards my fork because react flow has a build step. So I just packed the built library and included that file. Not the most elegant solution, but it is simple.


joeyballentine:
Works as a temporary solution until the next RF release is out, hopefully with this change

PR #292, opened 2022-05-19
Fixed node validity check

### LINKED ISSUE #290: Overlap validation does not work correctly
Base is required.

image

RunDevelopment:
fixes #290 

Honestly, I don't know what caused the bug in the old code. I just implemented the simplest validity check I could think of, and everything seemed to work.


joeyballentine:
This is so much simpler than what I had, which is awesome since mine was a mess. But it makes me wary that it covers every case. I'll play around with it first before merging just to be sure.

PR #296, opened 2022-05-19
"..." to indicate new window

RunDevelopment:
"..." is typically used to indicate that this menu entry will open a new window. E.g.

image


joeyballentine:
Vscode doesn't do this with anything in the help section though, I think it's only when it opens a new window that wants user input

RunDevelopment:
True, so only for file open/save dialogs.

PR #301, opened 2022-05-20
Refactored NumberInput and SliderInput

theflyingzamboni:
-Merged different number inputs into single NumberInput for ints and floats
    -clampNumber now enforces allowable values for both NumberInput and SliderInput
    -Uses offset, step, and precision to control allowed values
    -Added controlsSteps attribute/prop so that the increment/decrement arrows can differ from the actual step
    -Added optional units for Number/SliderInput
-Changed NumberInput and SliderInput in the frontend to enforce the rules as defined by the backend for setting/displaying numeric values
    -Added nullable units display


backend/nodes/image_dim_nodes.py
```diff
@@ -27,8 +27,14 @@ def __init__(self):
         )
         self.inputs = [
             ImageInput(),
-            BoundedNumberInput(
-                "Scale Factor (%)", minimum=0.1, maximum=None, default=100.0, step=25.0
+            NumberInput(
+                "Scale Factor",
+                precision=4,
+                step=0.0001,
```
RunDevelopment, line 33: Is it even necessary to specify a step here? Couldn't we just say that the default for step is `10 ** -precision`?

backend/nodes/image_dim_nodes.py
```diff
@@ -27,8 +27,14 @@ def __init__(self):
...
+                controls_step=25.0,
+                minimum=0,
```
RunDevelopment, line 35: Minimum=0 seems to be very common. Could we make this the default?

backend/nodes/image_dim_nodes.py
```diff
@@ -44,7 +50,10 @@ def run(self, img: np.ndarray, scale: float, interpolation: int) -> np.ndarray:
         logger.info(f"Resizing image by {scale} via {interpolation}")
 
         h, w = img.shape[:2]
-        out_dims = (math.ceil(w * (scale / 100)), math.ceil(h * (scale / 100)))
+        out_dims = (
+            math.ceil(w * (scale / 100)) or 1,
+            math.ceil(h * (scale / 100)) or 1,
```
RunDevelopment, line 55: Please use `min` here. Using the fact that 0 is falsey is quite hacky.

backend/nodes/image_filter_nodes.py
```diff
@@ -150,8 +152,8 @@ def run(
             # Make sure reference image dims are not resized to 0
             h, w = ref_img.shape[:2]
             out_dims = (
-                math.ceil(w * (scale_factor / 100)),
-                math.ceil(h * (scale_factor / 100)),
+                math.ceil(w * (scale_factor / 100)) or 1,
+                math.ceil(h * (scale_factor / 100)) or 1,
```
RunDevelopment, line 156: Same here, please use `min`.

backend/nodes/image_filter_nodes.py
```diff
@@ -325,9 +327,9 @@ def __init__(self):
             is guaranteed to be normalized."""
         self.inputs = [
             ImageInput("Normal Map 1"),
-            SliderInput("Strength 1", 0, 100, 100),
+            SliderInput("Strength 1", minimum=0, maximum=100, default=100),
```
RunDevelopment, line 330: Same here, please make minimum=0 the default, so we don't have to repeat it everywhere.

backend/nodes/image_util_nodes.py
```diff
@@ -26,9 +26,9 @@ def __init__(self):
         self.inputs = [
             ImageInput("Base"),
             ImageInput("Overlay A"),
-            SliderInput("Opacity A", default=50, min_val=0, max_val=100),
+            SliderInput("Opacity A", minimum=0, maximum=100, default=50),
             ImageInput("Overlay B ", optional=True),
-            SliderInput("Opacity B", default=50, min_val=0, max_val=100, optional=True),
+            SliderInput("Opacity B", minimum=0, maximum=100, default=50, optional=True),
```
RunDevelopment, line 31: @joeyballentine Does it even make sense for a slider to be optional? They have a default value (always, I think), so they will always have a value assigned to them.

backend/nodes/properties/inputs/numeric_inputs.py
```diff
@@ -3,21 +3,32 @@
...
+def clampNumber(
+    value: Union[float, int],
+    precision: int,
+    offset: Union[float, int],
+    step: Union[float, int],
+    min_value: Union[float, int],
+    max_value: Union[float, int],
+) -> Union[float, int]:
+    # Convert proper number type
+    if precision:
```
RunDevelopment, line 15: Again, I think number conditions are hacky.

```suggestion
    if precision > 0:
```

backend/nodes/image_adj_nodes.py
```diff
@@ -1,219 +1,225 @@
...
-            img,
-            real_maxval,
-            adaptive_method,
-            thresh_type,
-            block_size,
-            c,
-        )
-
-        return result.astype("float32") / 255
+from __future__ import annotations
```
RunDevelopment, line 1: I think you changed the line ends for this file. Please don't do that.

backend/nodes/properties/inputs/numeric_inputs.py
```diff
@@ -27,19 +38,32 @@ class NumberInput(BaseInput):
...
         self,
         label: str,
-        default=0.0,
-        minimum=0,
-        maximum=None,
-        step=1,
-        optional=False,
-        number_type="any",
+        precision: int = 0,
+        offset: Union[float, int] = 0,
```
RunDevelopment, line 42: This is problematic.

Let's use off numbers as an example. Since offset is 0 by default, odd numbers (min=some odd number, step=2) will not work. 

This is a pitfall that can be avoided if offset=minimum if minimum is defined.

Actually, maybe we could even get rid of this parameter completely if we say: `offset = minimum if minimum is not None else 0`.

src/renderer/components/inputs/NumberInput.tsx
```diff
@@ -46,36 +52,81 @@ const NumericalInput = memo(
...
         };
 
+        const onBlur = () => {
+            const valAsNumber =
+                precision > 0 ? parseFloat(inputString) : Math.round(parseFloat(inputString));
+
+            if (!Number.isNaN(valAsNumber)) {
+                const roundedVal = +`${Math.round(
+                    +`${Math.round((valAsNumber - offset) / step) * step + offset}e${precision}`
+                )}e-${precision}`;
```
RunDevelopment, line 66: Why don't we just use the same formula as we use on the python side instead of this abomination?
```suggestion
                const roundedVal = Math.round(
                    (Math.round((valAsNumber - offset) / step) * step + offset) * 10 ** precision
                ) * 10 ** -precision;
```

src/renderer/components/inputs/SliderInput.tsx
```diff
@@ -122,6 +130,17 @@ const SliderInput = memo(
...
                         </NumberInputStepper>
                     </NumberInput>
+                    <Text
+                        fontSize="xs"
+                        m={0}
+                        mb={-1}
+                        mt={-1}
+                        p={0}
+                        pb={-1}
+                        pt={-1}
```
RunDevelopment, line 140: Negative padding values are invalid and therefore ignored.

RunDevelopment (changes requested):

backend/nodes/image_dim_nodes.py
joeyballentine: Zamboni wants the step to be different sometimes. Like having it be 0.25 even though you could make the value 0.05 if you wanted. I kinda personally think that might be a bit confusing though

backend/nodes/properties/inputs/numeric_inputs.py
joeyballentine: The whole point of the offset is you set it to 1 to enforce that numbers be odd. Right now if you set step to 2 and min to 1, it'll still let you type any even number

joeyballentine: But yeah I guess technically you could just derive that from the minimum value

backend/nodes/image_dim_nodes.py
RunDevelopment: Not what I meant. Here, he made step its smallest possible value `10 ** -4`. Values smaller than this would conflict with precision. I don't mean the control step, that is a good addition IMO.

Honestly, the underlying problem here is that `precision` and `step` are very similar. They both limit the number of possible inputs within the input range, they just do it through different means. `precision` limits the number of possible inputs by allowing only a certain number of decimal places. `step` is more general and makes that sure all input values are of the form `offset + i * step, i∈Z` (ignoring minimum and maximum for now).

Any precision value can be represented by a step of `step = 10 ** -precision` with `offset = 0`, but not every step value can be represented by a precision. E.g. it is impossible to convert the step value 1/3 into a precision other than ∞.

It might be a good idea to get rid of precision completely, and just have a little helper function `precision_to_step`.

The only one that needs precision is the frontend, and precision can be computed from offset and step as `max(getPrecision(offset), getPrecision(step))`. The `getPrecision` function itself could be implemented like this:

```ts
function getPrecision(n: number): number {
	const fraction = n % 1;
	if (fraction === 0) return 0;
	if (Math.abs(fraction) < 1e-10) return 10;
	return Math.min(10, String(n).split('.')[1]?.length ?? 0);
}
```

Note that this implementation will return at most 10. 64bit floating numbers only have a precision of around 16 digits anyway, so I chose 10, because it's enough for most use cases and it hides away any floating point inaccuracies.

theflyingzamboni: I actually decided after I was done for the night that I should ditch precision and keep step, so I'm already planning to do this.

backend/nodes/image_adj_nodes.py
theflyingzamboni: I did not. This is just how git decided to treat the diff when I rebased and had to resolve the conflict with that one minor PR you made to this file.

backend/nodes/image_dim_nodes.py
theflyingzamboni: Yeah, that's fair.

src/renderer/components/inputs/NumberInput.tsx
theflyingzamboni: lol

backend/nodes/properties/inputs/numeric_inputs.py
theflyingzamboni: Makes sense, and was something else I'd been thinking about doing.

src/renderer/components/inputs/SliderInput.tsx
theflyingzamboni: I actually just copied this code from NodeBody, so I'll fix it there too.

backend/nodes/image_util_nodes.py
theflyingzamboni: I'll make this change in the PR with the other changes to this node.

backend/nodes/image_adj_nodes.py
RunDevelopment: > This is just how git decided to treat the diff 

I am like 95% certain that you accidentally changed the line ends here. I don't know the internals of how git handles line ends in merge conflicts, but you might have to renormalize the line ends for this file.

This is not normal behavior.

> when I rebased

Rebased or just merged the main branch?

backend/nodes/image_dim_nodes.py
RunDevelopment: You're right. I did mean `max`.

backend/nodes/image_adj_nodes.py
theflyingzamboni: Actually, you know what, that might have happened. I didn't like VSCode's interactive merger so I switched back to pycharm to do that specifically, but the line endings in PyCharm were never changed to LF.

backend/nodes/image_dim_nodes.py
theflyingzamboni: I think it would be better to leave precision as a derived attribute and calculate it in the backend, though, since the backend validation requires the value as well.

theflyingzamboni: `value = (
    round((round((value - offset) / step) * step + offset) * 10**precision)
    * 10**-precision
)`

It is used here. Because step could be something like 0.5, you can't simply multiply the value by step, so you need to perform mathematical operations on step regardless to use it.

theflyingzamboni: Also, it's just a one liner in python:
`self.precision = abs(Decimal(str(step)).as_tuple().exponent)`

backend/nodes/properties/inputs/numeric_inputs.py
theflyingzamboni: > Actually, maybe we could even get rid of this parameter completely if we say: offset = minimum if minimum is not None else 0.

This actually doesn't work in a scenario where you have no minimum or maximum, but you still want odd numbers. Then minimum == None, but you still need offset to be 1.

backend/nodes/image_dim_nodes.py
RunDevelopment: > It is used here.

I know that `precision` is used in the current formula... Me point is that this can be removed. 

> Because step could be something like 0.5, you can't simply multiply the value by step, so you need to perform mathematical operations on step regardless to use it.

I don't get what you mean. 

Also, "mathematical operations" are literally all that computers can do. You're gonna have to be a bit more specific.

> Given that, it makes more sense to do it once in one place and simply pass the value around.

It doesn't. From my perspective, it does not make sense. As I've shown above, `step`+`offset` is strictly more powerful than `precision`. 

You can literally convert all `precision`,`step`,`offset` triples into equivalent `step`,`offset` pairs. So why would we even bother with `precision`?

backend/nodes/properties/inputs/numeric_inputs.py
RunDevelopment: Since min=None effective means min=-∞, yes, offset is technically necessary.

However, are we ever going to have an input with an unbound range?

backend/nodes/image_dim_nodes.py
theflyingzamboni: > Also, "mathematical operations" are literally all that computers can do. You're gonna have to be a bit more specific.

The mathematical operations that have to be performed to transform step into a precision value, e.g. that getPrecision function that you wrote above. This needs to be done in both the frontend *and* the backend, as the backend performs its own value correction in case some invalid value slips through the frontend. If we only use step or step+offset, then the getPrecision function will need to be performed by the frontend, then again by an equivalent function in the backend. On the other hand, if we just include a precision attribute and calculate it from step or step+offset at object initialization in the backend, this only has to be calculated one time.

RunDevelopment: > as the backend performs its own value correction in case some invalid value slips through the frontend

As I said before, `precision` is not necessary on the backend. All these corrections can be done without `precision`.

backend/nodes/properties/inputs/numeric_inputs.py
theflyingzamboni: > However, are we ever going to have an input with an unbound range?

Well that was my point. We can have full flexibility, or simplify for the range of scenarios we think we might reasonably encounter. We just need to pick which of those we want.

backend/nodes/image_dim_nodes.py
RunDevelopment: `precision` is literally only needed in the frontend for the numeric input component. 

A `step`-`offset` pair is enough for the correction, again, as I said before. To be more specific, the formula is this:

```
corrected_value = round((value - offset) / step) * step + offset
```

`precision` is just unnecessary complexity.

theflyingzamboni: > As I said before, precision is not necessary on the backend. All these corrections can be done without precision.

Then what is that equation? There may be one, so if you know it, just write it out instead of repeating that it can be done.

backend/nodes/properties/inputs/numeric_inputs.py
RunDevelopment: I'm for simplifying. I don't think we will have much use for unbound ranges and if we did, we could approximate them by just choosing very large min-max values. offset can deal with large min-max values if we set it to `offset = min % step`, so there won't be any problems with floating point precision in the correction formula.

theflyingzamboni: Works for me.

RunDevelopment (approved):

PR #305, opened 2022-05-21
Implement NumberInput-equivalent changes in SliderInput

theflyingzamboni:
-Implemented onBlur addition from NumberInput in SliderInput
-Added onChange functions for both slider and number input to keep them synced with addition of onBlur functionality
-Added dynamic width for number input field beside slider that will alter field width based on precision and units (will need adjustment if we ever allow numbers >=1000)

### DIFF
--- a/src/renderer/components/inputs/SliderInput.tsx
+++ b/src/renderer/components/inputs/SliderInput.tsx
@@ -14,16 +14,17 @@ import {
     VStack,
 } from '@chakra-ui/react';
 import { memo, useEffect, useState } from 'react';
+import { getPrecision } from './NumberInput';
 import { InputProps } from './props';
 
 interface SliderInputProps extends InputProps {
     min: number;
     max: number;
-    precision: number;
+    offset: number;
     step: number;
     controlsStep: number;
     def?: number;
-    unit?: number | null;
+    unit?: string | null;
     accentColor: string;
     ends?: [string | number, string | number] | null;
     noteExpression?: string;
@@ -47,7 +48,7 @@ const SliderInput = memo(
         def,
         min,
         max,
-        precision,
+        offset,
         step,
         controlsStep,
         unit,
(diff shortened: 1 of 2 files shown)


src/renderer/components/inputs/SliderInput.tsx
```diff
@@ -57,17 +58,70 @@ const SliderInput = memo(
...
+            setInputString(String(sliderInput));
+            setSliderValue(sliderInput);
+        };
+
+        const onNumberInputChange = (numberAsString: string) => {
+            setInputString(numberAsString);
+            setSliderValue(Number(numberAsString));
+        };
+
+        const onBlur = () => {
```
RunDevelopment, line 84: This is just a copy of the onBlur logic from inputs/NumberInput, right?

Would it make sense to move all the logic around Chakra's `<NumberInput>` element into its own component?

src/renderer/components/inputs/SliderInput.tsx
```diff
@@ -117,15 +171,14 @@ const SliderInput = memo(
...
-                        }}
+                        value={addUnit(inputString)}
+                        onBlur={onBlur}
+                        onChange={onNumberInputChange}
                     >
                         <NumberInputField
                             m={0}
                             p={1}
-                            w="3.1rem"
+                            w={`${dynamicNumInputWidth}rem`}
```
RunDevelopment, line 181: Now sliders with different units will be misaligned. I think it would be easier to just set to "5em" or something.

src/renderer/components/inputs/SliderInput.tsx
theflyingzamboni: It is 99% similar. The only difference is something that I can actually take out now anyway in NumberInput because of the useEffect you added I believe.

src/renderer/components/inputs/SliderInput.tsx
theflyingzamboni: 5 was the first thing I tried, and I didn't like how much it crowded out the slider with negative space. Your point is a good one though, so I've changed it so a constant amount is added if a unit is present. The precision is still a dynamic factor though, so sliders won't have all that negative space. We can always tweak the unit constant later. Right now it is enough to accommodate mm or three narrower characters.

theflyingzamboni: I tried that initially, but really didn't like how much it crowded the sliders with negative space and made them smaller, which impedes their controllability. Your point is a good one though. For now, I'm keeping precision dynamic (I don't *expect* to have one node with sliders of different precision, but we can always revisit this if the situation arises), but I've made it so a static amount is added if unit is present. My admittedly a little hacky solution to one slider having units and another not is to just added unit="" (which does not impede number deletion like characters currently do, until that's fixed) to sliders without units. It's not a beautiful solution, but I think it's preferable to having a lot of negative space in slider nodes from a fixed width.

src/renderer/components/inputs/SliderInput.tsx
theflyingzamboni: There are actually some differences in how the NumberInput field is formatted in SliderInput, after looking at it again once I knew what you were describing (size, stepper m and p, input field width, etc.). This *could* I think be gotten around by adding extra inputs to the NumericalInput function after separating it out, but then these would need to be added to the Props interface as well. So I'm thinking leave it as is for this release, and refactor it later if we want.

PR #306, opened 2022-05-21
Swap undo/redo handlers from electron accelerators to handler hooks

joeyballentine:


src/renderer/components/HistoryProvider.tsx
```diff
@@ -122,6 +122,24 @@ export const HistoryProvider = ({ children }: React.PropsWithChildren<unknown>):
...
+        'ctrl+z, cmd+z',
+        () => {
+            historyObj.history = historyObj.history.undo();
+            apply(historyObj.history.current);
+        },
+        [apply]
+    );
+
+    useHotkeys(
+        'ctrl+y, cmd+y',
```
RunDevelopment, line 135: Please add 'ctrl+shift+z, cmd+shift+z' to this.

src/renderer/components/HistoryProvider.tsx
```diff
@@ -100,7 +100,7 @@ export const HistoryProvider = ({ children }: React.PropsWithChildren<unknown>):
 
         const id = setTimeout(() => {
             historyObj.history = historyObj.history.commit([getNodes(), getEdges()]);
-        }, 500);
+        }, 250);
```
RunDevelopment, line 103: Why?

src/renderer/components/HistoryProvider.tsx
joeyballentine: oh yeah, duh

src/renderer/components/HistoryProvider.tsx
joeyballentine: Because it's not needed anymore. I only increased it to 500 to try to fix the number input issues

RunDevelopment (approved):

PR #307, opened 2022-05-21
Add image dimensions node

joeyballentine:
image

### DIFF
--- a/backend/nodes/image_dim_nodes.py
+++ b/backend/nodes/image_dim_nodes.py
@@ -194,3 +194,36 @@ class EdgeCropNode(NodeBase):
         result = img[top : h - bottom, left : w - right]
 
         return result
+
+
+@NodeFactory.register("chainner:image:get_dims")
+class GetDimensionsNode(NodeBase):
+    """Node for getting the dimensions of an image"""
+
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
+        self.description = (
+            "Get the Height, Width, and number of Channels from an image."
+        )
+        self.inputs = [
+            ImageInput(),
+        ]
+        self.outputs = [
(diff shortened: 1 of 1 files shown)


backend/nodes/image_dim_nodes.py
```diff
@@ -194,3 +194,36 @@ def run(
...
+        ]
+        self.category = IMAGE_DIMENSION
+        self.name = "Get Dimensions"
+        self.icon = "BsRulers"
+        self.sub = "Utility"
+
+    def run(
+        self,
+        img: np.ndarray,
+    ) -> any:
```
RunDevelopment, line 225: Why an `any` type hint? 

I don't know anything about py type hints, but wouldn't this stop the IDE from inferring the return type?

backend/nodes/image_dim_nodes.py
```diff
@@ -194,3 +194,36 @@ def run(
...
+        super().__init__()
+        self.description = (
+            "Get the Height, Width, and number of Channels from an image."
+        )
+        self.inputs = [
+            ImageInput(),
+        ]
+        self.outputs = [
+            NumberOutput("Height"),
+            NumberOutput("Width"),
```
RunDevelopment, line 214: I think we should use the same order as the Resize node.

```suggestion
            NumberOutput("Width"),
            NumberOutput("Height"),
```

RunDevelopment (approved):

RunDevelopment:
Oops, I meant to request changes, not approve.

backend/nodes/image_dim_nodes.py
joeyballentine: Oh yeah, I forgot to finish that

PR #309, opened 2022-05-22
Added base opacity and changed sliders to float

theflyingzamboni:


joeyballentine:
This needs a migration now to fix the order of the inputs

RunDevelopment:
Man, I really need to implement the input/output id idea I suggested. This PR wouldn't need a migration if we had that.

theflyingzamboni:
Ah dang. I thought the changes to inputs already made that unnecessary, and that it would just load with the default value. RIP.

joeyballentine:
@theflyingzamboni you're good to start changing it over to this new system

backend/nodes/image_util_nodes.py
```diff
@@ -326,7 +350,7 @@ def run(
     ) -> np.ndarray:
         """Adjusts the position of an image"""
 
-        h, w, _ = get_h_w_c(img)
```
RunDevelopment, line 329: I think it's more explicit to destructure the tuple fully instead of doing `[:2]`.

backend/nodes/image_util_nodes.py
```diff
@@ -81,6 +103,8 @@ def run(
         center_x = imgout.shape[1] // 2
         center_y = imgout.shape[0] // 2
 
+        # Apply opacity to base, then overlay A and B sequentially at corresponding opacities
+        imgout[:, :, 3] *= opbase
```
RunDevelopment, line 107: What if it's an RGB image?

theflyingzamboni: Lines 68-69 explicitly adds an alpha channel to anything with c==3. Why didn't I use cv2.cvtcolor instead of a numpy operation? ¯\_(ツ)_/¯

RunDevelopment (approved):

PR #310, opened 2022-05-22
Use input group for units

joeyballentine:
image
This way it doesn't interrupt typing

### DIFF
--- a/src/renderer/components/inputs/NumberInput.tsx
+++ b/src/renderer/components/inputs/NumberInput.tsx
@@ -1,5 +1,7 @@
 import {
     HStack,
+    InputGroup,
+    InputLeftAddon,
     NumberDecrementStepper,
     NumberIncrementStepper,
     NumberInput,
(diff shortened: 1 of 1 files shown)


src/renderer/components/inputs/NumberInput.tsx
```diff
@@ -27,6 +29,22 @@ export const getPrecision = (n: number) => {
...
+};
+
+const clampMin = (value: number, min_val: number | undefined | null) => {
+    if (min_val !== undefined && min_val !== null) {
+        const valOrMin = Math.max(value, min_val);
+        return valOrMin;
+    }
+    return value;
+};
+
```
RunDevelopment, line 47: Could we combine this into a single `clmap` function?

joeyballentine: Yep

joeyballentine: I'll do that separately though since I'm pretty sure I'll need to do that in the slider as well

PR #311, opened 2022-05-22
Make favorite button a bit more obvious

joeyballentine:
image

### DIFF
--- a/src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
+++ b/src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
@@ -1,4 +1,3 @@
-import { StarIcon } from '@chakra-ui/icons';
 import {
     AccordionButton,
     AccordionIcon,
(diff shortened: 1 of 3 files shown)


src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
```diff
@@ -1,4 +1,3 @@
-import { StarIcon } from '@chakra-ui/icons';
```
RunDevelopment, line 1: Maybe we should promote unused variable to an error?

We could also say that the ESLint CI (and only the CI) fails when there are any warnings. That's how we do it [over at Prism](https://github.com/PrismJS/prism/blob/4c87d4186a27dcecd1786b90d31bdf109e4d57b0/package.json#L16).

RunDevelopment (approved):

src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
joeyballentine: Sure, to both of those

PR #321, opened 2022-05-23
Unique ids for input and output

RunDevelopment:
Changes:

- All backend inputs now extend the `BaseInput` class
- All backend outputs now extend the new `BaseOutput` class
- All inputs and outputs now guarantee that their dict representation is a super set of the dict representation of the class they extend.
- The processor now takes edge handles as `{ id: string, index: number }` (where id is the node id) instead of `{ id: string }` (where id is an edge handle that has to be parsed).
- Each input and output now has a unique id (unique per node class). 

### About ids

The input/output ids are only unique per node. If the id of input/output is not explicitly set using `with_id(number)`, then the id of the input/output will be its index in the inputs/outputs array. Since no input/output currently uses the `with_id` method, all ids will simply be the index right now. 

In the JSON returned by `http://backend/nodes`, ids look like this:

<details>
<summary>Long JSON</summary>

```json
    {
        "schemaId": "chainner:image:get_dims",
        "name": "Get Dimensions",
        "category": "Image (Dimensions)",
        "inputs": [
            {
                "id": 0,
                "type": "numpy::2d",
                "label": "Image",
                "optional": false,
                "hasHandle": true
            }
        ],
        "outputs": [
            {
                "id": 0,
                "type": "number::any",
                "label": "Width"
            },
            {
                "id": 1,
                "type": "number::any",
                "label": "Height"
            },
            {
                "id": 2,
                "type": "number::any",
                "label": "Channels"
            }
        ],
        "description": "Get the Height, Width, and number of Channels from an image.",
        "icon": "BsRulers",
        "subcategory": "Utility",
        "nodeType": "regularNode"
    },
    {
        "schemaId": "chainner:image:hue_and_saturation",
        "name": "Hue & Saturation",
        "category": "Image (Adjustments)",
        "inputs": [
            {
                "id": 0,
                "type": "numpy::2d",
                "label": "Image",
                "optional": false,
                "hasHandle": true
            },
            {
                "id": 1,
                "type": "number::slider",
                "label": "Hue",
                "optional": false,
                "hasHandle": true,
                "min": -180,
                "max": 180,
                "noteExpression": null,
                "def": 0,
                "offset": 0,
                "step": 1,
                "controlsStep": 1,
                "unit": null,
                "ends": [
                    -180,
                    180
                ]
            },
            {
                "id": 2,
                "type": "number::slider",
                "label": "Saturation",
                "optional": false,
                "hasHandle": true,
                "min": -255,
                "max": 255,
                "noteExpression": null,
                "def": 0,
                "offset": 0,
                "step": 1,
                "controlsStep": 1,
                "unit": null,
                "ends": [
                    -255,
                    255
                ]
            }
        ],
        "outputs": [
            {
                "id": 0,
                "type": "numpy::2d",
                "label": "Image"
            }
        ],
        "description": "Adjust the hue and saturation of an image.",
        "icon": "MdOutlineColorLens",
        "subcategory": "Adjustments",
        "nodeType": "regularNode"
    },
```

</details>

So why are ids useful? They allow us to reorder input/output without breaking save files.

#### Example: Load Image outputs

Here how you would reorder the output of Load Image. In this case, I want to put the "Image" output last:

```py
        self.outputs = [
            ImageOutput(),
            DirectoryOutput(),
            TextOutput("Image Name"),
        ]
```

First, you have to make the all ids explicit:

```py
        self.outputs = [
            ImageOutput().with_id(0),
            DirectoryOutput().with_id(1),
            TextOutput("Image Name").with_id(2),
        ]
```

Now, we can safely reorder them.

```py

        self.outputs = [
            DirectoryOutput().with_id(1),
            TextOutput("Image Name").with_id(2),
            ImageOutput().with_id(0),
        ]
```

We also have to adjust the actual `run` implementation. In this case, we change it like this:
```diff
-        self.result = (img, dirname, basename)
+        self.result = (dirname, basename, img)
         return self.result
```

(I'm ignoring here that `self.result` is also used elsewhere. That code also has to be adjusted.)

Here's how the result looks in a chain I made before changing the Load Image node:

image

I didn't need to write a migration and old saves still work. Yes, it even runs without problems :)

### Example: Overlay Images base opacity

This is how #309 can done using ids:

First, we add the new input at the end of the input list:

```py
        self.inputs = [
            ImageInput("Base"),
            ImageInput("Overlay A"),
            SliderInput("Opacity A", maximum=100, default=50),
            ImageInput("Overlay B ", optional=True),
            SliderInput("Opacity B", maximum=100, default=50),
            SliderInput("Base Opacity", maximum=100, default=50),
        ]
```

All slider input has a default value, so this addition does not break old save files, so no migration necessary.

Now, we make the ids explicit again:
```py
        self.inputs = [
            ImageInput("Base").with_id(0),
            ImageInput("Overlay A").with_id(1),
            SliderInput("Opacity A", maximum=100, default=50).with_id(2),
            ImageInput("Overlay B ", optional=True).with_id(3),
            SliderInput("Opacity B", maximum=100, default=50).with_id(4),
            SliderInput("Base Opacity", maximum=100, default=50).with_id(5),
        ]
```

And reorder:

```py
        self.inputs = [
            ImageInput("Base").with_id(0),
            SliderInput("Base Opacity", maximum=100, default=50).with_id(5),
            ImageInput("Overlay A").with_id(1),
            SliderInput("Opacity A", maximum=100, default=50).with_id(2),
            ImageInput("Overlay B ", optional=True).with_id(3),
            SliderInput("Opacity B", maximum=100, default=50).with_id(4),
        ]
```
 The order of arguments in `run` is still that of the inputs array (completely ignoring ids), so we change `run` to this:

```py
    def run(
        self,
        base: np.ndarray,
        op_base_int: int,
        ov1: np.ndarray,
        op1_int: int = 50,
        ov2: Union[np.ndarray, None] = None,
        op2_int: int = 50,
    ) -> np.ndarray:
```

And we get this:

image

Again, old save files will just work without a migration.


joeyballentine (approved):

theflyingzamboni:
Looks nice and straightforward. I'll update my overlay PR once this is merged.

PR #326, opened 2022-05-24
Changed scales of adjustment node sliders

theflyingzamboni:
-Changed uint scales of saturation, brightness, and contrast slider to 100 scale
-Changed existing 100 scales to float


src/common/migrations.js
```diff
@@ -346,5 +363,14 @@ export const migrate = (_version, data) => {
         }
     }
 
+    // v0.9.0 or whatever is next
+    if (semver.lt(version, '0.9.0')) {
```
RunDevelopment, line 367: Doesn't that mean that all our saves will be broken again until we release this version?

RunDevelopment: I would argue that saving and running chains in dev versions is very useful, because it means that we are more likely to detect bugs. It's also a requirement for the automated testing of chains.

RunDevelopment: > I also have a lot of "corrupted" chains that don't work properly because I saved them with WIP nodes

You would have none of them, had we used a migration counter independent of the app version from the start.

> I don't care about making sure those work

As I said, making chains work in dev versions is a requirement for automated testing, so I care a lot about that.

> Like changing the version immediately after release to be like 0.8.2-dev or something.

I dislike that because it means that we would have to decide whether to release a patch/minor/major next when releasing a version.

> someone can download the code off of master and think they're on the latest release

That's how it is in almost any software project, though. Unstable development branches have somewhat fallen out of favor, because you can just tag stable releases. If a user decides to check out the source code, they should expect to be on the bleeding edge.

However, you could make the download counter link to the latest release to make it easier for people to find.

joeyballentine: > You would have none of them, had we used a migration counter independent of the app version from the start.

This is not true, becsuse I would not have made migrations at all for the WIP changes I made, like my original NCNN node that had the load model and upscale steps combined into one solely for testing purposes. There's no reason to make migrations for that stuff since I was the only one it would affect. 

Anyway, I think I have a solution which is to just base it off of timestamps. That's how Knex migrations work for SQL stuff and I think it would work here. Just add a new migration with the UTC timestamp of when you made it, then the migration would be independent of version and we wouldn't need to keep a manual count. Plus we already store a timestamp in the save file. 

> As I said, making chains work in dev versions is a requirement for automated testing, so I care a lot about that.

For automated testing shouldn't we be updating the tests ourselves rather than relying on a migration system to keep it up to date for us?

joeyballentine: Could i merge this as-is now though?

RunDevelopment (approved):

PR #327, opened 2022-05-24
Update RF & change footer

joeyballentine:


src/renderer/components/node/NodeFooter.tsx
```diff
@@ -35,16 +20,7 @@ const NodeFooter = ({
...
-                <Icon
-                    as={isLocked ? LockIcon : UnlockIcon}
-                    color={iconShade}
-                    cursor="pointer"
-                    mb={-1}
-                    mt={-1}
-                    onClick={() => toggleNodeLock(id)}
-                />
-            </Center>
+            <Center className="nodrag" />
```
RunDevelopment, line 23: Are those centers on either side necessary?

joeyballentine: Probably not

RunDevelopment (approved):

PR #333, opened 2022-05-24
Improved favorites icon and section

RunDevelopment:
image

image


joeyballentine:
Should we add ", or by right clicking and using the context menu" to it? Mostly just to let people know that the context menu exists

RunDevelopment:
Good idea.

PR #336, opened 2022-05-25
Copy & Paste

joeyballentine:
Fully functional copying and pasting. The only issue is an existing issue where you can't duplicate or copy and paste something more than once due to the way we derive new unique ids from existing ones (which will result in the same id every time you duplicate/copypaste).

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -20,7 +20,6 @@ import {
     IteratorSize,
     Mutable,
     NodeData,
-    NodeSchema,
     Size,
 } from '../../common/common-types';
 import { ipcRenderer } from '../../common/safeIpc';
@@ -725,7 +724,7 @@ export const GlobalProvider = ({
                     .map<Node<NodeData>>((n) => {
                         const newId = deriveUniqueId(n.id);
                         if (!n.parentNode) {
-                            return {
+                            const returnData = {
                                 ...n,
                                 id: newId,
                                 position: {
@@ -738,10 +737,12 @@ export const GlobalProvider = ({
                                 },
                                 selected: false,
                             };
+                            delete returnData.handleBounds;
+                            return returnData;
                         }
 
                         const parentId = deriveUniqueId(n.parentNode);
-                        return {
+                        const returnData = {
                             ...n,
                             id: newId,
                             data: {
@@ -752,6 +753,8 @@ export const GlobalProvider = ({
                             parentNode: parentId,
                             selected: false,
                         };
+                        delete returnData.handleBounds;
+                        return returnData;
                     });
                 return [...nodes, ...newNodes];
             });
@@ -771,7 +774,6 @@ export const GlobalProvider = ({
                         }
                         target = deriveUniqueId(target);
                         targetHandle = targetHandle?.replace(e.target, target);
-                        console.log({ source, sourceHandle, target, targetHandle });
                         return {
                             ...e,
                             id: createUniqueId(),
@@ -817,7 +819,6 @@ export const GlobalProvider = ({
             const selectedNodes = getNodes().filter((n) => n.selected);
             const selectedEdges = getEdges().filter((e) => e.selected);
             const copyData = JSON.stringify({ nodes: selectedNodes, edges: selectedEdges });
-            console.log('🚀 ~ file: GlobalNodeState.tsx ~ line 808 ~ copyData', copyData);
             clipboard.writeText(copyData, 'clipboard');
         },
         [getNodes]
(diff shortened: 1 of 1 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -797,6 +813,33 @@ export const GlobalProvider = ({
         [modifyNode]
     );
 
+    useHotkeys(
+        'ctrl+c, cmd+c',
+        () => {
+            const selectedNodes = getNodes().filter((n) => n.selected);
+            const selectedEdges = getEdges().filter((e) => e.selected);
+            const copyData = JSON.stringify({ nodes: selectedNodes, edges: selectedEdges });
+            clipboard.writeText(copyData, 'clipboard');
```
RunDevelopment, line 822: This will allow users to just paste nodes as text. Worse, chaiNNer will interpret any text currently in the clipboard to be JSON it can parse. That's a massive usability issue.

As the MIME type, use something like `application/chainner.chain`.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -756,13 +753,19 @@ export const GlobalProvider = ({
...
                     });
                 return [...nodes, ...newNodes];
             });
 
             changeEdges((edges) => {
                 const newEdges = edges
-                    .filter((e) => nodesToCopy.has(e.target))
+                    .filter(
+                        (e) =>
+                            (nodesToCopy.has(e.target) && nodesToCopy.has(e.source)) ||
```
RunDevelopment, line 766: Why did you change how edges are copied?

joeyballentine: The target and source need to exist for the edge to work, and I only want to copy it if it was copied. I now realize this condition is wrong for that though

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: The only way to do that is to change this to writeBuffer

joeyballentine:
Fixed the issues. Also made it so it automatically selects the new duplicated or copy/pasted nodes and unselects the original.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1,3 +1,4 @@
+import { clipboard } from 'electron';
```
RunDevelopment, line 1: Why do you use an electron-specific `clipboard` API instead of the standard `clipboard` Web API?

joeyballentine: Because I didn't know that existed lol

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -800,6 +816,37 @@ export const GlobalProvider = ({
...
+                edges: Edge[];
+            };
+            if (nodes.length > 0) {
+                const nodesToCopy = new Set(nodes.map((n) => n.id));
+                const edgesToCopy = edges.length > 0 ? new Set(edges.map((e) => e.id)) : null;
+                copyNodesAndEdges(nodesToCopy, edgesToCopy);
+            }
+        },
+        [copyNodesAndEdges]
+    );
```
RunDevelopment, line 848: That's not a good way of making sure types align. Right now, if you changed this type to have one more field, or you removed a field from `writeBuffer` above, then would you not get a type error.

It's better to make an interface to make sure that both the write and read have no choice but to follow the interface.

```suggestion
interface ClipboardChain {
                nodes: Node[];
                edges: Edge[];
            }
            
    useHotkeys(
        'ctrl+c, cmd+c',
        () => {
            const selectedNodes = getNodes().filter((n) => n.selected);
            const selectedEdges = getEdges().filter((e) => e.selected);
            cosnt data: ClipboardChain = { nodes: selectedNodes, edges: selectedEdges }
            const copyData = Buffer.from(
                JSON.stringify(data)
            );
            clipboard.writeBuffer('application/chainner.chain', copyData, 'clipboard');
        },
        [getNodes]
    );

    useHotkeys(
        'ctrl+v, cmd+v',
        () => {
            const { nodes, edges } = JSON.parse(
                clipboard.readBuffer('application/chainner.chain').toString()
            ) as ClipboardChain;
            if (nodes.length > 0) {
                const nodesToCopy = new Set(nodes.map((n) => n.id));
                const edgesToCopy = edges.length > 0 ? new Set(edges.map((e) => e.id)) : null;
                copyNodesAndEdges(nodesToCopy, edgesToCopy);
            }
        },
        [copyNodesAndEdges]
    );
```

src/renderer/contexts/GlobalNodeState.tsx
RunDevelopment: I even linked to it in my previous comment...

joeyballentine: Whoops. I think I just just saw clipboard.write, tried to use it with electron, couldn't, and then used writeBuffer. Should have taken a look at the link instead, sorry.

Anyway, is there a benefit to using the web one instead of the electron one?

RunDevelopment: Mostly that it's more commonly used. Web APIs have the advantage that a lot of people use them, so they are more stable and better documented. Also, Electron started before the web clipboard API was introduced, so they might remove the Electron API at some point.

RunDevelopment (approved):

PR #346, opened 2022-05-26
Update eslint & make memos consistent

joeyballentine:


package.json
```diff
@@ -174,6 +173,8 @@
     "decompress": "^4.2.1",
     "electron-log": "^4.4.1",
     "electron-squirrel-startup": "^1.0.0",
+    "eslint-plugin-prefer-arrow-functions": "^3.1.4",
+    "eslint-plugin-react-memo": "^0.0.3",
```
RunDevelopment, line 177: Those are not dependencies but dev dependencies.

joeyballentine: Oops

PR #350, opened 2022-05-26
Useless union

RunDevelopment:
React.ReactNode contains string.


src/renderer/components/NodeSelectorPanel/TextBox.tsx
```diff
@@ -4,7 +4,7 @@ import { MouseEventHandler, memo } from 'react';
 export interface TextBoxProps {
     text: string;
     collapsed?: boolean;
-    toolTip?: React.ReactNode | string;
+    toolTip?: React.ReactNode | string | string;
```
joeyballentine, line 7: Did you mean to add another string instead of removing the existing one?

RunDevelopment: Lol, I was testing out an ESLint rule and forgot to remove this.

RunDevelopment:
Npm couldn't install deps. Could you please re-run the workflow @joeyballentine?

Issue #352, opened 2022-05-26
Regex replace node


RunDevelopment:
Add a node that replaces all occurrences of a regex with a given substitution pattern. This is useful for manipulating file names.

PR #355, opened 2022-05-26
Refactored how required py dependencies are stored

RunDevelopment:
This will make it easy to add new required dependencies if we ever need to (foreshadowing).


joeyballentine:
What required dependencies are you going to add?

RunDevelopment:
re2 for #352.

PR #359, opened 2022-05-26
Change how pip deps are handled

joeyballentine:


src/renderer/helpers/pip.ts
```diff
@@ -0,0 +1,82 @@
...
+        ipcRenderer
+            .invoke('get-python')
+            .then(({ python, version }) => {
+                this.pythonPath = python;
+                this.pythonVersion = version;
+            })
+            .catch((error) => {
+                log.error(error);
+            });
+    }
```
RunDevelopment, line 38: This is so hacky. Whether this promise resolves determines whether all of the instance methods actually work. This is not how class constructors are supposed to work. A class constructor is supposed to construct a valid instance of a class, but  `PipManager` only become valid when and if the promise resolves. You are essentially abusing this class to function as a namespace.

Just remove the class and make this what it actually is: a collection of functions. If a function need to know the python path, it just call a `getPythonKeys(): Promise<{ python: string, version: string }>` function.

joeyballentine: My goal was to avoid having to call the IPC method every time that was needed. Before we solved that just using react state. I suppose I could just add that stuff back and make it pass in the python path again

RunDevelopment: Just cache the promise. That's what I do [here](https://github.com/joeyballentine/chaiNNer/blob/b19dca17a448b73c8a3886378a82ef1837748031/src/main/main.ts#L122) too.

joeyballentine: Sounds good

PR #362, opened 2022-05-27
Added tests for save files

RunDevelopment:
I added tests for reading and writing save files.


.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
+name: Tests
+
+# Controls when the workflow will run
+on:
+    schedule:
+        - cron: '*/15 * * * *'
```
joeyballentine, line 6: Do we need the tests to run on a Cron? The only reason the lint one has that was for a workaround to get those inline code comments working, and even then I'm not 100% sure it was necessary

joeyballentine (approved):

PR #364, opened 2022-05-27
Add pytest

joeyballentine:
Doesn't add any useful tests yet, just adds the testing framework


.gitignore
```diff
@@ -1,3 +1,5 @@
+## JS
+
```
RunDevelopment, line 2: ?

joeyballentine: I combined the root gitignores and the backend gitignores, so I separated them out

joeyballentine: perhaps JS wasn't the best section name to give that

RunDevelopment:
Why did you use `tests/backend` for backend tests? Could you use `backend/tests` instead? That would also mean that you could remove a `__init__.py`s.

joeyballentine:
> Why did you use `tests/backend` for backend tests? Could you use `backend/tests` instead? That would also mean that you could remove a `__init__.py`s.

I could yes, but then I would be bundling the tests with chaiNNer, unless I find a way to specifically ignore a folder from the extraResources electron option

RunDevelopment:
> unless I find a way to specifically ignore a folder from the extraResources electron option

[Seems easy enough](https://www.electron.build/file-patterns#multiple-glob-patterns).

joeyballentine:
> > unless I find a way to specifically ignore a folder from the extraResources electron option
> 
> [Seems easy enough](https://www.electron.build/file-patterns#multiple-glob-patterns).

We don't use electron-builder

RunDevelopment (approved):

PR #371, opened 2022-05-28
Added test for minimum dependencies

RunDevelopment:


RunDevelopment:
Done @joeyballentine. And I even found an import that didn't work without cv2.

PR #374, opened 2022-05-29
Made Load Image preview constant dimensions

theflyingzamboni:
Image preview area will now always be 200x200 with image centered.

### DIFF
--- a/src/renderer/components/inputs/previews/ImagePreview.tsx
+++ b/src/renderer/components/inputs/previews/ImagePreview.tsx
@@ -42,6 +42,11 @@ const ImagePreview = memo(({ path, schemaId, id }: ImagePreviewProps) => {
     const [isCpu] = useIsCpu;
     const [isFp16] = useIsFp16;
 
+    const setWidth =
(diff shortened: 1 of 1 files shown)


src/renderer/components/inputs/previews/ImagePreview.tsx
```diff
@@ -42,6 +42,11 @@ const ImagePreview = memo(({ path, schemaId, id }: ImagePreviewProps) => {
     const [isCpu] = useIsCpu;
     const [isFp16] = useIsFp16;
 
+    const setWidth =
+        img && img.height > img.width ? `${200 * (img.width / img.height)}px` : '200px';
+    const setHeight =
+        img && img.width > img.height ? `${200 * (img.height / img.width)}px` : '200px';
```
RunDevelopment, line 48: Why? You just implemented the `maxW` & `maxH` the `Image` element previously used.

theflyingzamboni: True. I wrote that before I realized I needed to wrap the element, guess it's not doing anything different.

RunDevelopment (approved):

PR #377, opened 2022-05-30
Refactored pip usage

RunDevelopment:
This completely changes how pip is used. All usages of pip will now go through the `runPip` function.

Changes:
- Merged `get-gpu-name` and `get-has-nvidia` into `get-nvidia-gpu-name` IPC call.
- Python keys are now shared globally via `getPythonKeys` and `setPythonKeys`.
- Environment/platform variables are now exported from `env.ts`.
- Removed `pipInstallWithProgress`. Used the new `pipInstall` instead. However, `pipInstall` doesn't implement reporting progress yet. This is somewhat difficult to do, so I plan on adding this later.
- Simplified the python check.
- Setting up integrated python no longer installs required dependencies. Dependencies are checked (and installed) later on anyway, so there is no need to.
- DependencyManager now uses `useContext(AlertBoxContext)`.
- Refactored DependencyManager in general. I tested it by repeatedly installing and uninstalling some dependencies and everything worked fine.


joeyballentine:
One thing I personally wanted to change was python keys. That's from back when I used to save the location of both the python binary and pip binaries. I'd rather it just be separate pythonVersion and pythonPath variables/handlers if you wouldn't mind changing that while you're at it

RunDevelopment:
I think separating those is a bad idea, actually. The python version is specific to the python executable of that python path, so I'd rather have them be together.

RunDevelopment:
Then I'll go with Python info.

PR #384, opened 2022-05-31
Test different python versions in CI

RunDevelopment:
Run tests for all currently alive python versions.


RunDevelopment:
In this branch? Sure, why not.

joeyballentine:
> In this branch? Sure, why not.

I don't want CI failing for subsequent PRs, so might as well get it all passing here

PR #396, opened 2022-06-05
Blur fix

theflyingzamboni:
-Changed Blur node to use proper box filter instead of approximating gaussian. Can now accept non-odd values.
-Changed Blur so that amount is now the radius instead of kernel size. Kernel size is 2r+1.
-Blur and Gaussian Blur nodes now accept precision 1 decimals.
-Migrated original Blur node to Gaussian Blur node, since it was an approximation before. Used simple linear equation to roughly convert input values to maintain equivalent blur amount.
-Added basic Median Blur node using cv2.medianBlur. Amount is blur radius. Kernel size is 2r+1.
-Amount 0 for any blur node will output original image.

Closes joeyballentine/chaiNNer#139


src/common/migrations.js
```diff
@@ -298,6 +298,18 @@ const toV090 = (data) => {
...
+        if (node.data.schemaId === 'chainner:image:blur') {
+            // eslint-disable-next-line no-param-reassign, prefer-destructuring
+            node.data.schemaId = 'chainner:image:gaussian_blur';
+            // eslint-disable-next-line no-param-reassign, prefer-destructuring
+            node.data.inputData['1'] =
+                Math.round((1.16531 * node.data.inputData['1'] - 0.153601) * 10) / 10;
+            // eslint-disable-next-line no-param-reassign, prefer-destructuring
+            node.data.inputData['2'] =
+                Math.round((1.16531 * node.data.inputData['2'] - 0.153601) * 10) / 10;
+        }
```
RunDevelopment, line 312: This will cause save to break for people with a dev version of chainner (e.g. you and me). The new migration system I made was added specifically to prevent this.

Move your changes into its own migration function and add that function to the `migrations` array in line 338.

theflyingzamboni: I just added it to the existing current migration function with the changes for the adjustment nodes. Did you not integrate that into your new migration system?

theflyingzamboni: Yeah, I can confirm that both of these migrations function fine. If I save in dev version, clear, and load the save, the nodes are not changed.

RunDevelopment: >  I can confirm that both of these migrations function fine.

No, they don't. With the new migration, add a Blur node (the one you fixed), save the file, open the file, and see how your new blur node got converted into a gauss node.

Please you [the new migration system](https://github.com/joeyballentine/chaiNNer/pull/331) as intended. It guarantees that this will never be a problem ever again, no matter what the migration is.

RunDevelopment (approved):

PR #399, opened 2022-06-07
Image Metrics Node

theflyingzamboni:
Added node to calculate MSE, PSNR, and SSIM values for an image comparison. Outputs each value as a string to use with things like Add Caption.

Closes joeyballentine/chaiNNer#25


theflyingzamboni:
Check that I wrote the equations correctly, please (though SSIM was taken from elsewhere).

RunDevelopment (approved):

PR #401, opened 2022-06-07
Add collapse animation

joeyballentine:


RunDevelopment:
Things jump around a lot when opening the selector panel because the fav text gets expanded all at once.

image

Maybe we could fix its height at `1.5em` (or whatever the height of a single line is)?

RunDevelopment:
Also, there is a horizontal scrollbar during the open animation.

image

joeyballentine:
@RunDevelopment Those issues have now been fixed

PR #402, opened 2022-06-07
Collapse button changes

joeyballentine:
- Made the collapse button larger
- Made the show/hide functionality of the buttons use opacity rather than conditionals. This fixes the error from useColorModeValue as well as allows a less jarring transition when they show.

image


src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
```diff
@@ -262,31 +263,32 @@ const NodeSelector = memo(({ schemata, height }: NodeSelectorProps) => {
...
+            <Button
+                _hover={{
+                    bg: useColorModeValue('gray.400', 'gray.600'),
+                    opacity: 1,
+                }}
+                aria-label="collapse"
+                bg={useColorModeValue('gray.300', 'gray.700')}
+                borderRadius={0}
+                borderRightRadius="xl"
+                h="200px"
```
RunDevelopment, line 275: I think that's too much. IMO, the problem was that the button was too thin, and you fixed that. `h=100px` is big enough IMO.

image

RunDevelopment (approved):

PR #404, opened 2022-06-07
PIP progress WIP

RunDevelopment:
My progress towards PIP progress parsing so far.


joeyballentine:
I'm going to close this as I do not see it being possible to implement at any point in the near future. We can revisit this at a later date, if pip makes it possible.

PR #413, opened 2022-06-10
New type system

RunDevelopment:
I implemented a set theory-based type system. Each type is the set of its possible values. So e.g. `number` is the type of all possible numbers and `0..1` is the range of all real number 0 <= x <= 1. We can also do set union (e.g. `1 | 2`) and set intersection (e.g. `0..2 & 1..Infinity`) on types. Read `src/common/types/type-system.md` for more details.

Notably, the type system itself does not support functions. Functions are a special construct and not a type. Chainner's nodes are typed as functions with some number of inputs and outputs.

### Unanswered questions

- Who registers type definitions?
    The backend defines quite a few types. E.g. every dropdown input class is its own type definition. All type definitions have to be registered, but who is responsible for that?
    Because I couldn't find a good answer yet, I just defined all types the backend uses as built-in types.


src/common/types/type-system.md
```diff
@@ -0,0 +1,176 @@
...
+E.g. `Image { width: uint, height: uint, channels: uint }` and `null` are structure types.
+
+### Structure type definitions
+
+Structure types also have type definitions.
+These definitions specify the fields all structure types of that name have.
+
+E.g. `struct Image { width: uint, height: uint, channels: uint }` and `struct null` are structure type definitions.
+
+Not that all fields are optional when instantiating a type.
```
joeyballentine, line 74:
```suggestion
Note that all fields are optional when instantiating a type.
```

joeyballentine:
This is insane dude, the amount of work that must have gone into doing all this... I can't even imagine

RunDevelopment:
That was about a week of work and I had a lot of fun. They say that every programmer wants to make their own language, and I guess that's mine.

joeyballentine:
This will probably take me quite a while to thoroughly look through and play around with.

joeyballentine (approved):

PR #416, opened 2022-06-10
Blend mode node

theflyingzamboni:
Replaced Overlay Images and Difference Nodes with Blend Images node, which currently supports all blending modes available in paint.net.

### DIFF
--- a/backend/src/nodes/image_util_nodes.py
+++ b/backend/src/nodes/image_util_nodes.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import math
-from typing import List, Tuple
+from typing import List, Tuple, cast
 
 import cv2
 import numpy as np
+import numpy.typing as npt
 from sanic.log import logger
 
 from .categories import IMAGE_UTILITY
@@ -16,10 +17,12 @@ from .utils.image_utils import blend_images, calculate_ssim
 from .utils.pil_utils import *
 from .utils.utils import get_h_w_c
 
+ndarray32 = npt.NDArray[np.float32]
 
-@NodeFactory.register("chainner:image:overlay")
+
+@NodeFactory.register("chainner:image:blend")
 class ImBlend(NodeBase):
-    """OpenCV transparency overlay node"""
+    """Blending mode node"""
 
     def __init__(self):
         """Constructor"""
@@ -27,7 +30,7 @@ class ImBlend(NodeBase):
         self.description = """Blends overlay image onto base image using 
             specified mode and opacities."""
         self.inputs = [
-            ImageInput("Base").with_id(0),
+            ImageInput("Base"),
             SliderInput(
                 "Base Opacity",
                 maximum=100,
@@ -35,8 +38,8 @@ class ImBlend(NodeBase):
                 step=0.1,
                 controls_step=1,
                 unit="%",
-            ).with_id(5),
-            ImageInput("Overlay").make_optional().with_id(1),
+            ),
+            ImageInput("Overlay").make_optional(),
             SliderInput(
                 "Overlay Opacity",
                 maximum=100,
@@ -44,7 +47,7 @@ class ImBlend(NodeBase):
                 step=0.1,
                 controls_step=1,
                 unit="%",
-            ).with_id(2),
+            ),
             DropDownInput(
                 "Blend Mode",
                 [
@@ -63,7 +66,7 @@ class ImBlend(NodeBase):
                     {"option": "Screen", "value": 12},
                     {"option": "Xor", "value": 13},
                 ],
-            ).with_id(6),
+            ),
         ]
         self.outputs = [ImageOutput()]
         self.category = IMAGE_UTILITY
@@ -73,28 +76,31 @@ class ImBlend(NodeBase):
 
     def run(
         self,
-        base: np.ndarray,
+        base: ndarray32,
         opbase: float,
-        ov: np.ndarray | None,
+        ov: ndarray32 | None,
         op: float,
         blend_mode: int,
-    ) -> np.ndarray:
-        """Overlay transparent images on base image"""
+    ) -> ndarray32:
+        """Blend images together"""
+
+        # Return base image * base opacity if there is no overlay image.
+        if ov is None:
+            if get_h_w_c(base)[2] == 4:
+                base[:, :, 3] *= opbase / 100
+            return base
 
         # Convert to 0.0-1.0 range
         opbase /= 100
         op /= 100
 
-        # Return base image * base opacity if there is no overlay image.
-        if ov is None:
-            if get_h_w_c(base)[2] == 4:
-                base[:, :, 3] *= opbase
-            return base
-
         imgs = []
+        max_h = max_w = 0
         for img in (base, ov):  # Using loops in case variable inputs possible later.
             if img is not None:
                 h, w, c = get_h_w_c(img)
+                max_h = max(h, max_h)
+                max_w = max(w, max_w)
 
                 # All inputs must be BGRA for alpha compositing to work
                 if c == 1:
@@ -108,6 +114,19 @@ class ImBlend(NodeBase):
 
         imgout = imgs[0]
         imgs = imgs[1:]
+
+        # Pad base image with transparency if necessary to match size with overlay
+        h, w, _ = get_h_w_c(imgout)
+        tp = bm = lt = rt = 0
+        if h < max_h:
+            tp = (max_h - h) // 2
+            bm = max_h - h - tp
+        if w < max_w:
+            lt = (max_w - w) // 2
+            rt = max_w - w - lt
+        imgout = cv2.copyMakeBorder(
+            imgout, tp, bm, lt, rt, cv2.BORDER_CONSTANT, value=0
+        )
 
         center_x = imgout.shape[1] // 2
         center_y = imgout.shape[0] // 2
@@ -128,7 +147,7 @@ class ImBlend(NodeBase):
                 blend_mode,
             )
 
-            imgout = np.copy(img)
+            imgout[y_offset : y_offset + h, x_offset : x_offset + w] = img
         img = np.clip(imgout, 0, 1)
 
         return img
@@ -157,12 +176,12 @@ class StackNode(NodeBase):
 
     def run(
         self,
-        im1: np.ndarray,
-        im2: Union[np.ndarray, None],
-        im3: Union[np.ndarray, None],
-        im4: Union[np.ndarray, None],
+        im1: ndarray32,
+        im2: ndarray32 | None,
+        im3: ndarray32 | None,
+        im4: ndarray32 | None,
         orientation: str,
-    ) -> np.ndarray:
+    ) -> ndarray32:
         """Concatenate multiple images horizontally"""
         img = im1
         imgs = []
@@ -252,7 +271,7 @@ class CaptionNode(NodeBase):
         self.icon = "MdVideoLabel"
         self.sub = "Compositing"
 
-    def run(self, img: np.ndarray, caption: str) -> np.ndarray:
+    def run(self, img: ndarray32, caption: str) -> ndarray32:
         """Add caption an image"""
 
         return add_caption(img, caption)
@@ -279,12 +298,12 @@ class ColorConvertNode(NodeBase):
         self.icon = "MdColorLens"
         self.sub = "Miscellaneous"
 
-    def run(self, img: np.ndarray, color_mode: int) -> np.ndarray:
+    def run(self, img: ndarray32, color_mode: int) -> ndarray32:
         """Takes an image and changes the color mode it"""
 
         result = cv2.cvtColor(img, int(color_mode))
 
-        return result
+        return result  # type: ignore
 
 
 @NodeFactory.register("chainner:image:create_border")
@@ -306,7 +325,7 @@ class BorderMakeNode(NodeBase):
         self.icon = "BsBorderOuter"
         self.sub = "Miscellaneous"
 
-    def run(self, img: np.ndarray, border_type: int, amount: int) -> np.ndarray:
+    def run(self, img: ndarray32, border_type: int, amount: int) -> ndarray32:
         """Takes an image and applies a border to it"""
 
         amount = int(amount)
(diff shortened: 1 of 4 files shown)


backend/src/nodes/image_util_nodes.py
```diff
@@ -5,80 +5,98 @@
...
-                "Opacity A",
-                maximum=100,
-                default=50,
-                step=0.1,
-                controls_step=1,
-                unit="%",
-            ).with_id(2),
-            ImageInput("Overlay B").make_optional().with_id(3),
+            ),
+            ImageInput("Overlay").make_optional(),
```
RunDevelopment, line 42: Why is this optional? If this image is not given, then the node is a noop (expect for the opacity slider).

backend/src/nodes/image_util_nodes.py
```diff
@@ -5,80 +5,98 @@
...
 
     def __init__(self):
         """Constructor"""
         super().__init__()
-        self.description = "Overlay transparent images on base image."
+        self.description = """Blends overlay image onto base image using 
+            specified mode and opacities."""
         self.inputs = [
-            ImageInput("Base").with_id(0),
+            ImageInput("Base"),
```
RunDevelopment, line 33: Why is the bottom layer (base) above the top layer (overlay)?

backend/src/nodes/image_util_nodes.py
```diff
@@ -5,80 +5,98 @@
...
+                    {"option": "Add", "value": 4},
+                    {"option": "Color Burn", "value": 5},
+                    {"option": "Color Dodge", "value": 6},
+                    {"option": "Reflect", "value": 7},
+                    {"option": "Glow", "value": 8},
+                    {"option": "Overlay", "value": 9},
+                    {"option": "Difference", "value": 10},
+                    {"option": "Negation", "value": 11},
+                    {"option": "Screen", "value": 12},
+                    {"option": "Xor", "value": 13},
```
RunDevelopment, line 67: Please make class constants (or similar) for these values. These integer values are essentially just magic numbers that just happen to line up with the numbers used by image blender. That's way too brittle.

backend/src/nodes/image_util_nodes.py
```diff
@@ -5,80 +5,98 @@
...
         self.category = IMAGE_UTILITY
-        self.name = "Overlay Images"
+        self.name = "Blend Images"
         self.icon = "BsLayersHalf"
         self.sub = "Compositing"
 
     def run(
         self,
-        base: np.ndarray,
+        base: ndarray32,
```
RunDevelopment, line 79: Was it really a good idea to merge what could have been 2 PRs (better type annotations + blend node)?

RunDevelopment (changes requested):

backend/src/nodes/image_util_nodes.py
theflyingzamboni: It's the same way that overlay is. This is something we all probably need to hash out in the discord, as I am ambivalent and can think of reasons for both orderings.

backend/src/nodes/image_util_nodes.py
theflyingzamboni: Because if it's not optional, the node won't run with base only. It is desirable for it to run with base only because a) it could simply be used to adjust image opacity and b) in case a user wants to pass an image through while testing but doesn't want to delete the node.

backend/src/nodes/image_util_nodes.py
theflyingzamboni: Ehh, maybe not. It only happened because I was trying to see if there was a way to get rid of a pyright type error I was getting, and then I just made it consistent across the files I was working on.

RunDevelopment: > get rid of a pyright type error

Well, that didn't exactly work out...

backend/src/nodes/image_util_nodes.py
RunDevelopment: > it could simply be used to adjust image opacity

That's not the primary function of this node, and if people really want to repurpose the node, they can just pass in the same image as top and bottom and set the top opacity to 0.

> in case a user wants to pass an image through while testing but doesn't want to delete the node.

Just set the opacity of the top layer to 0.

backend/src/nodes/image_util_nodes.py
theflyingzamboni: Annoyingly.

backend/src/nodes/image_util_nodes.py
theflyingzamboni: That just makes noop more convoluted to achieve, so I don't really see the purpose behind doing that. We have a lot of nodes with noop conditions, all of which I just see as a temporary solutions to our lack of proper node disabling/passthrough. If we can make that work as a feature, we can get rid of all the noop conditions in various nodes.

backend/src/nodes/image_util_nodes.py
```diff
@@ -12,110 +12,112 @@
...
-                    img = np.dstack((img, np.ones((h, w), np.float32)))
-                elif c != 4:  # Explode if there are not 1, 3, or 4 channels
-                    logger.error(f"Number of channels ({c}) unexpected")
+            # All inputs must be BGRA for alpha compositing to work
+            if c == 1:
+                img = cv2.cvtColor(img, cv2.COLOR_GRAY2BGRA)
+            elif c == 3:
+                img = cv2.cvtColor(img, cv2.COLOR_BGR2BGRA)
+            elif c != 4:  # Explode if there are not 1, 3, or 4 channels
+                logger.error(f"Number of channels ({c}) unexpected")
```
RunDevelopment, line 83: Could you throw an error here? Or make an assertion before the if/elifs?

backend/src/nodes/image_util_nodes.py
```diff
@@ -12,110 +12,112 @@
...
 
         imgs = []
-        max_h, max_w = 0, 0
-        for img in base, ov1, ov2:
-            if img is not None:
-                h, w, c = get_h_w_c(img)
-                max_h = max(h, max_h)
-                max_w = max(w, max_w)
+        max_h = max_w = 0
+        for img in (base, ov):
```
RunDevelopment, line 72: The loop seems like overkill for 2 images.

backend/src/nodes/image_util_nodes.py
```diff
@@ -12,110 +12,112 @@
...
-        # Apply opacity to base, then overlay A and B sequentially at corresponding opacities
+        # Apply opacities to images, then blend overlay
         imgout[:, :, 3] *= opbase
-        for img, op in zip(imgs, (op1, op2)):
-            h, w, _ = get_h_w_c(img)
-
-            # Center overlay
-            x_offset = center_x - (w // 2)
-            y_offset = center_y - (h // 2)
+        img[:, :, 3] *= op
```
RunDevelopment, line 112: Are those the original image objects? If so, then you're modifying cached images, which is a bug.

RunDevelopment (changes requested):

backend/src/nodes/image_util_nodes.py
theflyingzamboni: I mostly just wanted to avoid doubling these lines:
```h, w, c = get_h_w_c(img)
max_h = max(h, max_h)
max_w = max(w, max_w)

# All inputs must be BGRA for alpha compositing to work
assert c != 4, f"Number of channels ({c}) unexpected"
if c == 1:
    img = cv2.cvtColor(img, cv2.COLOR_GRAY2BGRA)
elif c == 3:
    img = cv2.cvtColor(img, cv2.COLOR_BGR2BGRA)```

File: backend/src/nodes/image_util_nodes.py (hunk 7 above)
Comment (theflyingzamboni) [reply backend/src/nodes/image_util_nodes.py:112]: They're the objects defined at lines 86-87:
```
imgout = imgs[0]
img = imgs[1]
```

File: backend/src/nodes/image_util_nodes.py (hunk 6 above)
Comment (theflyingzamboni) [reply backend/src/nodes/image_util_nodes.py:72]: I can change it if the extra smidge of performance overhead is worth the extra clutter, but it just seemed more readable to leave it as a loop.

Comment (RunDevelopment) [reply backend/src/nodes/image_util_nodes.py:72]: > the extra smidge of performance overhead

I don't care about the performance overhead of a single loop when the numpy stuff we're doing right after takes like 1000x longer. It's not worth worrying about.

What I meant is that the loop is overcomplicating things. The whole

```py
# All inputs must be BGRA for alpha compositing to work
assert c != 4, f"Number of channels ({c}) unexpected"
if c == 1:
    img = cv2.cvtColor(img, cv2.COLOR_GRAY2BGRA)
elif c == 3:
    img = cv2.cvtColor(img, cv2.COLOR_BGR2BGRA)
```

 should really be its own function and getting the max of the sizes of 2 images shouldn't be too hard.

backend/src/nodes/image_util_nodes.py
RunDevelopment: In pretty much any OOP language, objects are passed by reference (technically by value of their pointer, but that's close to by reference). This means you can do stuff like this:

```py
a = SomeClass()
a.foo = 9
b = a
a.foo = 10
print(a.foo) # 10
print(b.foo) # 10
```

Since you only passed around the reference of the input image, I'm worried that you are modifying the cached image directly.

@joeyballentine Asking a python expert, is that what's happening here or is there some weird python magic I am unaware of?

theflyingzamboni: So a partial answer to this is that lists are mutable in python, so yes, that does modify the input images in the case of RGBA input.

So if I did `imgout[0 ,0, 0] = 3`, then that would alter base. If I then did `imgout = np.clip(imgout, 0, 1)`, that would not alter base, and it would still have that 3. The return value is a new object, so imgout now refers to the returned object, not base.

I'd be interested to hear more about what the bug is, as to my understanding this is considered normal behavior in python (and I'm pretty sure occurs many places in chaiNNer aside from this node). I don't know what issue it would cause to modify the object passed in vs. creating a new object (which would require either functions that return a new value or a deep copy), and the nodes always return a new object.

theflyingzamboni: It's all academic anyway, factoring out the BGRA conversion means that it will be returning new objects anyway, so whether or not this is an issue, it will not be an issue.

theflyingzamboni: Hmm, actually, returning itself may not create a new object like I thought. I still am not sure what bug this would create though.

RunDevelopment: The bug is not that lists (numpy arrays in this case) are mutable. That's by design. The problem is that the executor caches images. So an image passed to `run` might also be passed to other nodes afterward. So if you modify the image directly, you also modify that image for all nodes that get executed afterward. 

The executor is essentially doing this:
```py
img = some_node.run(...)
res1 = blend_node.run(img, ...)
res2 = some_other_node.run(img, ...)
```

So modifying any of the input objects is a big no no.

theflyingzamboni: Okay, if you check the latest commit, base is not modified (ov is once potentially, but it deadends in this node and does not move on to anywhere). If base is not BGRA, cvtColor returns a new object. Even if base is BGRA, when copyMakeBorder is used, it creates a new object that is assigned to imgout, and that is the object that is modified, so imgout never refers to base when it is modified.

theflyingzamboni: Gotcha. In that case, I need to make sure ov is never modified either. Didn't realize the executor handled it like that.

theflyingzamboni: The BGRA conversion function will always return a deep copy now, so neither cached image will ever be modified.

backend/src/nodes/utils/image_utils.py
```diff
@@ -90,6 +92,16 @@ def get_available_image_formats():
...
 
 
+def convert_to_BGRA(img: np.ndarray, c: int) -> np.ndarray:
+    assert c in (1, 3, 4), f"Number of channels ({c}) unexpected"
+    if c == 1:
+        img = cv2.cvtColor(img, cv2.COLOR_GRAY2BGRA)
+    elif c == 3:
+        img = cv2.cvtColor(img, cv2.COLOR_BGR2BGRA)
+
+    return deepcopy(img)
```
RunDevelopment, line 102: This will work, but you can also just use [copy](https://numpy.org/doc/stable/reference/generated/numpy.copy.html).

theflyingzamboni: So you can. I just assumed that was a shallow copy.

backend/src/nodes/properties/inputs/generic_inputs.py
```diff
@@ -166,3 +167,26 @@ def VideoTypeDropdown() -> DropDownInput:
...
+def BlendModeDropdown() -> DropDownInput:
+    """Blending Mode option dropdown"""
+    return DropDownInput(
+        "Blend Mode",
+        [
+            {"option": "Normal", "value": bm.NORMAL},
+            {"option": "Multiply", "value": bm.MULTIPLY},
+            {"option": "Darken", "value": bm.DARKEN},
+            {"option": "Lighten", "value": bm.LIGHTEN},
+            {"option": "Add", "value": bm.ADD},
```
RunDevelopment, line 181: Add and multiply should be grouped together IMO.

```suggestion
            {"option": "Add", "value": bm.ADD},
            {"option": "Multiply", "value": bm.MULTIPLY},
            {"option": "Darken", "value": bm.DARKEN},
            {"option": "Lighten", "value": bm.LIGHTEN},
```

theflyingzamboni: Oh whoops. I actually meant to sort these after I got them added and just forgot.

theflyingzamboni: Sorted based on categories a la Photoshop and GIMP: Normal, Darken, Lighten, Contrast, Inversion, Cancelation. (I'm not 100% sure exactly what Reflect/Glow qualify as, but they seem to fit in Contrast by the way they behave.)

RunDevelopment (approved):

PR #418, opened 2022-06-11
Fix node removal via context menu

joeyballentine:


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -451,6 +451,7 @@ export const GlobalProvider = memo(
 
         const removeNodeById = useCallback(
             (id: string) => {
+                changeEdges((edges) => edges.filter((e) => e.source !== id && e.target !== id));
```
RunDevelopment, line 454: `changeEdges` has to added to the deps array.

RunDevelopment (approved):

PR #419, opened 2022-06-11
Fix torch convert to onnx node not running when no connected outputs

joeyballentine:
Temporary fix. We will need to have a way of defining nodes with effects later, but for now this is fine


RunDevelopment:
Why don't we just give each node a field `side_effects: bool` that tells the executor this information? That would be a lot easier than this temporary workaround.

PR #428, opened 2022-06-14
Rotation node tweak

theflyingzamboni:
Changed rotation angle to 0-360 degree slider, and added options for interpolation method, dimension expansion, and fill.

Closes joeyballentine/chaiNNer#415


backend/src/nodes/image_util_nodes.py
```diff
@@ -353,12 +352,29 @@ def __init__(self):
         self.description = "Rotate an image."
         self.inputs = [
             ImageInput("Image"),
+            SliderInput(
+                "Rotation Angle",
+                default=0,
+                maximum=360,
+                step=0.1,
+                controls_step=1,
```
RunDevelopment, line 360: Could we make the control step larger? It would be nice to have something like 45° increments when clicking "up" in the number input box.

backend/src/nodes/properties/inputs/image_dropdown_inputs.py
```diff
@@ -93,7 +93,31 @@ def InterpolationInput() -> DropDownInput:
...
-            {"option": "Lanczos", "value": InterpolationMethod.LANCZOS},
+            {
+                "option": "Lanczos",
+                "value": InterpolationMethod.LANCZOS,
+            },
+        ],
+    )
+
+
+def ReducedInterpolationInput() -> DropDownInput:
```
RunDevelopment, line 104: Reduce?

backend/src/nodes/utils/pil_utils.py
```diff
@@ -47,6 +49,26 @@ def resize(
...
+    if fill == 0:
+        fill_color = (0,) * c
+    elif fill == 1:
+        fill_color = (0,) * c if c < 4 else (0, 0, 0, 255)
+    else:
+        img = convert_to_BGRA(img, c)
+        fill_color = (0, 0, 0, 0)
+
+    pimg = Image.fromarray((img * 255).astype("uint8"))
+    pimg = pimg.rotate(angle, interpolation, expand, fillcolor=fill_color)  # type: ignore
```
RunDevelopment, line 68: `interpolation` is incorrect-ish. You pass in a value from `InterpolationMethod` which are not guaranteed to be PIL interpolation methods. The only reason this works is that the values defined in `InterpolationMethod` happen to line up with PIL interpolation methods.

Please use the `INTERPOLATION_METHODS_MAP` to get a PIL interpolation method.

Since we no longer use OpenCV for resizing, we could also remove the InterpolationMethod class, but that should be its own PR.

backend/src/nodes/utils/pil_utils.py
```diff
@@ -47,6 +49,26 @@ def resize(
     return np.array(pimg).astype("float32") / 255
 
 
+def rotate(
+    img: np.ndarray, angle: float, interpolation: int, expand: int, fill: int
```
RunDevelopment, line 53: Please don't just use magic numbers for `expand` and `fill`. Add classes similar to `InterpolationMethod` that contain the integer constants.

RunDevelopment (changes requested):

backend/src/nodes/properties/inputs/image_dropdown_inputs.py
theflyingzamboni: Reduced, not Reduce, because PIL rotate accepts a reduced set of the interpolation methods. I renamed it to RotateInterpolationInput.

joeyballentine (approved):

PR #429, opened 2022-06-14
Startup template

joeyballentine:


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -449,6 +451,31 @@ export const GlobalProvider = memo(
         useIpcRendererListener('file-save-as', () => performSave(true), [performSave]);
         useIpcRendererListener('file-save', () => performSave(false), [performSave]);
 
+        const [firstLoad, setFirstLoad] = useSessionStorage('firstLoad', true);
+        const [startupTemplate] = useStartupTemplate;
+        useAsyncEffect(async () => {
```
RunDevelopment, line 456: This is not guaranteed to work. `useEffect` and `useLayoutEffect` (and all hooks are implemented using these hooks) behave inconsistently when the data they depend on doesn't cause a DOM update.

Basically, `useEffect` runs [after a DOM update](https://reactjs.org/docs/hooks-effect.html), but what if the current render doesn't cause a DOM update? If all of the elements created by this render are memoized, then React might decide that it doesn't need to update the DOM. In this case, the `useEffect` hook won't run.

Note that in the case of contexts, the DOM update can also be caused by any of the components using the context. This can make it very difficult to predict program behavior, since removing a `useContext` somewhere in the application might cause a hook to stop working. Our use of `useContextSelector` makes this even worse.

I know all of this because I had this exact bug when adding the edit history. I spend roughly 4 hours debugging only to find that React just doesn't run effect hooks sometimes.

joeyballentine: It seemed to work fine when I tested it. What do you recommend I do?

RunDevelopment: Yeah, it works because other stuff changes and causes DOM updates. This hook on its own doesn't work, but the side effects of the other stuff around it just so happen to make it work.

As for what to do: I don't know. This seems like a fundamental problem with `useEffect` and I haven't found a way around it yet.

I would say: let's just go with it for now.

RunDevelopment (approved):

PR #431, opened 2022-06-15
Another fix for blend node migration

theflyingzamboni:
The issues were:
a) Migration was not adding a parentNode id for nodes in an iterator
b) Migration was not connecting the first blend to the new blend when overlay had 3 inputs and was not output to another node

Now:
a) parentNode ID added to new node if original node has one
b) If overlay had 3 image inputs and an output, output source changes to new node and a new edge is created to connect the two blend nodes

Please test

### DIFF
--- a/src/common/migrations.js
+++ b/src/common/migrations.js
@@ -370,24 +370,35 @@ const addBlendNode = (data) => {
                     width: node.width,
                     zIndex: node.zIndex,
                 };
+                if (node.parentNode !== undefined) {
+                    newBlendNode.parentNode = node.parentNode;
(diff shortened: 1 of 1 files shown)


theflyingzamboni:
Just realized I didn't bother to account for edges connected to the opacity sliders, because I've never actually connected anything to them, and probably no one else has either. But someone could, technically, so I need to update that before this PR is merged.

theflyingzamboni:
Nevermind, forgot we still couldn't connect anything to sliders despite the handles.

joeyballentine (approved):

RunDevelopment (approved):

PR #432, opened 2022-06-16
Pane context menu mode list (also connection dragging to pane)

joeyballentine:
This works. But the code is bad and react complains. Hoping a better method for this can be found


src/renderer/components/ReactFlowBox.tsx
```diff
@@ -310,6 +327,220 @@ const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlowBoxPro
...
+                                            y: y - reactFlowBounds.top,
+                                        });
+                                        const nodeToMake: NodeProto = {
+                                            position,
+                                            data: {
+                                                schemaId: node.schemaId,
+                                            },
+                                            nodeType: node.nodeType,
+                                        };
+                                        createNode(nodeToMake, (newNodes) => {
```
RunDevelopment, line 461: The `onNodeCreate` function might be a problem because it is called from within the supplier function of a `setState` function (`setNodes` to be specific). It feels like doing that is an anti-pattern if not incorrect. However, the React docs don't seem to mention this.

joeyballentine:
I want to do the arrow key/tab stuff in a different PR

src/renderer/components/CustomIcons.tsx
```diff
@@ -143,148 +143,158 @@ export const IconFactoryOld = memo(({ category }: { category: string }) => {
...
+        return (
+            <Icon
+                alignContent="center"
+                alignItems="center"
+                as={libraryIcon}
+                color={accentColor}
+                height={boxSize}
+                transition="0.15s ease-in-out"
+                viewBox={`0 0 ${boxSize} ${boxSize}`}
+                width={boxSize}
```
RunDevelopment, line 294: Why do you use height, width, and viewBox here, while all the icons in the switch statement use boxSize?

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -105,19 +106,19 @@ export const GlobalVolatileContext = createContext<Readonly<GlobalVolatile>>({}
 export const GlobalContext = createContext<Readonly<Global>>({} as Global);
 
 const createNodeImpl = (
-    { position, data, nodeType }: NodeProto,
+    { id, position, data, nodeType }: NodeProto,
```
RunDevelopment, line 109: I can already see use using the wrong ID variable, so let's use a default value and remove `newId`.

```suggestion
    { id = createUniqueId(), position, data, nodeType }: NodeProto,
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -498,12 +499,16 @@ export const GlobalProvider = memo(
         );
 
         const createNode = useCallback(
-            (proto: NodeProto): void => {
+            (
+                proto: NodeProto,
+                onNodeCreate: (newNodes: Node<NodeData>[]) => void = () => {}
```
RunDevelopment, line 504: You don't use the callback anymore. Could you please remove it?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -343,6 +579,9 @@ const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlowBoxPro
                     borderRadius: '0.5rem',
                 }}
                 onConnect={createConnection}
+                // onConnectEnd={onConnectStop}
```
RunDevelopment, line 582: Please remove the commented out code.

RunDevelopment (changes requested):
I left you a few comments.

Also, ReactFlowBox is already complex enough as is. Please move your new menu either into its own component or into its own hook.

src/renderer/components/CustomIcons.tsx
joeyballentine: I have absolutely no idea why I did that originally. I assume it was when I was trying to get the icons to actually scale properly into 4x4 rather than being able to have transparent padding around them (which didn't work) and I just left the code. I can try to standardize it

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -318,6 +335,225 @@ const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlowBoxPro
...
+                    });
+                }
+                log.error(`Unknown handle type: ${connectingFrom.handleType!}`);
+                return true;
+            }),
+        [connectingFrom, connectingFromType, searchQuery, schemata.schemata]
+    );
+    const byCategories = useMemo(() => getNodesByCategory(matchingNodes), [matchingNodes]);
+    const menuRef = useRef<HTMLDivElement>(null);
+    const menu = useContextMenu(() => (
```
RunDevelopment, line 385: As I said, ReactFlowBox is already complex enough as is. Please move your new menu either into its own component or into its own hook.

RunDevelopment (changes requested):

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -50,7 +50,7 @@ type SetState<T> = React.Dispatch<React.SetStateAction<T>>;
 interface GlobalVolatile {
     nodeChanges: ChangeCounter;
     edgeChanges: ChangeCounter;
-    createNode: (proto: NodeProto) => void;
+    createNode: (proto: NodeProto, onNodeCreate?: (newNodes: Node<NodeData>[]) => void) => void;
```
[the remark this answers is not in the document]
joeyballentine: Whoops, forgot to remove the type. Still not entirely used to typescript lol

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -343,6 +466,8 @@ const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlowBoxPro
                     borderRadius: '0.5rem',
                 }}
                 onConnect={createConnection}
+                onConnectStart={onConnectStart}
+                onConnectStop={onConnectStop}
```
RunDevelopment, line 470: Sorry for not being clear enough. I meant that the whole code you added to this component should be its own hook. From what I see, the new pane is separate from the react flow box. The RFB only really needs the callbacks used here and another used below. All other code could be moved into the new hook.

joeyballentine: Oooh, ok

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+export const usePaneNodeSearchMenu = (
+    wrapperRef: React.RefObject<HTMLDivElement>
+): UsePaneNodeSearchMenuValue => {
+    const { createNode, createConnection } = useContext(GlobalVolatileContext);
+    const { closeContextMenu } = useContext(ContextMenuContext);
+    const { schemata } = useContext(GlobalContext);
+
+    const [connectingFrom, setConnectingFrom] = useState<OnConnectStartParams | null>();
+    const [connectingFromType, setConnectingFromType] = useState<string | null>();
```
RunDevelopment, line 40: If you use `useState<T>()` with a default, the type of the variable will be `T | undefined`. This is weird here because we already use `null` to express "no value."

Either make `null` the default value or remove `null` and use `undefined`.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+    const [isStoppedOnPane, setIsStoppedOnPane] = useState<boolean>(false);
```
RunDevelopment, line 41: The type can be inferred :)

```suggestion
    const [isStoppedOnPane, setIsStoppedOnPane] = useState(false);
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+    const { getNode, project } = useReactFlow();
+
+    const [searchQuery, setSearchQuery] = useState<string>('');
```
RunDevelopment, line 44: Same here.
```suggestion
    const [searchQuery, setSearchQuery] = useState('');
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+                data: {
+                    schemaId: node.schemaId,
+                },
+                nodeType: node.nodeType,
+            };
+            createNode(nodeToMake);
+            if (isStoppedOnPane && connectingFrom) {
+                if (connectingFrom.handleType === 'source') {
+                    const firstValidHandle = schemata
+                        .get(node.schemaId)!
```
RunDevelopment, line 94: The return value can never be null-ish. No idea why TSLint didn't catch this.
```suggestion
                        .get(node.schemaId)
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+                        )!.id;
+                    createConnection({
+                        source: connectingFrom.nodeId,
+                        sourceHandle: connectingFrom.handleId,
+                        target: nodeId,
+                        targetHandle: `${nodeId}-${firstValidHandle}`,
+                    });
+                } else if (connectingFrom.handleType === 'target') {
+                    const firstValidHandle = schemata
+                        .get(node.schemaId)!
```
RunDevelopment, line 106: Same here.
```suggestion
                        .get(node.schemaId)
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+                nodeType: node.nodeType,
+            };
+            createNode(nodeToMake);
+            if (isStoppedOnPane && connectingFrom) {
+                if (connectingFrom.handleType === 'source') {
+                    const firstValidHandle = schemata
+                        .get(node.schemaId)!
+                        .inputs.find(
+                            (input) => input.type === connectingFromType && input.hasHandle
+                        )!.id;
```
RunDevelopment, line 97: This is actually true? Are we always guaranteed to find a compatible input or is this just a reference error waiting to happen?

(Same for outputs)

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+    );
+    const byCategories = useMemo(() => getNodesByCategory(matchingNodes), [matchingNodes]);
+    const menuRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        setSearchQuery('');
+    }, [connectingFrom]);
+
+    const onPaneContextMenuNodeClick = useCallback(
+        (node: NodeSchema, position: { x: number; y: number }) => {
```
RunDevelopment, line 74: You are using this type below as well. Could you please give it a name?

Also, since it's used as React state, could you please make the fields readonly?

RunDevelopment:
I also tried out the PR and the right-click while making an edge is bugged.

1. When I'm making a new edge and right click, the selector pane opens, and when I release the left mouse button, it closes again. This happens on a cycle where the first time I try it, releasing the left mouse button doesn't close the pane, and the second try it does close the pane, on the third try it doesn't close it again, on the fourth it does close it again, and so on.
2. When I'm making a new edge, right click, and select a node, the selected node is created but no connection is made.
3. The position of newly created nodes is not where I clicked, but at the top left corner of the pane. This difference is noticeable when the pane is positioned above or to the right of where I clicked.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,299 @@
...
+                                        accentColor={accentColor}
+                                        boxSize={3}
+                                        icon={category}
+                                    />
+                                    <Text fontSize="xs">{category}</Text>
+                                </HStack>
+                                {[...categoryNodes].map((node) => (
+                                    <HStack
+                                        _hover={{
+                                            backgroundColor: useColorModeValue(
```
RunDevelopment, line 198: You are using a hook in a loop.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
joeyballentine: Since we filter by only nodes that have possible connections for this case, yes it should always find a compatible input/output. However, I could change this just to be safe

joeyballentine:
@RunDevelopment The way you do the auto-edge thing is to hold control or alt while dragging to the pane. Right clicking from the pane while dragging would be interesting to add though, I never thought of that.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
RunDevelopment: What do you mean? 

```ts
interface Position {
    readonly x: number;
    readonly y: number;
}
``` 
works fine.

joeyballentine: Ah, I was trying to do `x: readonly number;`

RunDevelopment:
Did you fix point 3 yet?

> The position of newly created nodes is not where I clicked, but at the top left corner of the pane. This difference is noticeable when the pane is positioned above or to the right of where I clicked.

RunDevelopment:
> the node will be placed on top of the source node

I did some more testing, and it will be placed to (0,0).

RunDevelopment:
2 more things:

1. Could you please make it so that right-clicking inside the pane doesn't close the pane? This should be relatively easy by adding `onContextMenu={(e) => e.stopPropagation()}` to your `MenuList`.
2. Maybe we could add a small favorite indicator just like in the node selector? I feel like that would be useful to find your usual nodes.

joeyballentine:
> Also, could you not use the Alt key? On Windows, the Alt key selects/deselects the main menu of the current window. So when you use the Alt key to open the selector pane, it will select the menu instead of the search box.

Sure

> Maybe we could just always open the pane if the user didn't connect to anything? I mean, it's easy to close the pane. I also fear the people might miss this incredible feature.

This is how I originally had it, and Kim complained about it being annoying when you miss a connection by accident, etc. But I honestly prefer it that way

RunDevelopment:
image

When there are no nodes, could you pelase display a little bit of text like "No compatible nodes found."

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,325 @@
...
+                                                icon={node.icon}
+                                            />
+                                            <Text>{node.name}</Text>
+                                        </HStack>
+                                    ))}
+                                </Box>
+                            );
+                        })
+                    ) : (
+                        <Center w="full">No compatible nodes found.</Center>
```
RunDevelopment, line 236: Could you make the text less bright? I feel like 

image

is a bit too bright for what is just a small note.

RunDevelopment (approved):

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+    useEffect(() => {
+        if (connectingFrom) {
+            const { nodeId, inOutId } = parseHandle(connectingFrom.handleId!);
+            const node: Node<NodeData> | undefined = getNode(nodeId);
+            if (node) {
+                if (connectingFrom.handleType === 'source') {
+                    const sourceType = functionDefinitions
+                        .get(node.data.schemaId)
+                        ?.outputDefaults.get(inOutId);
+                    setConnectingFromType(sourceType!);
```
RunDevelopment, line 350: `setConnectingFromType` accepts `null`, so why not use that?

```suggestion
                    setConnectingFromType(sourceType ?? null);
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+                } else if (connectingFrom.handleType === 'target') {
+                    const targetType = functionDefinitions
+                        .get(node.data.schemaId)
+                        ?.inputs.get(inOutId);
+                    setConnectingFromType(targetType!);
```
RunDevelopment, line 355: Same here.

```suggestion
                    setConnectingFromType(targetType ?? null);
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+            searchQuery,
+            schemata.schemata,
+            matchingNodes,
+            mousePosition,
+        ]
+    );
+
+    useEffect(() => {
+        if (connectingFrom) {
+            const { nodeId, inOutId } = parseHandle(connectingFrom.handleId!);
```
RunDevelopment, line 343: We can make the non-nullish assertion unnecessary like this.

```suggestion
        if (connectingFrom && connectingFrom.handleId) {
            const { nodeId, inOutId } = parseHandle(connectingFrom.handleId);
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+        [
+            connectingFrom,
+            connectingFromType,
+            byCategories,
+            onPaneContextMenuNodeClick,
+            searchQuery,
+            schemata.schemata,
+            matchingNodes,
+            mousePosition,
+        ]
```
RunDevelopment, line 338: `useContextMenu` is *really* efficient and it's easy to get these deps arrays wrong and cause hard-to-find bugs. I think it's best to just remove it.

```suggestion
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+    const byCategories = useMemo(() => getNodesByCategory(matchingNodes), [matchingNodes]);
+
+    const { favorites } = useNodeFavorites();
+
+    useEffect(() => {
+        setSearchQuery('');
+    }, [connectingFrom]);
+
+    const onPaneContextMenuNodeClick = useCallback(
+        (node: NodeSchema, position: Position) => {
```
RunDevelopment, line 135: It's not really a node. Could you rename the parameter or destructure it into `{ schemaId, nodeType }`?

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+                id: nodeId,
+                position: projPosition,
+                data: {
+                    schemaId: node.schemaId,
+                },
+                nodeType: node.nodeType,
+            };
+            createNode(nodeToMake);
+            if (isStoppedOnPane && connectingFrom) {
+                const targetTypes = functionDefinitions.get(node.schemaId);
```
RunDevelopment, line 153: I would like to get rid of the 4 non-nullish assertions in this code block.
```suggestion
            const targetTypes = functionDefinitions.get(node.schemaId);
            if (isStoppedOnPane && connectingFrom && targetTypes && connectingFromType) {
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -0,0 +1,417 @@
...
+                if (connectingFrom.handleType === 'source') {
+                    const firstPossibleTarget = [...targetTypes!.inputs].find(([number, type]) => {
```
RunDevelopment, line 155: How about calling the `number` parameter `inputId`?

RunDevelopment (changes requested):

RunDevelopment:
I just want to get rid of as many non-nullish assertions as possible to make the code more robust.

RunDevelopment (approved):
I don't like the `eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` comments, but it's not a dealbreaker.

joeyballentine:
> I don't like the `eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` comments, but it's not a dealbreaker.

IMO getting rid of the last if check in favor of just doing "else" isn't great. Sure now it's just source and target, but what if RF updates and adds another kind? Maybe we'd accidentally cause it to use the target logic for that new kind. I just prefer being explicit.

RunDevelopment:
This is why I suggested to use `switch` on Discord. Together with `assertNever`, we can let TS verify that all possible values have been handled. If new variants get added, TS no longer compiles and forces us to update the code. I use this pattern in other projects as well (e.g. [here](https://github.com/RunDevelopment/refa/blob/deaafbb9d8f8e5d3e2128747b2f24f17e8a8455e/src/nfa.ts#L961)).

PR #433, opened 2022-06-16
Opacity node

theflyingzamboni:
Separated opacity from the Blend Node and made it a standalone adjustment node.

Migration is intended to only create an Opacity Node from a Blend Node when Opacity != 100 to reduce clutter. Seems to be covering various possible conditions (e.g. opacity value, whether an edge is connected to an image input, iterator parent), but please try to break to make sure.

Closes joeyballentine/chaiNNer#425


backend/src/nodes/image_util_nodes.py
```diff
@@ -24,26 +24,10 @@ def __init__(self):
         """Constructor"""
         super().__init__()
         self.description = """Blends overlay image onto base image using 
-            specified mode and opacities."""
+            specified mode."""
         self.inputs = [
             ImageInput("Base Layer"),
-            SliderInput(
```
RunDevelopment, line 30: Is it really a good idea to remove them? I understand you can achieve the same effect by using the new opacity node, but I still think it's usefull to have them here.

Also, the opacity sliders make it very clear that an RGBA (well, BGRA actually) is returned. Without the sliders, users might expect that e.g. the diff of 2 1-channel images to be a 1-channel image.

theflyingzamboni: I've actually toyed with the idea that it should output an image with c = max(base, overlay). I wasn't sure whether it was actually worth it, but I might do that, at least if the node doesn't have its own sliders. As for whether blend should have its own opacity sliders, I think that might warrant some discussion on discord.

backend/src/nodes/image_adj_nodes.py
```diff
@@ -253,3 +253,44 @@ def run(
...
+        self.category = IMAGE_ADJUSTMENT
+        self.name = "Opacity"
+        self.icon = "MdOutlineOpacity"
+        self.sub = "Adjustments"
+
+    def run(self, img: np.ndarray, opacity: float) -> np.ndarray:
+        """Apply opacity adjustment to alpha channel"""
+
+        if opacity == 100:
+            return img
```
RunDevelopment, line 287: I think this is bad.

The return value of a node should be consistent. This `if` also makes it impossible to statically type the node correctly. I'd have to type the number of output channels as `4 | InputImage.channels` which is just not as useful of just `4`.

RunDevelopment: Please make the condition `opacity == 100 and c == 4`.

theflyingzamboni: Hmm. Was trying to avoid unnecessary conversions, but I also haven't looked into your typing system.

RunDevelopment (approved):

Issue #411, opened 2022-06-10
Usability: Collapsed node selector tooltips


RunDevelopment:
1. The tooltips for node selector are a bit too minimal when collapsed.
    ![image](https://user-images.githubusercontent.com/20878432/173068233-4bf5850a-627d-4f2c-9ceb-613b7cc798ef.png)
    When collapsed, node tooltips should include the name of the node.
2. When collapsed, categories should have tooltips that display their name.

PR #436, opened 2022-06-17
Tooltips for collapsed selector panel

theflyingzamboni:
When panel is collapsed, node categories and subcategories now have styled tooltips, and node tooltips display node name.

image
image
image

Closes joeyballentine/chaiNNer#411


src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
```diff
@@ -25,45 +27,67 @@ const RegularAccordionItem = memo(
...
-                    </HStack>
-                    <AccordionIcon />
-                </AccordionButton>
+                <Tooltip
+                    closeOnMouseDown
+                    hasArrow
+                    borderRadius={8}
+                    fontSize="1.05rem"
+                    isDisabled={!collapsed}
+                    label={<ReactMarkdown>{`**${category}**`}</ReactMarkdown>}
```
RunDevelopment, line 36: Could you please not use a whole Markdown renderer just to do `<strong>{category}</strong>`?

Same for the other non-Markdown usages as well.

RunDevelopment (changes requested):

src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
RunDevelopment: The point is that you don't need md to make text bold.

src/renderer/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
```diff
@@ -64,7 +63,13 @@ const RepresentativeNodeWrapper = memo(
                     closeOnMouseDown
                     hasArrow
                     borderRadius={8}
-                    label={<ReactMarkdown>{node.description}</ReactMarkdown>}
+                    label={
+                        <Box>
+                            <b>{collapsed ? node.name : ''}</b>
+                            {collapsed && <br />}
```
RunDevelopment, line 69: Can be simplified.

(Suggestion needs formatting.)

```suggestion
                            {collapsed && <><b>{node.name}</b><br /></>}
```

joeyballentine:
Why did you remove ReactMarkdown from the node description label? The whole point of having that was so that we can add markdown to the descriptions in the backend if needed. Please put that back

src/renderer/components/NodeSelectorPanel/RepresentativeNodeWrapper.tsx
theflyingzamboni: Ah, that's how you get it all into one thing. I tried something like that but it complained about multiple elements, didn't know I could wrap them in a single empty tag.

theflyingzamboni:
Augh, someone make up my mind! XD

RunDevelopment (approved):

src/renderer/components/NodeSelectorPanel/RegularAccordionItem.tsx
```diff
@@ -25,45 +26,67 @@ const RegularAccordionItem = memo(
...
+                </Tooltip>
                 <AccordionPanel
                     pb={2.5}
                     pt={0}
                 >
                     {[...subcategoryMap].map(([subcategory, nodes]) => (
                         <Box key={subcategory}>
-                            <Center
-                            // w="full"
+                            <Tooltip
```
RunDevelopment, line 72: Could you remove this tooltip please?

Sub categories aren't interactable, and it feels silly to have a tooltip for a line.

theflyingzamboni: They are not interactable, but they do provide relevant information to the user that is not visible when the panel is collapsed. I would prefer to have tooltips for them.

RunDevelopment: > relevant information

What relevant information? The name of the arbitrary sub category we decided?

theflyingzamboni: They aren't arbitrary though. Like the main categories, the groupings are based on related functionality. The collapsed panel should be able to provide the same information as the expanded panel.

RunDevelopment: > the same information

I would change that to "relevant information". I simply do not see the value of having tooltips for sub categories. In what scenario is having tooltips for sub categories useful? 

I made #411 because I couldn't tell which node is which when collapsed. I also included categories because they are interactable, and should know what you interact with.

theflyingzamboni: It has the same sort of informational usefulness as tooltips on categories when you're looking for a particular type of node and don't recall its exact position. Something I could find useful if I didn't already know where things were. If it isn't relevant information when the panel is collapsed, then it is not relevant when the panel is expanded either, which would suggest removing them altogether (something I definitely think would make for a worse UX). I get that you do not personally find it useful for yourself, but that doesn't mean it's not useful to anyone.

RunDevelopment: > It has the same sort of informational usefulness as tooltips on categories when you're looking for a particular type of node

Categories provide very little information because they are arbitrary. They can still be useful, but unless categories are derived from the (objective) properties of their members, they are arbitrary.

Examples: What's the difference between image adjustments, filters, and utilities? A: Who knows! We don't tell anyone, and neither do any other image editors that arbitrarily decided on similar categories.

I also want to point out that you might be overestimating the intuitiveness/clarity/usefulness of categories and sub categories since you created them. I also find categorizations I created useful, logical, and intuitive. However, for Chainner, I literally hard memorized the approximate position of nodes I regularly use and almost never look at categories.

> when you're looking for a particular type of node and don't recall its exact position

Frankly, I think that searching for nodes in *collapsed mode* is not a realistic use case. Who in their right mind would go through potentially dozens of nodes by slowly hovering over them to the tooltip, instead of just expanding the side panel for a second?

In my use case, I didn't search for nodes. I know roughly where all nodes are. My problem is that I can't tell the exact position when the nodes surrounding the node I'm looking for all have the same icon (see #411).

> If it isn't relevant information when the panel is collapsed, then it is not relevant when the panel is expanded either, which would suggest removing them altogether

Frankly, with sub categories such as:
- "Adjustments" (under "Image (Adjustments)"), 
- "Modifications" (the nodes in adjacent sub categories also modify images), 
- "Miscellaneous" (you know... the other stuff),
- "Modifications" (the nodes in adjacent sub categories also modify images), 
- "Processing", and
- "Utility"

we might as well remove them. What relevant information do these sub categories provide?

Note: The visual separation sub categories provide is useful. I only think not displaying the name wouldn't hurt anyone.

> but that doesn't mean it's not useful to anyone.

Just because some feature could potentially be useful in a narrow scenario to a small number of people, doesn't mean it's worth keeping.

E.g. Yesterday, we decided to remove the opacity slider of the blend node. This feature was useful to some people, but not wroth keeping.

theflyingzamboni: There's a degree of arbitrarity in that it is based on personal judgment, but there's still a rationale to the categories. It doesn't make them less useful either. Do I know why the GIMP creators placed the "Emboss" filter in the Distort category instead of, say, the Decor category? No, but having it categorized as a) a Filter and b) a member of the Distort subcategory still helps me find it. 

This is why I disagree that not displaying the name at all wouldn't be a problem. What's easier when helping a new user find a feature? "It's in Image (Filters) under Correction," or "It's in Image (Filters) under the second line"? Having lines for no rhyme or reason (to the end user) is not great.

I'll give you that people are generally not going to be hunting nodes in collapsed mode, but I don't see this as a reason not to include something that (to me at least) makes sense (re: informational parity), and is completely unnoticeable to someone who doesn't decide to use it unless they hover over the lines.

theflyingzamboni: At this point, though, I'm just going to say make a call @joeyballentine. Think this discussion has played out and just needs a tiebreak. It's too small an issue to spend more time talking about it.

RunDevelopment: > What's easier when helping a new user find a feature? "It's in Image (Filters) under Correction," or "It's in Image (Filters) under the second line"?

I don't understand what you mean by that. Do you mean users searching by themselves? In which case, sub categories aren't very helpful due to their arbitrary nature and the fact that we don't have many nodes per category. Do you mean other people helping a user? In which case, saying "the node is called X, search for in the search bar" is way easier. Or do you mean users trying to remember the name of node? In which case, the search bar is superior again.

> makes sense (re: informational parity)

I think that's the crux of the disagreement. I don't think informational parity makes sense for collapsed mode. Collapsed mode is a compacted version of the normal node selector, so you would expect that some information/functionality gets left behind. E.g. we don't show the favorite status of a node in collapsed mode and that's okay.

joeyballentine: I honestly don't think the information is important enough to have. all a user is really gonna care about when looking at these tooltips is the node name, and maybe the regular category. But for the subcategory, I don't really see any gain from it.

RunDevelopment (approved):
Please resolve the merge conflicts.

Issue #373, opened 2022-05-28
Node disable/passthrough


theflyingzamboni:
Allow individual nodes to be disabled and permit upstream inputs to be passed through if possible, else disable downstream nodes as well. One way this could potentially work at a high level is to check if the "primary" input type (e.g. an image being modified) matches the output type, then the input is just passed to the next node. If the "primary" input type does not match the output type, or there is some other factor that would prevent a sensible input/output connection, then all downstream nodes are disabled as well.

theflyingzamboni:
I have some more thoughts on how this could be implemented. I am basing this on how I tend to think about most of the nodes (*with some notable exceptions), and how I think other users likely think about chaiNNer as well.

*(Split/Merge Channels are probably the most notable exceptions)

The way I tend to think about nodes* is that you have an initial image that is loaded into the chain. This image then "flows" through the nodes from beginning to end, where it is outputted with whatever modifications that were made to it along the way. We sadly don't have those polls going, but this is a very layperson's understanding of how chains work that I suspect "average" end users use.

Given this, when it comes to disabling/passing through nodes, I believe users will expect it to work such that what they think of as their "main" image (or other input type) will go through a disabled node unmodified by whatever other inputs/calculations the node uses.

My suggestion is that we define this ourselves explicitly, based on what makes the most "common sense" of how the average layperson will see it (and regardless of whether someone thinks that is how it should work, they will know that's how it _does_ work).

This input would be defined as the "passthrough input", and would need to follow the following conditions:

1. There can only be one output.
2. The passthrough input type and output type must match.
3a. There is only one input of the same type as output.
3b. There are multiple inputs of the same type as output, but one modifies the other and is not an expected output (e.g. Color Transfer has a primary image and a reference image, where the reference image exists only to modify the primary image). There is obviously some disagreement over which nodes qualify under this condition, but we can make the behavior clear to users to resolve potential confusion.

Any nodes that do not meet these conditions would be disabled and disable all downstream nodes that relied on its output. Those that do would send the passthrough input on to the following node(s), if it/they exist(s).

If a node has an input that matches these conditions, we use markdown or some other visual indicator to express to the user what value they should expect to pass through. If a node does not meet these conditions, none of the inputs will be indicated as passthrough. A node can only have one passthrough input at most.

RunDevelopment:
[...]

PR #444, opened 2022-06-18
Added support for disabling nodes

RunDevelopment:
This implements the ability to disable nodes from #373.

A disabled node is effectively not present. When executing the chain, all disabled nodes are cut out before sending the chain to the backend, so this is a pure frontend feature.

If a node is disabled, then all of this child nodes are also disabled. If an edge connects an input of a node to a disabled node, then the node will also be disabled.

### In progress

We need to decide how we want the UI for this feature to look like. Right now, I added a context menu entry to disable/enable nodes alongside a little status icon that pops up when a node is disabled.

image
image


joeyballentine:
Is this ready to be reviewed/merged?

RunDevelopment:
I think it's ready.

joeyballentine (approved):

PR #445, opened 2022-06-19
Possibly optimize auto-split a bit

joeyballentine:
I did this when messing around with the memory estimation stuff

### DIFF
--- a/backend/src/nodes/utils/pytorch_auto_split.py
+++ b/backend/src/nodes/utils/pytorch_auto_split.py
@@ -53,7 +53,6 @@ def auto_split_process(
         d_img = None
         try:
             device = torch.device(os.environ["device"])
-            model = model.to(device)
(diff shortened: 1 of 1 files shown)


backend/src/nodes/utils/pytorch_auto_split.py
```diff
@@ -53,7 +53,6 @@ def auto_split_process(
         d_img = None
         try:
             device = torch.device(os.environ["device"])
-            model = model.to(device)
```
RunDevelopment, line 56: Why did you remove this?

joeyballentine: Because it's already on the GPU when it loads the model in. this is unnecessary. At best it's a no-op, at worst it just wastes time. It never moves it off the device, so doing it once is all it needs.

RunDevelopment (approved):

PR #448, opened 2022-06-21
More reliable dark mode setting

joeyballentine:
Fixes two things:
- Removes the code that made it back in that was causing a settings problem for fp16 and cpu
- fixes dark mode by using our storage for the setting. This has a side effect of making the transition from dark to light more abrupt unfortunately, but it should function properly now instead of relying on however chakra stores it


src/renderer/components/SettingsModal.tsx
```diff
@@ -251,18 +240,20 @@ const EnvironmentSettings = memo(() => {
             </SettingsItem>
             <Toggle
                 description="Use CPU for PyTorch instead of GPU."
-                isDisabled={!isNvidiaAvailable}
```
RunDevelopment, line 254: What problem did this cause?

RunDevelopment (approved):

PR #449, opened 2022-06-21
Give image previews a checkerboard bg if transparent

joeyballentine:
image

### DIFF
--- a/src/renderer/components/inputs/previews/ImagePreview.tsx
+++ b/src/renderer/components/inputs/previews/ImagePreview.tsx
@@ -83,6 +83,11 @@ const ImagePreview = memo(({ path, schemaId, id }: ImagePreviewProps) => {
                                     ? 'Image preview failed to load, probably unsupported file type.'
                                     : 'File does not exist on the system. Please select a different file.'
                             }
+                            backgroundImage={
+                                img?.channels === 4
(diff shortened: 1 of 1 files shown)


src/renderer/components/inputs/previews/ImagePreview.tsx
```diff
@@ -83,6 +83,11 @@ const ImagePreview = memo(({ path, schemaId, id }: ImagePreviewProps) => {
                                     ? 'Image preview failed to load, probably unsupported file type.'
                                     : 'File does not exist on the system. Please select a different file.'
                             }
+                            backgroundImage={
+                                img?.channels === 4
```
RunDevelopment, line 87: I don't think that's necessary. If the image isn't transparent, nobody will see the BG image. It's also more efficient without the conditional, since the CSS-in-JS library doesn't have to update the styles.

joeyballentine: You actually kinda can see it around the border... It's weird. Change it locally to do it all the time and you'll notice the difference

RunDevelopment: Good point.

RunDevelopment (approved):

PR #453, opened 2022-06-21
Added Combine RGBA node

RunDevelopment:
Added a Combine RGBA node with better semantics than Merge Channels.

I also renamed the "Split Channels" node to "Separate RGBA" and changed its outputs accordingly.


backend/src/nodes/image_chan_nodes.py
```diff
@@ -27,39 +27,91 @@ def __init__(self):
...
+        return [out[2], out[1], out[0], out[3]]
+
+
+@NodeFactory.register("chainner:image:combine_rgba")
+class CombineRgbaNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = (
+            "Merges the given channels together and returns an RGBA image."
+            " If RGB or RGBA image is given as a channel, it will be converted to grayscale first."
```
theflyingzamboni, line 66: The second sentence should be removed since the assert statement has been added.

backend/src/nodes/image_chan_nodes.py
```diff
@@ -27,39 +27,91 @@ def __init__(self):
...
+        )
+        self.inputs = [
+            ImageInput("R Channel"),
+            ImageInput("G Channel"),
+            ImageInput("B Channel"),
+            ImageInput("A Channel").make_optional(),
+        ]
+        self.outputs = [ImageOutput()]
+        self.category = IMAGE_CHANNEL
+        self.name = "Combine RGB"
```
theflyingzamboni, line 76: Is the name meant to be Combine RGB, or is it supposed to be RGBA? The alpha is optional, I realize, but the name should maybe use RGBA for symmetry with the Separate RGBA node.

RunDevelopment: Good find!

backend/src/nodes/image_chan_nodes.py
RunDevelopment: Ah, that documentation is outdated anyway.

joeyballentine (approved):

PR #454, opened 2022-06-22
Fix Torch model preview error

theflyingzamboni:
-Prints messages in node like image preview if pth file is missing or invalid.
-Nulls cache for Load Model node if an invalid pth is loaded.

### DIFF
--- a/backend/src/run.py
+++ b/backend/src/run.py
@@ -189,10 +189,14 @@ async def run_individual(request: Request):
     node_instance = NodeFactory.create_node(full_data["schemaId"])
     # Run the node and pass in inputs as args
     run_func = functools.partial(node_instance.run, *full_data["inputs"])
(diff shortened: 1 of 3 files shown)


backend/src/run.py
```diff
@@ -189,10 +189,14 @@ async def run_individual(request: Request):
     node_instance = NodeFactory.create_node(full_data["schemaId"])
     # Run the node and pass in inputs as args
     run_func = functools.partial(node_instance.run, *full_data["inputs"])
-    output = await app.loop.run_in_executor(None, run_func)
+    try:
+        output = await app.loop.run_in_executor(None, run_func)
+        extra_data = node_instance.get_extra_data()
+    except RuntimeError:
```
theflyingzamboni, line 195: These should maybe be in separate try blocks, I'm not sure.

PR #459, opened 2022-06-22
fix fp16 issue

joeyballentine:
Fixes this:
image


backend/src/nodes/pytorch_nodes.py
```diff
@@ -28,15 +28,6 @@ def check_env():
         "cuda" if torch.cuda.is_available() and os.environ["device"] != "cpu" else "cpu"
     )
 
-    if os.environ["isFp16"] == "True":
```
joeyballentine, line 31: this whole thing isn't really needed anymore

PR #466, opened 2022-06-23
Fixed generated search pattern

### LINKED ISSUE #451: Searching for "spl" doesn't find "Split Transparency"
Seems like there's some problem with the search function.

I'll fix this after #432 to avoid conflicts.

RunDevelopment:
Fixes #451.

The problem was that I originally intended the search regex to be a DFA. However, I made the regex more complex in #366 which means that the old way of ensuring a DFA is no longer correct. The fix is to remove my method of creating a DFA. This means that the generated regex new have polynomial worst-case runtime (with an arbitrarily large exponent), but that is a risk I am willing to take.


joeyballentine (approved):

joeyballentine:
image

PR #470, opened 2022-06-24
Conditional type expressions

RunDevelopment:
This is my first take on conditional type expressions. I essentially implemented a simplified version of [`match`](https://doc.rust-lang.org/rust-by-example/flow_control/match.html).

While I do think that `match` is the right way to go, I'm not too sure about its limitations right now. It doesn't even support numeric literals, which makes it a lot less useful.

---

I just wanted to have a PR open, so I don't forget about this.


RunDevelopment:
`match` is ready, and it's very powerful. I figured out how to do subtract (`without`) arbitrary structs, so `match` support arbitrary struct patterns now. So any conditional type can be done with `match` no matter how complex.

Example from the docs:

```
match 1 | 3 {
    1 => 1,
    _ as x => add(x, 1)
}
# evaluates to 1 | 4
```

For more details about `match`, see the docs.

---

This PR only adds `match` and a simple use case (see the Combine RGBA node). I'll use it in more nodes (notably PyTorch's upscale node) in follow-up PRs.

joeyballentine (approved):
Trusting you on this one lol

PR #483, opened 2022-06-27
Fix interpolation issue

joeyballentine:


RunDevelopment:
Why did you make this change?

Also, without knowing much about PyTorch upscaling, I think this is the wrong approach. The input no make the assumption that all nodes are going to follow the isFP16 setting. This means that the PyTorch nodes are now tightly coupled to the model input class. Even worse, it is simply wrong. E.g. the interpolation node should not work that way (and if it does, we should classify that as a bug; the interpolation of 2 fp32 models shouldn't be fp16 and that of 2 fp16 models shouldn't be fp32).

Instead, these lines of code should be a utility method that is called by the implementation.

joeyballentine:
The only reason I'm doing this is because something is causing a mismatch to happen to multiple people. This might be extreme but it would at least solve the problem. I'm not sure what else to do, since I do already do this in the places I thought mattered

RunDevelopment:
> something is causing a mismatch

Then wouldn't it be better to figure out what that something is instead of applying this band-aid solution?

If this is related to the model interpolation bug I reported yesterday: The bug is perfectly reproducible and consistent even after restarting chainner. If you want/need to chain/models, I'll gladly send them to you.

joeyballentine:
Sure, send them

RunDevelopment:
[Here you go.](https://drive.google.com/file/d/17Iy_9NTW4oKHl7I2OYZj647XKmtYPevS/view?usp=sharing)

RunDevelopment (approved):

PR #484, opened 2022-06-27
Type Colors & Redesign

joeyballentine:
- Handles and edges are now colored by type
- Dark and light mode are both a bit nicer now

Remaining questions:
- Is gray a good color for text, or should that be changed to something else?
- Is red a good color for directories? I feel like red should be reserved for the "not connectable" color

image


RunDevelopment:
Without having reviewed the code, here is a bit of feedback regarding the design:

1. I like that handles now get bigger when you hover over them. This not only makes them easier to hit, it also clearly communicates that they are interactable.
2. When connecting handles, I think they should not have a green/red fill to indicate compatibility. It's too much color. I think @theflyingzamboni's idea about hiding incompatible handles would fit nicely with this redesign. Making it so that incompatible handles don't grow on hover and mabye have 50% opacity would also communicate compatibility to the user. If the 50% opacity was not only on hover but the whole time the user is dragging the edge connection line, then it would even be more useful than the current system.
3. > Is red a good color for directories? I feel like red should be reserved for the "not connectable" color

   I agree. We already use red for the "node not valid icon" and the color red is generally used to communicate a problem. I think we should not use it for directories.

joeyballentine:
In regards to #1, since we have the hittable area surrounding the handle, it growing actually doesn't do anything for making them easier to hit. It's just nice visual feedback

joeyballentine:
image

strings are now light gray/dark gray (dark/light mode), the default color is now that neutral gray, and directories are now purple.

RunDevelopment:
Also, white on white isn't very visible.

image

src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -1,64 +1,189 @@
...
+                                height: '16px',
+                                borderWidth: isConnected ? '2px' : '0px',
+                                borderColor: parentTypeColor ?? 'none',
+                                transition: '0.15s ease-in-out',
+                                ...(handleColors.length > 1
+                                    ? { background: isConnected ? connectedColor : handleGradient }
+                                    : {}),
+                                backgroundColor: isConnected ? connectedColor : handleColors[0],
+                                boxShadow: '2px 2px 2px #00000014',
+                                opacity: showHandle ? 1 : 0.5,
```
RunDevelopment, line 157: I think going to down to 30% would be even better. And, I think it would be nice if we also applied a grayscale filter. Same for the output container too.

image


```suggestion
                                filter: showHandle ? undefined : 'grayscale(100%)',
                                opacity: showHandle ? 1 : 0.3,
```

src/renderer/helpers/colorTools.ts
```diff
@@ -0,0 +1,49 @@
...
+    const GG = G.toString(16).padStart(2, '0');
+    const BB = B.toString(16).padStart(2, '0');
+
+    return `#${RR}${GG}${BB}`;
+};
+
+// Below from https://codepen.io/njmcode/pen/NWdYBy
+
+// Converts a #ffffff hex string into an [r,g,b] array
+const h2r = (hex: string): number[] => {
```
RunDevelopment, line 27:
```suggestion
const h2r = (hex: string): [number, number, number] => {
```

RunDevelopment, line 27: Also, could you please give these functions real names?

RunDevelopment: This should also be done for all other `number[]`s, since the functions assume 3 elements anyway.

src/renderer/helpers/colorTools.ts
```diff
@@ -0,0 +1,49 @@
...
+    const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+    return [parseInt(result![1], 16), parseInt(result![2], 16), parseInt(result![3], 16)];
```
RunDevelopment, line 29: Assuming valid inputs is not a good strategy when it comes to parsing. Please throw an error when the hex is invalid/`result` is null.

src/renderer/helpers/colorTools.ts
```diff
@@ -0,0 +1,49 @@
...
+    // eslint-disable-next-line no-bitwise
+    `#${((1 << 24) + (rgb[0] << 16) + (rgb[1] << 8) + rgb[2]).toString(16).slice(1)}`;
+
+// Interpolates two [r,g,b] colors and returns an [r,g,b] of the result
+// Taken from the awesome ROT.js roguelike dev library at
+// https://github.com/ondras/rot.js
+const interpolateColorImpl = (color1: number[], color2: number[], factor = 0.5): number[] => {
+    const result = color1.slice();
+    for (let i = 0; i < 3; i += 1) {
+        result[i] = Math.round(result[i] + factor * (color2[i] - color1[i]));
```
RunDevelopment, line 43:
```suggestion
        result[i] = Math.round(color1[i] + factor * (color2[i] - color1[i]));
```

RunDevelopment, line 43: Also, it is typically not a good idea to interpolate in sRGB. I.e. let's look at the 50/50 between `#FF0000` and `#00FF00`. One would expect this to be yellow, but interpolating in sRGB yields `#808000`. On the other hand, interpolating in linear RGB yields `#C0C000`.

joeyballentine: Idk how to change that, I just got this code from the internet

RunDevelopment: Like this:
```suggestion
        const c1 = color1[i] ** 2.2;
        const c2 = color2[i] ** 2.2;
        const blend = c1 + factor * (c2 - c1);
        result[i] = Math.round(blend ** (1 / 2.2));
```

I used the gamma 2.2 trick to approximate sRGB to linear.

joeyballentine:
image

joeyballentine:
image

joeyballentine:
the teal/blue thing for string/number isn't ideal IMO, but green looked like it had even less contrast, so I'm not really sure what to do about that

RunDevelopment:
> I guess I could use a lime green?

Sounds like a good idea.

RunDevelopment:
Found another bug. It seems like the "fading out invalid handles" only considers types. It should also consider the difference between input and output, and account for cyclic edges.

Example: The string output here is shown to be allowed to connect to other string outputs and the source node itself. Trying to create any of these edges will not work, despite what the visuals say.
image

RunDevelopment:
image

Cyclic edges are still not visually represented correctly. In the above example, none of the image inputs can be actually be connected to the Resize (Res) output.

RunDevelopment:
I found no other bugs apart from that. While we should probably find criteria for the brightness and saturation of the type and node colors (they are a bit inconsistent rn), this can also be done later in follow-up PRs. Once that bug is fixed, I'll approve.

joeyballentine:
Fixed

src/renderer/helpers/getTypeAccentColors.ts
```diff
@@ -0,0 +1,28 @@
+import { evaluate } from '../../common/types/evaluate';
+import { NamedExpression } from '../../common/types/expression';
+import { intersect } from '../../common/types/intersection';
+import { TypeDefinitions } from '../../common/types/typedef';
+import { NumberType, StringType, Type } from '../../common/types/types';
+
+const defaultColor = () => '#718096';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const colorList = (typeDefinitions: TypeDefinitions, isDarkMode = true) => [
```
RunDevelopment, line 10: Please don't ignore the unused var.

src/renderer/components/outputs/OutputContainer.tsx
```diff
@@ -1,63 +1,172 @@
...
                             onContextMenu={noContextMenu}
                         />
-                    </div>
+                    </Center>
                 </HStack>
             );
         }
 
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        const [gray300, gray700]: string[] = useToken('colors', ['gray.300', 'gray.700']);
```
RunDevelopment, line 161: Please use `useToken() as string[]`. ESLint won't complain that way.

Same for the others `useToken()`s as well.

src/renderer/components/outputs/OutputContainer.tsx
```diff
@@ -1,63 +1,172 @@
...
                     sx={{
                         '.react-flow__handle-connecting': {
-                            background: '#E53E3E !important',
-                            // cursor: 'not-allowed !important',
+                            // background: '#E53E3E !important',
+                            opacity: showHandle ? 1 : 0,
                         },
                         '.react-flow__handle-valid': {
-                            background: '#38A169 !important',
+                            // background: '#38A169 !important',
```
RunDevelopment, line 109: Why is there so much commented out code everywhere?

src/renderer/components/node/Node.tsx
```diff
@@ -132,15 +132,31 @@ const Node = memo(({ data, selected }: NodeProps) => {
     const disabled = useDisabled(data);
     const menu = useNodeMenu(data, disabled);
 
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const [gray400, gray700, gray900, gray600]: string[] = useToken('colors', [
```
RunDevelopment, line 136:
```suggestion
    const [gray400, gray750, gray900, gray600]: string[] = useToken('colors', [
```

RunDevelopment (changes requested):
I gave the code a quick look and have some minor nits

src/renderer/helpers/getTypeAccentColors.ts
joeyballentine: I'm going to remove it from here but keep it (and ignore it) in the other function, as It could potentially be useful in the future and I don't want to remove all the other code that passes the dark mode value in.

src/renderer/components/outputs/OutputContainer.tsx
joeyballentine: Because I forgot to remove it all

src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -1,64 +1,208 @@
...
                     h="full"
                     sx={{
                         '.react-flow__handle-connecting': {
-                            background: '#E53E3E !important',
-                        },
-                        '.react-flow__handle-valid': {
-                            background: '#38A169 !important',
+                            opacity: showHandle ? 1 : 0,
                         },
+                        '.react-flow__handle-valid': {},
```
RunDevelopment, line 135: Does this even do anything?

src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -1,64 +1,208 @@
...
+        }, [connectingFrom, connectingFromType, type, id, inputId]);
+
+        const { typeDefinitions, functionDefinitions } = useContext(GlobalContext);
+        const { useIsDarkMode } = useContext(SettingsContext);
+        const [isDarkMode] = useIsDarkMode;
 
         let contents = children;
-        const handleColor = useColorModeValue('#EDF2F7', '#171923');
-        const borderColor = useColorModeValue('#171923', '#F7FAFC');
+        const handleColors = getTypeAccentColors(type, typeDefinitions, isDarkMode); // useColorModeValue('#EDF2F7', '#171923');
```
RunDevelopment, line 103: Still more commented out code. I think I saw this in a few other places as well. Maybe search with the regex pattern `// [a-z]\w*[(=:]` in the whole code base to find them all?

src/renderer/components/inputs/InputContainer.tsx
joeyballentine: No, I just don't wanna have to go look at the docs to find that again if I wanna use it for something in the future

joeyballentine: Never mind, I can just get rid of this whole style prop

RunDevelopment (approved):

PR #485, opened 2022-06-27
Shift fill

theflyingzamboni:
Added fill options from Rotate to Shift

### DIFF
--- a/backend/src/nodes/image_util_nodes.py
+++ b/backend/src/nodes/image_util_nodes.py
@@ -12,10 +12,12 @@ from .node_factory import NodeFactory
 from .properties.inputs import *
 from .properties.outputs import *
 from .utils.image_utils import (
+    FillColor,
     blend_images,
     calculate_ssim,
     convert_from_BGRA,
     convert_to_BGRA,
+    shift,
 )
 from .utils.pil_utils import *
 from .utils.utils import get_h_w_c
@@ -346,6 +348,15 @@ class ShiftNode(NodeBase):
             ImageInput(),
             NumberInput("Amount X", minimum=None, unit="px"),
             NumberInput("Amount Y", minimum=None, unit="px"),
+            DropDownInput(
+                input_type="FillColor",
+                label="Negative Space Fill",
+                options=[
+                    {"option": "Auto", "value": FillColor.AUTO},
+                    {"option": "Black Fill", "value": FillColor.BLACK},
+                    {"option": "Transparency", "value": FillColor.TRANSPARENT},
+                ],
+            ),
         ]
         self.outputs = [ImageOutput(image_type="Input0")]
         self.category = IMAGE_UTILITY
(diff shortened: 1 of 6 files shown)


backend/src/nodes/image_util_nodes.py
```diff
@@ -396,12 +405,12 @@ def __init__(self):
                 ],
             ),
             DropDownInput(
```
RunDevelopment, line 407: Could you extract this out into its own function? I would like to get rid of the copy paste.

backend/src/nodes/utils/image_utils.py
```diff
@@ -138,6 +144,39 @@ def normalize_normals(
     return x, y, z
 
 
+def select_fill(img: np.ndarray, fill: int, for_pil: bool = False):
```
RunDevelopment, line 147: Instead of having a strange `for_pil` parameter, couldn't `rotate` just multiply the numbers it gets by 255?

If we ever add more fill colors, the `for_pil` will make things more messy.

RunDevelopment: Also, why `select_fill`? Wouldn't `get_fill_color` be more descriptive? "Select" sounds like this function has side effects to me.

src/common/migrations.js
```diff
@@ -544,6 +544,16 @@ const fixDropDownNumberValues = (data) => {
     return data;
 };
 
+const updateShiftNode = (data) => {
```
RunDevelopment, line 547: Why did you add a migration? Added a new dropdown (or in general, any input with a default value) shouldn't require one.

backend/src/nodes/image_util_nodes.py
```diff
@@ -414,6 +423,8 @@ def __init__(self):
     def run(
         self, img: np.ndarray, angle: float, interpolation: int, expand: int, fill: int
     ) -> np.ndarray:
+        """Rotates an image by the desired angle."""
```
RunDevelopment, line 426: Why? This doc string adds no information. I mean, what else is the "RotateNode" with the description "Rotate an image." going to do?

I think is a general problem in the python code base. We have a bunch useless/trivial doc strings. I mean, what value does a "Constructor" doc string inside `__init__` have?

backend/src/nodes/utils/image_utils.py
```diff
@@ -138,6 +144,39 @@ def normalize_normals(
...
+    """Select how to fill negative space that results from rotation"""
+
+    c = get_h_w_c(img)[2]
+    if fill == FillColor.AUTO:
+        fill_color = (0,) * c
+    elif fill == FillColor.BLACK:
+        fill_color = (0,) * c if c < 4 else (0, 0, 0, 255 if for_pil else 1)
+    else:
+        img = convert_to_BGRA(img, c)
```
RunDevelopment, line 156: This has no effect. `convert_to_BGRA` is a pure function. Right now, `rotate` and `shift` will not convert to RGBA because of this.

Either move this call to `convert_to_BGRA` outside this function or return the adjusted image as well.

RunDevelopment (changes requested):

backend/src/nodes/image_util_nodes.py
joeyballentine: We should get rid of all those tbh. It's from the initial class I had set up with generic docstrings, and they really do nothing but clog up all the files now

backend/src/nodes/utils/image_utils.py
theflyingzamboni: Because it selects a fill color based on the option. 'Get' works too.

backend/src/nodes/image_util_nodes.py
theflyingzamboni: I only added it because the inconsistency bothered me haha

backend/src/nodes/utils/image_utils.py
theflyingzamboni: Oh, whoops. I was only half paying attention when I factored this out

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #490, opened 2022-07-02
Added "Copy File Name" and "Copy Full File Path" content menu entries

RunDevelopment:
As per Kim's suggestion, although I'm still waiting for feedback on the exact behavior of "Copy File Name".


joeyballentine:
Are we good with the current functionality of this?

RunDevelopment:
Ah right, I got a "that looks good to me" from Kim, so I assume it#s okay.

PR #491, opened 2022-07-02
Added Model Scale node

### LINKED ISSUE #474: Add a node to get the scale of a model

RunDevelopment:
Resolves #474.

image

---

To implement the node, I added a `PyTorchModel` python type that is the union of all currently supported model types. The type simply meant that I didn't have to add another `# type: ignore` comment.
While I was at it, I also used this type in a few other places, which made most `# type: ignore` comments useless.


joeyballentine:
Could we call it "Get Model Scale" like we do "Get Image Dimensions"? That way it's clear it's just extracting some information rather than performing an operation

PR #498, opened 2022-07-04
Add tile size target to ncnn upscale

joeyballentine:
This is to allow people with the "black result" issue to attempt to bypass auto-tiling


backend/src/nodes/ncnn_nodes.py
```diff
@@ -138,13 +139,19 @@ class NcnnUpscaleImageNode(NodeBase):
...
+            ImageInput(),
+            NumberInput("Tile Size Target", default=0, minimum=0, maximum=None),
+        ]
         self.outputs = [ImageOutput()]
         self.category = NCNN
         self.name = "Upscale Image"
         self.icon = "NCNN"
         self.sub = "Processing"
 
+        self.split_factor = None
```
RunDevelopment, line 153: I don't like that you added mutable state to the node. I can already see this causing bugs. Couldn't you give the `upscale` another parameter instead?

joeyballentine: Yeah

RunDevelopment (approved):

PR #500, opened 2022-07-04
Add eslint unused-imports plugin

joeyballentine:


RunDevelopment:
Doesn't ESLint already have such a rule?

joeyballentine:
> Doesn't ESLint already have such a rule?

It does, but without the plugin it isn't auto fixable

PR #501, opened 2022-07-04
Onnx nodes update

theflyingzamboni:
-Split Convert to ONNX node into Convert + Save Model nodes
-Convert to ONNX now outputs a byte string of the model instead of a file path and connects directly to Upscale Image rather than having to pass through Load Model first


backend/src/nodes/onnx_nodes.py
```diff
@@ -47,18 +48,32 @@ def run(self, path: str) -> Tuple[ort.InferenceSession, str]:
...
+        self.category = ONNX
+        self.name = "Save Model"
+        self.icon = "ONNX"
+        self.sub = "Input & Output"
+
+        self.side_effects = True
+
+    def run(self, onnx_model: bytes, directory: str, model_name: str) -> None:
+        full_path = f"{os.path.join(directory, model_name)}.onnx"
+        with BytesIO() as f:
```
joeyballentine, line 74: Is this with statement necessary? It doesn't seem like you use f anywhere

backend/src/nodes/onnx_nodes.py
```diff
@@ -47,18 +48,32 @@ def run(self, path: str) -> Tuple[ort.InferenceSession, str]:
...
-                "CPUExecutionProvider"
-                if os.environ["device"] == "cpu"
-                else "CUDAExecutionProvider"
-            ],
-        )
-
         basename = os.path.splitext(os.path.basename(path))[0]
 
-        return session, basename
+        return model_as_string, basename
```
RunDevelopment, line 53: The type of `run` still says `ort.InferenceSession`. Is that correct?

backend/src/nodes/pytorch_nodes.py
```diff
@@ -4,10 +4,12 @@
 
 from __future__ import annotations
 
+from io import BytesIO
 import os
 from typing import Any, OrderedDict, Union
 
 import numpy as np
+import onnx
```
RunDevelopment, line 12: Onnx is an optional dependency. Just importing it here will cause the backend to crash for all people (e.g. me) that have PyTorch installed but not Onnx.

RunDevelopment (changes requested):
I don't know how, you managed to make the diff algorithm super confused. The diff of `tests/common/__snapshots__/SaveFile.test.ts.snap` is a mess.

backend/src/nodes/onnx_nodes.py
theflyingzamboni: Whoops, good catch. Leftover code from a different approach.

backend/src/nodes/onnx_nodes.py
theflyingzamboni: Good catch.

backend/src/nodes/pytorch_nodes.py
theflyingzamboni: It's actually an unused import from a previous approach that ended up being unnecessary, just forgot to remove the import.

RunDevelopment (approved):

theflyingzamboni:
> I don't know how, you managed to make the diff algorithm super confused. The diff of `tests/common/__snapshots__/SaveFile.test.ts.snap` is a mess.

I honestly have no idea. When I tried to merge it, it gave me one conflict to resolve, but no matter how it was resolved, the result was somehow deleting the "Read image-iterator.chn" entry. I gave up trying to resolve that, copied the text from Joey's most recent snapshot on GH in, then just ran the update again to add the new chain and migration back in. I do not know why the diff is confused now, but it was definitely confused when it tried to merge main in.

RunDevelopment:
Could you try: 
1. removing the file, 
2. committing that, 
3. running `npm run test-js -- -u` to generate it again, and then 
4. committing that.

theflyingzamboni:
@RunDevelopment Nope, it looks exactly the same. GH is just being stupid. Like I said, VS Code has no issues interpreting this diff correctly as two simple insertions.

RunDevelopment:
Then it's a problem with the diff algorithm GitHub uses. That's a shame. Thanks for trying anyway.

I already approved, so this PR LGTM.

PR #502, opened 2022-07-05
Fix pane context menu node search typeState bug for inputs

joeyballentine:
image

I had noticed that the Image Stack node wasn't finding any nodes for the B/C/D inputs in the pane context menu. I fixed it by using the function definitions rather than the typestate when finding nodes based on inputs. I assume this has something to do with the type state being incomplete for these connections, and it'll be fine to leave it for outputs as those we actually would want to be type-narrowed in the future based on calculated types.

image


RunDevelopment:
This is something I didn't think of in #471. Basically, since those image input are nullable (optional) and aren't connected to anything yet, they get narrowed down to `null`. This is necessary for output types to be calculated correctly together with `match` (#470).

PR #505, opened 2022-07-05
Added small type tag

RunDevelopment:
This adds a small tag to some outputs.

Right now, tags are only shows for numbers and images with certain width and height. I also considered strings, but they can be quite long which messes with the layout.

It might also be useful to repeat this information at connected inputs, but that becomes a bit more tricky to layout with the "optional" tag.

Example:
image

### DIFF
--- a/src/common/types/util.ts
+++ b/src/common/types/util.ts
@@ -4,6 +4,7 @@ import {
     IntIntervalType,
     IntervalType,
     NumericLiteralType,
+    StringLiteralType,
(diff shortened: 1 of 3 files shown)


joeyballentine:
Is it just me or is the tag not vertically centered?

RunDevelopment:
It's top aligned, just like the "optional" tag.

joeyballentine (approved):

PR #510, opened 2022-07-05
Support onnx models with different tensor shape

joeyballentine:


backend/src/nodes/onnx_nodes.py
```diff
@@ -17,6 +18,11 @@
 from .utils.utils import get_h_w_c, np2nptensor, nptensor2np
 
 
+class TensorOrders(Enum):
```
RunDevelopment, line 21: Why did you make this an enum?

joeyballentine: Because I didn't wanna have magic numbers

RunDevelopment: I meant why you inherit (or whatever magic the `class ClassName(Name)` actually does) from `Enum`. Our other "enum classes" don't do that.

joeyballentine: Because I googled "python enum" and copied how to do it from there

RunDevelopment: That's what I did too when I first needed to add a dropdown input for Chainner. I quickly removed the `(Enum)` because enums aren't compatible with JSON and I don't want to add `.value` everywhere. That's why I just use static class variables nowadays.

joeyballentine: I see. I'll change it

RunDevelopment (approved):
I can't really review this properly since I don't have any expertise in this field, but the code LGTM.

PR #514, opened 2022-07-07
Node progress indicator

### LINKED ISSUE #243: Node progress indicator
Some nodes, e.g. upscale, can take quite a long time. It would be nice to have some progress indication for them.

We already do this for iterators (for obvious reasons), but it would be nice if nodes could yield progress updates while running.

joeyballentine:
Closes #243 for the most part. While it doesn't actually report how far into running a node is, it still shows that a node is running, which at least fixes this concern:

> Another example where this would be useful is the preview node. For some large images, it takes a good 10~20 seconds for the backend to save the png. During that time, there is no progress indication at all (if preview is the end of the chain) aside from the Start button not being enabled.

Also, while I was at it I changed how the valid/invalid icons work since I hate how I did that originally.

image
image


src/renderer/components/node/IteratorNode.tsx
```diff
@@ -163,6 +164,7 @@ const IteratorNode = memo(({ data, selected }: IteratorNodeProps) => {
                     schemaId={schemaId}
                 />
                 <NodeFooter
+                    animated={animated || false}
```
RunDevelopment, line 167: This barely works. Using the `||` on nullable booleans is pretty bad practice. Either use `??` or given them a default when destructuring `data` (so `const { ... , animated = false } = data`).

Same for all other occurrences of this pattern.

src/renderer/components/node/IteratorHelperNodeFooter.tsx
```diff
@@ -19,33 +20,53 @@ const NodeFooter = memo(({ validity }: NodeFooterProps) => {
...
-                gutter={24}
-                label={validity.isValid ? 'Node valid' : validity.reason}
-                px={2}
-                py={1}
-                textAlign="center"
-            >
-                <Center
-                    className="nodrag"
-                    my={-1}
+            {animated ? (
```
RunDevelopment, line 23: Do `IteratorHelperNodeFooter` and `NodeFooter` use the same code here? If so, could you please extract this out into its own component?

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #524, opened 2022-07-08
Factored out the common image upscaling code of the 3 upscaling nodes

RunDevelopment:
All the code that makes sure that the image has the right number of channels and handles alpha was shared between the different upscale nodes. I extracted this out into a function I called `convenient_upscale`. I choose this name because it doesn't follow any guidelines and just upscales the image in a way that this convenient most of the time.

I only tested the PyTorch and NCNN Upscale nodes, because I don't have ONNX installed. Could someone please help test ONNX?


joeyballentine:
Tested it, and it works perfectly fine. I did just run into an unrelated onnx bug though, so I guess its good I thoroughly tested all these.

RunDevelopment:
... why did you commit the fix for an unrelated onnx bug into this branch?

joeyballentine:
Sorry. I'll get rid of that. Was just me being lazy

joeyballentine (approved):

PR #525, opened 2022-07-08
Fix ONNX change_shape bug

joeyballentine:


RunDevelopment (approved):
I can't review this bug LGTM!

PR #526, opened 2022-07-08
Better output types for upscale nodes

RunDevelopment:
This adds an alias type that correctly calculates the number of output channels of an upscaled image.

### DIFF
--- a/src/common/types/typedef.ts
+++ b/src/common/types/typedef.ts
@@ -13,7 +13,14 @@ import {
     subtract,
     toString,
 } from './builtin';
-import { Expression, NamedExpression, UnionExpression } from './expression';
+import {
+    BuiltinFunctionExpression,
+    Expression,
+    MatchArm,
+    MatchExpression,
+    NamedExpression,
+    UnionExpression,
+} from './expression';
 import {
     assertValidFunctionName,
     assertValidStructFieldName,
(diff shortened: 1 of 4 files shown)


src/common/types/typedef.ts
```diff
@@ -272,6 +280,56 @@ const addBuiltinTypes = (definitions: TypeDefinitions) => {
...
+     *   outputChannels: int(1..inf),
+     * } = match imageChannels {
+     *   1 => 1,
+     *   4 => match inputChannels { 3 => add(outputChannels, 1), _ => outputChannels },
+     *   _ => outputChannels
+     * }
+     * ```
+     */
+    definitions.add(
+        new AliasDefinition(
```
RunDevelopment, line 297: I'm honestly considering adding a parser for type expressions and statements. I mean, look at this!

joeyballentine: That's probably a good idea. The one you did for text append was pretty gnarly as well

RunDevelopment: Yeah, I think I'll throw some ANTLR4 stuff together. Or do you know any other parser generators for JS/TS?

joeyballentine: I do not.

joeyballentine (approved):

PR #541, opened 2022-07-13
Disabled running fix

### LINKED ISSUE #539: Disabled nodes are animated as running
When you run a chain with disabled nodes, those nodes will be animated as running. Since they never actually run, we never get "node finished" message from the backend, so these nodes get animated as running until the whole chain finishes (or aborts).

image

joeyballentine:
Closes #539

### DIFF
--- a/src/renderer/components/CustomEdge.tsx
+++ b/src/renderer/components/CustomEdge.tsx
@@ -82,6 +82,8 @@ export const CustomEdge = memo(
 
         const chainHoleColor = useColorModeValue('#EDF2F7', '#1A202C');
 
+        const shouldShowAnimation = disabledStatus === DisabledStatus.Enabled && animateChain;
+
         return (
             <g
                 className="edge-chain-group"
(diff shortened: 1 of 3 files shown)


RunDevelopment:
I don't think that's a good solution. You made a fix around not *showing* that certain nodes are animated, when the fix should have been to not animate them in the first place.

joeyballentine:
> I don't think that's a good solution. You made a fix around not _showing_ that certain nodes are animated, when the fix should have been to not animate them in the first place.

That would require checking the disabled status of every node every time we update with the latest running status. That to me seems far worse than just not showing the spinner when it's disabled, even if it would be more technically correct. 

Ideally I'd like to change how we handle running status anyway. I could just deal with that then, if that's ok with you

RunDevelopment:
> That would require checking the disabled status of every node every time we update with the latest running status.

No, we only need to change [this line](https://github.com/joeyballentine/chaiNNer/blob/0592d978a3a2e054d19677dbf1fd4469d2e2098f/src/renderer/contexts/ExecutionContext.tsx#L242) to:

```ts
animate(nodes.map((n) => n.id));
```

This problem is that we animate all nodes when starting the execution instead of only the nodes that actually get executed.

joeyballentine:
Oh, duh. I'll do that then

src/renderer/components/CustomEdge.tsx
```diff
@@ -82,6 +82,8 @@ export const CustomEdge = memo(
 
         const chainHoleColor = useColorModeValue('#EDF2F7', '#1A202C');
 
+        const shouldShowAnimation = disabledStatus === DisabledStatus.Enabled && animateChain;
```
RunDevelopment, line 85: I think that is unnecessary too.

joeyballentine: Oh yeah, forgot about this

RunDevelopment (approved):

PR #545, opened 2022-07-13
Fix iterator crash and leftover edges

joeyballentine:


src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -108,12 +108,28 @@ export const InputContainer = memo(
...
-                    .outputDefaults.get(parentOutputId)!;
-                return getTypeAccentColors(parentType, typeDefinitions, isDarkMode)[0];
+                if (parentNode) {
+                    const parentDef = functionDefinitions.get(parentNode.data.schemaId);
+                    if (!parentDef) {
+                        return getTypeAccentColors(
+                            NeverType.instance,
+                            typeDefinitions,
+                            isDarkMode
+                        )[0];
```
RunDevelopment, line 120: This seems like a lot to get the default color. Maybe we could just export the default color and use it directly?

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -599,6 +599,21 @@ export const GlobalProvider = memo(
             [changeNodes, changeEdges]
         );
 
+        // TODO: This shouldn't be necessary, but it at least fixes the issue
```
RunDevelopment, line 602: So why is it necessary?

joeyballentine: Deleting an iterator doesn't delete an edge if the edge is connected to an outer node. I couldn't think of a way we could do this in removeNodeById, so I just did it here

RunDevelopment: Ah, I forget to account for iterator helper nodes in the code snippet, but the general idea should be clear.

src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -108,12 +108,20 @@ export const InputContainer = memo(
...
+                    if (!parentDef) {
+                        return defaultColor;
+                    }
+                    const parentType = parentDef.outputDefaults.get(parentOutputId);
+                    if (!parentType) {
+                        return defaultColor;
+                    }
+                    return getTypeAccentColors(parentType, typeDefinitions, isDarkMode)[0];
+                }
+                return getTypeAccentColors(NeverType.instance, typeDefinitions, isDarkMode)[0];
```
RunDevelopment, line 124: This should have been another `return defaultColor;` but whatever.

joeyballentine: Ah shit. Didn't notice it because of the formatting

joeyballentine: I knew I should have left this open for another check but I was confident I had fixed everything... Oh well

PR #546, opened 2022-07-14
Allow dragging already placed nodes into iterators

### LINKED ISSUE #395: [ Question ] Why is directory iterator a separate area?
I built up a workflow then setup a directory iterator, only to find I couldn't link my previous flow to the directory iterator.  I'm assuming this is because those nodes weren't "inside" the directory iterator.  Is that true?  If so, why the separate approach for this node?

joeyballentine:
Probably shouldn't merge this for 0.10 since it could introduce new bugs I didn't catch, but for 0.10.1 this should be a welcomed feature. (unless we wanna delay 0.10 a little more, which we could do)

I know iterators are going away relatively soon (hopefully) but this should still add convenience.

Also, closes #395 I guess

### DIFF
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -215,7 +215,7 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
     }, [nodes, edges]);
 
     const onNodeDragStop = useCallback(
-        (event: React.MouseEvent, _node: Node<NodeData>, nNodes: Node<NodeData>[]) => {
+        (event: React.MouseEvent, _node: Node<NodeData> | null, nNodes: Node<NodeData>[]) => {
             const newNodes: Node<NodeData>[] = [];
             const edgesToRemove: Edge[] = [];
             const allIterators = nodes.filter((n) => n.type === 'iterator');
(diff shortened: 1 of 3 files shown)


RunDevelopment:
I'm not sure whether I like this. It makes iterators this trap that nodes can't escape from. If you accidentally drag a node into an iterator, then that's it, you can't get it out anymore (ignoring Ctrl+Z). Would it be possible to allow dragging nodes out of iterators as well? Of course, this only affects nodes that can be dragged out (no inputs are connected to iterator nodes).

> unless we wanna delay 0.10 a little more

I think it has been delayed enough already.

joeyballentine:
> I'm not sure whether I like this. It makes iterators this trap that nodes can't escape from. If you accidentally drag a node into an iterator, then that's it, you can't get it out anymore (ignoring Ctrl+Z). Would it be possible to allow dragging nodes out of iterators as well? Of course, this only affects nodes that can be dragged out (no inputs are connected to iterator nodes).
> 
> > unless we wanna delay 0.10 a little more
> 
> I think it has been delayed enough already.

You must not have noticed you can right click and choose to release them

joeyballentine:
I agree though this is a bit weird, but so many people have tried to do this I figured it was worth it to implement. As for being able to drag stuff out, I'm not sure about that

RunDevelopment:
> You must not have noticed you can right click and choose to release them

I didn't lol.

But that's good enough for me then. Dragging nodes out is more of a nice-to-have feature IMO. The context menu entry is sufficient.

However, I think it is suboptimal that "Release" will just cut connects to iterator nodes. Maybe we could disable the context menu entry when the node to be released has some iterator node inputs?

joeyballentine:
Why is that suboptimal? I feel like it would be annoying to have to manually remove a connection first to be able to release something, especially since it preserves ones that are possible to keep. I don't think users are going to intuitively understand which ones they can keep and which ones to get rid of

PR #548, opened 2022-07-14
Update iterator header to new style

joeyballentine:


RunDevelopment:
I thought you intentionally didn't change iterators to make them more visually distinct. Is it really better to also have them in the new style?

Old:
image


New:
image

joeyballentine:
I previously thought they should be district but changed my mind. Do you think we should just leave them as is?

PR #549, opened 2022-07-14
Change the border style for iterator children

### LINKED ISSUE #382: Better visual indicator for iterator children
Right now, the visual indicator that a node is the child of an iterator is very subtle. Too subtle. Add a clearer visual indicator.

Current indicator:
image

joeyballentine:
closes #382
image


RunDevelopment (approved):
Not sure whether I like the design (maybe I just have to get used to it), but it's very clear. It's a clear improvement, so I'm all for merging this.

joeyballentine:
> Not sure whether I like the design (maybe I just have to get used to it), but it's very clear. It's a clear improvement, so I'm all for merging this.

I don't really like it either but yeah, it's much more visually apparent especially when zoomed out

RunDevelopment:
Yup, so let's merge it. We can still improve the design later, let's go for functionality first.

Issue #315, opened 2022-05-22
More general Math node


RunDevelopment:
Right now, the Math node is very limiting. It only supports a small selection of binary operations, and even simple formulas are difficult to implement.

Add a Math node with a variable number of inputs that produces one output. 

The formula itself is input using a simple text field. The input numbers are simply labeled lowercase a-z. A formula could look like this: `sqrt(a ** 2 + b ** 2) * c / 100`

joeyballentine:
Variable number of inputs is something not possible right now. But yeah, something like this would be cool.

Though, I'm not really sure what use case it would be for. Really all the math node is needed for right now is taking fractions of an image's size and stuff like that

RunDevelopment:
> Variable number of inputs is something not possible right now.

It's possible using a little trick. We don't actually need to support variable number of inputs, it just has to look like we do. We could just have a large-enough fixed number of inputs (e.g. 10) and hide all inputs except for the first one by default. If the user that "adds" new inputs, we just show them the next hidden one.

Hacky, yes, but functional.

> Though, I'm not really sure what use case it would be for.

We have a bunch of nodes with number inputs, **all of which** can benefit from a powerful Math node.

Even the simplest Math node (one that just returns a constant number) is useful because it allows you to control multiple connected values at once.

joeyballentine:
That's actually a pretty genius hack. We could use that for all the variable input nodes, like image stack. Just do a [ImageInput(label=f"Image {i}") for i in range(100))] and then have configurable option for whether it's variable or not, which we'd set to true (and probably have a starting index option), and then it would handle that on the frontend.

joeyballentine:
https://stackoverflow.com/questions/2371436/evaluating-a-mathematical-expression-in-a-string

PR #550, opened 2022-07-14
Remember note resize

### LINKED ISSUE #457: Note resize not preserved
The Note node allows you to resize the text box so that you can see the full text without scrolling. However, if the chain is saved, closed, and reopened, the node sizes revert to default instead of preserving the set dimensions.

(Cursor also currently doesn't change from text cursor when clicking and dragging the scroll bar)

joeyballentine:
I committed a grave sin with this. Nothing else uses this input though so I hope you can forgive me.

The other option if JSON stringifying this is a bit too much, is just using inOutId +1 and +2 to store the size. This would mean you could never have inputs come after a textarea though (not that anything else is using or will ever use this)

Closes #457


RunDevelopment:
I don't think this is a good solution. You essentially made the text area input exclusive to the Note node. There are several text inputs that could benefit for being multiline:

- The pattern input of Text Pattern.
- The formula input of the general math node (#315).
- The regex and pattern inputs of the Regex Replace node (#352).

It would be nice if all of those were also resizable.

Could we not just add a new node data property `inputSize: Record<InputId, Size>` to track this size instead?

joeyballentine:
@RunDevelopment So interestingly enough, after making the change to use a new inputSize property, it is doing the same thing that the iteratorSize property does when you resize it where it flashes between the previous and the next size. I have no idea what's causing that or how to fix it :/

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -755,6 +761,29 @@ export const GlobalProvider = memo(
             [modifyNode, schemata]
         );
 
+        const useInputSize = useCallback(
+            (
+                id: string,
+                inputId: InputId,
+                inputSize?: InputSize
+            ): readonly [Size, (size: Size) => void, () => void] => {
+                const currentSize = inputSize ? inputSize[inputId] : { width: 0, height: 0 };
```
RunDevelopment, line 770: This is incorrect. `inputSize` might not have a key `inputId`, so `currentSize` can be undefined.
```suggestion
                const currentSize = inputSize?.[inputId] ?? { width: 0, height: 0 };
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -755,6 +761,29 @@ export const GlobalProvider = memo(
...
+                    modifyNode(id, (old) => {
+                        const nodeCopy = copyNode(old);
+                        nodeCopy.data.inputSize = {
+                            ...nodeCopy.data.inputData,
+                            [inputId]: size,
+                        };
+                        return nodeCopy;
+                    });
+                };
+                const resetInputSize = () => setInputSize({ width: 0, height: 0 } as Size);
```
RunDevelopment, line 781: This type cast is unnecessary.

Also, please try to avoid type casts in general. A type cast is intended to be used can't correctly determine the type of something. Since you essentially say "trust me on that one", all safety TS provides is thrown out of the window. It's the same with non-nullish assertions. They are essentially hacks.

```suggestion
                const resetInputSize = () => setInputSize({ width: 0, height: 0 });
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -755,6 +761,29 @@ export const GlobalProvider = memo(
...
+                id: string,
+                inputId: InputId,
+                inputSize?: InputSize
+            ): readonly [Size, (size: Size) => void, () => void] => {
+                const currentSize = inputSize ? inputSize[inputId] : { width: 0, height: 0 };
+                const setInputSize = (size: Size) => {
+                    modifyNode(id, (old) => {
+                        const nodeCopy = copyNode(old);
+                        nodeCopy.data.inputSize = {
+                            ...nodeCopy.data.inputData,
```
RunDevelopment, line 775:
```suggestion
                            ...nodeCopy.data.inputSize,
```

src/renderer/components/inputs/TextAreaInput.tsx
```diff
@@ -1,18 +1,31 @@
...
+        const zoom = useContextSelector(GlobalVolatileContext, (c) => c.zoom);
+        const { useSnapToGrid } = useContext(SettingsContext);
+        const [isSnapToGrid, , snapToGridAmount] = useSnapToGrid;
+
         const [input, setInput] = useInputData<string>(inputId);
+        const [size, setSize] = useInputSize(inputId);
         const [tempText, setTempText] = useState('');
 
         useEffect(() => {
+            if (size.width === 0 && size.height === 0) {
```
RunDevelopment, line 26: We need to talk about the default size being 0;0. You are essentially treating 0;0 as a "size not defined" state. If you used `undefined` as the "size not defined" state, then you could get rid of the 0;0 default size completely. Even better, TypeScript would then be able to verify that you accounted for this state everywhere.

src/renderer/components/inputs/TextAreaInput.tsx
```diff
@@ -26,19 +39,54 @@ export const TextAreaInput = memo(
...
+                    bottom: !isLocked && resizable && true,
+                    left: false,
+                    topRight: false,
+                    bottomRight: !isLocked && resizable && true,
+                    bottomLeft: false,
+                    topLeft: false,
+                }}
+                grid={useMemoArray<[number, number]>(
+                    isSnapToGrid ? [snapToGridAmount, snapToGridAmount] : [1, 1]
+                )}
```
RunDevelopment, line 57: Why should the size of text boxes respect snap to grid?

src/common/common-types.ts
```diff
@@ -61,6 +61,7 @@ export interface Output {
 }
 
 export type InputData = Readonly<Record<InputId, InputValue>>;
+export type InputSize = Readonly<Record<InputId, Size>>;
```
RunDevelopment, line 64: Could you make it `Readonly<Size>` (here and in a few other places as well)?

RunDevelopment (changes requested):

src/renderer/contexts/GlobalNodeState.tsx
RunDevelopment: It's very strange that TS didn't catch this. For some reason, the type of the object comes out as `{}` (so any non-nullish value). This seems to be a bug in TS itself.

image

RunDevelopment: The problem seems to be that TS can't deal with branded types (e.g. `InputId`) as object keys. This seems to be a known issue.

src/renderer/components/inputs/TextAreaInput.tsx
joeyballentine: Idk, I did it for iterators so I just also did it here. I figured it made sense at the time that everything would just snap to the grid including resizing

RunDevelopment: That also seems a bit odd to me, but I think it's justifiable there. However, I don't see why inputs should snap to the grid. After all, the inputs don't directly determine the size of the whole node, so the size of the node isn't actually aligned with the grid most of the time.

joeyballentine: True. I guess I was thinking about it in terms of just the node node since it's the only input

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -102,6 +103,11 @@ interface Global {
         inputId: InputId,
         inputData: InputData
     ) => readonly [T | undefined, (data: T) => void, () => void];
+    useInputSize: (
+        id: string,
+        inputId: InputId,
+        inputSize?: InputSize
```
RunDevelopment, line 109: I don't like this. The parameter isn't optional, it can just be undefined.
```suggestion
        inputSize: InputSize | undefined
```

src/renderer/components/inputs/props.ts
```diff
@@ -1,10 +1,18 @@
...
+    SchemaId,
+    Size,
+} from '../../../common/common-types';
 import { Type } from '../../../common/types/types';
 
 export interface InputProps {
     readonly id: string;
     readonly inputId: InputId;
     readonly inputData: InputData;
+    readonly inputSize?: InputSize;
```
RunDevelopment, line 15: Is this even necessary? Why would any input use `inputSize` when we have `useInputSize`?

src/renderer/components/inputs/TextAreaInput.tsx
```diff
@@ -26,19 +35,47 @@ export const TextAreaInput = memo(
...
+                    }
                 }}
-            />
+            >
+                <Textarea
+                    className="nodrag"
+                    disabled={isLocked}
+                    draggable={false}
+                    h="100%"
+                    minW={240}
```
RunDevelopment, line 68: I think this isn't necessary.

src/renderer/components/inputs/TextAreaInput.tsx
```diff
@@ -26,19 +35,47 @@ export const TextAreaInput = memo(
...
+                    topLeft: false,
+                }}
+                minHeight={80}
+                minWidth={240}
+                scale={zoom}
+                onResizeStop={(e, direction, ref, d) => {
+                    if (!isLocked) {
+                        setSize({
+                            width: (size?.width || 0) + d.width,
+                            height: (size?.height || 0) + d.height,
```
RunDevelopment, line 58:
```suggestion
                            width: (size?.width ?? 0) + d.width,
                            height: (size?.height ?? 0) + d.height,
```

src/renderer/components/inputs/TextAreaInput.tsx
```diff
@@ -26,19 +35,47 @@ export const TextAreaInput = memo(
...
-                    setTempText(event.target.value);
-                    handleChange(event);
+                defaultSize={size}
+                enable={{
+                    top: false,
+                    right: !isLocked && resizable && true,
+                    bottom: !isLocked && resizable && true,
+                    left: false,
+                    topRight: false,
+                    bottomRight: !isLocked && resizable && true,
```
RunDevelopment, line 47: `&& true`?

src/renderer/components/inputs/props.ts
joeyballentine: I just copied how inputData/useInputData was, so you tell me lol

RunDevelopment: We inputs even use `inputData`....

RunDevelopment: Nope, they don't. Huh, well that's the next thing we can remove.

src/renderer/components/inputs/TextAreaInput.tsx
joeyballentine: lmao

RunDevelopment (approved):

PR #557, opened 2022-07-15
Guaranteed linear worst-case search predicate

RunDevelopment:
I replaced JavaScript's regex engine with [Rust's regex crate](https://docs.rs/regex/latest/regex/). This regex engine guarantees linear execution time, so it's always fast. To actually use the crate, I add [`rregex`](https://www.npmjs.com/package/rregex) as a dependency which is just some wasm bindings for the crate.

I used Rust's regex crate because I have previously used it and know that the library is battle-tested and mature. I initially wanted to use [re2](https://www.npmjs.com/package/re2), because it's more commonly used, but could neither get `re2` nor [`re2-wasm`](https://www.npmjs.com/package/re2-wasm) to work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "react-hotkeys-hook": "^3.4.6",
         "react-icons": "^4.4.0",
         "react-markdown": "^8.0.3",
+        "rregex": "^1.5.1",
         "scheduler": "^0.22.0",
         "semver": "^7.3.7",
         "systeminformation": "^5.11.16",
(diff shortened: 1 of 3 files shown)


RunDevelopment:
It's the only one I could get to work, so it seems more production ready to me than the re2 stuff.

joeyballentine:
It's probably fine

RunDevelopment:
Yeah, and it's not like we now forever depend on this package. It's a small implementation detail that we can change at any time if it causes problems.

PR #560, opened 2022-07-16
Output broadcasting (initial backend implementation)

joeyballentine:
This is my initial backend implementation for the "every output sends its values to the frontend after running" thing. I mainly just want to get this reviewed to see if there might be a better way of doing some of this, but it shouldn't be of any harm to keep this in even without the frontend portion.

The concept is simple: After a node finishes, it runs each output through a "broadcast" method, similar to what we do with input enforcement. This transforms the data into a format the frontend can actually use (like with "get_extra_data" right now). The difference is this gets put in its own cache, and gets sent along with the node finish progress. it then updates all the nodes with their data, and that data gets passed down into the outputs.

There is an issue with this approach though: we send this data to the frontend multiple times. I don't know what the best solution to this is though, because without doing this there isn't a guarantee the frontend will always get and process all the data due to the throttling I put in place. So we really have three options I can think of:
1. Remove the throttling (this will cause iterators to lag a lot if they process quickly)
2. Change how SSE events are handled in general so instead of sending an update when a node finishes we just send updates every X number of ms
3. We leave it as-is

I think option 2 might be the most preferable, but idk for sure. Regardless, for now I think this is fine due to the relatively small amount of data sent


backend/src/nodes/properties/outputs/base_output.py
```diff
@@ -34,3 +34,9 @@ def __repr__(self):
 
     def __iter__(self):
         yield from self.toDict().items()
+
+    def broadcast(self, value):
+        raise NotImplementedError()
+
+    def get_id(self):
```
RunDevelopment, line 41: Why is this necessary? `self.id` is public.

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -14,12 +18,37 @@ def AudioOutput():
...
+        label: str = "Image",
+        image_type: expression.ExpressionJson = "Image",
+    ):
+        super().__init__(expression.intersect(image_type, "Image"), label)
+
+    def broadcast(self, value: np.ndarray) -> dict:
+        img = value
+        h, w, c = get_h_w_c(img)
+
+        # resize the image, so the preview loads faster and doesn't lag the UI
```
RunDevelopment, line 33: Please move this into its own function instead of copy-pasting. This should include at least the resize+base64 part. I'm not sure about the dict.

backend/src/nodes/properties/outputs/pytorch_outputs.py
```diff
@@ -1,10 +1,51 @@
-from .. import expression
+try:
+    import torch
+except:
+    torch = None
+
+from typing import Union
+
+from ...properties import expression
```
RunDevelopment, line 8: Why did you change the expression import? Is that style preferred?

backend/src/nodes/properties/outputs/pytorch_outputs.py
```diff
@@ -1,10 +1,51 @@
-from .. import expression
+try:
+    import torch
+except:
+    torch = None
```
RunDevelopment, line 4: I hate this. Just remove `from .pytorch_outputs import *` from `__init__.py` and we should be able to make them all static imports.

Man, I hate that python imports suck so much.

backend/src/nodes/properties/outputs/pytorch_outputs.py
```diff
@@ -1,10 +1,51 @@
...
+    def __init__(
+        self,
+        model_type: expression.ExpressionJson = "PyTorchModel",
+        label: str = "Model",
+    ):
+        super().__init__(model_type, label)
+
+    def broadcast(self, value: PyTorchModel) -> dict:  # type: ignore
+        if torch is not None:
+            assert isinstance(value, torch.nn.Module), "Expected a PyTorch model"
```
RunDevelopment, line 32: Why this assertion?

backend/src/nodes/properties/outputs/pytorch_outputs.py
```diff
@@ -1,10 +1,51 @@
...
+
+class ModelOutput(BaseOutput):
+    def __init__(
+        self,
+        model_type: expression.ExpressionJson = "PyTorchModel",
+        label: str = "Model",
+    ):
+        super().__init__(model_type, label)
+
+    def broadcast(self, value: PyTorchModel) -> dict:  # type: ignore
```
RunDevelopment, line 30: Not typing the return type would probably be better here, since pyright and co could at least infer the type then.

backend/src/process.py
```diff
@@ -71,7 +72,10 @@ async def __process(self, node: UsableData) -> Any:
         logger.debug(f"Running node {node_id}")
         # Return cached output value from an already-run node if that cached output exists
         if self.output_cache.get(node_id, None) is not None:
-            finish_data = await self.check()
+            finish_data = {
+                "finished": [key for key in self.output_cache.keys()],
+                "broadcastData": self.broadcast_cache,
```
RunDevelopment, line 77: You send over **all** data every time something changes?

That's not good. The small image previews were designed to be around 1MB. This is fine for a single image, but not if you have potentially dozens of them.

backend/src/process.py
```diff
@@ -155,9 +159,38 @@ async def __process(self, node: UsableData) -> Any:
...
             run_func = functools.partial(node_instance.run, *enforced_inputs)
             output = await self.loop.run_in_executor(None, run_func)
+            node_outputs = node_instance.get_outputs()
+            broadcast_data = dict()
+            if len(node_outputs) > 0:
+                if len(node_outputs) == 1:
+                    try:
+                        broadcast_data[0] = node_outputs[0].broadcast(output)
+                    except NotImplementedError:
+                        pass
```
RunDevelopment, line 169: Please don't do exception-driven control flow.

How about this: `broadcast` is allowed to return `None` as a "nothing to broadcast" state. This would make for a very natural default implementation.

src/renderer/components/node/Node.tsx
```diff
@@ -11,6 +11,7 @@ import { shadeColor } from '../../helpers/colorTools';
 import { getSingleFileWithExtension } from '../../helpers/dataTransfer';
 import { DisabledStatus } from '../../helpers/disabled';
 import { getNodeAccentColor } from '../../helpers/getNodeAccentColor';
+
```
RunDevelopment, line 14: Please remove this line.

src/main/main.ts
```diff
@@ -591,7 +591,7 @@ const createWindow = async () => {
             contextIsolation: false,
         },
         // icon: `${__dirname}/public/icons/cross_platform/icon`,
-        show: false,
+        show: true,
```
RunDevelopment, line 594: Please don't forget to remove your debug code :)

src/common/common-types.ts
```diff
@@ -89,6 +90,7 @@ export interface NodeData {
     readonly isDisabled?: boolean;
     readonly isLocked?: boolean;
     readonly inputData: InputData;
+    readonly outputData?: OutputData;
```
RunDevelopment, line 93: I don't think this is a good idea. Nothing about output data is an intrinsic property of a node. These outputs are values computed from nodes. Just like types, I think output values should live outside of nodes and be tracked separately.

RunDevelopment (changes requested):

RunDevelopment:
> I think option 2 might be the most preferable, but idk for sure. Regardless, for now I think this is fine due to the relatively small amount of data sent

Or we just make a new event that says: "this one node finished, those are the outputs". This has no duplication, the executor doesn't need to cache the broadcast outputs, and all outputs are only sent once.

backend/src/nodes/properties/outputs/pytorch_outputs.py
joeyballentine: I had removed it by accident and added it again when I realized I needed it, and didn't realize it was different

backend/src/nodes/properties/outputs/pytorch_outputs.py
joeyballentine: Same

backend/src/process.py
joeyballentine: That's why I said it's concerning in my initial comment

src/main/main.ts
joeyballentine: Oops

joeyballentine:
> > I think option 2 might be the most preferable, but idk for sure. Regardless, for now I think this is fine due to the relatively small amount of data sent
> 
> Or we just make a new event that says: "this one node finished, those are the outputs". This has no duplication, the executor doesn't need to cache the broadcast outputs, and all outputs are only sent once.

This is what I tried originally and it didn't work at all. Though I think that might have been because I had been listening for it in node.tsx instead of in execution context

RunDevelopment:
> I think that might have been because I had been listening for it in node.tsx instead of in execution context

That's probably it. Right now, there is only one listener for SSEs. Maybe adding a second one broke some stuff?

src/renderer/components/node/NodeOutputs.tsx
```diff
@@ -12,6 +13,14 @@ interface NodeOutputsProps {
 
 export const NodeOutputs = memo(({ outputs, id, schemaId }: NodeOutputsProps) => {
     const { functionDefinitions } = useContext(GlobalContext);
+    const useOutputDataContext = useContextSelector(ExecutionContext, (c) => c.useOutputData);
+
+    const useOutputData = useCallback(
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        (outputId: OutputId) => useOutputDataContext(id, outputId),
+        [useOutputDataContext, id]
+    );
```
RunDevelopment, line 22: I added the `useInputData` "hook" for inputs because we have multiple input components. It was a means to avoid code duplication.

However, I don't think that this is worth it. It just adds more code. Do we expect to add more output components in the near future? If no, then this code should live in `GenericOutput`.

src/renderer/hooks/useMap.ts
```diff
@@ -0,0 +1,47 @@
+// From https://usehooks-ts.com/react-hook/use-map
```
RunDevelopment, line 1: Why did you copy a whole copy-on-write-map hook implementation? Do you expect to use more than `set` in the future?

Also, the implementation and types aren't even that good. If we decide to keep it, I'll do a followup PR to bring it in shape.

RunDevelopment (changes requested):

backend/src/process.py
```diff
@@ -177,9 +179,31 @@ async def __process(self, node: UsableData) -> Any:
...
+            node_outputs = node_instance.get_outputs()
+            broadcast_data = dict()
+            if len(node_outputs) > 0:
+                if len(node_outputs) == 1:
+                    try:
+                        broadcast_data[0] = node_outputs[0].broadcast(output)
+                    except Exception as e:
+                        logger.error(f"Error broadcasting output: {e}")
+                else:
+                    for idx, node_output in enumerate(node_outputs):
```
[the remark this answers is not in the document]
RunDevelopment: Ahh, I messed up. My suggestion isn't correct at all.

backend/src/process.py
```diff
@@ -177,9 +179,31 @@ async def __process(self, node: UsableData) -> Any:
             # Run the node and pass in inputs as args
             run_func = functools.partial(node_instance.run, *enforced_inputs)
             output = await self.loop.run_in_executor(None, run_func)
+            node_outputs = node_instance.get_outputs()
+            broadcast_data = dict()
+            if len(node_outputs) > 0:
+                if len(node_outputs) == 1:
+                    try:
+                        broadcast_data[0] = node_outputs[0].broadcast(output)
```
RunDevelopment, line 187: If you converted `output` into an iterable if `len(node_outputs) == 1`, then you could get rid of this `if` block.

You should probably store the iterable outputs in a new variable. So something like: `output_iter = [output] if len(node_outputs) == 1 else output`.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -147,6 +150,14 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
 
     const [isBackendKilled, setIsBackendKilled] = useState(false);
 
+    const [outputDataMap, outputDataMapActions] = useMap<string, OutputData>();
+    const useOutputData = useCallback(
+        (id: string, outputId: OutputId): OutputData | undefined => {
+            const currentInput = outputDataMap.get(id)?.[outputId] as OutputData | undefined;
```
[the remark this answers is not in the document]
joeyballentine: I think TS might just bug out every once in a while because this is the second time it's wanted me to do this when it wasn't necessary

src/renderer/components/node/NodeOutputs.tsx
joeyballentine: Yes I plan on adding more output components

src/renderer/hooks/useMap.ts
joeyballentine: I honestly just wasnt sure what the best way to store this data was

src/renderer/contexts/ExecutionContext.tsx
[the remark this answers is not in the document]
RunDevelopment: Was it TS or something ESLint suggested? ESLint seems to have problems. When imported files change, it sometimes fails to update the type from those files, which leads to really annoying errors. Reloading VSCode fixes that (F1 > "Reload Window").

src/renderer/components/node/NodeOutputs.tsx
RunDevelopment: Alright, then let's keep it. (Could you resolve this thread? I don't have the permissions to do so.)

src/renderer/hooks/useMap.ts
RunDevelopment: Storing it in a map is the right decision IMO, I just don't like the hook. Seems overkill to me if we use it for just the broadcast outputs.

joeyballentine: I'm willing to change it then. Should I just use a non-hook map?

src/renderer/contexts/ExecutionContext.tsx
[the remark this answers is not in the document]
joeyballentine: Eslint, yeah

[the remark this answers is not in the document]
joeyballentine: oh i know why, my type hint for the function's return is wrong

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -1,5 +1,11 @@
+import base64
+
+from ...utils.image_utils import preview_encode
+from ...utils.utils import get_h_w_c
 from .base_output import BaseOutput
 from .. import expression
+import numpy as np
+import cv2
```
RunDevelopment, line 8: What do you need this input for?

backend/src/process.py
```diff
@@ -177,9 +179,28 @@ async def __process(self, node: UsableData) -> Any:
             # Run the node and pass in inputs as args
             run_func = functools.partial(node_instance.run, *enforced_inputs)
             output = await self.loop.run_in_executor(None, run_func)
+            node_outputs = node_instance.get_outputs()
+            broadcast_data = dict()
```
RunDevelopment, line 183: Please add a type. Then you'll see the bug from the next comment.
```suggestion
            broadcast_data: Dict[int, Any] = dict()
```

backend/src/process.py
```diff
@@ -177,9 +179,28 @@ async def __process(self, node: UsableData) -> Any:
...
+            if len(node_outputs) > 0:
+                output_idxable = [output] if len(node_outputs) == 1 else output
+                for idx, node_output in enumerate(node_outputs):
+                    try:
+                        broadcast_data[node_output.id] = node_output.get_broadcast_data(
+                            output_idxable[node_output.id or 0]
+                        )
```
RunDevelopment, line 190: 1. Output values are accessed by index.
2. The output id defaults to the index if `None`.

```suggestion
                        output_id = node_output.id if node_output.id is not None else idx
                        broadcast_data[node_output.id] = node_output.get_broadcast_data(output_idxable[idx])
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -201,6 +213,17 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
         [setStatus, unAnimate, schemata]
     );
 
+    useBackendEventSourceListener(
+        eventSource,
+        'node-output-data',
+        (data) => {
+            if (data) {
+                setOutputDataMap((prev) => new Map({ ...prev, [data.nodeId]: data.data }));
```
RunDevelopment, line 221: We want to iterate the values of the map, not its object properties.

```suggestion
                setOutputDataMap((prev) => new Map([ ...prev, [data.nodeId]: data.data ]));
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -147,6 +149,16 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
 
     const [isBackendKilled, setIsBackendKilled] = useState(false);
 
+    const [outputDataMap, setOutputDataMap] = useState<Map<string, OutputData>>(
+        new Map<string, OutputData>()
+    );
```
RunDevelopment, line 154: Let's go full immutable. `EMPTY_MAP` is from util.

```suggestion
    const [outputDataMap, setOutputDataMap] = useState<ReadonlyMap<string, OutputData>>(EMPTY_MAP);
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -147,6 +149,16 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+    const useOutputData = useCallback(
+        (id: string, outputId: OutputId): unknown => {
+            const currentInput = outputDataMap.get(id)?.[outputId];
+            return currentInput;
+        },
```
RunDevelopment, line 159:
```suggestion
        (id: string, outputId: OutputId): unknown => outputDataMap.get(id)?.[outputId],
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -99,7 +100,10 @@ interface Global {
     changeEdges: SetState<Edge<EdgeData>[]>;
     useAnimate: () => readonly [
         (nodeIdsToAnimate?: readonly string[] | undefined) => void,
-        (nodeIdsToUnAnimate?: readonly string[] | undefined) => void
+        (
+            nodeIdsToUnAnimate?: readonly string[] | undefined,
+            outputData?: { [key: string]: OutputData }
```
RunDevelopment, line 105: Why?

RunDevelopment (changes requested):

src/renderer/contexts/ExecutionContext.tsx
joeyballentine: This doesn't work.

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: Forgot to remove this part

src/renderer/contexts/ExecutionContext.tsx
RunDevelopment: My bad, New suggestion.

PR #561, opened 2022-07-16
Added scopes to Navi

RunDevelopment:
This is a major update to Navi. Before I talk about scopes, here are the other changes this PR includes:

- Added user-defined functions. Functions can now be defined using the `def` keyword. E.g. `def inc(n: number) = add(n, 1);` is a function that takes one number `n` and returns `n+1`.
- Added variables. Variables are a way to get a name to an expression. E.g. `let foo = add(1, 1);` is a variable.
- Removed aliases. With variables and functions, aliases are no longer needed. Functions can do everything aliases can do, but functions are more convenient.
- All of the type definitions Chainner needs (e.g. the `Image` type) are now defined in as Navi code.

I removed aliases in favor of variables and functions because the latter is more in line with what Navi really is: a programming language. Aliases were already as powerful as functions, so not a lot changed. Functions and variables are just a more intuitive way of describing how a type is evaluated. With the addition of recursion (necessary for variable length inputs), Navi might even become turing complete (unless I only allow tail recursion, which might be enough). 

Okay, scopes:

A scope is pretty much what a scope in e.g. JS also is. It's a collection of variables and functions. 
More specifically, it's a collection of definitions. A scope can contain variable definitions, function definitions, and struct definitions. Scopes can also have a parent scope. If the parent scope has a definition with the same name as the child scope, then the definition of the child scope will be used (shadowing). 

This creates a hierarchy of scopes, which is super useful because it allows us to separate different contexts. I.e. there is now a global scope which contains all definitions defined by Navi, and a scope that defines all types specific to Chainner with the global scope as its parent scope. This finally allows us to cleanly separate Navi's built-in stuff from Chainner-specific stuff.

---

Sorry for the bad commit messages.


joeyballentine:
> Sorry for the bad commit messages.

Don't worry, they get squashed anyway.

src/common/types/antlr4/NaviLexer.js
```diff
@@ -3,79 +3,82 @@
 import antlr4 from 'antlr4';
 
 
-const serializedATN = [4,0,24,219,6,-1,2,0,7,0,2,1,7,1,2,2,7,2,2,3,7,3,2,
+const serializedATN = [4,0,26,227,6,-1,2,0,7,0,2,1,7,1,2,2,7,2,2,3,7,3,2,
```
joeyballentine, line 6: what even is this?

RunDevelopment: To my knowledge, an ATN is a collection of graphs (think DFAs) that match text. These numbers are a serialized form of this.

https://www.antlr.org/api/JavaTool/org/antlr/v4/codegen/model/SerializedATN.html
https://stackoverflow.com/questions/41306505/antlr4-what-does-atn-stand-for

joeyballentine: Neat

joeyballentine (approved):

PR #575, opened 2022-07-17
Fix iterator drag-in & releasing

joeyballentine:
what we talked about in discord

### DIFF
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -255,8 +255,7 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                             },
                         };
 
-                        edgesToRemove = [
-                            ...edgesToRemove,
+                        edgesToRemove.push(
                             ...edges.filter((e) => {
                                 if (e.source === node.id) {
                                     const target = nodes.find((n) => n.id === e.target);
@@ -266,7 +265,7 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                                 }
                                 return false;
                             }),
-                        ];
+                        );
 
                         newNodes.push(newNode);
                     }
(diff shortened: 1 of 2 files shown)


src/renderer/components/ReactFlowBox.tsx
```diff
@@ -255,14 +255,18 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
+                            ...edges.filter((e) => {
+                                if (e.source === node.id) {
+                                    const target = nodes.find((n) => n.id === e.target);
+                                    if (target) {
+                                        return target.parentNode !== iterInBounds.id;
+                                    }
+                                }
+                                return false;
+                            }),
+                        ];
```
RunDevelopment, line 269: While I commend your commitment to immutable data structures, using `push` would probably be easier.
```suggestion
                        edgesToRemove.push(
                            ...edges.filter((e) => {
                                if (e.source === node.id) {
                                    const target = nodes.find((n) => n.id === e.target);
                                    if (target) {
                                        return target.parentNode !== iterInBounds.id;
                                    }
                                }
                                return false;
                            }),
                        );
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -582,10 +582,14 @@ export const GlobalProvider = memo(
 
         const releaseNodeFromParent = useCallback(
             (id: string) => {
+                let parentNode: Node<NodeData> | undefined;
+                let node: Node<NodeData> | undefined;
+                let capturedNodes: Node<NodeData>[];
```
RunDevelopment, line 587: You assumed that state updates will be processed in order. While this guaranteed for a single memory cell, I don't think it's guaranteed for different memory cells. `changeNodes` and `changeEdges` ultimately use `useNodesState`'s and `useEdgesState`'s setters, which probably use different cells.

RunDevelopment (changes requested):

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: so what do you recommend I do here? use getNodes in the changeEdges callback?

RunDevelopment: I'm pretty sure using `getNodes` inside a state update isn't defined either. I think we need to `getNodes()` and figure out what edges to remove before `changeEdges`.

RunDevelopment (changes requested):
This fixes the 2 bugs, but I found another one. If you drag multiple nodes into an iterator, all edges between them will be removed.

RunDevelopment (approved):

PR #576, opened 2022-07-17
Add spritesheet iterator

### LINKED ISSUE #564: Sprite sheet iterator and generator
Add 2 nodes for iterating (splitting) and generating (combining) sprites.

1. An "Iterate Sprite Sheet" node that takes one image and its dimensions, and returns an image iterator of all sprites in the sheet.
2. A "Generate Sprite Sheet" node that takes an image iterator and dimensions, and returns one image that is the sprite sheet of all images in the iterator.

This requires the new iterator concept we've been talking about.

joeyballentine:
Sorta does what #564 wants, though this will need to be revisited when we change how iterators are. What this does, is it takes in a spritesheet and the number of rows and columns to split it into, and then iterates over them all, then takes the results and combines them back into a spritesheet.

This doesn't entirely solve the use case mentioned in discord, but it should help when you want to upscale individual sprites in a sheet rather than doing them all at once.

### DIFF
--- a/backend/src/nodes/image_iterator_nodes.py
+++ b/backend/src/nodes/image_iterator_nodes.py
@@ -22,6 +22,9 @@ IMAGE_ITERATOR_NODE_ID = "chainner:image:file_iterator_load"
 VIDEO_ITERATOR_INPUT_NODE_ID = "chainner:image:simple_video_frame_iterator_load"
 VIDEO_ITERATOR_OUTPUT_NODE_ID = "chainner:image:simple_video_frame_iterator_save"
 
+SPRITESHEET_ITERATOR_INPUT_NODE_ID = "chainner:image:spritesheet_iterator_load"
+SPRITESHEET_ITERATOR_OUTPUT_NODE_ID = "chainner:image:spritesheet_iterator_save"
+
 
 @NodeFactory.register(IMAGE_ITERATOR_NODE_ID)
 class ImageFileIteratorLoadImageNode(NodeBase):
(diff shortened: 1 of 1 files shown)


RunDevelopment (changes requested):

RunDevelopment:
Also, it doesn't work. I tried it on this chain and got this error:

image

image

joeyballentine:
> Also, it doesn't work. I tried it on this chain and got this error:
> 
> image
> 
> image

oh, to fix the unused variable thing i added a _ to percent not realized it was a named arg

RunDevelopment:
Please don't use closing keywords if you don't intend to close the linked issue. #564 is not resolved with this PR. The nodes users (e.g. I) wanted must return and consume iterators.

So, as is, this isn't really useful to me. So what is the use case for this? What does this enable that wasn't possible before?

> This doesn't entirely solve the use case mentioned in discord

.... Uhm, it doesn't solve it all... 
The user wanted to have a node that takes any number of images and creates a sprite sheet. That's not possible with this.

joeyballentine:
I know that it doesn't solve that particular issue, but I still thought it might be useful to be able to upscale individual items in an existing sprite sheet, for example if you want to add padding to each item before an upscale or add a mirrored border or something to each individual one, then put it back into the sprite sheet after cropping. Obviously, the intended thing isn't possible -- but it actually could be with a different iterator that takes in a directory of files and does the same kind of output thing I did for this one. Wouldn't be too difficult to add once this is in

RunDevelopment:
> but it actually could be with a different iterator that takes in a directory of files and does the same kind of output thing I did for this one.

I'm gonna stop you right there. Do you really want to add x-many highly specific iterators when we want to get rid of them soon anyway?

I say we merge this as is now and leave the issue open. It's not exactly what I wanted, but it's a step in the right direction, and it will probably be useful to some people until we transition to the new iterators.

RunDevelopment (approved):

PR #577, opened 2022-07-18
Fix iterator error when input not connected

joeyballentine:


joeyballentine:
> Instead of doing that, couldn't you just mark these helper nodes as having side effects? From the perspective of the sub flow, these helper nodes advance the iteration, so they have side effects.

I could do that yes, but I feel like this also just makes it safer in general, like if someone was somehow able to delete an iteratorHelper we shouldn't just assume they're there. But I'll update the PR

RunDevelopment:
Could you also fix the type error I previously ignored? Just search for all `TODO: Fix None key` comments.

RunDevelopment:
> wouldn't the fix for this be the thing you told me to remove?

I was more thinking of `assert`s. Them being None is an invalid program state, so it should fail instead of silently doing something wrong.

PR #579, opened 2022-07-18
Fixed color conversion for non-RGB formats

### LINKED ISSUE #571: Change colorspace returns HSV and YUV in the wrong order
Right now, RGB -> HSV will return an RGB image where (R,G,B)=(V,S,H) and RGB -> YUV will return (R,G,B)=(V,U,Y). This is because OpenCV uses BGR under the hood.

We should probably fix the order of for HSV and YUV, even though this is a breaking change.

### LINKED ISSUE #572: Change Colorspace RGB->HSV returns non-normalized hue values
The hue value is the range 0 to 360 (degrees), but pretty much all our nodes expect normalized (0-1) values. This makes working with HSV very difficult, if not impossible.

RunDevelopment:
Fixes #572. Fixes #571.

As it turns out, HSV was completely broken. Somehow, the computed H was **clipped** from 0-360 to 0-1. 

Changes:
- Normalize HSV hue value to 0-1.
- Reorder channels for HSV and YUV to match RGB.

This is a breaking change because of the channel reordering. I'm debating whether it is worth the effort to write a migration for this. We could, depending on the conversion mode, append or prepend nodes to switch channels around, which would prevent any breakage. However, that's 3 new nodes per Change Colorspace node, if the node convert to or from HSV or YUV.

### DIFF
--- a/backend/src/nodes/image_util_nodes.py
+++ b/backend/src/nodes/image_util_nodes.py
@@ -284,7 +284,27 @@ class ColorConvertNode(NodeBase):
     def run(self, img: np.ndarray, color_mode: int) -> np.ndarray:
         """Takes an image and changes the color mode it"""
 
-        result = cv2.cvtColor(img, int(color_mode))
+        def reverse3(image: np.ndarray) -> np.ndarray:
+            c = get_h_w_c(image)[2]
+            assert c == 3, "Expected a 3-channel image"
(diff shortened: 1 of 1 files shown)


joeyballentine:
I don't think a migration is worth it. Whatever is currently being done is a bug that we're fixing in my eyes. Also HSV was just completely broken so I doubt anyone was even using it

joeyballentine (approved):

PR #588, opened 2022-07-19
Interp number outputs

joeyballentine:
image

### DIFF
--- a/backend/src/nodes/ncnn_nodes.py
+++ b/backend/src/nodes/ncnn_nodes.py
@@ -254,7 +254,11 @@ class NcnnInterpolateModelsNode(NodeBase):
                 ends=("A", "B"),
             ),
         ]
-        self.outputs = [NcnnNetOutput()]
+        self.outputs = [
+            NcnnNetOutput(),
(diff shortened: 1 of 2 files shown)


backend/src/nodes/ncnn_nodes.py
```diff
@@ -312,7 +316,11 @@ def run(
                 "These NCNN models are not compatible and not able to be interpolated together"
             )
 
-        interp_bin_data = self.perform_interp(bin_data_a, bin_data_b, int(amount))
+        interp_bin_data = self.perform_interp(bin_data_a, bin_data_b, amount)
 
         # Put all this info with the net and disguise it as just the net
-        return [(param_path_a, interp_bin_data, input_name_a, output_name_a)]
+        return (
+            (param_path_a, interp_bin_data, input_name_a, output_name_a),
```
RunDevelopment, line 323: Unrelated to the PR: Shouldn't that tuple be its own class?

joeyballentine: Probably

RunDevelopment: Then I'll make a PR for that.

PR #591, opened 2022-07-19
Slight reformatting of #587

RunDevelopment:
I know this is super nitpicky, but...


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -927,16 +927,10 @@ export const GlobalProvider = memo(
         }, [changeNodes, changeEdges]);
         const selectAllFn = useCallback(() => {
             changeNodes((nodes) => {
-                return nodes.map((n) => ({
-                    ...n,
-                    selected: true,
-                }));
+                return nodes.map((n) => ({ ...n, selected: true }));
```
joeyballentine, line 930: Could also just get rid of the brackets and the return

RunDevelopment: Done.

joeyballentine (approved):

PR #595, opened 2022-07-20
Replace preview PNG encoding with WEBP

joeyballentine:
Smaller file size, still lossless. Could mean we could increase the preview resolution as well

The downside of course is it takes longer to encode, so I'm not 100% sure this is worth it

### DIFF
--- a/backend/src/nodes/utils/image_utils.py
+++ b/backend/src/nodes/utils/image_utils.py
@@ -247,6 +247,6 @@ def preview_encode(img: np.ndarray, target_size: int = 512) -> str:
         f = max(w / target_size, h / target_size)
         img = cv2.resize(img, (int(w / f), int(h / f)), interpolation=cv2.INTER_AREA)
 
-    _, encoded_img = cv2.imencode(".png", (img * 255).astype("uint8"))  # type: ignore
+    _, encoded_img = cv2.imencode(".webp", (img * 255).astype("uint8"))  # type: ignore
     base64_img = base64.b64encode(encoded_img).decode("utf8")
-    return f"data:image/png;base64,{base64_img}"
+    return f"data:image/webp;base64,{base64_img}"


joeyballentine:
This does remind me of that discussion we had about cv2 using "magic" with the default png settings which makes it super fast. maybe it really is worth it to just leave it as png

RunDevelopment:
> Smaller file size

About [20-30% smaller on average](https://developers.google.com/speed/webp/docs/webp_lossless_alpha_study#results).

> it takes longer to encode

How much longer? As you said, opencv2's default PNG is pretty fast. (That's because it skips a few PNG filters that likely don't yield much.) So it would be nice to have a comparison between cv2's PNG and WEBP compression speed.

RunDevelopment:
So we have a nicely symmetrical space-time tradeoff. So what's more important: going fast, or saving memory?

Couldn't we offload the task of encoding and sending the image to another thread? Broadcasts should be able to be sent in any order, so why block the executor? Then the increased encoding time wouldn't matter, and we would only benefit.

joeyballentine:
Yeah we could do that

RunDevelopment:
What are we going to do about this? If we are going to make it so that broadcast data is calculated in another thread, then let's merge this.

joeyballentine:
> What are we going to do about this? If we are going to make it so that broadcast data is calculated in another thread, then let's merge this.

I'd love to do that but I'm not entirely sure how. Threading in python is confusing. I wouldn't want to merge this without first confirming it's possible to do this stuff in another thread while also preserving the SSE queue and everything.

RunDevelopment:
It seems that numpy will often release the GIL, so parallelism is possible. Furthermore, it seems like **every** cv2 function releases the GIL. I suspect that `base64.b64encode` also release the GIL, so every expensive function in `preview_encode` can run in parallel to the rest of the program.

References:
- https://towardsdatascience.com/demystifying-python-multiprocessing-and-multithreading-9b62f9875a27
- https://stackoverflow.com/questions/50014026/is-the-gil-released-while-using-multithreading-with-python-opencv
- https://answers.opencv.org/question/182036/how-does-the-gil-release-happen-for-drawing-functions-exposed-to-python/
- https://news.ycombinator.com/item?id=25406030

joeyballentine:
I was moreso worried about the fact that we push this data to a queue that sends SSE stuff. We would need to process that data and send it in another thread. That requires passing the queue down. Python typically doesn't like that in my experience, at least when working with processes

joeyballentine:
We might be able to leverage some stuff from sanic for this. Like background tasks: https://sanic.dev/en/guide/basics/tasks.html#creating-tasks

Idk if those run in another thread or not though

RunDevelopment:
I don't think this will be a problem. I can see why other **processes** cause a problem (how do you serialize an event loop?), but I wanted to suggest multithreading, so everything will stay in the same process.

RunDevelopment:
Multiprocessing isn't viable for us anyway. Sending a serialized copy of an image to another process takes long and consume a lot of memory.

joeyballentine:
Ok. I just didn't know if multithreading would cause similar issues as multiprocessing. If it's going to be fine, then let's do it. I'll merge this. Do you want to attempt the threading part or should I?

RunDevelopment:
I can try.

PR #611, opened 2022-07-22
Add run/pause/kill shortcuts

### LINKED ISSUE #608: Keyboard shortcuts for running, pausing, stopping the chain.

joeyballentine:
Closes #608

### DIFF
--- a/src/renderer/components/Header.tsx
+++ b/src/renderer/components/Header.tsx
@@ -68,25 +68,23 @@ export const Header = memo(() => {
                         px={2}
                         py={1}
                     >
-                        <span>
-                            <IconButton
-                                aria-label="Start button"
-                                colorScheme="green"
-                                disabled={
-                                    !(
-                                        status === ExecutionStatus.READY ||
-                                        status === ExecutionStatus.PAUSED
-                                    )
-                                }
-                                icon={<IoPlay />}
-                                size="md"
-                                variant="outline"
-                                onClick={() => {
-                                    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-                                    run();
-                                }}
-                            />
-                        </span>
+                        <IconButton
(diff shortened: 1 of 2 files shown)


src/renderer/components/Header.tsx
```diff
@@ -90,7 +90,7 @@ export const Header = memo(() => {
                         closeOnClick
                         closeOnMouseDown
                         borderRadius={8}
-                        label="Pause"
+                        label="Pause (F5)"
```
RunDevelopment, line 93: I don't think "Pause" also using F5 is a good idea. Users might accidentally double tab the button. It might also be that chainner takes a while to start running a chain, so users might get impatient and press the button again.

src/renderer/components/Header.tsx
```diff
@@ -111,7 +111,7 @@ export const Header = memo(() => {
                         closeOnClick
                         closeOnMouseDown
                         borderRadius={8}
-                        label="Stop"
+                        label="Stop (Shift+F5)"
```
RunDevelopment, line 114: I don't like that I have to use two hands to use the shortcut for stopping a chain. On my keyword, the only key further away from the two controls is F6.

I think Run, Pause, and Stop should all have dedicated keys.

RunDevelopment (changes requested):

src/renderer/components/Header.tsx
joeyballentine: I took that from vscode but I can make them different

joeyballentine: Did you have specific shortcuts in mind?

RunDevelopment: Honestly, no, not really. Maybe just F5, F6, F7?

RunDevelopment (approved):

PR #614, opened 2022-07-22
Image preview outputs

joeyballentine:
Uses the new output broadcasting system to preview a new output type: LargeImageOutput. This completely replaces the ImagePreview component and allows a new node, _View Image_ to also show image outputs. There is also a new output component called DefaultImageOutput, which has a clickable button to create a new view node.

This also comes with a new bonus: Any node with no inputs with handles will automatically run once it is valid, and will re-run whenever their inputData changes. This means that certain nodes, like NCNN's _Load Model_ node, which were not previously able to be typed, are now typed correctly (and cached in the backend). This only applies to text outputs and directory outputs (as well as image outputs of course).

### Screenshots:

image
image


backend/src/nodes/utils/image_utils.py
```diff
@@ -247,6 +247,6 @@ def preview_encode(img: np.ndarray, target_size: int = 512) -> str:
         f = max(w / target_size, h / target_size)
         img = cv2.resize(img, (int(w / f), int(h / f)), interpolation=cv2.INTER_AREA)
 
-    _, encoded_img = cv2.imencode(".png", (img * 255).astype("uint8"))  # type: ignore
+    _, encoded_img = cv2.imencode(".webp", (img * 255).astype("uint8"))  # type: ignore
```
joeyballentine, line 250: We should decide on the webp PR before merging this

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -228,6 +236,23 @@ export const GlobalProvider = memo(
...
+        //     }, 500);
+        //     return () => clearTimeout(timerId);
+        // }, [outputDataMap]);
+        // useEffect(() => {
+        //     const cachedOutputData = getSessionStorageOrDefault<ReadonlyMap<string, OutputData>>(
+        //         'cachedOutputData',
+        //         EMPTY_MAP
+        //     );
+        //     setOutputDataMap(new Map([...cachedOutputData]));
+        // }, [setOutputDataMap]);
```
joeyballentine, line 254: Would it be a good idea to cache the outputData on refresh?

RunDevelopment: IMO, no.

backend/src/nodes/utils/image_utils.py
RunDevelopment: These feature as separate, so this PR should not include them. The difference between PNG and WEBP isn't that big anyway, so it's not like this change is required for this PR to work.

joeyballentine: Right. I cherry picked this commit to make that PR and I just didn't remove it after. I figured we'd have merged it by now

src/renderer/components/node/NodeOutputs.tsx
```diff
@@ -1,17 +1,83 @@
...
+import { DefaultImageOutput } from '../outputs/DefaultImageOutput';
+import { DirectoryOutput } from '../outputs/DirectoryOutput';
 import { GenericOutput } from '../outputs/GenericOutput';
+import { LargeImageOutput } from '../outputs/LargeImageOutput';
+import { OutputContainer } from '../outputs/OutputContainer';
+import { TextOutput } from '../outputs/TextOutput';
 
-interface NodeOutputsProps {
-    id: string;
+interface FullOutputProps extends Omit<Omit<Output, 'type'>, 'id'>, OutputProps {
```
RunDevelopment, line 18:
```suggestion
interface FullOutputProps extends Omit<Output, 'id' | 'type'>, OutputProps {
```

src/renderer/components/node/NodeOutputs.tsx
```diff
@@ -1,17 +1,83 @@
...
+                    definitionType={props.definitionType}
+                    hasHandle={props.hasHandle}
+                    id={props.id}
+                    key={`${props.id}-${props.outputId}`}
+                    label={props.label}
+                    outputId={props.outputId}
+                >
+                    <GenericOutput {...props} />
+                </OutputContainer>
+            );
```
RunDevelopment, line 54: I don't think you need to treat this as anything special.

```suggestion
            OutputType = GenericOutput;
            break;
```

RunDevelopment (changes requested):
I'm not done reviewing, but here's what I have so far.

src/renderer/components/outputs/LargeImageOutput.tsx
```diff
@@ -0,0 +1,176 @@
...
+        useEffect(() => {
+            if (value) {
+                setStale(false);
+            }
+        }, [value]);
+
+        useEffect(() => {
+            if (value && lastInputDataUpdatedId !== id) {
+                setStale(true);
+            }
```
RunDevelopment, line 48: This just doesn't work. If you create a simple chain with 2 Load Image nodes, and set an image for both of them, one will always be marked stale.

src/renderer/components/outputs/LargeImageOutput.tsx
```diff
@@ -0,0 +1,176 @@
...
+                            w="4rem"
+                            zIndex="99"
+                        >
+                            <WarningIcon
+                                boxSize={2}
+                                color={fontColor}
+                            />
+                            <Text
+                                color={fontColor}
+                                fontSize="xx-small"
```
RunDevelopment, line 119: For how important this information is, the font is mighty small.
image

src/renderer/components/node/Node.tsx
```diff
@@ -147,6 +155,68 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
...
+                                ...n.data,
+                                animated: true,
+                            },
+                        };
+                    }
+                    return n;
+                })
+            );
+
+            const result = await backend.runIndividual({
```
RunDevelopment, line 186: When changing whether a node is disabled or locked, the node will re-runs.

backend/src/process.py
```diff
@@ -181,7 +181,8 @@ async def __process(self, node: UsableData) -> Any:
             output = await self.loop.run_in_executor(None, run_func)
             node_outputs = node_instance.get_outputs()
             broadcast_data: Dict[int, Any] = dict()
-            if len(node_outputs) > 0:
+            # Only broadcast the output if the node has outputs and the output is not cached
+            if len(node_outputs) > 0 and self.output_cache.get(node_id, None) is None:
```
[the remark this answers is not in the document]
joeyballentine: This is so it doesn't broadcast when you run the chain. Without it, it gets from the cache and the broadcasts it again even though it doesn't need to, so you see the node flicker for a short while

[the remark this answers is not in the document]
RunDevelopment: Then that's a frontend bug. The backend should be allowed to repeat broadcasts for nodes. Child nodes have problems otherwise.

I would suggest doing a deep equal on received broadcast data to figure out whether the broadcast data map needs to be updated at all.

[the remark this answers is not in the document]
joeyballentine: How would child nodes have problems?

src/renderer/components/node/Node.tsx
```diff
@@ -147,6 +155,68 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
...
     const shadowColor = useColorModeValue('gray.600', 'gray.900');
 
+    const inputDataValues = useMemo(() => Object.values(inputData), [inputData]);
+
+    const [shouldRun, setShouldRun] = useState(false);
+    useEffect(() => {
+        setShouldRun(
+            validity.isValid &&
+                !inputs.some((i) => i.hasHandle) &&
+                inputDataValues.length >= inputs.filter((i) => !i.optional).length
```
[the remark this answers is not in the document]
joeyballentine: This determines if a node should make a run/individual request. The conditions are:
1) it has no inputs that have handles (meaning they don't require input from other nodes, we can just send all the inputData we have)
2) the input data we have covers all non-optional inputs

I'm not really sure why I did that second one, to be honest. It might have just been something I was trying when I had an unrelated bug and I just never removed it. But theoretically it wouldn't be necessary to keep that since the validity checker already checks this, right?

src/renderer/components/outputs/LargeImageOutput.tsx
joeyballentine: Oh yeah...... Got any ideas for how to do that better?

src/renderer/components/outputs/LargeImageOutput.tsx
joeyballentine: It's also misaligned for you for some reason

backend/src/nodes/image_nodes.py
```diff
@@ -197,3 +197,23 @@ def run(self, img: np.ndarray):
...
+class ImViewNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "See an inline preview of the image in the editor."
+        self.inputs = [ImageInput()]
+        self.outputs = [
+            LargeImageOutput("Preview", image_type="Input0", has_handle=False)
+        ]
+        self.category = IMAGE
+        self.name = "View Image"
```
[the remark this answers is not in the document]
joeyballentine: I was considering renaming preview image to "view image (external)", like it used to be before we removed the opencv preview

backend/src/process.py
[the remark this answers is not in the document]
joeyballentine: The backend doesn't cache anything in the main cache for iterator children, if that's what you mean. Those get treated as an entirely new executor, remember?

I don't disagree I should do a deep equals check when setting stuff though, but I also think we shouldn't be wasting IPC time sending data we don't need to send.

src/renderer/components/outputs/LargeImageOutput.tsx
joeyballentine: Ok, what i've done is just not show the stale indicator when it's a starting node. I think that's good enough

src/renderer/components/node/Node.tsx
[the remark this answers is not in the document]
joeyballentine: Ah, I know what this was for now: When you clear a node, it detects that an input data change happened before it recomputes validity, and therefore it tries to run it again and errors

src/common/common-types.ts
```diff
@@ -58,6 +61,8 @@ export interface Output {
      */
     readonly neverReason?: string | null;
     readonly label: string;
+    readonly kind: OutputKind;
+    readonly hasHandle: boolean;
```
RunDevelopment, line 65: The context pane now needs to check whether outputs have handles to connect to. It needs to be added to the `getFirstPossibleOutput`, I think. Please add that check.

backend/src/run.py
```diff
@@ -190,11 +193,45 @@ async def run_individual(request: Request):
         logger.info(f"Using device: {os.environ['device']}")
         # Create node based on given category/name information
         node_instance = NodeFactory.create_node(full_data["schemaId"])
+
+        # Enforce that all inputs match the expected input schema
+        enforced_inputs = []
```
RunDevelopment, line 198: I don't like that all this executor code is now here. We should probably make a follow up PR adding model outputs. Then we could get rid of this, because we wouldn't need `get_extra_data` anymore.

RunDevelopment: On that note: Is the `get_extra_data` in Load Image even used anymore?

src/renderer/components/node/Node.tsx
```diff
@@ -147,6 +158,70 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
     const bgColor = useColorModeValue('gray.400', 'gray.750');
     const shadowColor = useColorModeValue('gray.600', 'gray.900');
 
+    const inputDataValues = useMemo(() => Object.values(inputData), [inputData]);
+
+    const [shouldRun, setShouldRun] = useState(false);
+    useEffect(() => {
+        setShouldRun(
+            !isLocked &&
+                !disabled.isDirectlyDisabled &&
```
RunDevelopment, line 167: This didn't fix the disabled bug I reported. Try repeated disabling and enabling a Load Image node.

src/renderer/components/outputs/DirectoryOutput.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+        const { setManualOutputType, schemata } = useContext(GlobalContext);
+
+        const schema = schemata.get(schemaId);
+
+        const value = useOutputData(outputId) as string;
+
+        useEffect(() => {
+            // Run this only if this is a "starting" node
+            if (!schema.inputs.some((i) => i.hasHandle)) {
```
RunDevelopment, line 24: I don't like this definition of a starting node. It's very implicit. Could you please make a function `isStartingNode(schema: NodeSchema)` (maybe somewhere in util, idk)? That way, you wouldn't need the comment either.

src/renderer/components/outputs/GenericOutput.tsx
```diff
@@ -1,62 +1,43 @@
...
+        c.typeState.functions.get(id)?.outputs.get(outputId)
+    );
 
-export const GenericOutput = memo(
-    ({ label, id, outputId, definitionType, useOutputData }: GenericOutputProps) => {
-        const type = useContextSelector(GlobalVolatileContext, (c) =>
-            c.typeState.functions.get(id)?.outputs.get(outputId)
-        );
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const value = useOutputData(outputId);
```
RunDevelopment, line 14: Pretty sure you can remove that now.

src/renderer/components/outputs/DirectoryOutput.tsx
RunDevelopment: And please use the function everywhere, obviously.

src/renderer/components/outputs/OutputContainer.tsx
```diff
@@ -1,20 +1,23 @@
...
 import { noContextMenu } from '../../hooks/useContextMenu';
+import { TypeTag } from '../TypeTag';
 
 interface OutputContainerProps {
     hasHandle: boolean;
     outputId: OutputId;
     id: string;
     definitionType: Type;
+    label: string;
+    generic: boolean;
```
RunDevelopment, line 20: What does "generic" even mean here?

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -220,10 +221,13 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
         'node-output-data',
         (data) => {
             if (data) {
-                setOutputDataMap((prev) => new Map([...prev, [data.nodeId, data.data]]));
+                const existingData = outputDataMap.get(data.nodeId);
+                if (!existingData || !isJsonEqual(existingData, data.data)) {
+                    setOutputDataMap((prev) => new Map([...prev, [data.nodeId, data.data]]));
+                }
             }
```
RunDevelopment, line 228: To not read stale data (`outputDataMap`), move all of this into `setOutputDataMap`. React won't update a `useState` if the value return by the set-supplier function is that same as the input value.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -220,10 +221,13 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
         },
-        [setOutputDataMap]
+        [unAnimate, outputDataMap, setOutputDataMap]
```
RunDevelopment, line 230: These dependencies seem weird.

src/common/common-types.ts
joeyballentine: Oh yeah, forgot about that.

backend/src/run.py
joeyballentine: I've already finished the model outputs stuff locally. Just waiting for this to be finished before I make that PR so I can rebase. And no, that get extra data is no longer used. I'll remove it.

src/renderer/components/outputs/OutputContainer.tsx
joeyballentine: It's basically whether or not to put the label/type props next to the handle vs above the input

src/renderer/contexts/ExecutionContext.tsx
joeyballentine: Yeah this is kinda a problem -- without putting something here that forces it to update (which is what I used unAnimate for here), it doesn't properly update. I have no idea why, but using outputDataMap doesnt just work.

src/renderer/components/node/Node.tsx
```diff
@@ -71,7 +80,10 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
         [selected, accentColor, regularBorderColor]
     );
 
-    const [validity, setValidity] = useState(VALID);
+    const [validity, setValidity] = useState({
+        isValid: false,
+        reason: 'Validating nodes...',
+    } as Validity);
```
RunDevelopment, line 86: I might have mentioned this before already, but it bares repeating: `as` should be a last resort.

```suggestion
    const [validity, setValidity] = useState<Validity>({
        isValid: false,
        reason: 'Validating nodes...',
    });
```

src/renderer/components/outputs/TextOutput.tsx
```diff
@@ -0,0 +1,58 @@
+import { Center, Flex, Spacer, Text } from '@chakra-ui/react';
+import { memo, useEffect } from 'react';
+import { useContext, useContextSelector } from 'use-context-selector';
+import { StringLiteralType } from '../../../common/types/types';
+import { isStartingNode } from '../../../common/util';
+import { GlobalContext, GlobalVolatileContext } from '../../contexts/GlobalNodeState';
+import { TypeTag } from '../TypeTag';
+import { OutputProps } from './props';
+
+export const TextOutput = memo(({ label, id, outputId, schemaId, useOutputData }: OutputProps) => {
```
RunDevelopment, line 10: What's the reason for this component? The only difference between TextOutput, DirectoryOutput, and GenericOutput is how they handle broadcast data. They are identical otherwise.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -220,10 +216,16 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+                    const existingData = prev.get(data.nodeId);
+                    if (!existingData || !isDeepEqual(existingData, data.data)) {
+                        return new Map([...prev, [data.nodeId, data.data]]);
+                    }
+                    return prev;
+                });
             }
         },
-        [setOutputDataMap]
+        [[...outputDataMap], setOutputDataMap]
```
RunDevelopment, line 228: ..... This is a deps array.

RunDevelopment: Also, why do you still depend on `outputDataMap` at all?

src/renderer/components/node/Node.tsx
```diff
@@ -147,6 +159,72 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
...
+    if (!isDeepEqual(inputDataRef.current, inputData)) {
+        inputDataRef.current = inputData;
+    }
+    const [shouldRun, setShouldRun] = useState(false);
+    useEffect(() => {
+        setShouldRun(
+            !isLocked &&
+                !disabled.isDirectlyDisabled &&
+                validity.isValid &&
+                !inputs.some((i) => i.hasHandle) &&
```
RunDevelopment, line 172: isStartingNode

RunDevelopment (changes requested):
I think we are pretty much there. I just have a few more suggestions and nits.

src/renderer/components/outputs/OutputContainer.tsx
RunDevelopment: Huh, the input container does the same thing. Strange, but alright.

src/renderer/contexts/ExecutionContext.tsx
joeyballentine: Try it yourself without doing this. It doesn't work.

src/renderer/components/outputs/TextOutput.tsx
joeyballentine: Exactly

src/renderer/contexts/ExecutionContext.tsx
RunDevelopment: You are creating a new object with render. So this hook constantly updates.

joeyballentine: Oh

joeyballentine: Well something is wrong with this regardless

src/renderer/components/outputs/TextOutput.tsx
RunDevelopment: "Exactly" what lol 

You just duplicated a bunch of code everywhere. Why didn't you just give GenericOutput a prop like `convertOutputToType?: (value: unknown) => Type | undefined`. Or to be more compatible with the whole pick component thingy, you could also have made a higher kind component that is generic over the prop I suggested. So e.g.
```ts
const OutputComponent = (convertOutputToType?: (value: unknown) => Type | undefined) => {
  return memo(({ ... }: OutputProps) => { 
    // code of GenericOutput, but handling broadcasts like TextOutput with convertOutputToType 
  })
}

const TextOutput = OutputComponent(toStringType);
const DirectoryOutput = OutputComponent(toDirectoryType);
const GenericOutput = OutputComponent();
```

joeyballentine: I like that idea. I'll do that.

joeyballentine: Ended up not doing it that way because I felt like it was too complicated and we already pass the kind into the full props, so might as well just handle it in the type setter function itself

src/renderer/components/node/NodeOutputs.tsx
```diff
@@ -1,17 +1,70 @@
...
+            break;
+        case 'directory':
+            OutputType = GenericOutput;
+            break;
+        case 'text':
+            OutputType = GenericOutput;
+            break;
+        case 'generic':
+            OutputType = GenericOutput;
+            break;
```
RunDevelopment, line 41:
```suggestion
        case 'directory':
        case 'text':
        case 'generic':
            OutputType = GenericOutput;
            break;
```

src/renderer/components/outputs/GenericOutput.tsx
```diff
@@ -1,62 +1,74 @@
...
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const { setManualOutputType, schemata } = useContext(GlobalContext);
+
+        const schema = schemata.get(schemaId);
+
         const value = useOutputData(outputId);
 
+        useEffect(() => {
+            if (isStartingNode(schema)) {
+                if (value) {
```
RunDevelopment, line 25: The value might an empty string.

```suggestion
                if (value !== undefined) {
```

src/renderer/contexts/ExecutionContext.tsx
RunDevelopment: You what? Fuck it. I can't even fix it using my usual invisible `div` hack.

Just make the deps array `[{}, setOutputDataMap]`. This should make it obvious that something is wrong.

RunDevelopment (approved):

PR #617, opened 2022-07-23
Added support for input conversions

RunDevelopment:
image

### DIFF
--- a/backend/src/nodes/properties/inputs/base_input.py
+++ b/backend/src/nodes/properties/inputs/base_input.py
@@ -22,6 +22,7 @@ class BaseInput:
         has_handle=True,
     ):
         self.input_type = input_type
+        self.input_conversion: Union[expression.ExpressionJson, None] = None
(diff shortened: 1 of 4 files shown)


joeyballentine:
Do you think it would be a good idea to have some sort of indicator that the data is being converted? I just thought of that today. I still think it's much better to convert floats to ints, but it might be a good idea to like "soft" warn them that it's happening. Like doing a gray (!) validity with a message saying that the node doesn't support decimal numbers but it has been converted by rounding to the nearest whole number, or something like that. Like an eslint warning rather than an error. 

Idk, it's not necessary or anything. Just an idea.

RunDevelopment:
While I do like the idea, it seems very narrow at the moment. Input conversions are way more general than just rounding numbers, so generating good warning will be non-trivial. I also don't know whether we should generate warning at all.

In any case, these warning should be its own PR (if at all).

joeyballentine (approved):

PR #619, opened 2022-07-23
Sanitize process.env for spawned processes to remove external PYTHONHOME

### LINKED ISSUE #583: PYTHONHOME env var appears to mess with integrated python
This was the cause of #543. This probably isn't fixable by us, but if we can somehow suppress that env var maybe it would be

joeyballentine:
Fixes #583, I think. Can't really test it myself.

### DIFF
--- a/src/common/env.ts
+++ b/src/common/env.ts
@@ -4,3 +4,7 @@ export const isMac = process.platform === 'darwin';
 export const isM1 = isMac && (os.cpus()[0]?.model.includes('Apple M1') ?? false);
 
 export const isRenderer = typeof process !== 'undefined' && process.type === 'renderer';
+
+const env = { ...process.env };
+delete env.PYTHONHOME;
+export const sanitizedEnv = env;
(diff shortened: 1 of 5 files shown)


src/common/env.ts
```diff
@@ -4,3 +4,7 @@ export const isMac = process.platform === 'darwin';
 export const isM1 = isMac && (os.cpus()[0]?.model.includes('Apple M1') ?? false);
 
 export const isRenderer = typeof process !== 'undefined' && process.type === 'renderer';
+
+const env = { ...process.env };
+delete env.PYTHONHOME;
+export const sanitizedEnv = env;
```
RunDevelopment, line 10: Does this actually work? I though envs were inherited System > User > Program, so doing deleting it from the program envs wouldn't do anything.

joeyballentine: It only matters for the spawned process, which you can control the env vars for. 

https://nodejs.org/api/child_process.html#:~:text=the%20child%20process.-,env%20%3CObject%3E%20Environment%20key%2Dvalue%20pairs.%20Default%3A%20process.env.,-argv0%20%3Cstring%3E

RunDevelopment (approved):

PR #620, opened 2022-07-24
Use session storage for save path

### LINKED ISSUE #586: ChaiNNer loses name of current file on reload
~If you hit a runtime error while running a chain in a chain that you have already saved, the filename in the header will revert to unnamed and Save will require you to Save As, because the association between the chain and the file gets broken.~

Using View -> Reload causes chain to lose association to its filename, the header reverts to "Untitled", and Save As is required to save chain again.

joeyballentine:
Fixes #586


RunDevelopment:
How will this affect the saved-change status? If I make changes to a file and reload, will Chainner forget that I made changes? Before this PR, this wasn't a problem, but it might be now.

joeyballentine:
> How will this affect the saved-change status? If I make changes to a file and reload, will Chainner forget that I made changes? Before this PR, this wasn't a problem, but it might be now.

it's actually the opposite. If you have no changes and reload, it will think you have changes. similarly, if you have changes it will still think you have changes, but that's probably just due to it always thinking you have unsaved changes on reload.

I guess this is fine?

src/renderer/hooks/useSessionStorage.ts
```diff
@@ -1,11 +1,19 @@
...
+    }
+};
+
 export const getSessionStorageOrDefault = <T>(key: string, defaultValue: T): T => {
     const stored = sessionStorage.getItem(key);
     if (!stored) {
         return defaultValue;
     }
-    return JSON.parse(stored) as T;
+    return safeJsonParse(stored) as T;
```
RunDevelopment, line 16: How could there ever be invalid JSON in there?

joeyballentine: the string "undefined" apparently does not like being json parsed

joeyballentine: remove this and you'll crash when you reload (without having a file opened/saved)

RunDevelopment: Well, that's because you are using `useSessionStorage` incorrectly. This hook only supports JSON values, and `undefined` is not a JSON value. Use `null` instead.

joeyballentine: I still think it would be good to wrap the json parse around this. If you ever somehow get an invalid state in your session, it'll just continue to crash without a way to fix it (which is very bad)

RunDevelopment: With this "fix", the JSON parse won't crash anymore, yes. Now the first consumer of the returned value will crash because `undefined` is guaranteed to not be returned by the API. With this fix, the API lies.

RunDevelopment: Same for `null`. The consumer controls the `T`, not `useSession`.

RunDevelopment: @joeyballentine

joeyballentine: I'll get rid of it then

PR #621, opened 2022-07-24
Don't highlight AMD GPUs red

### LINKED ISSUE #556: AMD GPUs still get highlighted in red in the dependency manager
This is a relic from the era when pytorch was the only option. Now, pytorch and onnx should be highlighted red instead of the gpu, to let users know they should use ncnn

joeyballentine:
Fixes #556 

I think there maybe should also be some kind of description for each dependency that explains what it is and why it would be needed, and shows stuff like "Recommended for Nvidia users or for those who want to convert pytorch models (.pth) to other formats" for pytorch and "Recommended for AMD users" for ncnn, and maybe gives links to their main pages or something. But that can be a separate ticket

### DIFF
--- a/src/renderer/contexts/DependencyContext.tsx
+++ b/src/renderer/contexts/DependencyContext.tsx
@@ -237,7 +237,6 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
                                     w="full"
                                 >
                                     <Text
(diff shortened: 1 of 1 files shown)


RunDevelopment (approved):

RunDevelopment:
Regarding the AMD UX, you might want to make an issue TODO list with everything you just mentioned. Then we can slowly work towards making the AMD experience as good as the Nvidia one.

PR #622, opened 2022-07-24
Disable items from menu when alert is open

### LINKED ISSUE #229: Disable electron menu and shortcuts while an alert is open
Right now, our custom alert have the problem that they can be undermined. E.g. the "Discard unsaved changes?" alert can be undermined by just opening a new file (see #219) or saving the current file. The same is also true for lots of other alerts.

The basic problem is that the user can still do other inputs by using the menu or shortcuts. This essentially bypasses the input lock imposed by the alert.

joeyballentine:
Closes #229

### DIFF
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -250,3 +250,131 @@ export const setMainMenu = ({ mainWindow, openRecentRev = [] }: MainMenuArgs) =>
     const menu = Menu.buildFromTemplate(template);
     Menu.setApplicationMenu(menu);
 };
+
+export const setEmptyMenu = () => {
+    const template = [
+        ...(isMac ? [{ role: 'appMenu' }] : []),
+        {
+            label: 'File',
+            submenu: [
+                {
+                    label: 'New',
+                    enabled: false,
+                },
+                {
+                    label: 'Open...',
+                    enabled: false,
+                },
+                {
+                    label: 'Open Recent',
+                    enabled: false,
+                },
+                { type: 'separator' },
+                {
+                    label: 'Save',
+                    enabled: false,
+                },
+                {
+                    label: 'Save As...',
+                    enabled: false,
+                },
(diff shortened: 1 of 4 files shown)


src/main/menu.ts
```diff
@@ -250,3 +250,131 @@ export const setMainMenu = ({ mainWindow, openRecentRev = [] }: MainMenuArgs) =>
     const menu = Menu.buildFromTemplate(template);
     Menu.setApplicationMenu(menu);
 };
+
+export const setEmptyMenu = () => {
```
RunDevelopment, line 254: I hate it....

You didn't disable the menu, you made a new menu that is disabled. Now we have to maintain 2 menus. If we change the main menu, we now also have to remember to update the second disabled menu. That will be fun...

Instead, could you please add a `enabled: boolean = true` property to the arguments object? Then you'll only have to add the property `enabled` for all menu items you want to disable.

src/renderer/contexts/AlertBoxContext.tsx
```diff
@@ -161,6 +162,7 @@ export const AlertBoxProvider = memo(({ children }: React.PropsWithChildren<unkn
     );
     const sendAlert = useCallback<AlertBox['sendAlert']>(
         (type: AlertType | AlertOptions, title?: string | null, message?: string) => {
+            ipcRenderer.send('disable-menu');
```
RunDevelopment, line 165: I would move this down to the effect where the alert is actually opened.

src/main/menu.ts
joeyballentine: Oh yeah, duh. For a while I was trying to dynamically get the current menu and map the items to make them all disabled, but that didn't work at all, so I just thought I needed to make a new menu. Didn't think to just add the enabled stuff to the original lol

RunDevelopment (approved):

PR #624, opened 2022-07-24
Torch preview outputs

joeyballentine:


RunDevelopment:
This removed less code than I expected. Everything related to `get_extra_data` is now unnecessary and can be removed.

joeyballentine:
> This removed less code than I expected. Everything related to `get_extra_data` is now unnecessary and can be removed.

oh yeah, forgot about that

backend/src/run.py
```diff
@@ -232,9 +232,8 @@ async def run_individual(request: Request):
             )
         # Cache the output of the node
         request.app.ctx.cache[full_data["id"]] = output
-        extra_data = node_instance.get_extra_data()
         del node_instance, run_func
-        return json({"success": True, "data": extra_data})
+        return json({"success": True, "data": {}})
```
RunDevelopment, line 236: I think this will make a cleaner API. Also, please change the signature of `Backend#runIndividual` to `runIndividual(data: BackendRunIndividualRequest): Promise<BackendResult<null>>`.

```suggestion
        return json({"success": True, "data": None})
```

RunDevelopment (approved):

PR #626, opened 2022-07-24
Allow iterator children with image view buttons to make view nodes inside iterator

joeyballentine:


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -551,12 +551,13 @@ export const GlobalProvider = memo(
         );
 
         const createNode = useCallback(
-            (proto: NodeProto): void => {
+            (proto: NodeProto, parent?: Node<NodeData>): void => {
```
RunDevelopment, line 554: I think it would be better if this was a node id. This would probably play together nicer with `hoveredNode` since you could do something like `const parentId = hoveredNode ?? parentIdParameter` and then search for that.

joeyballentine:
There's sorta an issue with this where the position for the created node isn't correct, but I think it's good enough. We can figure out how to fix that later

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -551,11 +551,12 @@ export const GlobalProvider = memo(
...
-            (proto: NodeProto): void => {
+            (proto: NodeProto, parentId?: string): void => {
                 changeNodes((nodes) => {
-                    const parent = hoveredNode
-                        ? nodes.find((n) => n.id === hoveredNode)
-                        : undefined;
+                    const parent =
+                        parentId || hoveredNode
+                            ? nodes.find((n) => n.id === (parentId ?? hoveredNode))
+                            : undefined;
```
RunDevelopment, line 559: [I like my code how I like my beer](https://www.youtube.com/watch?v=yKPA5l9pvFo).

```suggestion
                    const searchId = parentId ?? hoveredNode;
                    const parent =
                        searchId
                            ? nodes.find((n) => n.id === searchId)
                            : undefined;
```

RunDevelopment (approved):

PR #628, opened 2022-07-25
Add tab support to node selector panel

joeyballentine:
- move down the list with tab, or up with shift-tab (uses built-in tab focus system)
- create node with enter

### DIFF
--- a/src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
+++ b/src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
@@ -1,6 +1,7 @@
 import { StarIcon } from '@chakra-ui/icons';
 import { Box, Center, Flex, HStack, Heading, Spacer, useColorModeValue } from '@chakra-ui/react';
-import { memo, useState } from 'react';
+import { memo, useRef, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
 import { SchemaId } from '../../../common/common-types';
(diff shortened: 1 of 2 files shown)


src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
```diff
@@ -40,8 +43,26 @@ export const RepresentativeNode = memo(
...
+
+        useHotkeys(
+            'enter',
+            () => {
+                if (document.activeElement === ref.current) {
+                    createNodeFromSelector();
+                }
+            },
+            [ref]
+        );
```
RunDevelopment, line 56: This seems so wrong to me. Why not use `onKeydown`?

src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
```diff
@@ -40,8 +43,26 @@ export const RepresentativeNode = memo(
...
             <Center
+                _active={{
+                    outlineColor: accentColor,
+                }}
+                _focus={{
+                    outlineColor: accentColor,
+                }}
                 _hover={{
                     outlineColor: accentColor,
                 }}
```
RunDevelopment, line 68: Could we make these a single line each?

```suggestion
                _active={{ outlineColor: accentColor }}
                _focus={{ outlineColor: accentColor }}
                _hover={{ outlineColor: accentColor }}
```

RunDevelopment (approved):

PR #630, opened 2022-07-25
Extracted the code for running a single node out into a hook

RunDevelopment:
What the title says. Also:

-  I changed how it is decided when to run a node. Instead of using `useEffect` and some setState functions, I remember the inputs with which the node was last run. This makes it easy to not run nodes again.
- `!isLocked && !disabled.isDirectlyDisabled` seems to be a left over from the disabled bug I reported. I removed this because I don't see why disabled or locked nodes shouldn't have previews.

`useRunNode` is a controlled hook, so the caller decides whether the nodes actually runs. Since the hook has a small delay before actually running, the caller may also change its mind with `useEffect`.

### DIFF
--- a/src/common/util.ts
+++ b/src/common/util.ts
@@ -179,3 +179,9 @@ export const topologicalSort = <T>(
 export const isStartingNode = (schema: NodeSchema) => {
     return !schema.inputs.some((i) => i.hasHandle);
 };
+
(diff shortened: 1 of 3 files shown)


joeyballentine:
It all looks good, but I wanna test it before merging

joeyballentine (approved):

PR #638, opened 2022-07-26
More torch type fields

joeyballentine:


backend/src/nodes/properties/expression.py
```diff
@@ -199,3 +199,15 @@ def Image(
         fields["width"] = field(size_as, "width")
         fields["height"] = field(size_as, "height")
     return named("Image", fields)
+
+
+def PyTorchModel(
```
joeyballentine, line 204: I should probably get rid of this since I don't use it at all, yeah?

src/common/types/chainner-scope.ts
```diff
@@ -29,6 +29,8 @@ struct PyTorchModel {
     scale: int(1..),
     inputChannels: int(1..),
     outputChannels: int(1..),
+    modelType: string, // "ESRGAN" | "RealESRGAN" | "RealSRGANv2" | "Swift-SRGAN"
```
joeyballentine, line 32: I wasn't sure if it was preferable to use just "string" or if defining the specific strings was better

backend/src/nodes/properties/expression.py
RunDevelopment: If you don't use it, then yes please. I only added `Image` because I used it a lot before Navi parsing. `Image` is still used a bit, but I also contemplated removing it.

src/common/types/chainner-scope.ts
RunDevelopment: I would probably leave it as `string` because then you'll have to change less when adding support for another model type.

src/renderer/components/outputs/PyTorchOutput.tsx
```diff
@@ -63,6 +72,14 @@ export const PyTorchOutput = memo(
                                 'outputChannels',
                                 new NumericLiteralType(value.outNc)
                             ),
+                            new NamedExpressionField(
+                                'modelType',
+                                new StringLiteralType(convertToInternalModelType(value.modelType))
```
RunDevelopment, line 77: This seems weird. Why do you map `RealESRGAN` to `ESRGAN`?

backend/src/nodes/pytorch_nodes.py
```diff
@@ -73,7 +73,7 @@ def __init__(self):
             Real-ESRGAN's SRVGG architecture, and Swift-SRGAN."""
         self.inputs = [PthFileInput()]
         self.outputs = [
-            ModelOutput(kind="pytorch", should_broadcast=True),
+            ModelOutput(kind="PyTorchModel", should_broadcast=True),
```
RunDevelopment, line 76: Is this correct?

src/renderer/components/outputs/PyTorchOutput.tsx
joeyballentine: Because those model types are compatible. We talked about this in discord. Internally we convert RealESRGAN models to esrgan models

backend/src/nodes/pytorch_nodes.py
joeyballentine: This was the only one I saw using "pytorch" instead of "PyTorchModel" (the default)

RunDevelopment: I think you mixed up type and [kind](https://github.com/joeyballentine/chaiNNer/blob/f3ce6b3dbbe4770bd487a1a29cabc3090190a5cb/src/common/common-types.ts#L55).

src/renderer/components/outputs/PyTorchOutput.tsx
RunDevelopment: Okay, then why does the backend send "RealESRGAN" then?

joeyballentine: Because it starts as a RealESRGAN type model and mainly people think of RealESRGAN as something different (which I guess technically it is even though it can be converted to the same format). 

Though even though technically they can be interpolated with regular ESRGAN models, I don't think any exist which would actually work.

RunDevelopment: > it can be converted to the same format

Where does this conversion happen?

joeyballentine: Inside RRDB.py

RunDevelopment: So models get converted on load. This means that the `modelType` field is ill-defined right now. Does model type represent the original (unconverted) model type or the actual (converted) mode type?

(Also, the mapping "RealESRGAN" -> "ESRGAN" is not enough, because we also have "ESRGAN+". You probably don't need to add this mapping though. My end goal is to get rid of the need to do this mapping altogether.)

joeyballentine: ESRGAN+ models are not compatible with ESRGAN or RealESRGAN models, and RealESRGAN models cannot be +

Anyway, it represents the unconverted model. I suppose it should probably represent the converted model though

RunDevelopment: > I suppose it should probably represent the converted model though

Sounds good.

RunDevelopment (approved):

PR #639, opened 2022-07-27
Added event types and minor improvements to executor

RunDevelopment:
This is the third time I wanted to implement parallelism for broadcast data, but ended up doing refactoring instead...

Well, this time around, I did some minor refactoring around `Executor` and `ExecutorContext`, I added an `AppContext` class to strictly type `app.ctx` in `run.py`, and I completely typed server-sent events on the python side (I also found and fixed a small and inconsequential error in the TS types for events).

### DIFF
--- a/backend/src/events.py
+++ b/backend/src/events.py
@@ -1,0 +1,75 @@
+import asyncio
+from typing import (
+    Any,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    TypedDict,
+    Union,
+)
+
+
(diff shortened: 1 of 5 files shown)


RunDevelopment:
`asyncio.Queue` was [only recently](https://github.com/python/cpython/pull/30784) made generic in 3.10.something, so I had to write a wrapper around `Queue` to make it type safe.

joeyballentine (approved):

PR #640, opened 2022-07-27
Compute broadcasts in parallel to regular node execution

RunDevelopment:
What the title says. I did this by using a ThreadPoolExecutor to compute the broadcast in a separate thread. Since `run_in_executor` does not guarantee any number of threads (it could be just a single thread for all that we know), I explicitly created my own pool with 4 threads. It would be more, but 4 should be more than enough for now.

However, there is one bug(?) I noticed. The `node-output-data` event may now arrive later than `node-finish` event, which causes the frontend to display a "Preview not available" instead of a loading spinner. I don't know whether this is a problem in the frontend (because it assumes that the broadcast arrives before `node-finish`) or in the backend (because it doesn't guarantee the event order).

### DIFF
--- a/backend/src/events.py
+++ b/backend/src/events.py
@@ -27,17 +27,14 @@ class ExecutionErrorData(TypedDict):
 
 class NodeFinishData(TypedDict):
     finished: List[str]
+    nodeId: str
+    data: Optional[Dict[int, Any]]
 
 
 class IteratorProgressUpdateData(TypedDict):
     percent: float
     iteratorId: str
     running: Optional[List[str]]
-
-
-class NodeOutputDataData(TypedDict):
-    nodeId: str
-    data: Dict[int, Any]
 
 
 class FinishEvent(TypedDict):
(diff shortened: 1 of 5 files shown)


joeyballentine:
I'll see if I can figure out if it's fixable from the frontend

RunDevelopment:
This might not be fixable in the frontend/should be fixed there. This is more of the question: is "broadcasts arrive before `node-finish`" something the backend guarantees? If we say that the backend has to guarantee this, then that's it. We could also make this guarantee very explicit by removing the broadcast event and attaching the (optional) broadcast data to the `node-finish` event.

joeyballentine:
> This might not be fixable in the frontend/should be fixed there. The question is mostly: is "broadcasts arrive before `node-finish`" something the backend guarantees? If we say that the backend has to guarantee this, then that's it. We could also make this guarantee very explicit by removing the broadcast event and attaching the (optional) broadcast data to the `node-finish` event.

lmao, I originally had the broadcast data attached to the node-finish event and you made me remove it. I'm perfectly fine with putting that back, we just need to change how the node finish event works to not send the historical data and remove the throttling in the frontend so we guarantee all messages get processed.

RunDevelopment:
Oh shit, sorry.

> remove the throttling in the frontend

Why do we throttle in the first place? It's not like we don't trust the backend.

joeyballentine:
It's because I noticed for iterators if you had a sub-flow that was composed of really simple stuff, it would go through events (and therefore state updates) so fast it would lag everything pretty badly

joeyballentine (approved):

PR #641, opened 2022-07-28
Added `useBatchedCallback`

RunDevelopment:


joeyballentine:
Did you test this on a fast iterator?

RunDevelopment:
Not yet.

joeyballentine:
> Not yet.

If you try just a load -> view or load -> save without any processing that should be a good test. The thing I was testing when I found the performance problem was using a video iterator on a very small video that was just in -> out

RunDevelopment:
I tested this iterator (129 copies of the same image).
image

Throttle is signifcantly faster for a simple reason: it's incorrect. Throttle will drop events if there are too many. This is obviously incorrect for broadcasts. However, it's okay for updating iterator percentage, so I still use it there.

Updating dozens of broadcasts takes its toll of the UI and creates lag. To combat this, I increased the wait time on `node-finish`. Since the first `node-finish` event (after a full wait time) will always be processed immediately, this shouldn't make the UI feel delayed/slow.

joeyballentine (approved):

PR #644, opened 2022-07-28
Block Size -> Block Radius

RunDevelopment:


src/common/migrations.js
```diff
@@ -666,6 +666,17 @@ const removeEmptyStrings = (data) => {
     return data;
 };
 
+const blockSizeToRadius = (data) => {
+    data.nodes.forEach((node) => {
+        if (node.data.schemaId === 'chainner:image:threshold_adaptive') {
+            const size = node.data.inputData[4] ?? 3;
+            node.data.inputData[4] = Math.floor((size - 1) / 2);
```
RunDevelopment, line 673: The floor is just in case the chain had an invalid input. It should never do anything, but better safe than sorry.

joeyballentine (approved):

PR #648, opened 2022-07-29
Cache fix fix

RunDevelopment:
I was just about to push this commit when #646 got merged


RunDevelopment:
The whole thing is based on a counter. Each run individual increments the counter while it is executing its node. Run simply waits until the counter is zero. This waiting is implemented via a dumb spin lock, but it'll work.

RunDevelopment:
Not sure about the name "ZeroCounter" btw.

joeyballentine (approved):

PR #651, opened 2022-07-30
Fix integer inputs

joeyballentine:
Something recently changed how rounding works which made floats no longer turn into integers when precision was 0. This is fixed by making the default `None`, as `None` is what the documentation says to use to make `round` turn a float into an int.


RunDevelopment:
Soooo, instead of adding an `if` into the `clampNumber` function, you changed the type and default of `precision`, breaking the JSON API in the process (because our numeric inputs can't handle `precision == null`) and not even fixing the bug (because `precision` can still be 0)?

PR #671, opened 2022-08-01
Add "Export For Sharing" option

### LINKED ISSUE #658: Save blank chains
When making templates, I would be useful to have some way to save a blank version of the current chain. So with all file and directory inputs cleared.

joeyballentine:
Closes #658

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -17,6 +17,7 @@ import {
     EdgeData,
     InputData,
     InputId,
+    InputKind,
     InputSize,
     InputValue,
     IteratorSize,
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -445,7 +445,7 @@ export const GlobalProvider = memo(
         }, [hasRelevantUnsavedChanges, changeNodes, changeEdges, setSavePath, setViewport]);
 
         const performSave = useCallback(
-            (saveAs: boolean, clearInputData = false, updateSavePath = true) => {
+            (saveAs: boolean, clearInputData = false, saveSeparately = false) => {
                 (async () => {
                     try {
                         const saveData = dumpState();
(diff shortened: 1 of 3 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -445,23 +445,46 @@ export const GlobalProvider = memo(
         }, [hasRelevantUnsavedChanges, changeNodes, changeEdges, setSavePath, setViewport]);
 
         const performSave = useCallback(
-            (saveAs: boolean) => {
+            (saveAs: boolean, clearInputData = false, saveSeparately = false) => {
```
RunDevelopment, line 448: I think these 2 parameters should be one `isTemplate: false`. `clearInputData` doesn't actually clear all input data, and `saveSeparately` doesn't really explain what it does.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -445,23 +445,46 @@ export const GlobalProvider = memo(
...
                 (async () => {
                     try {
                         const saveData = dumpState();
+                        if (clearInputData) {
+                            saveData.nodes = saveData.nodes.map((n) => {
+                                const inputData = { ...n.data.inputData } as Mutable<InputData>;
+                                const nodeSchema = schemata.get(n.data.schemaId);
+                                nodeSchema.inputs.forEach((input) => {
+                                    if (['file', 'directory'].includes(input.kind)) {
```
RunDevelopment, line 457: This array is only typed as `string`.
```suggestion
                                    const clearKinds = new Set<InputKind>(['file', 'directory']);
                                    if (clearKinds.has(input.kind)) {
```

Also, since the `clearKinds` set could be moved outside the `nodes.map`.

RunDevelopment (changes requested):

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: Good point

RunDevelopment (approved):

PR #672, opened 2022-08-01
Clear all output data on state clear

joeyballentine:
Noticed this when working on #671. If you save, press File>New, then open the saved chain, it will still have the old output data, but say it's all stale.

### DIFF
--- a/src/renderer/hooks/useOutputDataStore.ts
+++ b/src/renderer/hooks/useOutputDataStore.ts
@@ -18,6 +18,7 @@ export interface OutputDataActions {
         data: OutputData | undefined
     ): void;
     delete(nodeId: string): void;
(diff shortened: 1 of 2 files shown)


src/renderer/hooks/useOutputDataStore.ts
```diff
@@ -57,6 +58,9 @@ export const useOutputDataStore = () => {
             },
             [setMap]
         ),
+        clearAll: useCallback(() => {
```
RunDevelopment, line 61: The map operation is called `clear`.
```suggestion
        clear: useCallback(() => {
```

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #679, opened 2022-08-01
Blend node performance improvement

RunDevelopment:
I refactored the blend node to drastically improve performance.

There are a number of optimizations the implementation now makes:
1. `blend_images` will now be called directly if the 2 images have the same size. This skips all the size conversions and at least 1 image copy.
2. Channels conversions (if any) are now done by `blend_images`. This enables all the following optimizations.
3. If the 2 images have the same number of channels, no channel conversion and copying will be done.
4. If neither of the images is RGBA, then no alpha blending will be done.
5. Alpha blending caches intermediate results more efficiently.

All these optimizations together mean that if the base and overlay image are both RGB images of the same size (common case), the node will basically just do `apply_blend(overlay, base, blend_mode)` and nothing else (aside from some clipping). Can't get any more efficient than that.

### DIFF
--- a/backend/src/nodes/image_util_nodes.py
+++ b/backend/src/nodes/image_util_nodes.py
@@ -62,37 +62,38 @@ class ImBlend(NodeBase):
 
         if (b_w, b_h) == (o_w, o_h):
             # we don't have to do any size adjustments
-            blend_images(ov, base, blend_mode)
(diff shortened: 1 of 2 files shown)


RunDevelopment:
@theflyingzamboni Could you please test that everything still works?

joeyballentine (approved):

PR #680, opened 2022-08-01
Let the node-finish handler un-animate individual node runs

joeyballentine:
This fixes the issue I mentioned about the node un-animating before the output data was set. Now, the SSE handler handles un-animating the node, and it does so after the output data has been set. The node only un-animates when there is an error.

### DIFF
--- a/src/renderer/contexts/ExecutionContext.tsx
+++ b/src/renderer/contexts/ExecutionContext.tsx
@@ -235,11 +235,7 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                     data ?? undefined
                 );
 
-                if (finished.length > 0) {
-                    unAnimate(finished);
-                } else if (nodeId) {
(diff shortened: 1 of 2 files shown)


src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -237,6 +234,12 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                     inputHash,
                     data ?? undefined
                 );
+
+                if (finished.length > 0) {
+                    unAnimate(finished);
+                } else if (nodeId) {
+                    unAnimate([nodeId]);
+                }
```
RunDevelopment, line 242:
```suggestion
                unAnimate([nodeId, ...finished]);
```

RunDevelopment (changes requested):
Looks okay as a whole. Just one nit.

PR #681, opened 2022-08-02
Improve descriptions on some nodes

joeyballentine:


backend/src/nodes/ncnn_nodes.py
```diff
@@ -252,7 +252,8 @@ def upscale(i: np.ndarray) -> np.ndarray:
 class NcnnInterpolateModelsNode(NodeBase):
     def __init__(self):
         super().__init__()
-        self.description = "Interpolate two NCNN models of the same type together."
+        self.description = """Interpolate two NCNN models of the same type together. Note: models must share a common 'pretrained model' ancestor"
```
RunDevelopment, line 255: Random quote.

```suggestion
        self.description = """Interpolate two NCNN models of the same type together. Note: models must share a common 'pretrained model' ancestor
```

RunDevelopment (changes requested):

PR #685, opened 2022-08-03
Delete cache entry when individual node run errors

joeyballentine:
Fixes an issue kim found

### DIFF
--- a/backend/src/run.py
+++ b/backend/src/run.py
@@ -238,8 +238,9 @@ class RunIndividualRequest(TypedDict):
 async def run_individual(request: Request):
     """Runs a single node"""
     ctx = AppContext.get(request.app)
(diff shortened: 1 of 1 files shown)


backend/src/run.py
```diff
@@ -294,6 +294,7 @@ async def run_individual(request: Request):
         return json({"success": True, "data": None})
     except Exception as exception:
         logger.error(exception, exc_info=True)
+        del ctx.cache[full_data["id"]]
```
RunDevelopment, line 297: Instead of removing it *after* something went wrong, it's a lot more robust to remove the cached value *before* running that node. Actually, even before validating inputs.

joeyballentine: Whats the point of catching it if we're just going to remove it before running it?

PR #686, opened 2022-08-03
Reset inputData when file causes error

joeyballentine:
This is just to show the alternate fix I did. I didn't like it since it causes an infinite undo-lock where you can't undo since it undoes to the invalid state, then immediately resets it.


RunDevelopment:
1. I don't like chainner just reverting my inputs willy-nilly.
2. If the input that causes the error is not stored in `inputData` (e.g. because the node in a helper node or because the input came from an edge), this hack will make arbitrary changes to `inputData`. While this isn't a problem right now (we only run nodes that can't have connections), this will limit us if we ever want to add a handle to e.g. Load Image.

PR #693, opened 2022-08-04
Let node-finish handler handle un-animating rather than final finish

joeyballentine:
Since broadcasts are done in another thread, this would cause the chain to be finished before the data was sent to the frontend. This would cause view nodes to become unanimated before being ready to show their new previews.

I can't think of any downsides of removing this, as it just lets each node handle the un-animating rather than every node un-animating, but that doesn't mean there couldn't be some hidden issue that would need to be fixed by this.

### DIFF
--- a/src/renderer/contexts/ExecutionContext.tsx
+++ b/src/renderer/contexts/ExecutionContext.tsx
@@ -185,7 +185,6 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
         eventSource,
         'finish',
         () => {
-            unAnimate();
             setStatus(ExecutionStatus.READY);
         },
         [setStatus, unAnimate]


RunDevelopment:
> Since broadcasts are done in another thread, this would cause the chain to be finished before the data was sent to the frontend.

It would not. This is what the [`__broadcast_tasks` list](https://github.com/joeyballentine/chaiNNer/blob/d76cba16ff74b839c6cfdbc4670ef3aa81d9e2ef/backend/src/process.py#L96) is for. This list holds a task for each broadcast we make, so we can [await all of them](https://github.com/joeyballentine/chaiNNer/blob/d76cba16ff74b839c6cfdbc4670ef3aa81d9e2ef/backend/src/process.py#L305) before finishing processing.

> I can't think of any downsides of removing this

Same.

RunDevelopment (approved):

joeyballentine:
@RunDevelopment well, it was still sending the chain finish event before the data was displayed, because this fixed the issue.

RunDevelopment:
> it was still sending the chain finish event before the data was displayed

That might be, but it wouldn't matter. Events arrive in order. Here's an example where it broadcasts a 8k texture (so it takes some time to create the broadcast):

image

Since `node-finish` un-animates the node before `finish` even arrives, this PR effectively just removed code that didn't do anything.

So what issue did you fix, @joeyballentine?

joeyballentine:
@RunDevelopment check out the commit before this, wire up a load image -> view image, select an image, run, cha he the image, run again. The view node will inanimate, show the old image, then show the new one.

RunDevelopment:
You're right. I also understand why now. `node-finish` events get batched, so they will arrive on time, but they might be processed with some delay.

PR #710, opened 2022-08-08
Added hints for missing dependencies

RunDevelopment:
As suggested and shown on discord.

This is a branch of #709.

### DIFF
--- a/src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
+++ b/src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
@@ -112,6 +112,7 @@ export const NodeSelector = memo(({ schemata }: NodeSelectorProps) => {
                         <TabPanels>
                             <TabPanel
                                 m={0}
+                                overflowX="hidden"
                                 p={0}
                             >
                                 <InputGroup borderRadius={0}>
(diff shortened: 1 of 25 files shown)


joeyballentine:
Merge conflicts

joeyballentine:
Should we also get rid of that "missing nodes? Open the dependency manager" thing at the bottom of the list now?

src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
```diff
@@ -34,11 +35,44 @@ import {
 } from '../../helpers/nodeSearchFuncs';
 import { useNodeFavorites } from '../../hooks/useNodeFavorites';
 import { FavoritesAccordionItem } from './FavoritesAccordionItem';
-import { RegularAccordionItem } from './RegularAccordionItem';
+import { PackageHint, RegularAccordionItem, Subcategories } from './RegularAccordionItem';
 import { TextBox } from './TextBox';
 
+const packageHints: Partial<Record<string, () => ReactNode>> = {
```
joeyballentine, line 41: Instead of defining these here, could we define them in dependencies.js? This doesn't seem very DRY. Then we could just build each one of these from the information there

RunDevelopment:
I would keep it. We might have categories without hints in the future. Also, the information "you should look at the dependency manager" is important, so it bears repeating.

src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
RunDevelopment: Because dependencies.ts is in src/common and that doesn't do React.

RunDevelopment: Hints are a frontend thing.

joeyballentine: Oh right I forgot this uses the categories from the backend. Maybe we could define this stuff there? I don't mean defining the react stuff but defining the strings used, and then just {description} them, ya know. That can be handled in a different ticket though.

joeyballentine (approved):

PR #715, opened 2022-08-09
Refactored category information to come from backend

joeyballentine:
This accomplishes a few things:
- The category names used by nodes are now 100% guaranteed to be correct and matching to the actual categories 
- The category icons and colors are no longer hardcoded in the frontend as they are now defined in the backend. The only caveat to this is that custom icons (PyTorch, NCNN, and ONNX) still have to be defined in the frontend. 
- The install hints are also no longer hardcoded in the frontend

### DIFF
--- a/backend/src/nodes/categories.py
+++ b/backend/src/nodes/categories.py
@@ -1,24 +1,3 @@
-# Image
-IMAGE = "Image"
-IMAGE_DIMENSION = "Image (Dimensions)"
-IMAGE_ADJUSTMENT = "Image (Adjustments)"
-IMAGE_FILTER = "Image (Filters)"
-IMAGE_UTILITY = "Image (Utilities)"
-IMAGE_CHANNEL = "Image (Channels)"
-
-# PyTorch
-PYTORCH = "PyTorch"
-
-# ONNX
-ONNX = "ONNX"
-
-# NCNN
-NCNN = "NCNN"
-
-# Utility
-UTILITY = "Utility"
-
-
 class Category:
     def __init__(
         self, name: str, description: str, icon: str, color: str, install_hint: str = ""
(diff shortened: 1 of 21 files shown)


backend/src/nodes/categories.py
```diff
@@ -18,15 +18,105 @@
...
-    IMAGE_UTILITY,
-    IMAGE_CHANNEL,
-    PYTORCH,
-    ONNX,
-    NCNN,
-    UTILITY,
+
+class Category:
+    def __init__(
+        self, name: str, description: str, icon: str, color: str, install_hint: str = ""
```
RunDevelopment, line 24: Please type the install hint as nullable instead of using the empty string as sentinel value.

src/common/common-types.ts
```diff
@@ -150,3 +150,11 @@ export interface WindowSize {
     readonly width: number;
     readonly height: number;
 }
+
+export interface Category {
+    name: string;
+    description: string;
+    icon: string;
+    color: string;
+    installHint: string;
```
RunDevelopment, line 159: This type also need to be nullable.

src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
```diff
@@ -246,24 +212,25 @@ export const NodeSelector = memo(() => {
...
                                                         category={category}
                                                         collapsed={collapsed}
-                                                        key={category}
+                                                        key={category.name}
                                                     >
                                                         <PackageHint
                                                             collapsed={collapsed}
-                                                            description={hint}
+                                                            hint={category.installHint}
+                                                            // TODO: Somehow link categories to deps
```
RunDevelopment, line 225: TODO: Open issue instead of using a TODO.

---

Yeah, I think this should be in issue rather than a TODO comment.

RunDevelopment (changes requested):
I think the general approach is good. Just a few nits and suggestions.

> The category names used by nodes are now 100% guaranteed to be correct and matching to the actual categories

No, they aren't because all nodes use these constants and not `MyCategory.name`. Those constants just so happen to be used by categories.

IMO, there shouldn't be any category name constants. This will require more code changes, though.

> The category [...] colors [...] are now defined in the backend.

I don't like this. Color is such a pure frontend thing. What if light and dark mode require slightly different category colors in the future? What if we want to add themes that can also change category colors (e.g. a color-blind or high-contrast mode)? It's just not practical IMO.

joeyballentine:
> Color is such a pure frontend thing. What if light and dark mode require slightly different category colors in the future?

I started typing up a big rebuttal to this, which I'll still post if you disagree with this idea, but I had an idea that's kind of a middle ground:

We define a color system similar to chakra's, where you use a string of a pre-defined color variable for the category (in the backend) like "category.red.500". Then, have frontend code to take that value and apply an actual color. I would be completely ok with doing this (in a separate PR ideally). What I'm not ok with is keeping the big hardcoded switch statement for category information.

RunDevelopment:
I like the idea of category color variables in the frontend, but I think e.g. "category.red.500" is the wrong approach. This will just lead to a bunch of colors that aren't used (i.e. I doubt we'll use all "category.red.X00" colors). IMO we should just have something like CSS variables for our categories.

Anyway, this is nothing that should block this PR, so let's continue this discussion in an issue after merging this PR.

backend/src/nodes/image_adj_nodes.py
```diff
@@ -38,7 +38,7 @@ def __init__(self):
             ),
         ]
         self.outputs = [ImageOutput(image_type="Input0")]
-        self.category = IMAGE_ADJUSTMENT
+        self.category = ImageAdjustmentCategory.name
```
RunDevelopment, line 41: Looking at this, I have another idea: Let's make it `self.category = ImageAdjustmentCategory`.

This would make `self.category` an actual category object and not just some string, and we could save the `.name` everywhere. We would just have to change the type of `self.category` and make it `"category": node_object.get_category().name` in `run.py`.

What do you think?

joeyballentine: Sure

joeyballentine:
> This will just lead to a bunch of colors that aren't used

I do not want to hardcode any of the categories in the frontend anywhere, period. I know you see it differently, but I want chainner to move in the direction of being able to use any backend with the same frontend. As if the frontend was a closed source project and we were developing a backend for it. That would mean we can't rely on changing the frontend to pick node colors, and would instead mean that the frontend provides essentially an API for how the backend picks what color a node is. Does that make sense?

RunDevelopment (approved):

PR #718, opened 2022-08-10
use grab cursor for representative nodes

joeyballentine:
Yes, I know this is excessive. But, it was the only way to override the global cursor CSS for every part of the component

### DIFF
--- a/src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
+++ b/src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
@@ -46,12 +46,13 @@ export const RepresentativeNode = memo(
             <Center
                 _active={{ outlineColor: accentColor }}
                 _focus={{ outlineColor: accentColor }}
-                _hover={{ outlineColor: accentColor }}
+                _hover={{ outlineColor: accentColor, cursor: 'grab' }}
                 bgGradient={bgGradient}
                 borderColor={bgColor}
                 borderRadius="lg"
                 borderWidth="0px"
                 boxShadow="lg"
+                className="representative-node"
                 outline="1px solid"
                 outlineColor={bgColor}
                 overflow="hidden"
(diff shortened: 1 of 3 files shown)


RunDevelopment:
This isn't quite there yet.

1. The cursor doesn't update correctly after dropping a node. When I drop a node into the RF editor while holding my mouse still as I let go of the left mouse button, the cursor will still be `grab` after the node has been added. It seems like the UI thinks that my mouse still hovers over the node selector.
2. When holding down the mouse and dragging a node into the RF editor, the cursor should change to `grabbing`.

joeyballentine:
> When holding down the mouse and dragging a node into the RF editor, the cursor should change to grabbing.

It should. But it didn't work when I tried it.

joeyballentine:
And idk what to do about that first thing. I'm not entirely convinced it's something I did incorrectly rather than a chromium quirk

joeyballentine:
> When holding down the mouse and dragging a node into the RF editor, the cursor should change to grabbing.

I don't think this is actually possible, since we aren't dragging around a draggable object (like we do when we move around nodes). We are doing the equivalent of dragging an image from a webpage. This has built-in cursors that are not changeable since while dragging, it is no longer interacting with the dom.

RunDevelopment:
> I don't think this is actually possible

Uhm, then why did you use the `grab` cursor in the first place? Isn't the whole point of this cursor to be used with the `grabbing` cursor? It's not like that cursor combination is commonly used either. The last piece of software that I used that had it was Google Earth, so ~8 years ago.

Honestly, do we even need this change? Nowadays, it seems like software simply changes the cursor when you start dragging (e.g. the map on Google Maps) and we already do that with the cursor we can't get rid of.

joeyballentine:
I just wanted something that made it a bit more obvious you could drag and drop onto the canvas. Right now they just look like buttons rather than something you can drag

RunDevelopment:
> I just wanted something that made it a bit more obvious you could drag and drop onto the canvas.

Then it might be worth a user survey whether this actually helps. The grab cursor is so uncommon nowadays that it wouldn't have helped me if I had not known the name of this cursor ahead of time. 

I think it would be better if the cursor changed to grabbing (because that give feedback), but it seems like that's not possible.

> Right now they just look like buttons rather than something you can drag

This might sound like a stupid question, but: How does something you can drag look like?

joeyballentine:
> This might sound like a stupid question, but: How does something you can drag look like?

That's not a stupid question. I don't know, lol

joeyballentine:
I'll just close this. I no longer think it's necessary

PR #721, opened 2022-08-12
ONNX to NCNN conversion

theflyingzamboni:
Alright, this should finally be ready for testing, and at the least implements enough to convert/load ESRGAN models.

This PR:
1) Implements onnx2ncnn as a converter class
2) Adds a Convert to NCNN onnx node
3) Implements an NcnnModel class and various supporting classes
4) Alters all NCNN nodes to pass/receive NCNN model objects rather than net tuples
5) Makes use of the load_x_mem pybinds in current ncnn-vulkan-python main branch (though I don't think on pypi yet)

Closes joeyballentine/chaiNNer#235


backend/src/nodes/ncnn_nodes.py
```diff
@@ -199,9 +154,10 @@ def upscale(
             raise RuntimeError("An unexpected error occurred during NCNN processing.")
 
     def run(
-        self, net_data: NcnnNetData, img: np.ndarray, tile_size_target: int
+        self, model: NcnnModel, img: np.ndarray, tile_size_target: int
     ) -> np.ndarray:
         h, w, _ = get_h_w_c(img)
+        model_c = self.get_model_channels(model)
```
RunDevelopment, line 160: In the below code, you assume that `model_c` is either 1 or 3. Could you please add an assertion for this?

backend/src/nodes/utils/parse_ncnn_param.py
```diff
@@ -0,0 +1,58 @@
...
+                layer_specific_params = layer_specific_params_temp
+            except IndexError:
+                layer_specific_params = []
+
+            print(
+                f"{layer_type=} {layer_name=} {input_count=} {output_count=} {input_blobs=} {output_blobs=} {layer_specific_params=}"
+            )
+
+
+parse_ncnn_param(r"D:/Upscaling/models/LoD/New folder/4x_BSRGAN-opt.param")
```
RunDevelopment, line 58: Typical production code :)

backend/src/nodes/utils/parse_ncnn_param.py
```diff
@@ -0,0 +1,58 @@
...
+                for layer_param in layer_specific_params:
+                    layer_param_key, layer_param_value = layer_param.split("=")
+                    layer_param_key = int(layer_param_key)
+                    if layer_param_key >= 0:
+                        if "." in layer_param_value:
+                            layer_param_value = float(layer_param_value)
+                        else:
+                            layer_param_value = int(layer_param_value)
+                    else:
+                        layer_param_key = (layer_param_key * -1) - 23300
```
RunDevelopment, line 40: Ah, yes. 23300.

backend/src/nodes/utils/parse_ncnn_param.py
```diff
@@ -0,0 +1,58 @@
+import json
+
+with open("./ncnn_param_schema_converted.json") as f:
+    param_schema = json.load(f)
+
+print(f"{param_schema['Convolution']=}")
+
+
+def parse_ncnn_param(param_path):
```
RunDevelopment, line 9: I won't say much about this function. It seems somewhat unpolished, with a few antipatterns sprinkled in here and there. 

Please thoroughly test and type this code. You seem to make a lot of assumptions about the parsed text, so please add assert statements where appropriate.

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,524 @@
...
+            f.write(f"{self.MAGIC}\n")
+            f.write(f"{self.node_count} {self.blob_count}\n")
+
+            for layer in self.layer_list:
+                layer_str = (
+                    f"{layer.op_type:<16} "
+                    f"{layer.name:<24} "
+                    f"{layer.num_inputs} "
+                    f"{layer.num_outputs} "
+                    f"{self.stringify_list(layer.inputs)}"
```
RunDevelopment, line 473: Please just use `.join`...

Also, I didn't know that I could just ignore the `@staticmethod` decorator.

```suggestion
                    f"{" ".join(layer.inputs)} "
```

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,524 @@
...
+                self.blob_count = int(counts[1])
+
+                for line in paramf:
+                    op_type, layer = self.parse_param_layer(line)
+                    layer.weight_data = self.load_layer_weights(binf, op_type, layer)
+                    self.add_layer(layer)
+
+    def write_param(self, filename: str) -> None:
+        with open(filename, "w", encoding="utf-8") as f:
+            f.write(f"{self.MAGIC}\n")
```
RunDevelopment, line 464: This feels like it should be a static constant and not a mutable instance field.

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,524 @@
...
+                    f"{layer.name:<24} "
+                    f"{layer.num_inputs} "
+                    f"{layer.num_outputs} "
+                    f"{self.stringify_list(layer.inputs)}"
+                    f"{self.stringify_list(layer.outputs)}"
+                    f"{str(layer.params)}".rstrip()
+                )
+                f.write(layer_str + "\n")
+
+    def write_param_to_mem(self) -> str:
```
RunDevelopment, line 479: This function is identical to `write_param` with the only difference being the API you use to append strings. I'm sure python has something like a string buffer that will allow you to abstract over the append method.

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,524 @@
...
+            )
+            param_str += layer_str + "\n"
+
+        return param_str
+
+    def write_bin(self, filename: str) -> None:
+        with open(filename, "wb") as f:
+            f.write(self.weights_bin)
+
+    def interpolate_ncnn(self, model_b, alpha):
```
RunDevelopment, line 499: Types please.
```suggestion
    def interpolate_ncnn(self, model_b: "NcnnModel", alpha: float) -> "NcnnModel":
```

RunDevelopment (changes requested):
Awesome work!

I'm not gonna review `onnx_to_ncnn.py` though... At least not now.

backend/src/nodes/ncnn_nodes.py
```diff
@@ -19,103 +16,49 @@
...
-
-    def run(self, param_path: str, bin_path: bytes) -> Tuple[NcnnNetData, str]:
+    def run(self, param_path: str, bin_path: str) -> Tuple[NcnnModel, str]:
         assert os.path.exists(
             param_path
         ), f"Param file at location {param_path} does not exist"
         assert os.path.exists(
             bin_path
-        ), f"Bin file at location {param_path} does not exist"
+        ), f"Bin file at location {bin_path} does not exist"
```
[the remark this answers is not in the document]
joeyballentine: Yeah it probably should. And yes these checks are no longer necessary

backend/src/nodes/ncnn_nodes.py
```diff
@@ -167,6 +103,25 @@ def __init__(self):
         self.icon = "NCNN"
         self.sub = "Processing"
 
+    def get_model_channels(self, model: NcnnModel) -> int:
+        num_filters = model.layer_list[1].params[0].value
+        kernel_w = model.layer_list[1].params[1].value
```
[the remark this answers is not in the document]
joeyballentine: Just the param

backend/src/nodes/ncnn_nodes.py
```diff
@@ -167,6 +103,25 @@ def __init__(self):
...
+            kernel_h = kernel_w
+        weight_data_size = model.layer_list[1].params[6].value
+
+        assert (
+            isinstance(num_filters, int)
+            and isinstance(kernel_w, int)
+            and isinstance(kernel_h, int)
+            and isinstance(weight_data_size, int)
+        ), "Out nc, kernel width and height, and weight data size must all be ints"
+        in_nc = weight_data_size // num_filters // kernel_w // kernel_h
```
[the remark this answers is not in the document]
joeyballentine: That's not an assumption, it's literally how weight data size is calculated

backend/src/nodes/utils/parse_ncnn_param.py
joeyballentine: Lol

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,524 @@
...
+from sanic.log import logger
+
+param_schema_file = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), "ncnn_param_schema.json"
+)
+with open(param_schema_file) as f:
+    param_schema = jload(f)
+
+DTYPE_FP32 = b"\x00\x00\x00\x00"
+DTYPE_FP16 = b"\x47\x6b\x30\x01"
```
joeyballentine, line 19: Do these have to be binary strings? Or could you put them in here as hex? (i.e. 0x00000000)

backend/src/nodes/utils/ncnn_model.py
joeyballentine: Should be able to do something like 
```py
    @property
    def magic(cls):
        return "7767517"
```
Which would make it immutable

backend/src/nodes/utils/ncnn_model.py
joeyballentine: Is that correct? Why did you surround the type with quotes?

backend/src/nodes/utils/utils.py
```diff
@@ -358,7 +358,7 @@ def convenient_upscale(
     else:
         # Add extra channels if not enough (i.e single channel img, three channel model)
         gray = False
-        if img.ndim == 2:
+        if input_channels > 1 and img.ndim == 2:
```
joeyballentine, line 361: Why did you add this?

backend/src/nodes/utils/ncnn_model.py
RunDevelopment: But that's still on instance thing.

backend/src/nodes/utils/ncnn_model.py
RunDevelopment: Because Python can't handle self-referential types. Remember, types aren't static in python, they are a runtime construct. So I can't reference the class `NcnnModel` until after it is defined. Quotes are the only way around this.

backend/src/nodes/utils/ncnn_model.py
joeyballentine: (deleted the original message since it put it on its own in my review)
Should be able to do something like 
```py
    @property
    def magic(cls):
        return "7767517"
```
Which would make it immutable

backend/src/nodes/utils/ncnn_model.py
joeyballentine: (deleted the original, so reposting) 
Why did you put this in quotes?

backend/src/nodes/utils/ncnn_auto_split.py
```diff
@@ -66,14 +68,20 @@ def ncnn_auto_split_process(
         # ex.set_light_mode(True)
         try:
             lr_img_fix = fix_dtype_range(lr_img.copy())
+            lr_c = get_h_w_c(lr_img)[2]
+            pixel_type = (
+                ncnn.Mat.PixelType.PIXEL_RGB
+                if lr_c == 3
+                else ncnn.Mat.PixelType.PIXEL_GRAY
+            )
```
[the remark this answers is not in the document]
theflyingzamboni: Added, but I'm also going to see if we can expand that to include 4-channel as well.

backend/src/nodes/onnx_nodes.py
```diff
@@ -173,3 +175,33 @@ def run(
...
+                ],
+            ),
+        ]
+        self.outputs = [NcnnModelOutput("NCNN Model")]
+
+        self.category = "ONNX"
+        self.name = "Convert To NCNN"
+        self.icon = "NCNN"
+        self.sub = "Utility"
+        self.side_effects = True
```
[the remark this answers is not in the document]
theflyingzamboni: Leftover test code I forgot was there, because I didn't want to need to add a preview to run it.

backend/src/nodes/ncnn_nodes.py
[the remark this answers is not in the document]
theflyingzamboni: I'll remove them then.

backend/src/nodes/ncnn_nodes.py
```diff
@@ -19,103 +16,49 @@
...
-        input_name, output_name, _out_nc = self.get_param_info(param_path)
+        assert os.path.isfile(bin_path), f"Path {bin_path} is not a file"
 
+        model = NcnnModel()
+        model.load_model(param_path, bin_path)
         with open(bin_path, "rb") as f:
-            bin_file_data = f.read()
-        bin_data = parse_ncnn_bin_from_buffer(bin_file_data)
-
+            model.weights_bin = f.read()
```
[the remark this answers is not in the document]
theflyingzamboni: The blank constructor exists because when converting ONNX to NCNN, the layers are being converted and added to it one at a time. I could overload the constructor to pass files to it directly, though.

As for load_model, it's loading the param layer information, as well as the corresponding numerical data arrays from the bin. For our purposes, we need the weights as arrays so we can interpolate them, but the actual ncnn load_model needs raw byte data. The whole having a separate attribute for the bin in bytes is a speed hack on my part, because converting the numerical arrays back to bytes takes time (especially with python for loops being so slow). When loading the model from file, the fastest way to do that is just read it straight from the bin. When converting from onnx or interpolating, it converts the arrays to bytes as it goes, then concatenates the byte strings at the end.

backend/src/nodes/ncnn_nodes.py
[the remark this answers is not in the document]
theflyingzamboni: What Joey said. I used integer division because the type needs to be int.

backend/src/nodes/ncnn_nodes.py
```diff
@@ -279,57 +229,23 @@ def __init__(self):
...
-        assert len(a.bin_data) == len(
-            b.bin_data
-        ), "The provided model bins are not compatible as they are not the same size."
-        assert (
-            a.bin_data.dtype == b.bin_data.dtype
-        ), "The provided model bins are not compatible as they are not the same datatype."
+        self, a: NcnnModel, b: NcnnModel, amount: int
+    ) -> Tuple[NcnnModel, int, int]:
+        f_amount = 1 - amount / 100
+        interp_model = a.interpolate_ncnn(b, f_amount)
```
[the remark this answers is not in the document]
theflyingzamboni: Yeah, that's a holdover from when it was outside the class.

backend/src/nodes/utils/parse_ncnn_param.py
theflyingzamboni: Don't look at me, that line was in this file when I received it lol. I only changed the path to test it. XD

Plus, that file is extraneous and has been removed.

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: Oh interesting. I left it untyped because it caused an error. Didn't realize you could get around that with quotes.

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: Is there a reason to have them as numbers? Then they have to be converted every time they're written to file or buffer.

backend/src/nodes/utils/utils.py
theflyingzamboni: Ah, guess it's not necessary. Missed that it wouldn't duplicate dims if there was only one input channel.

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: @joeyballentine Do you know if there's a way to abstract this and combine these methods?

backend/src/nodes/ncnn_nodes.py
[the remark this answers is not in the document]
theflyingzamboni: Apparently python does not do overloading lol

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: Nevermind, I think I've got it.

backend/src/nodes/utils/ncnn_model.py
joeyballentine: I was gonna say for clarity, but it doesn't matter tbh

backend/src/nodes/onnx_nodes.py
```diff
@@ -173,3 +175,22 @@ def run(
...
+        self.outputs = [NcnnModelOutput("NCNN Model")]
+
+        self.category = ONNXCategory
+        self.name = "Convert To NCNN"
+        self.icon = "NCNN"
+        self.sub = "Utility"
+
+    def run(self, onnx_model: bytes, is_fp16: int) -> NcnnModel:
+        converter = Onnx2NcnnConverter(onnx.load_model_from_string(onnx_model))
+        ncnn_model = converter.convert(is_fp16, False)
```
RunDevelopment, line 195: Please convert `is_fp16` to a bool before passing it in. The correctness of `Onnx2NcnnConverter.convert` depends on the values of `OnnxFpDropdown` right now. That's not good.

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,540 @@
...
+                a.num_outputs,
+                a.inputs,
+                a.outputs,
+                a.params,
+                weights_interp,
+            ),
+            layer_bytes,
+        )
+
+    def get_model_channels(self) -> int:
```
RunDevelopment, line 385: Those are the input channels, right?

RunDevelopment (changes requested):

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -0,0 +1,540 @@
...
+
+
+class NcnnModel:
+    def __init__(
+        self,
+        node_count: int = 0,
+        blob_count: int = 0,
+        param_path: str = "",
+        bin_path: str = "",
+    ) -> None:
```
RunDevelopment, line 284: Okay, let's not do overloading ever again.

Sorry about this, but could you please please separate this into a simple constructor and a static factory method `load_from_files` that essentially does this:

```py
        model = NcnnModel(param_path=param_path, bin_path=bin_path)
        with open(bin_path, "rb") as f:
            model.weights_bin = f.read()
```

Then we could do `NcnnModel.load_from_file(param_path, bin_path)` and `ncnn_model = NcnnModel(node_count, blob_count)`, which are both clean APIs.

backend/src/nodes/utils/ncnn_auto_split.py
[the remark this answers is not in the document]
theflyingzamboni: I'm going to do this as a separate PR, because I need to change convenient_upscale, which pytorch and onnx use too.

backend/src/nodes/onnx_nodes.py
theflyingzamboni: Isn't it going to do that regardless? If I convert it to a bool, it will still depend on the values of OnnxFpDropdown.

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: Yes. I'll rename to clarify.

backend/src/nodes/utils/ncnn_model.py
theflyingzamboni: ... Isn't this very similar to what I had in the first place, minus reading from bin_path being in the load_model method?

backend/src/nodes/onnx_nodes.py
RunDevelopment: > Isn't it going to do that regardless?

No. The contract `convert` defines now (after your recent commits) is decoupled from the `OnnxFpDropdown`. If we change the values of `OnnxFpDropdown`, we won't have to change `convert`'s implementation, but the `bool()` cast. So future changes are going to be a lot more localized.

I didn't like that `convert` made assumptions about the values of `OnnxFpDropdown` (global reasoning). These two things should be independent of each other.

backend/src/nodes/utils/ncnn_model.py
RunDevelopment: Not quite. I suggested a [static factory method](https://stackoverflow.com/questions/929021/what-are-static-factory-methods). Static factory methods are essentially static methods that create an instance of their class. They are pretty much constructors with a name, so we don't have the overloading problem.

RunDevelopment (approved):

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -174,14 +174,22 @@ def __init__(
         )
 
     def __getitem__(self, key: int) -> NcnnParam:
-        return self.param_dict[key]
+        try:
+            return self.param_dict[key]
+        except KeyError as e:
+            logger.error(f"Op {self.op} does not have param {key}, please report")
+            raise e
```
RunDevelopment, line 181: This should be [just `raise`](https://docs.python.org/3/tutorial/errors.html#raising-exceptions). Otherwise, we lose the stack trace. The stack trace isn't hugely important here, but, you know, best practices.

```suggestion
            raise
```

backend/src/nodes/utils/ncnn_model.py
```diff
@@ -174,14 +174,22 @@ def __init__(
...
         self, pid: int, value: Union[float, int, List[Union[float, int]]]
     ) -> None:
         idstr = str(pid)
         param_dict = param_schema[self.op]
-        param = param_dict[idstr]
+        try:
+            param = param_dict[idstr]
+        except KeyError as e:
+            logger.error(f"Op {self.op} does not have param {idstr}, please report")
+            raise e
```
RunDevelopment, line 192: Same here please.

backend/src/nodes/onnx_nodes.py
```diff
@@ -180,3 +183,31 @@ def run(
...
         )
+
+
+@NodeFactory.register("chainner:onnx:convert_to_ncnn")
+class ConvertOnnxToNcnnNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = """Convert an ONNX model to NCNN."""
+        self.inputs = [OnnxModelInput("ONNX Model"), OnnxFpDropdown()]
+        self.outputs = [NcnnModelOutput("NCNN Model"), TextOutput("FP Mode")]
```
RunDevelopment, line 194: Could you type the FP mode output? This should be fairly simple and would be a nice introduction into working with the type system.

This will be pretty much analogous to the [`StackOrientationDropdown`](https://github.com/joeyballentine/chaiNNer/blob/3d57f4ddd379b1df32f04c1f3573cf0827ef0a7e/backend/src/nodes/properties/inputs/generic_inputs.py#L133). You need to [define](https://github.com/joeyballentine/chaiNNer/blob/3d57f4ddd379b1df32f04c1f3573cf0827ef0a7e/src/common/types/chainner-scope.ts#L65) an [enum](https://github.com/joeyballentine/chaiNNer/blob/main/src/common/types/navi.md#enum-definitions) for the Fp dropdown and then [match](https://github.com/joeyballentine/chaiNNer/blob/main/src/common/types/navi.md#match) the dropdown variants in the text output.

joeyballentine:
@theflyingzamboni could you fix merge conflicts and also target this PR to the v0.12.0 branch?

joeyballentine (approved):

PR #724, opened 2022-08-12
Color rework

### LINKED ISSUE #719: Get rid of `useColorMode`

RunDevelopment:
This PR completely reworks how themes and colors work.

The main idea is that all colors are now defined in `colors.css`. Chakra variables still exist, and they just reference the CSS variables.

It's not done yet.

---

Resolves #719.


src/renderer/colors.css
```diff
@@ -0,0 +1,53 @@
...
+    --gray-200: #e2e8f0;
+    --gray-300: #cbd5e0;
+    --gray-400: #a0aec0;
+    --gray-500: #718096;
+    --gray-600: #4a5568;
+    --gray-650: #3e485a;
+    --gray-700: #2d3748;
+    --gray-750: #252d3c;
+    --gray-800: #1a202c;
+    --gray-850: #191d28;
```
RunDevelopment, line 14: I had to hard-code the interpolation between 800 and 900. I'm thinking about introducing SCSS, so I can use [`mix`](https://sass-lang.com/documentation/modules/color#mix).

src/renderer/colors.css
```diff
@@ -0,0 +1,53 @@
...
+    --gray-900: #171923;
+
+    /* backgrounds */
+    --bg-800: var(--gray-800);
+    --bg-700: var(--gray-700);
+    --bg-600: var(--gray-600);
+
+    /* node colors */
+    --node-border-color: var(--gray-800);
```
joeyballentine, line 23: Is there a point to having the word "color" in these?

src/renderer/colors.css
```diff
@@ -0,0 +1,53 @@
...
+    --tag-fg: var(--gray-300);
+
+    /* specific colors */
+    --window-bg: var(--gray-900);
+    --header-bg: var(--bg-800);
+    --chain-editor-bg: var(--bg-800);
+    --node-selector-bg: var(--bg-800);
+}
+
+:root[data-theme='light'] {
```
joeyballentine, line 37: I kinda want to not use chakra's light/dark theme switching. I'd rather treat light mode as just an alternative to any other theme, but with light mode colors.

src/renderer/colors.css
RunDevelopment: CSS variables can be anything. So `--something-border: 1px solid #FFF` is what I would expect from a border.

joeyballentine: that makes sense

RunDevelopment:
I think this PR is in a pretty good place now. All `useColorModeValue` calls have been removed, and (almost) nothing depends on the color mode value anymore.

joeyballentine (approved):
you know, i actually like the darker tags

PR #727, opened 2022-08-14
Added support for SCSS

RunDevelopment:
I want to use SCSS in #724, so this PR adds SCSS.

Changes:
- Installed SCSS deps and added SCSS loader webpack config.
- Make ESLint ignore SCSS files.
- Removed unnecessary references to `global.css`.
- Renamed `global.css` to `global.scss` and used a few SCSS features to simplify the styles a little.


src/renderer/global.scss
```diff
@@ -0,0 +1,145 @@
+/* :not(.chakra-slider__track):not(.chakra-menu__menu-button):not(.react-flow__handle)
+:not(.chakra-slider__track):not(.chakra-menu__menu-button):not(.react-flow__handle)
```
joeyballentine, line 2: Can you put the `.chakra-slider__track` rule back? This was there because without it, it makes the slider track not have the pointer hand, even though the surrounding area has the pointer hand.

RunDevelopment: This was already like this before, I didn't comment it out. I only grouped selectors, so no functional changes were made.

joeyballentine: Oh..... Then why tf did I get rid of that?? Lol. Guess it's fine fine

joeyballentine (approved):

PR #729, opened 2022-08-14
process iterator files alphabetically

### LINKED ISSUE #728: Image File Iterator processes files in seemingly random order
We can probably fix this by just sorting the files alphabetically or something

joeyballentine:
Closes #728

### DIFF
--- a/backend/src/nodes/image_iterator_nodes.py
+++ b/backend/src/nodes/image_iterator_nodes.py
@@ -108,7 +108,7 @@ class ImageFileIteratorNode(IteratorNodeBase):
             if context.executor.should_stop_running():
                 return
 
-            for name in files:
+            for name in sorted(files):
                 filepath = os.path.join(root, name)
                 _base, ext = os.path.splitext(filepath)
                 if ext.lower() in supported_filetypes:


backend/src/nodes/image_iterator_nodes.py
```diff
@@ -108,7 +108,7 @@ def walk_error_handler(exception_instance):
             if context.executor.should_stop_running():
                 return
 
-            for name in files:
+            for name in sorted(files):
```
RunDevelopment, line 111: Out of curiosity: what comparison function does this use?

joeyballentine: Not sure. I can't find any documentation about the actual function it uses, but I've found posts saying it's O(nlogn) time complexity

joeyballentine: Aha! It uses timsort

RunDevelopment: That's not what I meant. I want to know the comparison functions. There are a bunch of ways in which you can compare strings (e.g. by byte code, code point, character, etc.). I couldn't find any docs on that, so I thought you might know.

RunDevelopment: > [The sort algorithm uses only < comparisons between items.](https://docs.python.org/3/library/functions.html#sorted)
> [Lexicographical ordering for strings uses the Unicode code point number to order individual characters.](https://stackoverflow.com/questions/4806911/how-are-strings-compared)

Found it. Python's sorting uses the elements `<` like C++ (even in the same duck-typing manner).

Issue #661, opened 2022-07-30
Undo gets "stuck" in text boxes


kimberly990:
Occasionally while using CTRL+Z to undo actions, it will continuously undo text edits within a node (such as the Resize (Factor) node) rather than actually undo node arrangements or modifications.

RunDevelopment:
So I just implemented the fix and I think it's worse. The problem is that you can now undo text edits (basically, it'll uses Chainner's undo and will undo whatever text you just typed + the change before that), but you can't redo it. Since the text edit wasn't committed to history as a change, it can be redone.

RunDevelopment:
I made a PR, so feel free to check it out and see for yourself whether it's better. #745

kimberly990:
In my case, that fix would absolutely be of use. I'm not sure it'd benefit the majority of users though, for the reasons you stated.

RunDevelopment:
@joeyballentine @kimberly990 Can we close this issue now? I just closed the fix PR. As I said before, I think the fix (as in, the behavior desired by this issue) is worse than what we currently have.

kimberly990:
While I agree the existing fix should be closed, this is an ongoing issue. It basically makes the undo function useless if you've edited any text boxes. 

If you believe there are no alternative solutions, then I agree this should be closed.

vekien:
Can I ask why it seems the fix is "worse".

Currently if you edit a textbox, then move the node, then undo to put the node back, it'll also undo the textbox, this seems insane to me? This doesn't happen in blender, or unreal or any other graph node system I've used... It counts a "move" as a separate history action in all other programs. 

This has been causing me huge headache when random input fields keep resetting because I undo a movement.

joeyballentine:
We can probably revisit this soon. I think we have a few issues with undo that are mostly just caused by how we store our state. If we redid that, we could probably fix multiple issues

PR #745, opened 2022-08-17
Undo gets "stuck" in text boxes

RunDevelopment:
This applies the fix for #661 to number inputs. I didn't apply it to text boxes so far.

### DIFF
--- a/src/common/safeIpc.ts
+++ b/src/common/safeIpc.ts
@@ -173,3 +173,6 @@ export interface BrowserWindowWithSafeIpc extends BrowserWindow {
 
 export const ipcMain = unsafeIpcMain as SafeIpcMain;
 export const ipcRenderer = unsafeIpcRenderer as SafeIpcRenderer;
+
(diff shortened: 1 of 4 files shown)


joeyballentine:
@RunDevelopment what is this waiting on?

RunDevelopment:
Feedback on whether we actually want to fix, [because I think that the fix makes the UI worse](https://github.com/joeyballentine/chaiNNer/issues/661#issuecomment-1217789505).

joeyballentine:
Ah, that does seem worse

joeyballentine:
Could we commit text box changes to the undo history? I think that would solve the issue and make this not as bad of a change. Otherwise I think we should just close this PR

RunDevelopment:
> Could we commit text box changes to the undo history

No. As Kim rightly pointed out, that's just a bad UX. We don't do that for other inputs (number) either, so it would be inconsistent on top of that.

> Otherwise I think we should just close this PR

I agree. Let's just close it.

PR #763, opened 2022-08-19
Fix PYTHONHOME variable deletion causing crash

### LINKED ISSUE #749: after installed the 0.11.1 exe on windows,failed to fetch

### LINKED ISSUE #752: critical error - failed to fetch with ChaiNNER 0.11.1 on Mac OS 10.14.6 Mojave
"chaiNNer  - has encountered a critical error - failed to fetch" error message.

I clicked "allow" network collection.
After downloading app components and loading dependencies etc
A screen opened with an error message.
Screenshot 2022-08-18 at 16 32 41

joeyballentine:
For sure fixes #752 and fixes #749. Might fix the others as well, but I'm not entirely sure.

### DIFF
--- a/src/common/env.ts
+++ b/src/common/env.ts
@@ -6,5 +6,7 @@ export const isM1 = isMac && (os.cpus()[0]?.model.includes('Apple M1') ?? false)
 export const isRenderer = typeof process !== 'undefined' && process.type === 'renderer';
 
 const env = { ...process.env };
-delete env.PYTHONHOME;
+if (env.PYTHONHOME) {
+    delete env.PYTHONHOME;
+}
 export const sanitizedEnv = env;


RunDevelopment:
Could you explain what this does?

joeyballentine:
> 

#583. Turns out the fix only works when the variable exists, otherwise it crashes

RunDevelopment:
That doesn't make any sense. If you delete a key that doesn't exist in the object, then [`delete` won't do anything](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete). So why does it crash?

joeyballentine:
You're right. I just retested it and it turns out this does nothing. I must have accidentally switched between doing npm run dev and npm start, because run dev still crashes like that (but start doesn't)

PR #766, opened 2022-08-19
Update react-flow

joeyballentine:


RunDevelopment:
So far, each update of RF has broken *something*. Is there a reason for updating?

joeyballentine:
> So far, each update of RF has broken _something_. Is there a reason for updating?

We don't need to merge this yet, but I want to clean up the node search context menu now that the onConnectStop state thing is fixed

RunDevelopment:
Okay, then let's merge this after the macOS fixes have been released.

PR #767, opened 2022-08-19
Fix pytorch install on mac

### LINKED ISSUE #747: Pytorch not installing on macos
Clicking "Pytorch not installed" opens up dependency manager, clicking button next to "Pytorch Not installed" in dependency manager outputs Looking in links download.pytorch.org/blahblah in console but nothing happens for hours. You can restart as many times as you want, modules are not being downloaded.

joeyballentine:
Fixes #747

This also updates the PyTorch dep to use the new "--extra-index-url" flag rather than "--find-links" (-f).

### DIFF
--- a/src/common/dependencies.ts
+++ b/src/common/dependencies.ts
@@ -18,51 +18,46 @@ export interface Dependency {
     packages: PyPiPackage[];
 }
 
-const getTorchKind = (isNvidiaAvailable: boolean) => {
-    if (isMac) return '';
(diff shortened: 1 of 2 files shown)


joeyballentine:
> Why does this work?

The +CPU we had there before didn't work with macos for some reason. Now that I've switched to the extra index url thing, +CPU isn't necessary at all

RunDevelopment (approved):

PR #771, opened 2022-08-19
Ensure we only handle dom-ready once

joeyballentine:


RunDevelopment:
This won't actually fix the problem. The `'dom-ready'` listener is called twice. With this "fix", you just moved the error. Now `registerEventHandlers()` will fail.

RunDevelopment (approved):

PR #809, opened 2022-08-22
Update opencv, numpy, and pillow

joeyballentine:
Also, run the pip update on every open. it will install whatever isn't installed, and it will update whatever needs updating, but it will do nothing if there isn't anything to do, so there's no point in not doing it (plus its easier than writing update check logic)

### DIFF
--- a/src/common/safeIpc.ts
+++ b/src/common/safeIpc.ts
@@ -62,8 +62,7 @@ export interface SendChannels {
     'file-save': SendChannelInfo;
     'file-export-template': SendChannelInfo;
     'finish-loading': SendChannelInfo;
-    'installing-deps': SendChannelInfo;
-    'updating-deps': SendChannelInfo;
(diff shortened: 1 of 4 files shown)


src/common/safeIpc.ts
```diff
@@ -63,6 +63,7 @@ export interface SendChannels {
     'file-export-template': SendChannelInfo;
     'finish-loading': SendChannelInfo;
     'installing-deps': SendChannelInfo;
+    'updating-deps': SendChannelInfo;
```
RunDevelopment, line 66: Instead of adding a new channel and duplicating the `.on(handler)` code, could you add a parameter to `installing-deps` instead?
```suggestion
    'installing-deps': SendChannelInfo<[onlyUpdating: boolean]>;
```

RunDevelopment (approved):

Issue #811, opened 2022-08-23
Use bfloat16 if RTX gpu


joeyballentine:
I just learned torch has a special dtype for RTX GPUs called "Boosted FP16" which apparently has much better performance.,

We should check whether the GPU name contains "RTX", and if it does send a "use_bfp16" flag to the backend's execution options. Then we can pick that for the autocasting instead of regular fp16.

PR #812, opened 2022-08-23
SwinIR Support

joeyballentine:
- Adds SwinIR support for all models
- The gray models are currently broken, but will be fixed by #721
- The "classical sr" models are unable to use fp16, hence the new flags in each model. However, they are able to use bfloat16, so we should add support for that, which is why I suggested #811 
- I also switched from manually casting the model and image to half/float to just using the autocast context manager pytorch provides. This ensures that only operations that are actually able to be cast to fp16 become fp16. Unfortunately, this did not fix the issue with the "classical sr" models, but it is still good to have.


backend/src/nodes/properties/outputs/pytorch_outputs.py
```diff
@@ -21,6 +21,19 @@ def get_broadcast_data(self, value: PyTorchModel):
...
         if "SRVGG" in value.model_type:  # type: ignore
             size = [f"{value.num_feat}nf", f"{value.num_conv}nc"]
+        elif "SwinIR" in value.model_type:  # type: ignore
+            head_length = len(value.depths)  # type: ignore
+            if head_length <= 4:
+                size_tag = "small"
+            elif head_length < 9:
+                size_tag = "medium"
+            else:
+                size_tag = "large"
```
RunDevelopment, line 31: This seems arbitrary. Could you explain this?

RunDevelopment (commented):
Lots and lots of additions in this PR. Frankly, all that stuff is over my head anyway, so I'm just gonna say that it's probably fine.

backend/src/nodes/properties/outputs/pytorch_outputs.py
joeyballentine: https://github.com/JingyunLiang/SwinIR/releases/tag/v0.0

The S, M and L in the model names

RunDevelopment: I guess it's fine then.

joeyballentine:
Just realized I need to update the readme with this

RunDevelopment (approved):
Should be fine.

Issue #813, opened 2022-08-23
Error report for chaiNNer 0.11.3


blindorange:
Error:
```UnhandledRejection TypeError: Object has been destroyed
    at n.loadURL (node:electron/js2c/browser_init:33:1843)
    at /Applications/chaiNNer.app/Contents/Resources/app/.webpack/main/index.js:2:77327
```
OS: macOS 10.17

blindorange:
Worked fine second attempt

PR #815, opened 2022-08-23
Safe window load

### LINKED ISSUE #800: Error report for chaiNNer 0.11.3
Error:
```UnhandledRejection TypeError: Object has been destroyed
    at n.loadURL (node:electron/js2c/browser_init:33:1843)
    at /Applications/chaiNNer.app/Contents/Resources/app/.webpack/main/index.js:2:77327
```
OS: macOS 10.13

joeyballentine:
This is an attempt to fix #800 and #813. Replacing the splash screen will take some time, so for now let's wrap the destroy calls with a try/catch as well as an if statement to check if the windows have been destroyed already. Theoretically, it will fix the issues if the problem really is that it is running the splash 'onclose' handler after the window has been destroyed

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -581,7 +581,11 @@ const doSplashScreenChecks = async () =>
             splash.webContents.send('finish-loading');
             splash.on('close', () => {});
             await sleep(500);
-            splash.destroy();
+            try {
+                splash.destroy();
+            } catch (error) {
+                log.error(error);
+            }
             mainWindow.show();
             if (lastWindowSize?.maximized) {
                 mainWindow.maximize();
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -536,7 +536,11 @@ const doSplashScreenChecks = async () =>
         });
 
         splash.on('close', () => {
-            mainWindow.destroy();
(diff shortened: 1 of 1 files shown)


RunDevelopment:
> This is an attempt to fix #800 and #813

The problem with the linked PRs is **not** that we fail to destroy a window, but that we already destroyed the window and then try to use it again (in this case, we try to load a URL).

joeyballentine:
Oh well in that case we could surround the loads with a !isDestroyed check, right?

RunDevelopment:
Will that help? We only load because we want to show the window.

The root cause is that the splash gets closed before `dom-ready` calls `resolve()`. No idea how to fix that though.

Maybe you could remove `splash.on('close', () => {
            mainWindow.destroy();
        });`? I mean, why did you even add it to begin with?

RunDevelopment (approved):
Let's try it out, I guess.

modulator:
Thanks Joey! FYI, ChaiNNer version 0.11.4 seems to be working now!  The dependencies took a while (prob around 1hr) to fully install, but the app seems to be running very stable now! Thanks SO much for this great tool!

PR #869, opened 2022-08-30
Add Canny Edge Detection node

JumpyJacko:
Please tell me if there's a problem with having to un-normalise and then normalise the values. It's my first time working with OpenCV so I'm still learning.


backend/src/nodes/image_util_nodes.py
```diff
@@ -554,3 +554,37 @@ def run(
...
+
+    def run(
+        self,
+        img: np.ndarray,
+        t_lower: int,
+        t_upper: int,
+    ) -> np.ndarray:
+        
+        def unnormalize(img: np.ndarray) -> np.ndarray:
+            img *= 255/img.max()
```
joeyballentine, line 582: Why do you divide 255 by img.max here?

theflyingzamboni: Yeah, this is not correct. Dividing by img.max() sets the maximum pixel value to 1 and scale all the pixel values, shifting the image color. This should be `img = (img * 255).astype(np.uint8)`, as cv2.Canny simply requires the image to be in 8-bit color. I would just write this as a single line of code as well and skip the overhead on a method call.

theflyingzamboni:
Oh, this should be PR'd to the 0.12.0 branch, too.

JumpyJacko:
Oh alright, thanls for letting me know

JumpyJacko:
Oh alright, thanks for letting me know

backend/src/nodes/image_util_nodes.py
```diff
@@ -554,3 +554,32 @@ def run(
...
+
+@NodeFactory.register("chainner:image:canny_edge_detection")
+class CannyEdgeDetectionNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Detect the edges of the input image and output as black and white image."
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Lower Threshold", minimum=0, default=100),
+            NumberInput("Higher Threshold", minimum=0, default=300),
```
theflyingzamboni, line 566: For grammatical context reasons, I would change 'Higher' to 'Upper', but other than that, the code itself looks good.

theflyingzamboni (commented):
Forgot to mention, there's one more thing you need to do for this to be pulled. You need to save a sample chain with this node in it and save it to `chaiNNer/tests/data`. After you do that, run `npm run test-js -- -u` to update the save test snapshot.

theflyingzamboni (approved):

joeyballentine (approved):

PR #873, opened 2022-08-31
Use imwrite when possible

joeyballentine:
Solves a bit of ram overhead

### DIFF
--- a/backend/src/nodes/image_nodes.py
+++ b/backend/src/nodes/image_nodes.py
@@ -6,9 +6,11 @@ from __future__ import annotations
 
 import os
 import platform
+import random
 import subprocess
 import time
 from tempfile import mkdtemp
+import string
 
 import cv2
 import numpy as np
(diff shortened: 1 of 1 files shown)


backend/src/nodes/image_nodes.py
```diff
@@ -135,10 +135,15 @@ def run(
 
         os.makedirs(base_directory, exist_ok=True)
 
-        status, buf_img = cv2.imencode(f".{extension}", img)
-        with open(full_path, "wb") as outf:
-            bytes_written = outf.write(buf_img)
-            status = status and bytes_written == len(buf_img)
+        # Write image with opencv if path is ascii, since imwrite doesn't support unicode
+        # This saves us from having to keep the image buffer in memory, if possible
+        if full_path.isascii():
```
0x4E69676874466F78, line 140: What if on windows try to convert path in UTF-8?
https://github.com/opencv/opencv/issues/4292#issuecomment-601686965

joeyballentine: image

PR #886, opened 2022-09-03
Discord RPC integration

BigBoyBarney:
A rudimentary js Discord Rich Presence integration


src/renderer/components/SettingsModal.tsx
```diff
@@ -348,6 +370,7 @@ export const SettingsModal = memo(({ isOpen, onClose }: SettingsModalProps) => {
                             <Tab>Environment</Tab>
                             <Tab>Python</Tab>
                             <Tab>Advanced</Tab>
+                            <Tab>DiscordRPC</Tab>
```
joeyballentine, line 373: I would probably put this setting in an existing tab rather than making it its own tab. After all, there's only ever going to be one setting for it. Maybe just put it in Advanced?

src/discordRPC.js
```diff
@@ -0,0 +1,46 @@
...
+export function updateDiscordRPC(detailsArg, largeImageKeyNameArg, startTimestampArg){ // can be expanded with other params. Call this when you want to update the RPC.
+	detailsArg = detailsArg || config.details;
+	largeImageKeyNameArg = largeImageKeyNameArg || config.largeImageKeyName;
+	startTimestampArg = startTimestampArg || startTimestamp;
+	client.setActivity({
+		details: detailsArg,
+		largeImageKey: largeImageKeyNameArg,
+		startTimestamp: startTimestampArg,
+	});
+	console.log("updated RPC")
```
joeyballentine, line 31: Please remove the console logs

src/discordRPC.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function hideDiscordRPC(){
+	client.clearActivity();
+	console.log("hidden RPC")
```
joeyballentine, line 36: Ditto

src/discordRPC.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function showRPC(show){
```
joeyballentine, line 39: This should probably be called toggleRPC or something, since it does more than just show it

src/discordRPC.js
```diff
@@ -0,0 +1,46 @@
+const rpc = require('discord-rpc');
+const config = {
+	// Joey pls upload the assets to (https://discord.com/developers/applications -> Rich Presence -> Art Assets)
+	appId: "1015217737724874803", //ID
+	details: "ChaiNNing something cool!", //Large text under ChaiNNer
```
joeyballentine, line 5: Do most apps start with a message like this? Or do they just have a blank message unless something is actually happening?

BigBoyBarney: You don't have to set anything at all, it can run without them just fine. You could technically start it without an explicit largeImage too, but then it defaults to whatever you have set on the Discord Devportal. This is the most barebones version that runnable:
image

PR #890, opened 2022-09-04
Fix in_nc detection for static axes

theflyingzamboni:
Forgot to actually PR this. Please test it out on some models (non-chaiNNer-converted ones too) to make sure it doesn't break for any of them. It's worked on all the ones I've tried, but it was only six or so.


joeyballentine:
I'm fine with just merging it tbh

PR #897, opened 2022-09-05
Discord RPC Integration 

joeyballentine:
My personal additions and fixups on top of @BigBoyBarney's PR. I figured instead of making them learn TS and how we do electron IPC stuff, I'd just handle it. Plus, their PR is borked. So now it's all just in this one.

Additions:
- Swapped to my app ID
- The description gets updated when you clear/open/save a chain
- Properly use async/await and TypeScript
- RPC toggle remembers the previous description

image
image

Open questions:
- Should the time reset when opening/clearing a chain?
- Should we have a different description for when the chain is running?


joeyballentine:
CI fails on this one too... Going to see if rebasing & squashing those commits fixes it

BigBoyBarney:
- I think having a global timer on uptime (not resetting) looks better in general, especially when swapping chains a lot.
- Most of the time chains don't run for very long so changing description for that might be redundant. Having a timer / current status for _iterators_ would be really nice though. The iterator node already tells us where it's at at any given time, so having that as description, like: "Finished 53/98 files" could be cool :D

Thanks for cleaning up my code Joey!! <3

joeyballentine:
Doing stuff for iterators would be cool, especially if we made one of the images some sort of iterator icon and whatnot. That's gonna be a bit complicated though (since I'd have to worry about reverting to the previous "idle" state), so I'll do that in a separate PR.

PR #908, opened 2022-09-06
Add modulo to math node

joeyballentine:
image


joeyballentine:
Requesting a review from @RunDevelopment since I probably did something wrong related to the Navi changes

joeyballentine:
You know, it's good enough for now I guess. Can always fix it later

Issue #798, opened 2022-08-20
Fit image to (edge)


CharonM72:
Requesting a feature to fit the image to a specific width or height, or to fit the image within specific boundaries - in simpler terms, like in Windows etc. when setting a desktop wallpaper, there are various options to:
- Fill: resizes the image so the entire image fits inside the specified resolution, creating empty borders on the edge that are too small
- Fit: resizes the image so the entire image fills the specified resolution, cutting off the edges of the image that are too large
- Stretch: I think this is already implemented in resize (resolution)
- Tile: I think? already implemented in tile fill
- Center: already implemented in a sense in crop (border)

This is related to #777 but it does not necessarily need to make a canvas of a different size than the image. For instance:

I have a large folder of images of various sizes. I want it to be as large as possible, without exceeding the dimensions of 3840x2160. Let's say I have one image that is 1024x1024, and another that is 1920x768.
First, I upscale the images by 4x, so the first becomes 4096x4096 and the second becomes 7680x3072.
I use this node, with the "Fit" setting. The first image is resized to 2160x2160, and the second is resized to 3840x1536. No excess canvas border would necessarily have to exist.
Now let's say I want the image to fill the _whole_ resolution, by cutting off excess edges.
By doing the same thing with the "Fill" setting, the first image is resized to 3840x2160 (cutting off the top and bottom edges), and the second one is also resized to 3840x2160 (cutting off the left and right edges). Additionally, there could be an option to resize to those resolutions and _not_ cutting off the edges.

Here's a visual representation of what I am thinking (with GIMP, dotted red line is the canvas border):
<details>
  <summary>Click to expand</summary>
Fill (without cropping):

![image](https://user-images.githubusercontent.com/1299220/185769371-446d0156-313d-4282-99ba-6945ef45f575.png)

Fill (with cropping):
![image](https://user-images.githubusercontent.com/1299220/185769155-a9c253ac-c098-48d1-8147-0e286b9dde1b.png)

Fit:
![image](https://user-images.githubusercontent.com/1299220/185769174-fd853666-942f-4188-a1e0-a81dac63de90.png)

Stretch (already implemented):
![image](https://user-images.githubusercontent.com/1299220/185769264-9663e139-257a-41fa-8bdb-782612ad01d3.png)

Tile (already implemented):
![image](https://user-images.githubusercontent.com/1299220/185769330-a4f4abe4-b06e-4daf-865b-ea81c6743d51.png)

Center (kind of implemented?):
[...]

PR #910, opened 2022-09-06
Resize to side node

BigBoyBarney:
Added a new node that resizes an image while keeping aspect ratio. Resize targets are width, height, shorter side and longer side.
Previously this was only possible through the combination of 3 math nodes and a resize (factor) node


BigBoyBarney:
Also this partially fulfils #798

BigBoyBarney:
build checks fail after my recent "Resize Condition" push. It was fine before. Hmmm

joeyballentine:
Holy shit that's a large type definition... Maybe we should start putting these in their own files? @RunDevelopment would need to take care of that though. Maybe we can worry about that when we refactor how node files work.

Btw, there's a lot of repeated code in this. Consider extracting out the repeated stuff into their own methods.

joeyballentine (approved):

Issue #687, opened 2022-08-03
Add a "Copy to Clipboard" node


RunDevelopment:
When I edit images, it's sometimes easier to do certain tasks in chainner. E.g. tiling a texture to a certain size. To get the final image into my image editing software, I use the View Image (External) node and then copy the opened image to paste into a new layer in the image I'm working on.

It would be nice to have a node that copies an image to clipboard to speed up my workflow.

joeyballentine:
I think this should be possible by having an image clipboard output, and using a full image broadcast where instead of a preview it gets copied to clipboard. That is, unless python supports copying to clipboard as well

RunDevelopment:
I'd rather not send a 100MB PNG via JSON to the frontend just to write it to the clipboard. This is something the backend has to do.

joeyballentine:
Yeah, I just didn't know if python could copy to clipboard. Since it can, we definitely should do it through that

RunDevelopment:
This has been solved by #920.

PR #920, opened 2022-09-08
Add Copy to Clipboard node

Sryvkver:
Added a Node to copy a Image to the clipboard.

Addresses Issue #687


requirements.txt
```diff
@@ -5,11 +5,13 @@ onnx
 onnxoptimizer
 onnxruntime
 opencv-python
+pasteboard ; sys_platform=="darwin"
 Pillow
 protobuf==3.16.0
 pylint
 pyright==1.1.249
 pytest
+pywin32 ; sys_platform=="win32"
```
joeyballentine, line 14: You'll also need to add the extra dependencies to dependencies.ts in order for them to be auto-installed on startup. Requirements.txt is just for local development installation.

README.md
```diff
@@ -72,6 +72,8 @@ You can right-click in the editor viewport to show an inline nodes list to selec
 
 - Arch Linux users may need to manually install libxcrypt before chaiNner's integrated Python will correctly start up.
 
+- Linux users need to have xclip or, for wayland users, wl-copy installed to be able to use the clipboard functionality.
```
joeyballentine, line 75: I'm wondering if this should maybe note that it's specific to the clipboard functionality that's performed by nodes, otherwise people might assume that regular copying and pasting doesn't work

backend/src/nodes/utils/clipboard/clipboard_win32.py
```diff
@@ -0,0 +1,15 @@
...
+
+class WindowsClipboard(ClipboardBase):
+    def __init__(self) -> None:
+        if win32clipboard is None:
+            raise Exception("pywin32 must be installed to use this library on Windows platform.")
+
+    def copy_image(self, img: bytes) -> None:
+        win32clipboard.OpenClipboard()
+        win32clipboard.EmptyClipboard()
+        win32clipboard.SetClipboardData(49472, img)  # type: ignore # 49472 -> PNG format
```
joeyballentine, line 14: Could you make this a constant at the top of the file rather than a magic number?

joeyballentine (commented):
This is awesome, thanks for working on this! Just a few minor things.

requirements.txt
joeyballentine: It's [this](https://github.com/joeyballentine/chaiNNer/blob/main/src/common/dependencies.ts) file by the way.

backend/src/nodes/utils/clipboard/clipboard_win32.py
Sryvkver: Thanks for the feedback.
Updated all the mentioned issues, hopefully it all fits now.

joeyballentine:
So I just tried this out. It doesn't seem to actually work on windows. I can't paste the image into anything. I'm assuming that's because you're encoding the image as a PNG rather than a raw bitmap?

Sryvkver:
Mhhh that's odd, I did test it on my machines (Mac and Windows 11), but maybe something broke with my last commit.

Will check it out, once I'm home.
Sorry about that.

Sryvkver:
Not home yet, but I figured something out.
You are right it is because I am encoding it as a PNG, which is not a offically supported Windows clipboard format.
The magic number I got from my machine, was a previously registerd PNG format.

I could try to convert it to a bitmap, however I read that this could destroy the alpha channel for some applications, which i would like to avoid.
Another option would be to [register a custom clipboard format](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerclipboardformata), but this would problably take some time to understand and implement.

0x4E69676874466F78:
@Sryvkver use CF_DIBV5 it supports png.
https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapv5header

joeyballentine:
By the way, is there any reason in particular you're using PIL to encode rather than opencv? IIRC opencv tends to be faster but I don't know for sure

Sryvkver:
No particular reason.
I just saw it being used in another file and tried using it. Didn't know that opencv was faster.

joeyballentine:
I haven't done a comparison myself for encoding but I know for saving specifically it's faster. Also, less time taken needing to convert to a PIL array and whatnot

joeyballentine:
We use PIL pretty much only in places where opencv is not suitable, such as for resizing (since opencv's resign algorithms do not do antialiasing)

Sryvkver:
Good to know, will definitely look into that.

Sryvkver:
Hopefully the last commit needed to get this done, everything seemed to work fine on my end.
@joeyballentine If you notice anything wrong, please let me know.

joeyballentine:
I know this is a bit of a nitpick, but could you change the subcategory (**not** the regular category) to "utility"? I just don't like that alphabetically it goes above "Load Image" when it's in I/O lol. Also, pyright is complaining about broken imports. I think the only way you'll be able to fix that is by wrapping them in a try/catch.

Sryvkver:
thb, it did annoy me aswell, but I wasn't to sure what the best way was to change it.
Ye saw the complain from pyright, kinda sad about it. Will fix asap.

joeyballentine:
Whoops, I accidentally created a conflict in the README. Fix that and this should be good to go 👍. 

Only thing that's too bad about this is that (at least on windows) copying and pasting into a Discord message doesn't work, but that might be unavoidable. It at least works pasting into image processing stuff, which is good enough imo.

joeyballentine:
I'm good with this, but I would like to get @RunDevelopment's opinion before merging.

RunDevelopment (changes requested):
Amazing work. Just 2 minor nits.

Also, it is ridiculous that this amount of work is even necessary to achieve something so simple. There isn't a good library we could use. The best one is pyperclip which only supports text.

While this shouldn't be part of this PR, we might want to move all the clipboard code into its own cross-platform clipboard package that supports text and images. That way, other people can also use it and even contribute to it (gotta leverage the open source community, you know).

backend/src/nodes/properties/inputs/generic_inputs.py
```diff
@@ -85,6 +86,29 @@ def toDict(self):
...
+    """Input for pasting from clipboard"""
+
+    def __init__(self, label: str = "Clipboard input"):
+        super().__init__(["Image", "string", "number"], label, kind="text-line")
+        self.input_conversion = """
+            match Input {
+                Image as i => "Copying image",
+                _ as i => i,
+            }
+        """
```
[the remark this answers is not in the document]
Sryvkver: Good Idea, definitely look better this way.

joeyballentine (approved):

Sryvkver:
Thank you.

Good Idea, for this to become its own seperate package, might look into it.
At the moment I wouldn't really have the time to do that tho.

But if you got some free time, please go for it.

RunDevelopment (approved):

PR #934, opened 2022-09-12
Refactor dep installing on startup to only run when needed, and support auto-updating specific optional dependencies

joeyballentine:
This was needed for two reasons: 
1. The "updating deps..." step of the splash screen sometimes took a while, even when it wasn't doing anything.
2. NCNN really should be auto-updated since implementation details change surrounding things added to it. This means that at any point in time an update could break NCNN functionality unless the user manually updates.


src/main/main.ts
```diff
@@ -341,23 +346,83 @@ const checkPythonEnv = async (splashWindow: BrowserWindowWithSafeIpc) => {
...
-        const pending = requiredDependencies.filter((dep) => {
-            return dep.packages.some((pkg) => {
-                if (installedPackages.has(pkg.packageName)) return false;
-                log.info(`Dependency ${dep.name} (${pkg.packageName}) not found.`);
-                return true;
-            });
+        let requiredIndividualPackages: PyPiPackage[] = [];
+        requiredDependencies.forEach((dependency) => {
+            requiredIndividualPackages = [...requiredIndividualPackages, ...dependency.packages];
+        });
```
RunDevelopment, line 367:
```suggestion
        const requiredIndividualPackages = requiredDependencies.flatMap((dep) => dep.packages);
```

Same for `optionalInvividualPackages` below.

src/main/main.ts
```diff
@@ -341,23 +346,83 @@ const checkPythonEnv = async (splashWindow: BrowserWindowWithSafeIpc) => {
...
+
+        let optionalInvividualPackages: PyPiPackage[] = [];
+        getOptionalDependencies(hasNvidia).forEach((dependency) => {
+            optionalInvividualPackages = [...optionalInvividualPackages, ...dependency.packages];
+        });
+
+        let isInstallingRequired = false;
+        let isUpdating = false;
```
RunDevelopment, line 375: Please move those two further down and make them this:
```suggestion
        const isInstallingRequired = missingRequiredPackages.length > 0);
        const isUpdating = outOfDateRequiredPackages.length > 0 || outOfDateOptionalPackages.length > 0);
```

RunDevelopment (changes requested):

src/main/main.ts
```diff
@@ -576,8 +643,8 @@ const doSplashScreenChecks = async () =>
             await sleep(250);
 
             splash.webContents.send('checking-deps');
-            await checkPythonDeps(splash);
-            await checkNvidiaSmi();
+            const hasNvidia = await checkNvidiaSmi();
```
RunDevelopment, line 646: This check plus its side effects don't need python at all, right? If not, then we could start the promise at the top of the function and only await the result here. That way, the result is likely already ready when we need its result, and we won't have to wait.

src/main/main.ts
joeyballentine: This is why I've missed your reviews :)

src/main/main.ts
joeyballentine: That is correct, we could do that

RunDevelopment (approved):

PR #937, opened 2022-09-12
Fix ONNX saving

joeyballentine:
Messed up in the last merged PR. Did not type properly using ort.InferenceSession and instead used Any just to avoid having another import, plus the session is not needed to save.

### DIFF
--- a/backend/src/nodes/model_save_nodes.py
+++ b/backend/src/nodes/model_save_nodes.py
@@ -1,5 +1,6 @@
 # These sad files have to be all on their own :(
 from __future__ import annotations
+from typing import Any
 from sanic.log import logger
 
 from .node_base import NodeBase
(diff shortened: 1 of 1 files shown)


backend/src/nodes/model_save_nodes.py
```diff
@@ -32,11 +33,14 @@ def __init__(self):
 
         self.side_effects = True
 
-    def run(self, onnx_model: bytes, directory: str, model_name: str) -> None:
+    def run(
+        self, onnx_model: Tuple[Any, bytes], directory: str, model_name: str
```
RunDevelopment, line 37: How is this correct? The `OnnxLoadModelNode` nodes returns `bytes`, so why is this suddenly `Tuple[Any, bytes]`?

joeyballentine: It does not, anymore 

https://github.com/joeyballentine/chaiNNer/blob/00cde1256fa1d07f0c3cfa012a8a7bbec95c5ee5/backend/src/nodes/onnx_nodes.py#L106

You must be looking at outdated code

RunDevelopment: Indeed, I was.

RunDevelopment:
Just an idea. Instead of changing everything to tuples and then changing everything again the next time you change anything with ONNX, how about making a class for ONNX models?

joeyballentine:
@theflyingzamboni wants to make model wrapper classes for all the model stuff, so I'll let him do that

Issue #928, opened 2022-09-10
Prevent restricted types from being connected until valid type is determined


joeyballentine:
This may or may not be a good idea, but right now you can connect a node without a type to a node with a restricted type, then do things that determine that node's type to something that would normally not let you connect it. 

So basically I just think we shouldn't allow connections unless the type is determined to be a valid type.

This may cause problems for untypable things though, so I don't know if it's actually a good idea to do.

RunDevelopment:
Could you give an example?

Right now, I image a case with 3 nodes A, B, and C. C only accepts images with exactly 3 channels. B outputs an image with the same number of channels as its input image. A produces an image with 1 channel. If you connect B to C, and after that A to B, then chainner will allow the connections, but the chain will now error. Is this what you mean?

joeyballentine:
Yes exactly. If you connected A to B first, then you wouldn't be allowed to connect B to C. But in the way you described, you can bypass the restriction and get a runtime error

RunDevelopment:
> and get a runtime error

No, the type system will give a static error.

![image](https://user-images.githubusercontent.com/20878432/189674033-1bcfbade-190b-4278-9bf0-1438521a1946.png)

joeyballentine:
That only applies when the output type is determined by the input given (and therefore has a never type).

If for example you were doing a restriction on model type and the output is merely determined by the models scale, then this would not apply.

PR #939, opened 2022-09-12
Report all type mismatches

RunDevelopment:
Fixes https://github.com/joeyballentine/chaiNNer/issues/928

---

As discussed on discord.

All input type error will now be reported. Even if a nicely formatted error message is not available, an error message will be generated.

Since type errors are quite technical in nature, I expect that we will likely tweak the `printErrorTrace` method in the future to further improve user-friendliness.

Example: Here is the same error, once with a generated error trace and once with a nice error message.

image
image


RunDevelopment:
@joeyballentine, does this fix #928?

joeyballentine:
Yes I would say so.

joeyballentine (approved):

PR #940, opened 2022-09-12
Increase max VRAM when estimating (PyTorch)

joeyballentine:
I originally made it split one more time when the usage was calculated to be more than half of the usable VRAM. This was to account for any inaccuracies in the estimation. 

However, I'm pretty sure this can be increased now, especially since we now allow it to OOM & split again once before truly erroring.


RunDevelopment (approved):
Let's see how it goes.

PR #947, opened 2022-09-13
Fix NCNN "Fatal blob allocator error" and use new acquire allocator methods

### LINKED ISSUE #913: vkQueueSubmit failed on most images larger than 512px
**Information:**
- Chainner version: 12.2 Alpha
- OS: Windows 10

**Description**
Errors occur after processing images with iteration, or when processing a single image by itself. Usually only happens when the image is larger than 512 pixels on at least one dimension. If I downscale the images first so they are never larger than 512px in any direction I can process all images without problem. Sorry if this is a duplicate issue reported elsewhere. I looked and saw several people reporting similar problems, but with slightly different context, and mostly on MacOS. This is occurring for me on Windows10.

An error occurred in a Image File Iterator node:

```
Errors occurred during iteration: 
• An unexpected error occurred during NCNN processing.
• vkQueueSubmit failed
```

**Logs**
[renderer.log](https://github.com/joeyballentine/chaiNNer/files/9501076/renderer.log)
[main.log](https://github.com/joeyballentine/chaiNNer/files/9501077/main.log)

joeyballentine:
Didn't get an answer from the ncnn people (of course) about what's better to use, so I'm going to try out the new ones and hope they work.

As for the other fix, all I had to do was not forcefully clear vram until the very end

I'm hoping that this will fix #913 but I'm not sure it will


backend/src/nodes/ncnn_nodes.py
```diff
@@ -87,8 +93,8 @@ def upscale(
                 staging_vkallocator=staging_vkallocator,
                 max_depth=tile_mode if tile_mode > 0 else None,
             )
-            # blob_vkallocator.clear() # this slows stuff down
-            # staging_vkallocator.clear() # as does this
+            blob_vkallocator.clear()
+            staging_vkallocator.clear()
```
RunDevelopment, line 97: Memory won't be cleared if `ncnn_auto_split_process` throws. Can you use `with` on the allocators?

joeyballentine: I don't think so, but I'll check. We could make a custom context manager for it

joeyballentine: Confirmed it does not have one in the pybind. We can either make our own context manager for it, or just check if they are not none in the except and call clear on them there

RunDevelopment: I'd rather have a context manager. There's no need to do `try except` when python has a language feature for exactly this purpose.

joeyballentine: Well luckily that's what I did :)

RunDevelopment: Oh

backend/src/nodes/ncnn_nodes.py
```diff
@@ -25,6 +25,24 @@
 from .model_save_nodes import NcnnSaveNode
 
 
+class AllocatorContextManager:
```
RunDevelopment, line 28: This still isn't correct. If the second allocator fails for some reason, then the first one doesn't get cleared. If either one fails, then its field will be None and `__exit__` will deref null.

I would do it like this:

```py
from contextlib import contextmanager

@contextmanager
def managed_blob_vkallocator(vkdev: ncnn.VulkanDevice):
    try:
        blob_vkallocator = vkdev.acquire_blob_allocator()
    except:
        blob_vkallocator = ncnn.VkBlobAllocator(vkdev)
    try:
        yield blob_vkallocator
    finally:
        blob_vkallocator.clear()

@contextmanager
def ncnn_allocators(vkdev: ncnn.VulkanDevice):
    with managed_blob_vkallocator(vkdev) as blob_vkallocator:
        try:
            staging_vkallocator = vkdev.acquire_staging_allocator()
        except:
            staging_vkallocator = ncnn.VkStagingAllocator(vkdev)
        try:
            yield blob_vkallocator, staging_vkallocator
        finally:
            staging_vkallocator.clear()
```

More verbose, yes, but it's guaranteed to call `clear`.

Also, what does `clear` do anyway? I assumed that it's leaking memory if it doesn't get called.

joeyballentine: Correct. It holds onto vram without it

RunDevelopment (approved):

Issue #926, opened 2022-09-10
Notify used before downloading integrated python


joeyballentine:
**Motivation**
Many people do not want to use the integrated python, plus it tends to fail to download. 

**Description**
We should have a message like "Integrated python not found. Would you like to download?" which would allow them to bypass it and turn "use-system-python" on. Then we should also notify after they click that saying "chainner will now install it's required dependencies to your system python installation, would you like to proceed?" With an option to switch back to integrated, just in case.

We also should detect when integrated python fails to download, and show an appropriate error message with a retry option. 

Basically we should make it optional but still sorta suggest that the best option is to use integrated, but then also make it so downloading integrated is less likely to cause confusion.

Also, we need to check if the python version is undefined and warn the user that the python instance was corrupt or not found, or something.

**Alternatives**
At the very least we should warn the user that it failed to download properly rather than silently failing and giving them a cryptic error with no relevant logs.

RunDevelopment:
I don't think we should do this. To understand the question "Would you like to download integrated Python?", users have to know at least the following things:

1. That chainner needs a python installation to function.
2. Whether they can even use the alternative python environment (not every one has python installed).
3. That integrated and system are the only choices. Not having a python installation isn't possible.

I do not want to give new users this question. A new user shouldn't have to know what integrated Python. I mean, they shouldn't even have to know that we use Python in the backend at all. Installing a program shouldn't have to require knowing how that program works.

> Many people do not want to use the integrated python, plus it tends to fail to download.

The UX problem of the download failing has been solved by #2768. Advanced users that do not want to use the integrated Python can install chainner normally (with integrated) and then switch to their System python in the settings.

joeyballentine:
I completely agree. I never really wanted to do this, it was just requested a lot

PR #949, opened 2022-09-14
Refactored how system and integrated python are detected

RunDevelopment:
This is the start of the python/backend refactoring we talked about on Discord. I stared things by factoring out how the `PythonInfo` objects for system python and integrated python are obtained. I also made the error checking a lot stricter around integrated python and added its own failure dialog. Basically, instead of the previous "it's probably fine" approach, it now has to prove that the integrated python is a valid python version. Since some of the mac bug reports seem to have been caused by chainner being unable to install its integrated python, I suspect that we will see this new error message a few times in the future.


RunDevelopment:
Not yet IMO. #926 is more about UX, while this PR is more about restructuring how we interact with python.

joeyballentine:
Makes sense

joeyballentine (approved):

PR #951, opened 2022-09-14
Removed global python info

RunDevelopment:
This removes the `[sg]etPythonInfo` methods. Removing this bit of global state is necessary for the changes to come, and generally improves the code a little.


src/main/main.ts
```diff
@@ -572,16 +572,16 @@ const doSplashScreenChecks = async () =>
...
 
             splash.webContents.send('checking-deps');
             const hasNvidia = await nvidiaSmiPromise;
-            await checkPythonDeps(splash, hasNvidia);
+            await checkPythonDeps(splash, pythonInfo, hasNvidia);
             await sleep(250);
 
             splash.webContents.send('spawning-backend');
-            await spawnBackend(port);
+            spawnBackend(port, pythonInfo);
```
RunDevelopment, line 584: One of the improvements is here. We can now clearly see the data dependency between `checkPythonEnv` and the below functions. We have to know which python to use before we can install dependencies and spawn the backend. Clearly expressed in code. Global state used to hide this from us.

joeyballentine (approved):

PR #968, opened 2022-09-15
Added identity node

### LINKED ISSUE #781: Dummy-proxy node
Even with the substance designer, I noticed a problem when I need to switch one source image for many nodes, then I have to change links for each, so I began to use a doing nothing proxy node there.
Here the situation is similar, now I began to use HUE & SATURATION for this.
I can continue to do this, but it doesn't seem right in terms of architecture/design.

Example:
изображение

RunDevelopment:
Fixes #781.

This adds support for the requested dummy/proxy node. I call it "identity" because (1) it implements the identity function and (2) because I don't have a good name for it yet. This will likely be the hardest part of the PR: How do we communicate the function of this node?

To describe the behavior and function of the node a bit: While this node my implement the identity function, if chainner is to looked at as a programming language, then this node enables the aliasing of values. It behaves like a variable, so to say. 

Example: The following chain behaves like this program:

image

```py
loadImageResult = LoadImage(...)
variable = loadImageResult
SaveImage(variable)
ViewImage(variable)
```

---

Changes:
- Added identity node (all labels and icons are still WIP).
- Changed type accent colors to account for the `any` type.
- Changed how handles and edges are colored. They used to use the definition type, but that's useless for the identity node, so I changed them to use the current type (typeState).
- Changed optional tag. We used to assumed that `null ⊆ type`, then it's optional, but that's not true. While all optional inputs accept `null`, not all inputs that accept `null` are optional.
- Fixed a bug in the context selector. When creating a connection from an existing node, it used the current type (typeState) of the node to filter all schemata, but it used the definition type of the existing node to find the first compatible input when creating the connection. This made it possible to create invalid connections.

With those changes, the technical requirement for the identity node are met. In my tests, it behaved exactly as expected.


joeyballentine:
I think a good name might just be "Passthrough", since all it does is pass through the input to the output

joeyballentine:
image
This could be a good icon

RunDevelopment:
Actually, you're right. I initially thought that it might be confusing, but this node is literally the best example of what the pass through feature from #373 does. So let's call it "Pass through".

RunDevelopment:
Any idea on a sub category name for the node? I'm thinking of "Value" and adding stuff like constant nodes (nodes that just return a constant number/string/image) to that category later.

joeyballentine:
Yeah I think that works

RunDevelopment:
Looks kinda nice actually.

image

joeyballentine (approved):

PR #973, opened 2022-09-16
Clear manual output type

### LINKED ISSUE #927: Clear manually set types on node clear
Right now, a manually set type (from a starter node running) does not get cleared when you right-click> clear.

RunDevelopment:
Fixes #927.

While I say that this fixes the linked issue, I wasn't actually able to see any issues caused by the output type not being cleared. This is because the output elements (`renderer/components/outputs/*`) that set the manual output type also clear it again when their stored output data is reset. Since `clearNode` already resets the output data, I'm not sure why clearing the output type is also necessary. What was the reason you opened #927 @joeyballentine?

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -206,6 +206,15 @@ export const GlobalProvider = memo(
             },
             [setManualOutputTypes]
         );
+        const clearManualOutputTypes = useCallback(
+            (nodeId: string): void => {
+                setManualOutputTypes(({ map }) => {
+                    map.delete(nodeId);
+                    return { map };
+                });
+            },
+            [setManualOutputTypes]
+        );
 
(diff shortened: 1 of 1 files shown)


joeyballentine:
@RunDevelopment I had noticed that type previews when you had things like string connections did not get reset

RunDevelopment:
Could you give an example chain?

joeyballentine:
Load image image name -> text append

RunDevelopment:
> Load image image name -> text append

This works as intended in main. The manual output type is cleared as expected.

joeyballentine:
Huh. I'm not sure what it was I did when I made that issue then

RunDevelopment:
Hmmm, try to reproduce the issue again. If you can't, then something else might have already fixed it and we can close the issue and this PR.

RunDevelopment:
@joeyballentine Do we still need this?

joeyballentine:
I don't think so

PR #974, opened 2022-09-16
Fixed starting nodes not running after clearing

RunDevelopment:
If you cleared a Load Image node with a selected image and then selected the same image again, then the node wouldn't run again and display no image preview.

This was because the lastInputHash was only updated when running a node. Since nodes without all required inputs can't be run, the `runRunNode` hook basically ignored that the node was cleared.

The fix is pretty simple. We just always update lastInputHash, even if we aren't going to run the node.


joeyballentine:
Sweet, this should make it so I can finish my file watcher stuff

PR #978, opened 2022-09-16
Clear starting node cache on clear

joeyballentine:
This realistically would never be a problem, but probably still good to do anyway

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -37,6 +37,7 @@ import {
     createUniqueId,
     deepCopy,
     deriveUniqueId,
+    isStartingNode,
     parseSourceHandle,
     parseTargetHandle,
 } from '../../common/util';
@@ -154,7 +155,7 @@ interface GlobalProviderProps {
 export const GlobalProvider = memo(
     ({ children, reactFlowWrapper }: React.PropsWithChildren<GlobalProviderProps>) => {
         const { sendAlert, sendToast, showAlert } = useContext(AlertBoxContext);
-        const { schemata, functionDefinitions } = useContext(BackendContext);
+        const { schemata, functionDefinitions, backend } = useContext(BackendContext);
         const { useStartupTemplate } = useContext(SettingsContext);
 
         const [nodeChanges, addNodeChanges, nodeChangesRef] = useChangeCounter();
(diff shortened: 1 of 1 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1028,6 +1029,15 @@ export const GlobalProvider = memo(
                 });
                 outputDataActions.delete(id);
                 addInputDataChanges();
+                const nodeSchemaId = getNode(id)?.data.schemaId;
+                if (nodeSchemaId && isStartingNode(schemata.get(nodeSchemaId))) {
```
RunDevelopment, line 1033: I don't like the `isStartingNode` check here:

1. Even if the node isn't a starting node, then it won't do anything. We'd just send a small request to the backend. That's basically no cost.
2. I don't like that we cement the assumption that only starting nodes get executed here as well.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1028,6 +1029,15 @@ export const GlobalProvider = memo(
...
+                    backend
+                        .clearNodeCacheIndividual(id)
+                        .then(() => {})
```
RunDevelopment, line 1036: This doesn't do anything and can be removed.

```suggestion
```

joeyballentine: That's there (as well as in other places) because eslint yells at you if you dont await a promise unless you have a .then and a .catch.

Well, now that I think about it maybe it only needs the catch and I just misread the error.... Will try

RunDevelopment: It only yells at you when you don't handle the error a promise could "throw". So `.then(fn)` doesn't fix the lint, but `.then(fn, fn)` and `.catch(fn)` will (`.catch(fn)` is the same as `.then(x => x, fn)` btw).

joeyballentine: yeah i definitely misread the error. it must end in a call to .then OR .catch, not both

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1028,6 +1028,9 @@ export const GlobalProvider = memo(
                 });
                 outputDataActions.delete(id);
                 addInputDataChanges();
+                backend.clearNodeCacheIndividual(id).catch((error) => {
+                    log.error(error);
+                });
```
RunDevelopment, line 1033: Super minor nit: Could we make this a single line?

```suggestion
                backend.clearNodeCacheIndividual(id).catch((error) => log.error(error));
```

RunDevelopment (approved):

PR #981, opened 2022-09-17
Lower vram amount again

joeyballentine:


RunDevelopment (approved):
And the binary search continues.

PR #988, opened 2022-09-18
Reverted change in ncnn model that broke conversion

theflyingzamboni:
Fixed dumb change I made


joeyballentine (approved):

RunDevelopment (approved):
I'm just gonna trust you that this works. I don't use NCNN anyway...

PR #997, opened 2022-09-19
Executor rewrite

RunDevelopment:
This is the executor "rewrite" I mentioned on Discord.

Major changes:
- The `/run` API is now a lot simpler. The frontend only has to submit the current chain. No caching stuff. The shape of the JSON also changed.
- `NewType` for node ids and input/output ids. This allows us to verify that the implementation is correct using types.
- Simpler API for `NodeBase` inputs. I removed `get_inputs` and made `.inputs` a getter/setter pair instead. The setter will automatically fill in any missing input ids. This makes it way easier to use the input ids of inputs. Same for outputs.
- `Chain` is the new Python class to describe the whole graph of the chain. Notably, this does not include inputs, they are separate. Since we now have the full graph in a convenient representation on the python side, I also implemented a simple chain optimizer. It only removes useless nodes and outlines child nodes right now, but we can add more in the future.
- New input system. Inputs are now their own data structure and not part of the nodes anymore. This allows iterators to set inputs without modifying the chain itself. Further, inputs can inherit from a parent input map. This makes the input maps for iterators very cheap.
- New caching system. Everything related to caching is now in `chain.cache`. The basic data structure is a multi-level cache where each level can lookup but not modify parent caches. This makes it quite easy to implement iterators and how they cache stuff. 
   Iterators also cache more efficiently now. Previously, all child node outputs were cached until the end of the current iteration. They are now reference-counted and will be cleared from cache ASAP. I didn't explicitly add this, btw. This just naturally emerged from the new caching system.
- Iterators got simpler. With the new cache and input system, iterators got a lot simpler, both in their `Executor` code and their `run` methods. The executor doesn't have to preemptively execute required free nodes anymore, and most of the shared code between iterators either became unnecessary or was moved into `IteratorContext` (formerly `ExecutorContext`).

I did some basic testing, and it worked well so far.

### DIFF
--- a/backend/src/base_types.py
+++ b/backend/src/base_types.py
@@ -1,0 +1,5 @@
+from typing import NewType
+
+NodeId = NewType("NodeId", str)
+InputId = NewType("InputId", int)
+OutputId = NewType("OutputId", int)
--- a/backend/src/chain/cache.py
+++ b/backend/src/chain/cache.py
@@ -1,0 +1,109 @@
+from typing import Any, Dict, Iterable, Optional, Set
+import gc
+from sanic.log import logger
+
+from .chain import NodeId, Chain
+
+
+class CacheStrategy:
+    STATIC_HITS_TO_LIVE = 1_000_000_000
+
+    def __init__(self, hits_to_live: int) -> None:
+        assert hits_to_live >= 0
+        self.hits_to_live = hits_to_live
+
+    @property
+    def static(self) -> bool:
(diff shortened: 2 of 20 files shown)


joeyballentine:
On initial skim the code seems fine. I'll test it out a bunch tonight

joeyballentine:
Played around with it and it doesn't seem to have broken anything. Worst case we'll run into something when testing other things, and we can fix it.

joeyballentine (approved):

PR #1004, opened 2022-09-20
Correct image iterator sorting

theflyingzamboni:
Should fix image iterator not iterating in a sensible order. Both subdirectories and files are now sorted in proper alphanumeric order.


backend/src/nodes/image_iterator_nodes.py
```diff
@@ -100,12 +100,13 @@ def walk_error_handler(exception_instance):
...
 
         just_image_files: List[str] = []
-        for root, _dirs, files in os.walk(
+        for root, dirs, files in os.walk(
             directory, topdown=True, onerror=walk_error_handler
         ):
             await context.progress.suspend()
 
-            for name in sorted(files):
+            dirs.sort(key=numerical_sort)
```
RunDevelopment, line 108: We don't use `dirs`, so why are you sorting it?

backend/src/nodes/utils/utils.py
```diff
@@ -454,3 +456,12 @@ def compare_conditions(a, b):
         raise RuntimeError(f"Unknown side selection {side}")
 
     return w_new, h_new
+
+
+def numerical_sort(value: str):
```
RunDevelopment, line 461: Please add a return type. I had problems understanding this functions, and a return type would have helped.

```suggestion
def numerical_sort(value: str) -> List[Union[str, int]]:
```

RunDevelopment (changes requested):
Really good idea. Just 2 nits.

backend/src/nodes/image_iterator_nodes.py
theflyingzamboni: We don't, os.walk does. Try walking a tree with numbered directories. If you don't sort the directories, they'll come out e.g. 1, 10, 11, 12, 7, 8, 9. With that sort, it will walk them 1, 7, 8, 9, 10, 11, 12.

backend/src/nodes/utils/utils.py
theflyingzamboni: Whoops. Looks like there's another function that was added to utils without a return type too.

backend/src/nodes/image_iterator_nodes.py
RunDevelopment: So I looked up the behavior of `os.walk` and you are right. I did not expect that `dirnames` can be modified. I would proclaim that iterators changing their behavior based on what happens to their output is an antipattern (or as some people call it: pythonic).

(Could you please resolve this thread? I can't do that myself...)

backend/src/nodes/utils/utils.py
theflyingzamboni: Had to add a type: ignore. For some reason, pyright thinks the type should be have str | Any instead of str | int, but as far as I know those are the only two possible types.

backend/src/nodes/image_iterator_nodes.py
theflyingzamboni: It's definitely weird. I can't think of another function that does that.

0x4E69676874466F78: In addition, sorting is different in linux and windows.

backend/src/nodes/utils/utils.py
RunDevelopment: @theflyingzamboni What happens when you add `List[Union[str, int]]` to `parts`?

theflyingzamboni: Then the return value is fine, but it complains that list[str | Any] cannot be assigned to List[str | int].

joeyballentine (approved):

Issue #777, opened 2022-08-20
Change canvas size


CharonM72:
A node to change canvas size (extendable with transparency for RGBA, or maybe a color or something for other colorspaces) would be useful.

RunDevelopment:
Just to clarify: Using Paint.net as an example, you basically want this, right?
![image](https://user-images.githubusercontent.com/20878432/185740597-6e09d02b-d9c3-4daa-bbbe-656d4cb63cab.png)

theflyingzamboni:
This is already (largely) accomplished by Create Border. We should probably expand on what that node can do. Right now it lacks a version to control border sizes independently (like with crop), and the idea of a color-picker for constant color is still unimplemented as well.

CharonM72:
Yes, something like that from paint.net.
I'm trying to see if it's possible to automate a process I currently do manually for converting a large number of images into a specific aspect ratio and size without loss in quality. Right now it has a lot of the tools needed but it still missing a few. I submitted a somewhat broader but kind of different issue in #798.

simolg:
"canvas size" is a different method to change the image size:
- canvas to bigger size is similar to "create border" method
- canvas to smaller size is similar to "crop" method

they are similar methods but not the same!

in many cases "canvas size" is very useful:
the destination of the new canvas must be specified in pixels,
after setting an anchor, the image is refitted (**automatically**)

if it comes out it creates edges, if it shrinks it crops (or even a mix of the two)

this feature would be very welcome
Thanks again

Tarnyd:
What you are looking for can be achieved using the following:
"**COLOR**" connected to "**CREATE COLOR**" and a "**LOAD IMAGE**"

"COLOR" specifies the background color for your canvas. It can be transparent if you wish. Use RGBA in that case.
In "CREATE COLOR" you specify the canvas size.

Connect them both to a "**BLEND IMAGES**" node. 
```
"CREATE COLOR" to the Base layer 
"LOAD IMAGE" to the Overlay Layer.
```

Connect the output to "VIEW IMAGE" or "SAVE IMAGE"

You can use normal crop or resize before or after this flow to change the size of the image before or after adding them together.

PR #1009, opened 2022-09-21
Added Create Edges node

RunDevelopment:
Changes:
- Added a Create Edges node. This node is basically the same as Create Border, but it lets specify the amount for each side separately.
- Changed the behavior of the "Transparent" border type. This is technically a breaking change, but @theflyingzamboni agreed that this is a bug.
- Added channel information to the output type.

image


RunDevelopment:
Hmmm, this technically covers the functionality of what was requested in #777, but I don't think it's in the spirit of #777. I'd rather have a node where you input an image, the target size, and an anchor point like in Paint.net for #777.

joeyballentine (approved):

PR #1018, opened 2022-09-22
Add output scale input for Face Upscale (GFPGAN)

joeyballentine:
This was talked about on Discord. Basically, since regular GFPGAN allows you to define an output scale, we figure the majority of users familiar with existing implementations will find it odd that it doesn't allow different scales. Since the scaling is done internally via the `face_helper` stuff, we can just utilize that to do the scaling for us and it will be consistent with existing implementations.

image

### DIFF
--- a/backend/src/nodes/pytorch_alt_nodes.py
+++ b/backend/src/nodes/pytorch_alt_nodes.py
@@ -37,14 +37,17 @@ class FaceUpscaleNode(NodeBase):
             ),
             ImageInput(),
             ImageInput("Upscaled Background").make_optional(),
+            NumberInput(
+                label="Output Scale", default=8, minimum=1, maximum=8, unit="x"
+            ),
         ]
         self.outputs = [
             ImageOutput(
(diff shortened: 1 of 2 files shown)


backend/src/nodes/pytorch_alt_nodes.py
```diff
@@ -37,14 +37,17 @@ def __init__(self):
             ),
             ImageInput(),
             ImageInput("Upscaled Background").make_optional(),
+            NumberInput(
+                label="Output Scale", default=8, minimum=1, maximum=8, unit="x"
```
RunDevelopment, line 41: Are non-power-of-two inputs allowed?

If not, then I would suggest making the output type this:
```
let scale = Input3 & pow(2, int(0..4));
Image {
    width: scale * Input1.width,
    height: scale * Input1.height,
    channels: 3
}
```

And adding a never reason.

RunDevelopment (approved):

PR #1020, opened 2022-09-22
Add timing support for iterators

joeyballentine:


RunDevelopment:
This will also measure the time during which an iterator is paused.

Honestly, I'm not quite sure how to get around this. My only idea would be to somehow track paused time and subtract that from the execution time, but even that doesn't work, because the iterator might not pause the moment the executor is paused.

joeyballentine:
I think that's fine tbh. It might not be ideal but at least it's kinda understandable behavior

PR #1033, opened 2022-09-25
TGA support

emarron:
Literally just adds TGA option to 'SAVE IMAGE'.
I checked with chinese and emojis and it works fine.
doesn't screw up the opencv imwrite.

Could potentially add other [pillow supported types.](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html)


backend/src/nodes/image_nodes.py
```diff
@@ -135,12 +135,12 @@ def __init__(self):
...
-        base_directory: str,
-        relative_path: Union[str, None],
-        filename: str,
-        extension: str,
+            self,
+            img: np.ndarray,
+            base_directory: str,
+            relative_path: Union[str, None],
+            filename: str,
+            extension: str,
```
RunDevelopment, line 143: Please revert these formatting changes.

Also, we use black for formatting, so please run it.

backend/src/nodes/image_nodes.py
```diff
@@ -155,22 +155,32 @@ def run(
...
 
         os.makedirs(base_directory, exist_ok=True)
-
-        # Write image with opencv if path is ascii, since imwrite doesn't support unicode
-        # This saves us from having to keep the image buffer in memory, if possible
-        if full_path.isascii():
-            status = cv2.imwrite(full_path, img)
+        # Any image not supported by cv2, will be handled by pillow.
+        if extension not in ["png", "jpg", "gif", "tiff", "webp"]:
+            status = 1  # spoof
```
RunDevelopment, line 160: @joeyballentine Looking at this, why do we return anything at all? This node doesn't have any outputs, and no other code in the backend is using this node otherwise (I looked). Also, instead of returning a status code that nobody checks, the OpenCV code should probably just throw if an image couldn't be saved.

@emarron I'm sorry for spamming your PR. This is not something you have to change. It's a problem with our code.

backend/src/nodes/image_nodes.py
```diff
@@ -155,22 +155,32 @@ def run(
...
+            if img.shape[2] == 4:
+                img = cv2.cvtColor(img, cv2.COLOR_BGRA2RGBA)
+            else:
+                img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
```
RunDevelopment, line 164: Chainner also supports and creates single-channel images, so `img.shape` might only have 2 elements. 

1. Use `get_h_w_c(img)[2]` to reliably get the number of channels.
2. Add support for single-channel images.

We should probably also make it explicit that only single-channel, RGB, and RGBA images are supported for PIL. Maybe like this:

```py
c = get_h_w_c(img)[2]
if c == 1:
    img = ??? # idk how to do this in PIL
elif c == 3:
    img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
elif c == 4:
    img = cv2.cvtColor(img, cv2.COLOR_BGRA2RGBA)
else:
    assert False, f"Unsupported number of channels. Saving .{extension} images is only supported for grayscale, RGB, and RGBA images."
```

RunDevelopment (changes requested):
Thank you for this PR @emarron!

The overall approach looks good. Just 2 nits.

backend/src/nodes/image_nodes.py
joeyballentine: I did that when I first made this node thinking at some point we might wanna have branching on if it failed or not, and I've just never removed this

RunDevelopment: *If* we want to allow failable nodes in the future, we should probably do it in a way that all nodes can use it.

backend/src/nodes/image_nodes.py
emarron: good point. I added the check. It now only supports 1,3,4 channels. fortunately grayscale 'L' channels just pass through in pillow so we don't need to do anything,

emarron:
@RunDevelopment @joeyballentine
Made the requested changes. Can you take a look at it again? 😄

backend/src/nodes/image_nodes.py
```diff
@@ -155,22 +155,37 @@ def run(
...
+            channels = get_h_w_c(img)[2]
+            # if array is BGRA or BGR it must be converted to RGBA or RGB, any grayscale should pass through fine.
+            # if it has neither 4, 3, nor 1 channels we can't use pillow.
+            if channels not in [4, 3, 1]:
+                raise RuntimeError(f"Unsupported number of channels. Saving .{extension} images is only supported for "
+                                   f"grayscale, RGB, and RGBA images.")
+            elif channels == 3:
+                img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+            elif channels == 4:
+                img = cv2.cvtColor(img, cv2.COLOR_BGRA2RGBA)
```
RunDevelopment, line 170: I don't think this is optimal. This kinda feels like bike-shedding, but there are 3 things I don't like about this:
1. You needed to comment that single-channel images are passed through, when that should be written as code instead. IMO, comments shouldn't say what the program does (code should do that), but why the program does things.
2. There are now 2 places in code that check for RGB and RGBA images.
3. The `if channels not in [4, 3, 1]` needs to know all supported channel formats.

I would suggest a simple if-elif chain.

```suggestion
            if channels == 1:
                # PIL supports grayscale images just fine, so we don't need to do any conversion
                pass
            elif channels == 3:
                img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
            elif channels == 4:
                img = cv2.cvtColor(img, cv2.COLOR_BGRA2RGBA)
            else:
                raise RuntimeError(f"Unsupported number of channels. Saving .{extension} images is only supported for "
                                   f"grayscale, RGB, and RGBA images.")
```

This makes it obvious why we don't do a conversion for single-channel images, we only check each channel count once, and the error naturally appears in the `else` branch.

RunDevelopment (changes requested):
One last comment.

emarron:
@RunDevelopment yeah that looks better, thanks for all your work with me. 😄

emarron:
@RunDevelopment done 😺

RunDevelopment (approved):

PR #1044, opened 2022-09-30
Added support other type models (-1..1) for ONXX

0x4E69676874466F78:
Added new node "Upscale Image (other)" with color transformation from 0..1 (chaiNNer) to -1..1 (like ArmorLab) and back to 0..1 (chaiNNer).

### DIFF
--- a/backend/src/nodes/onnx_nodes.py
+++ b/backend/src/nodes/onnx_nodes.py
@@ -141,6 +141,85 @@ class OnnxImageUpscaleNode(NodeBase):
         )
 
 
+@NodeFactory.register("chainner:onnx:upscale_image_other")
+class OnnxImageUpscaleOtherNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Other processing for models like ArmorLab (0..1 → -1..1)"
+        self.inputs = [
+            OnnxModelInput(),
+            ImageInput(),
+            TileModeDropdown(has_auto=False),
+        ]
+        self.outputs = [
+            ImageOutput(
+                "Upscaled Image",
+                image_type=expression.Image(channels="Input1.channels"),
+            )
+        ]
+
+        self.category = ONNXCategory
+        self.name = "Upscale Image (other)"
+        self.icon = "ONNX"
+        self.sub = "Processing"
+
+    def upscale(
+        self,
+        img: np.ndarray,
+        session: ort.InferenceSession,
+        tile_mode: Union[int, None],
+        change_shape: bool,
+    ) -> np.ndarray:
+        logger.info("Upscaling image")
+        is_fp16_model = session.get_inputs()[0].type == "tensor(float16)"
+        img = np2nptensor(img, change_range=False)
+        logger.info(img.shape)
+        out, _ = onnx_auto_split_process(
+            img.astype(np.float16) if is_fp16_model else img,
+            session,
+            max_depth=tile_mode,
+            change_shape=change_shape,
+        )
+        logger.info(out.shape)
+        out = nptensor2np(out, change_range=False, imtype=np.float32)
+
+        # Like ArmorLab models, where from -1 to 1.
+        np.multiply(out, 0.5, out=out)
+        np.add(out, 0.5, out=out)
+
+        del session
+        logger.info("Done upscaling")
+        return out
+
+    def run(
+        self,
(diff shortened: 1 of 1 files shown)


0x4E69676874466F78:
image

joeyballentine:
I don't really like the idea of having an entirely separate node for this, especially since "(other)" doesn't describe at all what's going on here. Is there really nothing contained in the model graph itself that could determine what data range it's made for?

0x4E69676874466F78:
@joeyballentine 
>I don't really like the idea of having an entirely separate node for this, especially since "(other)" doesn't describe at all what's going on here. Is there really nothing contained in the model graph itself that could determine what data range it's made for?

but...
image

I made this PR so it wouldn't get lost.

0x4E69676874466F78:
I can try to add modes for the Upscale node if you want.

0x4E69676874466F78:
@joeyballentine 
> Is there really nothing contained in the model graph itself that could determine what data range it's made for?
I don't have the experience to know.
You can view the models here: https://github.com/armory3d/armorlab_models/releases/download/2022.8/models.zip

theflyingzamboni:
To answer whether the expected data range can be gotten from the model, no. I looked through the start of the graph and didn't find anything of value. The model doesn't care, it just operates on the data given, and we give it data that is technically within its range.

RunDevelopment:
Just an idea: instead of adding a new upscaling node that assume a specific data range, couldn't we add a node that adds the range information to the model class we keep internally? The problem is that we can't know this range by just looking at the model. Why don't we just assume it's 0..1 by default (as we currently do), and let the user specify it otherwise with that new node?

That would also allow us to detect more invalid operations (e.g. interpolating a 0..1 model with a -1..1 model) since it's something the type system can track.

joeyballentine:
> That would also allow us to detect more invalid operations (e.g. interpolating a 0..1 model with a -1..1 model) since it's something the type system can track.

I doubt you'd be able to interpolate two models that expect a different type range, unless someone specifically went out of their way to change the range of something existing (i.e. esrgan), which IMO is super unreasonable to expect.

This is still an ok idea I guess. Could be useful in NCNN as well for supporting those cugan models that want 0-255. But, what I'd much rather have is something that checks for certain layer patterns that are unique to this model specifically and just auto-adjusts for it. That's obviously too complicated and too much work for something like this though, so having that other node is an ok compromise. It just requires the user to know that they need to use it.

theflyingzamboni:
Two totally different models would never interpolate unless they _somehow_ happened to coincidentally have all of their weights line up exactly. You just wouldn't be able to make that happen. The supported data range is irrelevant, it just won't work regardless.

I'm not sure where I fall on the two suggestions. The data range is not something users are likely to know unless they really dig into model documentation, and those are not always well-documented. Asking a user to input a custom data range might be a big ask for a lot of users. On the other hand, I really don't like the idea of us having to hard-code patterns for every new arch someone wants to use in three different network types. The fact that we have to hard-code stuff in for pytorch is already kind of a pain.

joeyballentine:
> This is why I was leaning towards user-created plugin functionality being the route used to support these sorts of non-standard models.

I agree. I'm kinda working on this already (with #1047 being the start of it) so I think we should just go with the user-defined plugin route

joeyballentine:
@0x4E69676874466F78 If you're still willing, this could be converted to a "Generate Normal Map" node. If you could find a way to detect the armorlab architecture.

RunDevelopment:
> If you could find a way to detect the armorlab architecture.

If you could detect those models, then adding the range information to the model class would be even better, because the normal upscale node would just work.

0x4E69676874466F78:
I've been busy with other things, but I'm slowly getting back into it. 
I'm not sure if I can detect the model, but I'll see what I can do.
Still no plugin support?

joeyballentine:
> Still no plugin support?

Correct. But I'm still working on the necessary things for it.  I'm basically treating our pytorch/onnx/ncnn packages as third party plugins and making sure everything those need, someone else would be able to do as a plugin. It's mostly there now, there's just a few missing things left, like plugins being able to define their own migrations

joeyballentine:
There's still no plugin support, but I am going to finally close this PR. Once we finally have everything in place for plugins, I'd hope that you would consider making an armorlab plugin.

PR #1047, opened 2022-10-02
Node import refactor

joeyballentine:
This completely refactors how nodes are imported. This accomplishes a few things:

- Node files are now not extremely long and take forever to find what you're looking for
- It's a lot easier to not do wildcard importing for inputs and outputs now
- A single node having an error won't cause the entire file to fail to import and therefore the other nodes will import properly
- Now we can show a "missing dependencies" package hint thing if one or more of the nodes fail. This is a bit better than the "excluded nodes" thing I did before to allow the NCNN and ONNX save nodes to still show up.
- We are now one step closer to being able to support plugins. We're obviously a long way away from that, but a plugin system will have to use something similar to this auto-importing system anyway.

Here's an example of what it looks like when you don't have NCNN or ONNX installed:
image

This was pretty important to do, as right now we give no indication to the user if they are missing something like facexlib (which is required for face upscaling), so the node just isn't there. Now it would show that PyTorch is missing dependencies.


backend/src/run.py
```diff
@@ -7,81 +7,69 @@
...
+from nodes.utils.exec_options import set_execution_options, ExecutionOptions
+
 from base_types import NodeId, InputId, OutputId
 from chain.cache import OutputCache
 from chain.json import parse_json, JsonNode
 from chain.optimize import optimize
 from events import EventQueue, ExecutionErrorData
 from process import Executor, NodeExecutionError, timed_supplier
-from progress import Aborted
+from progress import Aborted  # type: ignore
```
RunDevelopment, line 38: Why?

RunDevelopment (commented):
While I do approve of the approach of this PR, I think that the node file structure you went with is optimal. I don't like that "image" has a bunch of sub folders and files. I think it would be easier to find stuff if we just put each node into its category. E.g. `nodes/image/load_image.py` and `nodes/image_dimensions/resize_factor.py`. This creates a nice 1:1 relation between path and category.

backend/src/run.py
joeyballentine: My vscode thinks this import is from a separate progress package and I'm tired of seeing this error

RunDevelopment: Uhmm, then are you sure that we aren't? Does it import the right package if you use system python?

joeyballentine: I assume so, I've never seen it actually have a runtime error, and usually local files take import predenence over global ones

joeyballentine: For example, I've run into issues before when testing things where I name a file something like `ncnn.py` and then doing `from ncnn import ncnn` results in a circular dependency error since it tries to import itself rather than the actual ncnn package I have installed. So should be fine I think. But maybe it would be safer to just rename our progress package. Unfortunately you can't do relative imports from run.py because python sucks. But anyway I think that would be better resolved in a separate PR anyway.

backend/src/nodes/nodes/image_iterator/image_file_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+        self.inputs = [IteratorInput().make_optional()]
+        self.outputs = [
+            ImageOutput(broadcast_type=True),
+            DirectoryOutput("Image Directory"),
+            TextOutput("Subdirectory Path"),
+            TextOutput("Image Name"),
+            NumberOutput("Overall Index"),
+        ]
+
+        self.category = ImageCategory
```
RunDevelopment, line 41: Iterators are still in the "Image" category, so could we also move them into that folder?

joeyballentine: Oh yeah, sure

RunDevelopment (approved):

PR #1051, opened 2022-10-03
Refactor how categories are defined

joeyballentine:
Since the node folders and categories have a 1:1 relationship now, I decided to just define categories in each of these folders in the `__init__.py` files. I think ideally this would use the decorator `@register` syntax with a category factory or something, and maybe the nodes would somehow automatically know what category to use based on the folder they're in rather than having to define it in the constructor. 

However, I think this is fine for now.


RunDevelopment (changes requested):
I like the idea of this PR, but I don't like what you did with `category_order`. The new `.order` prop just isn't a good idea IMO. Adding or removing categories now means changing the `.order` prop of other categories. Category order is an inherently global property, so any local property will come with downsides.

Maybe we could remove the `.order` prop and bring back `categories.py` with `categories` and `category_order` lists? Essentially, `categories.py` would import all built-in categories while ``category.py`` defines the `Category` class.

We have to define the global category order somewhere, and I'd rather do it in one file.

joeyballentine:
So the reason I did it this way is planning ahead to plugins. We'll need a way for people who want to make plugins to decide what order their own defined categories go in. And this was just the only way I could think of for accomplishing that

RunDevelopment:
I don't think that plugin author should be able to decide that their plugins are so important that they are to be at the top of the node list. This also creates the problems that plugin authors now have to coordinate for their plugin order...

How about we define the order of built-in categories explicitly, sort plugin categories alphabetically, and show builtins first, plugins second?

joeyballentine:
That sounds like a good plan

backend/src/run.py
```diff
@@ -62,14 +56,36 @@
...
+                category = getattr(importlib.import_module(module), "category")
+                categories.add(category)
+            except:
+                # Case when it is a folder with multiple modules that has a category order
+                try:
+                    category_order.extend(
+                        getattr(importlib.import_module(module), "category_order")
+                    )
+                except:
+                    pass
```
RunDevelopment, line 86: All of this code when you could have just made another file (e.g. `src/nodes/categories.py`) for `category_order` and import that file.

joeyballentine: True

backend/src/nodes/nodes/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+from .image_dimension import category as ImageDimensionCategory
+from .image_channel import category as ImageChannelCategory
+from .image_utility import category as ImageUtilityCategory
+from .utility import category as UtilityCategory
+from .pytorch import category as PyTorchCategory
+from .ncnn import category as NCNNCategory
+from .onnx import category as ONNXCategory
+
+
+builtin_categories = [
```
RunDevelopment, line 13: I think you forgot to remove that.

backend/src/nodes/nodes/builtin_categories.py
```diff
@@ -0,0 +1,25 @@
...
+from .image_dimension import category as ImageDimensionCategory
+from .image_channel import category as ImageChannelCategory
+from .image_utility import category as ImageUtilityCategory
+from .utility import category as UtilityCategory
+from .pytorch import category as PyTorchCategory
+from .ncnn import category as NCNNCategory
+from .onnx import category as ONNXCategory
+
+
+builtin_categories = [
```
RunDevelopment, line 13: Should this file really be in `src/nodes/nodes`? This folder only contains node definitions otherwise. It feels weird that this file is in there as well.

RunDevelopment (changes requested):

backend/src/nodes/nodes/builtin_categories.py
joeyballentine: It also contains the category definitions though

RunDevelopment (approved):

PR #1063, opened 2022-10-08
Upgrade to React-Flow v11

joeyballentine:
This gives us a few things
- A cool grabbing cursor when panning
- The ability to middle click drag to pan, even on nodes
- Much better performance when dragging around nodes, for some reason

Problems I've found so far:
- ~Panning has much worse performance now, for some reason. Going to guess it has something to do with a viewport related hook or something.~
- Occasionally the grab cursor won't change while dragging, but changes after


package.json
```diff
@@ -217,10 +217,11 @@
     "re-resizable": "^6.9.9",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
-    "react-flow-renderer": "^10.3.16",
+    "react-flow": "^1.0.3",
```
RunDevelopment, line 220: [Uhmm](https://www.npmjs.com/package/react-flow)

src/renderer/global.scss
```diff
@@ -1,4 +1,5 @@
 @use "colors.scss";
+@import "reactflow/dist/style.css";
```
RunDevelopment, line 2: Does this even work? Can you use npm-style imports in SCSS files?

src/renderer/global.scss
```diff
@@ -129,12 +130,12 @@ body {
     width: 8px;
 }
 
-[data-theme='dark'] ::-webkit-scrollbar-thumb {
+[data-theme="dark"] ::-webkit-scrollbar-thumb {
```
RunDevelopment, line 133: I'm pretty sure we use single-quotes everywhere, so why? Did your default formatter do that? If so, then maybe we should add stylelint. Since we plan on moving away from CSS-in-JS, adding stylelint might be a good move regardless.

package.json
joeyballentine: Whoops

src/renderer/global.scss
joeyballentine: Yes. Not only does it work, it's necessary.

src/renderer/global.scss
joeyballentine: Different pr please, but yes. I didn't do this on purpose so it has to have been formatter

joeyballentine:
I think we should just merge this. It doesn't seem to break anything and having to keep fixing merge conflicts on it is getting annoying. Plus, we probably won't release anything for a while considering the tile size issues.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -425,7 +425,7 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
             >
                 <Background
                     gap={16}
-                    size={0.5}
+                    size={1}
```
RunDevelopment, line 428: Why was this changed?

joeyballentine: Because they changed something internally that made this necessary to get the old size.

RunDevelopment (approved):

Issue #1057, opened 2022-10-03
unexpected error with real esrgan ncnn model


HACCKKER:
**Information:**
- Chainner version: [0.13.0]
- OS: [Manjaro Linux]
- RAM: 8GB

**Description**
Even if gpu is selected for use with ncnn it uses cpu and fails

![Screenshot_20221003_225237](https://user-images.githubusercontent.com/49511278/193682981-eac125f7-eb9f-474e-b9e8-fd3aebaec45b.png)
I used real esrgan anime model because even if i use regular x4plus model how it is described in README.md
./realesrgan-ncnn-vulkan -i inputs -o outputs -n realesrgan-x4plus
I always get vkQueueSubmit failed and can only run its pytorch or onnx version on cpu
and anime model is working absolutely correct when used in that way
./realesrgan-ncnn-vulkan -i inputs -o outputs -n realesrgan-x4plus-anime

but when I use it in chainner I get vkQueueSubmit failed error in command line even with anime model that runs fine, when I changed tiling settings from auto to 4096 it stopped saying unexpected error in chainner, but it started asking for more extreme tiling, but there is nothing more extreme to set than 4096

![Screenshot_20221003_225223](https://user-images.githubusercontent.com/49511278/193682569-c49c0edb-accf-44ab-834d-03e488bc0f97.png)
I set my integrated gpu as ncnn device

![Screenshot_20221003_225201](https://user-images.githubusercontent.com/49511278/193682325-dfb4e653-421a-419e-86b1-783a8882ccbc.png)
I started chainner from command line to see logs and it says device: cpu and fp16 is set to False, so I think defaulting to cpu can be the root of the problem, if it just says cpu because my gpu is integrated, then I don't know what causes the problem, the only other thing I can think of is that real esrgan's ncnn runtime is modified

**Logs**
[Archive.zip](https://github.com/chaiNNer-org/chaiNNer/files/9700961/Archive.zip)

joeyballentine:
[...]

Issue #861, opened 2022-08-29
0.11.5 on macos Big Sur 11.6


dem10com:
In this version, the installation went smoothly and I downloaded the ncnn packages. but this time the model I am using on windows and I am sure it works (4x-UltraSharp-opt-fp16.bin ) gives a black screen. Anyone have experience with this before?

![Screen Shot 2022-08-29 at 12 24 43](https://user-images.githubusercontent.com/111634201/187170318-69ad5b72-32cc-4bd0-a5ef-c9864506a180.png)

dem10com:
By the way, the reason for using ncnn is that the others are not using the gpu. pytorch is using only the cpu. 

thank you

joeyballentine:
It seems like ncnn on macos is still broken unfortunately. It has something to do with vulkan, and I'm not sure what I can do to fix it. But I'll try

dem10com:
> It seems like ncnn on macos is still broken unfortunately. It has something to do with vulkan, and I'm not sure what I can do to fix it. But I'll try

thanks a lot Joey

joeyballentine:
Onnx does not have an arm64 (m1) version, therefore it is unable to be installed.

joeyballentine:
Also this is the first I've heard of the ncnn m1 version I just released working. I was last told it produced black outputs as well. So this is good news I guess :)

dem10com:
oh, Now I noticed that if it is more than 512x512 image size it gives a black screen.

joeyballentine:
Thats probably due to the auto tiling then. If you set a target tile size of 512 does it always work?

joeyballentine:
Hey @dem10com, I think I fixed this issue with 0.11.6 and the ncnn update that came with it. Could you give that a try?

dem10com:
i try, install and update ncnn (2022.8.29). But i am getting this error.

An error occurred in a NCNN Upscale Image node:

An unexpected error occurred during NCNN processing.

Input values (partial):
• Image: RGB image 2048x2048
• Tile Size Target: 0
![Screen Shot 2022-08-30 at 13 15 34](https://user-images.githubusercontent.com/111634201/187411903-90381867-3597-4acb-9999-c100086b17b7.png)


I change the tile size target but nothing change.


> Hey @dem10com, I think I fixed this issue with 0.11.6 and the ncnn update that came with it. Could you give that a try?

joeyballentine:
Can you set the target tile size to 0 and try again?

[...]

PR #1068, opened 2022-10-08
Add NCNN VRAM usage estimation

joeyballentine:
Similar to the PyTorch estimation, this gets the heap budget (a built-in NCNN property, they use it for tile size estimation in realesrgan-ncnn-vulkan) and the size of the image and returns a tile size for it.

This should hopefully reduce vkQueueSubmit errors (like in #861, #913, and #1057) since that error seems to mainly happen whenever an OOM error happens.

I used 80% as the "max allowed" value because the heap budget is already smaller than the actual total VRAM, and because it's what I remembered realesrgan-ncnn-vulkan using.

From my tests, it seems to work pretty well, but @theflyingzamboni should probably test it first before we merge this.

Also, I added a special except case for vkQueueSubmit while I was at it.

### DIFF
--- a/backend/src/nodes/nodes/ncnn/upscale_image.py
+++ b/backend/src/nodes/nodes/ncnn/upscale_image.py
@@ -16,6 +16,7 @@ from ...properties.inputs import NcnnModelInput, ImageInput, TileSizeDropdown
 from ...properties.outputs import ImageOutput
 from ...properties import expression
 from ...utils.exec_options import get_execution_options
+from ...utils.auto_split import estimate_tile_size
 from ...utils.ncnn_auto_split import ncnn_auto_split
 from ...utils.ncnn_model import NcnnModel
 from ...utils.ncnn_session import get_ncnn_net
@@ -81,34 +82,13 @@ class NcnnUpscaleImageNode(NodeBase):
             vkdev = ncnn.get_gpu_device(exec_options.ncnn_gpu_index)
             heap_budget = vkdev.get_heap_budget() * 1024 * 1024
 
-            def estimate_tile_size() -> Union[int, None]:
-                free = heap_budget
-
-                element_size = 4
-
-                h, w, c = get_h_w_c(img)
-                img_bytes = h * w * c * element_size
-                model_bytes = model.bin_length
-                mem_required_estimation = (model_bytes / (1024 * 52)) * img_bytes
-
-                # Attempt to avoid using too much vram at once
-                free_allowed_usage = 0.8
-                tile_pixels = (
-                    w * h * (free * free_allowed_usage) / mem_required_estimation
+            max_tile_size = (
+                tile_size
+                if tile_size is not None
+                else estimate_tile_size(
+                    heap_budget, heap_budget, model.bin_length, img, 4, 0.8
                 )
-                # the largest power-of-2 tile_size such that tile_size**2 < tile_pixels
-                tile_size = 2 ** (int(tile_pixels**0.5).bit_length() - 1)
-
-                GB_AMT = 1024**3
-                required_mem = f"{mem_required_estimation/GB_AMT:.2f}"
-                free_mem = f"{free/GB_AMT:.2f}"
-                total_mem = f"{heap_budget/GB_AMT:.2f}"
-                logger.info(
-                    f"Estimating memory required: {required_mem} GB, {free_mem} GB free, {total_mem} GB total."
-                    f" Estimated tile size: {tile_size}"
-                )
-
-                return tile_size
+            )
 
             with ncnn_allocators(vkdev) as (
                 blob_vkallocator,
@@ -121,9 +101,7 @@ class NcnnUpscaleImageNode(NodeBase):
                     output_name=output_name,
                     blob_vkallocator=blob_vkallocator,
                     staging_vkallocator=staging_vkallocator,
-                    max_tile_size=tile_size
-                    if tile_size is not None
-                    else estimate_tile_size(),
+                    max_tile_size=max_tile_size,
                 )
         except ValueError as e:
             raise e
--- a/backend/src/nodes/nodes/ncnn/upscale_image.py
+++ b/backend/src/nodes/nodes/ncnn/upscale_image.py
@@ -69,15 +69,50 @@ class NcnnUpscaleImageNode(NodeBase):
     def upscale(
         self,
         img: np.ndarray,
-        net,
+        model: NcnnModel,
         input_name: str,
         output_name: str,
-        max_tile_size: Union[int, None],
+        tile_size: Union[int, None],
     ):
         exec_options = get_execution_options()
+        net = get_ncnn_net(model, exec_options)
         # Try/except block to catch errors
         try:
             vkdev = ncnn.get_gpu_device(exec_options.ncnn_gpu_index)
+            heap_budget = vkdev.get_heap_budget() * 1024 * 1024
+            logger.info(heap_budget)
+            logger.info(f"{model.blob_count=}, {model.node_count=}")
+            logger.info(model.bin_length)
+
+            def estimate_tile_size() -> Union[int, None]:
+                free = heap_budget
+
+                element_size = 4
+
+                h, w, c = get_h_w_c(img)
+                img_bytes = h * w * c * element_size
+                model_bytes = model.bin_length
+                mem_required_estimation = (model_bytes / (1024 * 52)) * img_bytes
+
+                # Attempt to avoid using too much vram at once
+                free_allowed_usage = 0.8
+                tile_pixels = (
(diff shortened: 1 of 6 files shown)


backend/src/nodes/utils/ncnn_model.py
```diff
@@ -299,35 +299,35 @@ def __init__(
...
         self.weights_bin: bytes = b""
+        self.bin_length = 0
 
     @property
     def magic(self):
         return "7767517"
 
-    @staticmethod
-    def load_from_file(param_path: str = "", bin_path: str = "") -> "NcnnModel":
+    def load_from_file(self, param_path: str = "", bin_path: str = "") -> "NcnnModel":
```
RunDevelopment, line 308: That was supposed to be a static constructor. You shouldn't call this method on anything by an empty NcnnModel anyway.

joeyballentine: Oh whoops, I meant to undo this once I finally was able to get the bin length. 

But, why is this supposed to be a static method? I don't really understand the reasoning

RunDevelopment: A constructor is supposed to create a valid instance of a class. "Valid" here meaning "in a valid state." However, that's not what the NcnnModel constructor does. This constructor doesn't create a valid NCNN model. 

In other languages, you might make the empty constructor private and expose this method as a constructor overload, but Python supports neither of those features, so we have to resort to static constructor/factory methods instead.

backend/src/nodes/utils/auto_split.py
```diff
@@ -12,6 +12,34 @@ class Split:
     pass
 
 
+def estimate_tile_size(
+    free: int,
```
RunDevelopment, line 16: I like NCNN's term for it: budget. That way, we could also get rid of `allowed_usage`. We simply say that we are allowed some budget and want a tile size for that.

backend/src/nodes/utils/auto_split.py
```diff
@@ -12,6 +12,34 @@ class Split:
...
+    img_element_size: int = 4,
+    allowed_usage: float = 0.8,
+) -> Union[int, None]:
+    h, w, c = get_h_w_c(img)
+    img_bytes = h * w * c * img_element_size
+    mem_required_estimation = (model_size / (1024 * 52)) * img_bytes
+
+    tile_pixels = w * h * (free * allowed_usage) / mem_required_estimation
+    # the largest power-of-2 tile_size such that tile_size**2 < tile_pixels
+    tile_size = 2 ** (int(tile_pixels**0.5).bit_length() - 1)
```
RunDevelopment, line 29: We talked about this on discord: I also rounded it down to a power of 2 is because I wanted to avoid OOM. I judged that an OOM is more expensive than a smaller tile size.

If you want to remove the power of 2, make it `tile_size = int(tile_pixels**0.5)`.

backend/src/nodes/utils/auto_split.py
```diff
@@ -12,6 +12,34 @@ class Split:
...
 
 
+def estimate_tile_size(
+    free: int,
+    total: int,
+    model_size: int,
+    img: np.ndarray,
+    img_element_size: int = 4,
+    allowed_usage: float = 0.8,
+) -> Union[int, None]:
```
RunDevelopment, line 22: We always return a tile size.

```suggestion
) -> int:
```

backend/src/nodes/utils/auto_split.py
```diff
@@ -12,6 +12,34 @@ class Split:
...
+    h, w, c = get_h_w_c(img)
+    img_bytes = h * w * c * img_element_size
+    mem_required_estimation = (model_size / (1024 * 52)) * img_bytes
```
RunDevelopment, line 25: This factor is very specific to PyTorch. Is that even applicable to NCNN?

backend/src/nodes/nodes/ncnn/upscale_image.py
```diff
@@ -69,16 +70,26 @@ def __init__(self):
...
         try:
             vkdev = ncnn.get_gpu_device(exec_options.ncnn_gpu_index)
-            # logger.info(vkdev.get_heap_budget())
+            heap_budget = vkdev.get_heap_budget() * 1024 * 1024
+
+            max_tile_size = (
+                tile_size
+                if tile_size is not None
+                else estimate_tile_size(
+                    heap_budget, heap_budget, model.bin_length, img, 4, 0.8
```
RunDevelopment, line 89: [Are you sure we need the 80% usage?](https://github.com/Tencent/ncnn/blob/bbbe17c5b502d4128968cf755ab533265289d620/src/gpu.cpp#L2907)

RunDevelopment (changes requested):

backend/src/nodes/utils/ncnn_model.py
joeyballentine: Why not just move this code to the constructor then?

backend/src/nodes/utils/auto_split.py
```diff
@@ -12,6 +12,34 @@ class Split:
     pass
 
 
+def estimate_tile_size(
+    free: int,
+    total: int,
```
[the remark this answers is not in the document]
joeyballentine: I guess not, but imo the logging from it (for pytorch) is useful

backend/src/nodes/utils/auto_split.py
joeyballentine: I haven't really tested it enough to know for sure, but it at least seems to work the same for ncnn

backend/src/nodes/nodes/ncnn/upscale_image.py
joeyballentine: That's only if `(!info.support_VK_EXT_memory_budget())`

backend/src/nodes/utils/ncnn_model.py
RunDevelopment: I believe we need the empty constructor somewhere else. Zamboni actually implemented "overloading" before moving to a static constructor method. It was really messy since we had to deduce what to do depending on which args we got. This might be doable for empty constructor + load from file, but it gets worse quickly. Just try adding another "overload"... 
There's also the problem that it's not terribly descriptive. Guess what `NcnnModel(someString)` does. `NcnnModel.load_from_file(someString)` tells you want it does.

For those reasons (and others), there are even some programming languages that don't do constructors at all. E.g. in Rust, there are only static constructors. This makes it really easy to understand what's going on. Ever spent 10mins reading through C++ docs, because your IDE couldn't figure out which of the 20+ vector overloads was called?

joeyballentine: makes sense

backend/src/nodes/nodes/ncnn/upscale_image.py
```diff
@@ -69,16 +70,24 @@ def __init__(self):
     def upscale(
         self,
         img: np.ndarray,
-        net,
+        model: NcnnModel,
         input_name: str,
         output_name: str,
-        max_tile_size: Union[int, None],
+        tile_size: Union[int, None],
```
RunDevelopment, line 76: It's called max tile size, cause that's what it is. auto_split will choose tiles that are at most this large, not exactly this large.

backend/src/nodes/nodes/ncnn/upscale_image.py
```diff
@@ -94,15 +103,14 @@ def upscale(
                 )
         except ValueError as e:
             raise e
+        except RuntimeError as e:
+            raise e
```
RunDevelopment, line 107: I think just `raise` would be better here.

```suggestion
        except (RuntimeError, ValueError):
            raise
```

RunDevelopment (changes requested):

backend/src/nodes/nodes/ncnn/upscale_image.py
RunDevelopment: Good point.

backend/src/nodes/nodes/ncnn/upscale_image.py
joeyballentine: didn't know you could do that

backend/src/nodes/nodes/ncnn/upscale_image.py
RunDevelopment: Ahh, whatever.

joeyballentine: We can just change it back later. Sorry but I can't change it right this second

RunDevelopment: Nah, it's fine. We'll likely have to change that variable later anyway, and it's not important enough to hold the PR.

PR #1071, opened 2022-10-09
Separate backend & frontend workflows, run only when each changes

joeyballentine:
Added path specifiers so that we don't run backend-related stuff when nothing in the backend changed, and the same for frontend.

Reasoning: There's no reason to waste our time having backend tests run when only frontend code changed, and vice versa.
Also, running everything when I increment the version number or make a readme change is also just a waste of time


RunDevelopment:
Is there actually a reason for this? Python and JS lints and tests all take about the same amount of time, so you don't really gain anything since they all run in parallel.

Also, you need to include the various config files. E.g. JS needs package.json, ESLint config and that sort of stuff. You should probably include all top-levels in the repo just to be sure.

joeyballentine:
> Is there actually a reason for this?

We've had backend-only PRs fail in the past because frontend workflows started failing for no reason. Also, there is no reason to run any CI when stuff like the README is the only change. 

I could switch these to being path exclusions instead, that way it would by default include everything at the top level.

RunDevelopment:
> We've had backend-only PRs fail in the past because frontend workflows started failing for no reason.

Good point. Making that less likely is worth it.

> I could switch these to being path exclusions instead, that way it would by default include everything at the top level.

I didn't know you could that, but that sounds better, yes.

joeyballentine:
There, now the frontend workflows include everything but the backend stuff, and the backend workflows just include the backend stuff.

RunDevelopment (approved):

PR #1073, opened 2022-10-09
Add resize node tests

joeyballentine:
- Adds tests for resize by factor and resize by resolution nodes
- Adds some test images of varying sizes
- fixes some things that weren't relative imports

These tests are kinda useless, but I wanted to use them as a starting off point for writing actual tests


joeyballentine:
closing this because i'll probably just have to redo it anyway to get it to not fail

PR #1083, opened 2022-10-11
Improve the Nvidia-SMI check

joeyballentine:
- On windows, we weren't just checking if running `nvidia-smi` worked. I think that's because the `where nvidia-smi` command should only work if `nvidia-smi` works, but might as well check anyway.
- Added a check for if nvidia-smi.exe is in System32. We've seen cases where this is true, so might as well check for it.

### DIFF
--- a/src/main/nvidiaSmi.ts
+++ b/src/main/nvidiaSmi.ts
@@ -28,13 +28,10 @@ const hasOutput = (command: string): Promise<boolean> => {
 };
 
 const getNvidiaSmiCommandFromPath = async (): Promise<string | undefined> => {
-    if (
-        platform === 'win32' &&
-        ((await hasOutput('where nvidia-smi')) || (await hasOutput('nvidia-smi')))
-    ) {
+    if (await hasOutput(`${platform === 'win32' ? 'where' : 'which'} nvidia-smi`)) {
         return 'nvidia-smi';
     }
-    if ((await hasOutput('which nvidia-smi')) || (await hasOutput('nvidia-smi'))) {
+    if (await hasOutput('nvidia-smi')) {
         return 'nvidia-smi';
     }
     return undefined;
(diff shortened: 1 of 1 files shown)


src/main/nvidiaSmi.ts
```diff
@@ -28,7 +28,10 @@ const hasOutput = (command: string): Promise<boolean> => {
 };
 
 const getNvidiaSmiCommandFromPath = async (): Promise<string | undefined> => {
-    if (platform === 'win32' && (await hasOutput('where nvidia-smi'))) {
+    if (
```
RunDevelopment, line 31: I think the whole function can be simplified to this:
```ts
const getNvidiaSmiCommandFromPath = async (): Promise<string | undefined> => {
    if (await hasOutput(`${platform === 'win32' ? "where" : "which"} nvidia-smi`)) {
        return 'nvidia-smi';
    }
    if (await hasOutput('nvidia-smi')) {
        return 'nvidia-smi';
    }
    return undefined;
};
```

src/main/nvidiaSmi.ts
```diff
@@ -72,6 +75,11 @@ const getNvidiaSmiCommand = async (): Promise<string | undefined> => {
         if (targetDir) {
             return path.join(basePath, targetDir, 'nvidia-smi.exe');
         }
+        // check if smi is in System32
+        const smiPath = `${WINDIR}\\System32\\nvidia-smi.exe`;
+        if ((await fs.stat(smiPath)).isFile()) {
```
RunDevelopment, line 80: Doesn't `fs.stat` throw an error if the file/dir doesn't exist? Is this function intended to throw an error?

You might want to use our `checkFileExists` function.

joeyballentine: It is intended for that

RunDevelopment (approved):

PR #1085, opened 2022-10-11
Always convert outputs to list

RunDevelopment:
Broadcasts and the executor were using 2 different methods to convert the raw outputs of a node into a list. This previously created problems when Joey tried to return a tuple value in NCNN nodes.

To prevent this from happening again, I made it so that outputs are now always converted by looking at the number of `.outputs` a node has. I also added assertions to make sure that nodes actually return what their `.outputs` say they return. At least, the number of outputs is checked, right now. We should probably also have some kind of output validation to check the correctness of node implementations at runtime. We already had the Canny edge detection node that returned an image with the wrong number of channels. If we also give `ImageOutput`s a `channels` parameters like for inputs, we'll be able to detect such errors in the future.

Other changes:
- Set images to readonly.
- Made the cache class generic over its items. I did this to decouple the cache implementation from the `Output` type.

### DIFF
--- a/backend/src/chain/cache.py
+++ b/backend/src/chain/cache.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Iterable, Optional, Set
+from typing import Any, Dict, Generic, Iterable, Optional, Set, TypeVar
 import gc
 from sanic.log import logger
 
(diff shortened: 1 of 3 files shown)


RunDevelopment:
About the added output assertions: I expect some nodes to fail. I haven't tested all nodes, so there might be one or two that will just stop working now because their outputs don't match up with their `.outputs`. Since this is a runtime check, there isn't really a good way to verify the correctness of nodes statically.

joeyballentine (approved):

Issue #1034, opened 2022-09-25
Paste images into chainner


RunDevelopment:
**Motivation**
I sometimes move back and forth between my image editing software and chainner. So tasks are easier to do with chainner and others are not, so I sometimes switch between the two.

The Copy to Clipboard node makes it easy to get images *out* of chainner, but there's no easy to get them into chainner without manually creating a temporary file.

**Description**
We already have listen for Ctrl+C, so let's do the following when we see an image in the clipboard:

- Save the image as a temporary PNG.
- Create a new Load Image node with the temporary file.

PR #1088, opened 2022-10-12
Fix clipboard paste crash

joeyballentine:
I will eventually make the image handler write a temp file and make a new node referencing it. However, for now I just fixed the crash and added some boilerplate for standard pasting formats.

Basically, I'll work on the image stuff tomorrow, but figured I'd just PR my crash fix now


RunDevelopment:
So those changes are prep work for #1034, right?

joeyballentine:
Yes

joeyballentine:
For now though, really all it does is prevents the frontend from erroring if you paste anything that isn't a chain

RunDevelopment (approved):

PR #1095, opened 2022-10-13
New tiling options

RunDevelopment:
This implements the tiling options we've been talking about on discord.

The main new abstraction is a `Tiler` (I'm not too keen on the name btw). This interface defines (1) the initial tile size and (2) how splits are implemented (=how the tile size decreases).  This makes it trivial to implement something like a no tiling mode. The interface will also allow us to implement tile size memorization.

image

Notes:
- I implemented the "Maximum" as just a very large tile size (2**31).
- If CUDA isn't available, pytorch will default to "Maximum" in "Auto (estimate)" mode.
- Since the Tile Size dropdown now contains 2 magic numbers, I added a method to parse this number into a tiler.


joeyballentine:
> I implemented the "Maximum" as just a very large tile size (2**31).

why not make this a more manageable magic number like -2 or something?

RunDevelopment:
Sure.

joeyballentine (approved):

PR #1102, opened 2022-10-16
Add pasting an image from clipboard

### LINKED ISSUE #1034: Paste images into chainner

joeyballentine:
Closes #1034

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -1037,7 +1037,7 @@ export const GlobalProvider = memo(
             copyToClipboard(getNodes(), getEdges());
         }, [getNodes, getEdges]);
         const pasteFn = useCallback(() => {
-            pasteFromClipboard(changeNodes, changeEdges);
+            pasteFromClipboard(changeNodes, changeEdges, createNode);
         }, [changeNodes, changeEdges]);
         const selectAllFn = useCallback(() => {
             changeNodes((nodes) => nodes.map((n) => ({ ...n, selected: true })));
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -166,6 +166,7 @@ export const GlobalProvider = memo(
             setNodes: rfSetNodes,
             setEdges: rfSetEdges,
             viewportInitialized,
+            project,
         } = useReactFlow<NodeData, EdgeData>();
 
         const currentViewport = useViewport();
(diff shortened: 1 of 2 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1037,7 +1037,7 @@ export const GlobalProvider = memo(
             copyToClipboard(getNodes(), getEdges());
         }, [getNodes, getEdges]);
         const pasteFn = useCallback(() => {
-            pasteFromClipboard(changeNodes, changeEdges);
+            pasteFromClipboard(changeNodes, changeEdges, createNode);
         }, [changeNodes, changeEdges]);
```
RunDevelopment, line 1041:
```suggestion
            pasteFromClipboard(changeNodes, changeEdges, createNode);
        }, [changeNodes, changeEdges, createNode]);
```

src/renderer/helpers/copyAndPaste.ts
```diff
@@ -98,9 +102,37 @@ export const pasteFromClipboard = (
...
+                case 'image/bmp':
+                case 'image/tiff':
+                case 'image/png': {
+                    const tempDir = os.tmpdir();
+                    const imgData = clipboard.readImage().toPNG();
+                    const randName =
+                        (Math.random() + 1).toString(36).substring(7) +
+                        (Math.random() + 1).toString(36).substring(7) +
+                        (Math.random() + 1).toString(36).substring(7);
+                    const imgPath = path.join(tempDir, `chaiNNer-clipboard-${randName}.png`);
```
RunDevelopment, line 116: Why not use a UUID as the random name?

```suggestion
                    const imgData = clipboard.readImage().toPNG();
                    const imgPath = path.join(os.tmpdir(), `chaiNNer-clipboard-${uuid4()}.png`);
```

src/renderer/helpers/copyAndPaste.ts
```diff
@@ -98,9 +102,37 @@ export const pasteFromClipboard = (
...
+                    writeFile(imgPath, imgData)
+                        .then(() => {
+                            log.debug('Clipboard image', imgPath);
+                            createNode({
+                                nodeType: 'regularNode',
+                                position: { x: 0, y: 0 },
```
RunDevelopment, line 122: Maybe we could place the node at the center of the current viewport instead of the origin? Right now, users might not even see the new node.

RunDevelopment (changes requested):

src/renderer/helpers/copyAndPaste.ts
joeyballentine: That's a good point

RunDevelopment (approved):

PR #1108, opened 2022-10-19
Add more encoding options to video frame iterator

joeyballentine:
Adds MKV and Webm as types. Also adds the encoding presets as well as a CRF slider

image

### DIFF
--- a/backend/src/nodes/properties/inputs/generic_inputs.py
+++ b/backend/src/nodes/properties/inputs/generic_inputs.py
@@ -223,8 +223,28 @@ def VideoTypeDropdown() -> DropDownInput:
         label="Video Type",
         options=[
             {"option": "MP4", "value": "mp4"},
+            {"option": "MKV", "value": "mkv"},
+            {"option": "Webm", "value": "webm"},
             {"option": "AVI", "value": "avi"},
             {"option": "None", "value": "none"},
+        ],
+    )
+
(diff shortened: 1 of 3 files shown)


backend/src/nodes/properties/inputs/generic_inputs.py
```diff
@@ -223,12 +223,32 @@ def VideoTypeDropdown() -> DropDownInput:
...
+        label="Preset",
+        options=[
+            {"option": "ultrafast", "value": "ultrafast"},
+            {"option": "superfast", "value": "superfast"},
+            {"option": "veryfast", "value": "veryfast"},
+            {"option": "fast", "value": "fast"},
+            {"option": "medium", "value": "medium"},
+            {"option": "slow", "value": "slow"},
+            {"option": "slower", "value": "slower"},
+            {"option": "veryslow", "value": "veryslow"},
```
RunDevelopment, line 247: Why not use nice labels here?

joeyballentine: Because these are standard ffmpeg terms. Unless you mean capitalize them or something

RunDevelopment: That's alright then.

RunDevelopment (approved):

PR #1110, opened 2022-10-19
Manually specify codecs for formats, add GIF

joeyballentine:
Fixes AVI being encoded with a terrible quality codec. Adds support for saving video as GIF

### DIFF
--- a/backend/src/nodes/nodes/image/video_frame_iterator.py
+++ b/backend/src/nodes/nodes/image/video_frame_iterator.py
@@ -45,7 +45,6 @@ if not has_ffmpeg():
 codec_map = {
     "mp4": "libx264",
     "avi": "libx264",
-    "mov": "libx264",
     "mkv": "libx264",
     "webm": "libvpx-vp9",
     "gif": "gif",
--- a/backend/src/nodes/nodes/image/video_frame_iterator.py
+++ b/backend/src/nodes/nodes/image/video_frame_iterator.py
@@ -42,6 +42,16 @@ if not has_ffmpeg():
     static_ffmpeg.add_paths()
 
 
+codec_map = {
+    "mp4": "libx264",
+    "avi": "libx264",
+    "mov": "libx264",
+    "mkv": "libx264",
+    "webm": "libvpx-vp9",
+    "gif": "gif",
+}
+
+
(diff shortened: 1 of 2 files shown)


backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -136,6 +145,7 @@ def run(
                         r=fps,
                         crf=crf,
                         preset=video_preset if video_preset != "none" else None,
+                        vcodec=codec_map[video_type],
```
RunDevelopment, line 148: There is no "none" video preset, but there is a "none" video type. Did you add this check for the wrong variable?

RunDevelopment: > Using none won't even get this far in the code

Oops, I didn't see the if at the start of the function. Yes, that's fine.

> And no, this is supposed to be video type

Video preset can't have the value `"none"` though. The dropdown just doesn't have this as an option.

joeyballentine: There's no way to not specify a name, directory, or quality either. It just gets ignored when you set it to none. 

I wanna get rid of none anyway. The only reason it exists is so you can export to an image sequence using the save image node, but since ffmpeg can export image sequences I figure we can just make that an option instead and make it a bit easier and intuitive

RunDevelopment: Not what I meant. I mean `video_preset if video_preset != "none" else None` is useless because `video_preset` can't be `"none"`.

joeyballentine: Oooooh, I didn't notice that because it wasn't in the little diff in the comment. Yeah I think I just forgot to remove that when I removed none as an option. But also that's not really related to this PR

RunDevelopment: > that's not really related to this PR

True.

RunDevelopment (approved):

backend/src/nodes/nodes/image/video_frame_iterator.py
joeyballentine: Actually, it _is_ in the diff in the comment.... I just didn't notice it lol. I'll fix that when I add the PNG sequence stuff

PR #1113, opened 2022-10-20
Use while/break instead of for loop for video iterator

joeyballentine:
Relies on reading the stream until its end rather than on the frame count. The frame count of a video can be inaccurate as it's just metadata encoded onto a video rather than an actual computed value. It is however a good enough estimate that we can still use it for progress.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -168,6 +168,37 @@ class IteratorContext:
                 "Errors occurred during iteration: \n• {}".format("\n• ".join(errors))
             )
 
+    async def run_while(
+        self,
+        length: int,
+        before: Callable[[int], Union[None, Literal[False]]],
+    ):
+        errors: List[str] = []
+        index = 0
+        while True:
+            try:
+                await self.progress.suspend()
+
+                result = before(index)
+                if result is False:
+                    break
+
+                await self.run_iteration(index, length)
+
+                if index < length:
+                    index += 1
+            except Aborted:
+                raise
+            except Exception as e:
+                logger.error(e)
+                errors.append(str(e))
+
+        if len(errors) > 0:
+            raise Exception(
+                # pylint: disable=consider-using-f-string
+                "Errors occurred during iteration: \n• {}".format("\n• ".join(errors))
+            )
+
 
 def timed_supplier(supplier: Callable[[], T]) -> Callable[[], Tuple[T, float]]:
     def wrapper():
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -170,28 +170,43 @@ class IteratorContext:
 
     async def run_while(
         self,
-        length: int,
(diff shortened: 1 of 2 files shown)


backend/src/process.py
```diff
@@ -168,6 +168,37 @@ async def run(
...
+                await self.progress.suspend()
+
+                result = before(index)
+                if result is False:
+                    break
+
+                await self.run_iteration(index, length)
+
+                if index < length:
+                    index += 1
```
RunDevelopment, line 189: Right now, progress is not correctly calculated if the number of actual iterations is different from `length`.

If there are more iterations, then `index` will be equal to `length` and since `run_iteration` sends out `(index+1)/length` as progress, this will result in progress >100%. This is not a problem with `run_iteration` since that function assumes valid indexes (as you would).

If there are fewer iterations, then the progress never reaches 100%. This should be easy to fix by sending out 100% at the end of the while loop if `index < length`.

Also, the current `index += 1` doesn't run if either `before(index)` or `run_iteration` errors. This is a huge problem because `before` might require the index to be correct.

RunDevelopment (changes requested):

backend/src/process.py
```diff
@@ -168,6 +168,37 @@ async def run(
                 "Errors occurred during iteration: \n• {}".format("\n• ".join(errors))
             )
 
+    async def run_while(
+        self,
+        length: int,
```
RunDevelopment, line 173: It might be more accurate to call this `length_estimate`.

backend/src/process.py
```diff
@@ -168,6 +168,52 @@ async def run(
...
+        errors: List[str] = []
+        index = -1
+        while True:
+            try:
+                await self.progress.suspend()
+
+                if index < length_estimate:
+                    index += 1
+                else:
+                    length_estimate = int(length_estimate * 1.1)
```
RunDevelopment, line 185: What's the idea here?

Also, you still skip incrementing the index sometimes, so the index of some frames will be wrong.

joeyballentine: Oh yeah, oops. 

The idea here is if we pass the estimate, the estimate is obviously too small so let's just increase it a bit and keep going

RunDevelopment: Couldn't we do something like keep it at 99% until we exit the while loop? Basically, we could always do `run_iteration(min(index, length - 1), length)`.

joeyballentine: We could I guess, but this way shows that there's still progress going on since you can see it still increasing

joeyballentine: Idk which is better tbh. On the one hand, sitting at 99% is more accurate, but going backwards a bunch and then slowly going up is a bit more indicative of something actively happening.

RunDevelopment (approved):

PR #1120, opened 2022-10-21
Increase PyTorch's VRAM estimation budget

joeyballentine:
I think 0.6 is just a bit too low


RunDevelopment (approved):
I see you're starting the next round of binary search ;)

PR #1123, opened 2022-10-22
Replace static-ffmpeg package with custom ffmpeg downloader

joeyballentine:


src/main/ffmpeg/ffmpeg.ts
```diff
@@ -0,0 +1,147 @@
...
+        case 'win32':
+        case 'linux':
+        case 'darwin':
+            return platform;
+        default:
+            throw new Error(
+                `Unsupported platform: Our FFMPEG is not supported on ${platform}. Please report this to us and we may add support.`
+            );
+    }
+};
```
RunDevelopment, line 37: Pretty sure I used this somewhere else too. Maybe we could move this in a `src/main/platform.ts`?

RunDevelopment (changes requested):
Sorry, I didn't read the conv on Discord. Was spinning up our own ffmpeg installation really the best option? Why didn't static ffmpeg work?

joeyballentine:
The static ffmpeg package was crashing for Zamboni. No clue why. This works better and gives us more control anyway.

src/main/ffmpeg/ffmpeg.ts
joeyballentine: That's a good idea. And yeah a lot of this stuff is similar to the python downloading stuff. If we ever need a third thing like this we should probably abstract it away a bit more

PR #1126, opened 2022-10-23
Add compression level to save image

joeyballentine:
The default is 30 because that translates to 3, which we determined was the fastest for png compression. It also translates to 70% quality for jpeg and webp, which I think makes sense enough.

It's a bit odd the way it is, since ideally it would change depending on what format was selected, but we all know we cant do that right now. This was the best solution I had.

image

### DIFF
--- a/backend/src/nodes/nodes/image/save_image.py
+++ b/backend/src/nodes/nodes/image/save_image.py
@@ -18,6 +18,7 @@ from ...properties.inputs import (
     DirectoryInput,
     TextInput,
     ImageExtensionDropdown,
+    SliderInput,
 )
 from ...utils.pil_utils import *
 from ...utils.utils import get_h_w_c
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I really *really* don't like this.
1. I think the general approach of having an input that gets repurposed or ignored depending on the image format is bad. 
2. This specific compression level slider has numerous problems:
    - PNG compression level is unintuitive and most users probably won't even know what it is, because most program that save as PNG don't expose this option.
    - PNG compression ignores 90% of all inputs because of the `/ 10`.
    - JPEG/WEBP quality is unintuitive.
    - The default only makes sense for PNG. 70% JPEG quality really doesn't look good.

You also didn't add a migration. This is a breaking change because chains that save as JPEG right now will produce noticeable worse-looking images after this PR gets merged because of the 70% quality default.

So while I really don't like this approach in general, I also understand the need for a temporary solution. From what I saw, most users requested this to change the JPEG quality setting. So why not have one slider "Quality (JPEG/WEBP only)" and make it the 0-100% slider people already know and love from other image editors? 
I have never seen people request PNG compression levels before, and the setting doesn't do much anyway given how good OpenCV's default is. So I don't see why we should support it.

joeyballentine:
That's a good point, I think that's a much better idea

RunDevelopment (approved):

PR #1127, opened 2022-10-23
Allow selecting custom python path + fall back to integrated python

### LINKED ISSUE #828: Option to use homebrew's python?
**Motivation**
Is it possible to use homebrew's python instead the bundled python (looks like 3.9?) or the system python (3.8.9)?

The reason I ask is I can only get ONNX to load and work by using the system python and despite having NCNN saying it's install via the dependancies manager I can't use any of the nodes. That's only after switching to the system's python. I tried to use the bundled python, but then neither ONNX nor NCNN would install at all. However, via homebrew I have the latest python (3.10.6), NCNN, ONNX, PyTorch, etc. and it's easier to keep everything updated and manage in case things get screwy. Which with me is almost a certainty.

**Description**
If it's at all possible, in addition to having option for the bundled python or the system python maybe an option to use homebrew's python. Homebrew is saved in slightly different locations between the Intel Macs and the silicon Macs though and I've no idea how difficult it would be to implement something like that… Or maybe a way to manually interface the bundled python to troubleshoot… It's not a fully fleshed out idea. Just throwing it out there. 

**Alternatives**
I'm not really sure of alternatives. I've used homebrew's python to run your forked BlueAmulet/ESRGAN and other upscalers without much issue outside of having to use cpu and run outta of memory or whatever.

Oh, speaking of memory I have a side note question of sorts… Could this relate to this project at all? https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

Totally random question/request and I don't mind if it's not acted on.
Again, just throwing it out there. (＾ᴗ＾;;)

joeyballentine:
Closes #828

Gonna be honest I didn't actually test this aside from regression testing system python, only because I don't have another python instance. But I see no reason for this not to work. And if the selected file isn't a valid python instance, it'll skip it and fall back to 'python3' or 'python'. Speaking of which, we should add the integrated python path to that loop, so it can fall back to that instead of just erroring and dying when system python isn't available at all (I can do that in a different PR though)


src/main/python/systemPython.ts
```diff
@@ -1,12 +1,18 @@
...
 
-export const getSystemPython = async (): Promise<PythonInfo> => {
-    for (const py of ['python', 'python3']) {
-        // eslint-disable-next-line no-await-in-loop
-        const version = await getPythonVersion(py).catch(() => null);
-        if (version && isSupportedPythonVersion(version)) {
-            return { python: py, version };
+export const getSystemPython = async (localStorageLocation: string): Promise<PythonInfo> => {
+    const localStorage = new LocalStorage(localStorageLocation);
+    const systemPythonLocation = localStorage.getItem('system-python-location');
```
RunDevelopment, line 7: I don't like this API. Why does `getSystemPython` have to know how we store settings?

Instead, how about we pass in a string array of python locations to check? The calling code would look something like this:
```ts
    const localStorage = new LocalStorage(localStorageLocation);
    const systemPythonLocation: string | null = localStorage.getItem('system-python-location');
    pythonInfo = await getSystemPython(systemPythonLocation ? [systemPythonLocation] : []);
```

RunDevelopment (changes requested):

src/main/python/systemPython.ts
joeyballentine: That's a good idea

joeyballentine:
Decided to go ahead and add the integrated python fallback + warning as well.

src/main/python/systemPython.ts
```diff
@@ -1,12 +1,14 @@
...
 
-export const getSystemPython = async (): Promise<PythonInfo> => {
-    for (const py of ['python', 'python3']) {
-        // eslint-disable-next-line no-await-in-loop
-        const version = await getPythonVersion(py).catch(() => null);
-        if (version && isSupportedPythonVersion(version)) {
-            return { python: py, version };
+export const getSystemPython = async (pythonsToCheck: string[]): Promise<PythonInfo> => {
+    for (const py of pythonsToCheck) {
+        if (py) {
```
RunDevelopment, line 6: Please filter out invalid paths before passing them into the function.

RunDevelopment (approved):

PR #1128, opened 2022-10-23
Prevent context node selector cross-iterator connections (& more)

### LINKED ISSUE #1028: Context selector can create invalid connections
It is not allowed to create downstream connections from child nodes to free nodes.

However, the context selector can currently create these invalid connections. If you start from the input of a free node and drag the handle into the iterator, the context selector pane will pop up and allow you to create a child node.

image

image

For some reason, this can even be made into a valid chain.

image

So this report includes 2 bugs that need to be fixed:

1. The context selector should not create invalid connections from child nodes to free nodes.
2. Our node validity checks should catch this obviously invalid change.

joeyballentine:
Closes #1028

~~Just realized this only covers the cross-iterator portion. will also fix input connections~~

I think I handled every condition. I had to split these up into so many boolean variables just to keep everything straight. It seems to work correctly now.

Note that this doesn't handle condition 2 of the issue, which is making the node validity checks check for this. But TBH I don't really think that's necessary so long as we handle preventing those cases correctly.

### DIFF
--- a/src/renderer/hooks/usePaneNodeSearchMenu.tsx
+++ b/src/renderer/hooks/usePaneNodeSearchMenu.tsx
@@ -473,16 +473,30 @@ export const usePaneNodeSearchMenu = (
                     ? firstClass.slice('iterator-editor='.length)
                     : undefined;
 
-            if (isStoppedOnPane || stoppedIteratorId) {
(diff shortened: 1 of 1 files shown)


RunDevelopment (approved):
I'm not gonna even try to understand that.

PR #1133, opened 2022-10-23
Replaced weights_bin with serialize method

theflyingzamboni:
Fixes bug where interpolating ncnn models would cause an empty bin file to be saved due to clearing the weights_bin attribute. Resolved this by discarding the attribute entirely and replacing it with a serialize method to be called to convert the weights to a byte string when needed. Any performance overhead seems to be indistinguishable in terms of runtime, and regardless is only incurred when saving or the first time an upscale is run using a model. Also prevents the need for the full byte string to be stored in RAM in the model object in the first place.


RunDevelopment (approved):
LGTM;

RunDevelopment (approved):

PR #1136, opened 2022-10-24
Prevent face sr models from being converted to ncnn and onnx

joeyballentine:


RunDevelopment (approved):

RunDevelopment (changes requested):
Actually, no. Remember, the type system is a best-effort, it cannot be trusted for correctness. Please also add runtime assertions to ensure correctness.

If it's common to disallow face arch models, then maybe we could even add a `allow_face: bool` parameter to the model input that controls both the type and a runtime check.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -12,25 +12,34 @@
...
         self.icon = "ONNX"
         self.sub = "Utility"
 
-    def run(self, model: PyTorchModel) -> OnnxModel:
+    def run(self, model: PyTorchSRModel) -> OnnxModel:
         exec_options = to_pytorch_execution_options(get_execution_options())
 
+        assert isPyTorchSRModel(
+            model
+        ), "Only normal SR models can be converted to ONNX at this time."
```
RunDevelopment, line 41: From a type perspective, this doesn't make any sense. You already declared that the type of `model` is an SR torch model, so `isPyTorchSRModel` should always return true. Of course, that's not what actually happens because the type of `model` is simply wrong. `model` is actually a `PyTorchModel` (no SR).

There are 2 ways to fix this:
1. Move the runtime type into the `enforce` method of the input. `model: PyTorchSRModel` will now be correct and the assertion in this `run` method can be removed.
2. Type narrowing using [type guards](https://peps.python.org/pep-0647/). I *think* you used those wrong type annotations because python didn't understand that `isPyTorchSRModel` and `isPyTorchFaceModel` are just `isinstance` checks, and so it didn't narrow the type of `model: PyTorchModel` correctly.

    This can be fixed by using a type guard like this:
    ```py
    def isPyTorchFaceModel(model: PyTorchModel) -> TypeGuard[PyTorchFaceModel]:
        return isinstance(model, PyTorchFaceModels)
    ```

    Note that while type guards are a relatively standard feature for type checkers, it's a recent addition (py 3.10) to python, so we'll need [type-extensions](https://pypi.org/project/typing-extensions/). While this is another dep, I don't think it's a bad idea to add it. I already encountered a few situations where I needed a py 3.11 type and had to find another solution instead.

RunDevelopment (changes requested):

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
joeyballentine: I'll just add to the enforce method

backend/src/nodes/properties/inputs/pytorch_inputs.py
```diff
@@ -5,6 +5,15 @@
 
 from typing import Union
 
+from ...utils.torch_types import (
```
RunDevelopment, line 8: This file optionally imports torch, but the archs imported by `torch_types.py` require torch. Won't this cause all inputs to fail if torch isn't installed since torch inputs are exported in `properties/inputs/__init__.py`?

backend/src/nodes/properties/inputs/pytorch_inputs.py
```diff
@@ -20,9 +29,46 @@ def __init__(
     ):
         super().__init__(input_type, label)
 
-    def enforce(self, value):
+    def enforce(self, value: PyTorchModel):
+        if torch is not None:
+            assert isinstance(value, torch.nn.Module), "Expected a PyTorch model."
+            assert isPyTorchModel(value), "Expected a supported PyTorch model."
```
RunDevelopment, line 35: Again, this doesn't make sense from a type perspective. You declared that `value` is a `PyTorchModel`, so all these assertions are supposed to always be true. Of course, they aren't because the type of `value` is wrong. We don't know the type of `value`, that's why we have these checks.

Please remove the type annotation on `value` in all `enforce` methods.

backend/src/nodes/properties/inputs/pytorch_inputs.py
```diff
@@ -20,9 +29,46 @@ def __init__(
...
+        return value
+
+
+class SrModelInput(ModelInput):
+    def __init__(
+        self,
+        label: str = "Model",
+        input_type: Union[
+            str, ExpressionJson
+        ] = "PyTorchModel { arch: invStrSet(PyTorchModel::FaceArchs) }",
```
RunDevelopment, line 45: Woops. `ExpressionJson` already includes `str`, so this union is useless. Seems like I did this wrong when I added `ExpressionJson`.

backend/src/nodes/utils/torch_types.py
```diff
@@ -18,3 +19,23 @@
...
+def isPyTorchFaceModel(model: PyTorchFaceModel):
+    return isinstance(model, PyTorchFaceModels)
+
+
+PyTorchModels = (*PyTorchSRModels, *PyTorchFaceModels)
+PyTorchModel = Union[PyTorchSRModel, PyTorchFaceModel]
+
+
+def isPyTorchModel(model: PyTorchModel):
+    return isinstance(model, PyTorchModels)
```
RunDevelopment, line 41: From its type, you already know that `model` is a pytorch model, so this function should always return true. Of course, that's not what the function is supposed to do so let's fix its type. 

In general, the point of these type check functions is that they take a value of some type and return whether the current value is actually a most specific type. The best example is `isinstance(x, t)`. This function tells you whether a value x (of any type) is also of type `t`. The key point here is that the input value `x` has a broader type than the type being checked. E.g. `isinstance("foo", str)` doesn't make sense to check in code because we already know that `"foo"` is a `str`.

So how broad can we go for `model`? As far the implementation of `isPyTorchModel` model allows us.
Since we only use `isinstance` to implement `isPyTorchModel`, we can allow anything that `isinstance` allows and `isinstance` allows everything. Literally any value.

Same for `isPyTorchFaceModel` and `isPyTorchSRModel` as well.

```suggestion
def isPyTorchModel(model: object):
    return isinstance(model, PyTorchModels)
```

backend/src/nodes/properties/inputs/pytorch_inputs.py
```diff
@@ -20,9 +29,46 @@ def __init__(
...
+        return value
+
+
+class FaceModelInput(ModelInput):
+    def __init__(
+        self,
+        label: str = "Face SR Model",
+        input_type: Union[
+            str, ExpressionJson
+        ] = "PyTorchModel { arch: PyTorchModel::FaceArchs }",
```
RunDevelopment, line 62: There is a better way for doing this. We always want the input type to be a subset of `PyTorchModel { arch: PyTorchModel::FaceArchs }`, but the dev can just specify any type right now. Instead, I suggest using an intersection.

```py
input_type: ExpressionJson = "PyTorchModel"
super().__init__(label, intersect(input_type, "PyTorchModel { arch: PyTorchModel::FaceArchs }"))
```

RunDevelopment (approved):

PR #1138, opened 2022-10-24
RGB-like color conversion

RunDevelopment:
I recently had to upscale a set of mixed grayscale and RGB images. It was quite annoying because there as no single-node way to convert a grayscale or RGB image to RGB. So I added one.

This PR adds the Gray/RGB/RGBA color space detector. The "From" direction of Change Colorspace now allows color space detectors. These detectors look at the number of channels of an image and deduce the color space from that. Right now, there is only one detector, but we may expand this in the future for things like HSLA.

Note that detectors are only available for the "From" input.

image
image

---

Open questions:

I made Gray/RGB/RGBA the new default for the "From" direction. Should it be?
image

Since Gray/RGB/RGBA is a super set of the members, should we remove Gray, RGB, and RGBA from the "From" input? I tend towards keeping them for consistency, but they also feel a bit like dead weight now.

### DIFF
--- a/backend/src/nodes/nodes/image_utility/convert_color.py
+++ b/backend/src/nodes/nodes/image_utility/convert_color.py
@@ -11,7 +11,11 @@ from ...properties.inputs import (
 )
 from ...properties.outputs import ImageOutput
 from ...properties import expression
-from ...utils.color.convert import convert, color_space_from_id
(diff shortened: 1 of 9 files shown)


RunDevelopment:
I removed them
image

joeyballentine (approved):

PR #1139, opened 2022-10-24
Groups

RunDevelopment:
This PR adds groups. Groups are a new concept for inputs in chainner.

A group is a non-empty list of contiguous inputs. A group is able to influence both the behavior and display of inputs. This PR includes examples of both of this:
- The NCNN group keeps the .param and .bin file paths of NCNN model file inputs in sync.
- The From-To group alters how 2 dropdowns are displayed.

However, there are a few limitations:
- No nested groups.
- No group state. While I do plan to add state for groups to implement pseudo-variable-length input lists, I haven't done so in this PR because no group uses it rn.
- Handles. Right now, groups are not allowed to create handles (the things RF edges connect to). This is a big limitation because it makes things like linked inputs I had planned for Crop (Edges) impossible.

Other changes:
- Since groups decouple inputs from how they are displayed, input elements are no longer tied to a specific input ids. This allows groups to completely control input elements.
- Input props (e.g. min/max of number inputs) are now in `common-types.ts`. This makes it easier to work with node schemata.
- I fixed a bug related to Text Input defaults. TS found this because of the new types.


src/renderer/components/groups/NcnnFileInputsGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+    checkFileExists(file)
+        .then((exists) => {
+            if (exists) {
+                then();
+            }
+        })
+        .catch((reason) => log.error(reason));
+};
+
+export const NcnnFileInputsGroup = memo(
```
joeyballentine, line 19: Groups are going to be super difficult to use if they require whole frontend changes just to make a single group. Is there not a way you can make it dynamically create a group from a backend group definition?

RunDevelopment: No. Groups are kinda like inputs: you have to define a component for each group.

However, groups are reusable. E.g. any number of nodes can use the NCNN group.

joeyballentine: In that case, then it shouldn't just be an "ncnn" group, it should be a "linked file" group or something a little more generic. 

I still would rather have had something that just lets you combine any number of inputs however you want. Was there a technical limitation preventing that?

RunDevelopment: > In that case, then it shouldn't just be an "ncnn" group, it should be a "linked file" group or something a little more generic.

The logic that the group implements is specific to NCNN and doesn't make much sense in any other context.

I also want to warn of over abstraction here. I'd rather start with groups that fulfill a very specific purpose, and then generalize later as the need arises.

> I still would rather have had something that just lets you combine any number of inputs however you want.

Could you be more specific? I don't know what "however you want" may imply. I.e. grouping non-adjacent inputs together is kinda out of the question because this doesn't mix with the tree structure of React (assuming that we want groups to be able to determine the layout of multiple inputs (I want that, at least)).

joeyballentine: I mean like being able to define any two inputs and say you want the two to be next to each other, and one to have its value determined by the other -- surely this is something possible without having to make a "two-number-inputs-next-to-each-other-where-the-right-one-is-half-of-the-left-one" group

joeyballentine: Ultimately you're right though, it's probably better to just get it working right now and worry about abstraction later. But then of course, we risk our future selves wishing we had just abstracted it from the beginning because we're stuck with all these hardcoded groups and we find it annoying to have to make new ones all the time ;)

joeyballentine (dismissed):

joeyballentine:
Fix the merge conflicts and I'm good to merge this

joeyballentine (approved):

PR #1141, opened 2022-10-24
Improve missing dependency error logging & UI

joeyballentine:
These logs just end up confusing people who think they have a major problem by bot having the extra deps installed

### DIFF
--- a/backend/src/run.py
+++ b/backend/src/run.py
@@ -36,9 +36,32 @@ from response import (
     successResponse,
 )
 
+
+class AppContext:
+    def __init__(self):
+        self.executor: Optional[Executor] = None
+        self.cache: Dict[NodeId, Output] = dict()
+        # This will be initialized by setup_queue.
+        # This is necessary because we don't know Sanic's event loop yet.
+        self.queue: EventQueue = None  # type: ignore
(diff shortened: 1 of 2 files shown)


RunDevelopment:
I would argue that knowing which deps fail to import is pretty important for us to diagnose errors. This PR will make it so that we also can't see these logs anymore, right?

joeyballentine:
True, unless there's a critical error. The main problem is it logs for every node that fails. Maybe I just add the failing dep to a set and print that at the end

RunDevelopment:
> unless there's a critical error

We define what a critical error is, and I would argue that a dependency failing to import is pretty critical. If the dependency is optional, then it's not critical, of course, but non-optional deps can also fail to import.

> The main problem is it logs for every node that fails. Maybe I just add the failing dep to a set and print that at the end.

Instead, could we log import error (immediately) unless we've seen already? Delaying errors until the end might make it harder to understand the logs.

joeyballentine:
Now, if a critical import error happens for a category that doesn't have an install hint, it will look like this on the frontend:
image

meanwhile optional ones still look like they did before:
image

As for logging, this is what it looks like now:
image

RunDevelopment (approved):

PR #1142, opened 2022-10-25
Presets & "Experimental Features" toggle

joeyballentine:
The implementation for this is working, but I want to get some feedback. Mainly, I'd like to collect some ideas on what presets would be good to include.

The UI for the presets tab is pretty basic at the moment, but function over form, for now.

It is now hidden behind a toggle, so we can merge and deal with the actual presets later


src/common/common-types.ts
```diff
@@ -256,3 +256,5 @@ export interface Category {
     installHint?: string;
     excludedFromCheck: string[];
 }
+
+export type SetState<T> = React.Dispatch<React.SetStateAction<T>>;
```
RunDevelopment, line 260: Common shouldn't include React-specific types IMO

RunDevelopment (commented):
One comment, otherwise gtg

RunDevelopment (approved):

PR #1144, opened 2022-10-25
Fixed image iterator sort

theflyingzamboni:
Python by default does an ASCII-like sort, so upper and lowercase letters get separated. Added .lower() to the key method so that case would be ignored when sorting alphanumerically.

### DIFF
--- a/backend/src/nodes/nodes/image/image_file_iterator.py
+++ b/backend/src/nodes/nodes/image/image_file_iterator.py
@@ -19,7 +19,7 @@ from ...properties.outputs import (
     NumberOutput,
 )
 from ...utils.image_utils import get_available_image_formats
-from ...utils.utils import numerical_sort
+from ...utils.utils import alphanumeric_sort
 
 IMAGE_ITERATOR_NODE_ID = "chainner:image:file_iterator_load"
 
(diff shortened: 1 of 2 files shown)


joeyballentine (approved):

RunDevelopment:
Uhhmmm, while `.lower` and `.upper` aren't really all that good for ignore-case comparing text, `.upper` should be preferred. E.g. in German we have "ß" which is a lower-case only character and use "SS" for upper case (e.g. "heiß" → "HEISS").

joeyballentine:
I knew it was a bad idea to merge this... Even simple things that seem fine always have some caveat. 

That being said though, we can't possibly make sure we do everything correctly for every language. There's probably a case where .lower would be preferred for another language, and just none of us know it. 

Anyway, what would be the preferred way without using .upper or .lower?

theflyingzamboni:
> Uhhmmm, while `.lower` and `.upper` aren't really all that good for ignore-case comparing text, `.upper` should be preferred. E.g. in German we have "ß" which is a lower-case only character and use "SS" for upper case (e.g. "heiß" → "HEISS").

Why would a language do this? O_o

RunDevelopment:
> simple things

Text is hugely complex... [Here's a one-hour talk](https://www.youtube.com/watch?v=_mZBa3sqTrI) about just a few complexities with text if you're interested. And that's just a surface-level introduction to the wonderful world of Unicode!

> There's probably a case where .lower would be preferred for another language, and just none of us know it.

It's a Unicode thing, so this has nothing to do with python.

> Anyway, what would be the preferred way without using .upper or .lower?

There is a Unicode thing that gives you a case-invariant version of a string, but I forgot what it was called.

PR #1157, opened 2022-10-31
Add "Image Pairs Iterator"

joeyballentine:
Not the prettiest thing in the world, but it works.

image


backend/src/nodes/nodes/image/paired_image_file_iterator.py
```diff
@@ -0,0 +1,151 @@
...
+            directory_b, topdown=True, onerror=walk_error_handler
+        ):
+            await context.progress.suspend()
+
+            dirs.sort(key=alphanumeric_sort)
+            for name in sorted(files, key=alphanumeric_sort):
+                filepath = os.path.join(root, name)
+                _base, ext = os.path.splitext(filepath)
+                if ext.lower() in supported_filetypes:
+                    just_image_files_b.append(filepath)
```
RunDevelopment, line 137: This begs to be extracted into a function.

RunDevelopment (changes requested):
Mostly okay, but please define "pairs". I only figured out that this node performs a zip after reading the implementation. Please clearly describe that this node does not go through all n*m ordered or n*m/2 unordered pairs of all possible combinations.

I also don't like that we zip the two lists by order. This is really brittle and means that any change to the way we sort paths is potentially a breaking change. That being said, we also have to do it by order, so I guess this is a problem we can't do anything about.

joeyballentine:
Idk what else to call it... Pairs usually refers to things bring matched up, like a pair of socks. I don't think it has to be that intuitive tbh, it just has to be there so when people ask how to do this we have a solution

RunDevelopment:
Fair point.

joeyballentine:
Dammit, will fix the linter tomorrow

backend/src/nodes/utils/utils.py
```diff
@@ -316,3 +317,21 @@ def alphanumeric_sort(value: str) -> List[Union[str, int]]:
...
     return parts  # type: ignore
+
+
+def walk_error_handler(exception_instance):
+    logger.warning(
+        f"Exception occurred during walk: {exception_instance} Continuing..."
+    )
+
+
+def walk_sorted(path: str, ext_filter: Union[List[str], None] = None) -> List[str]:
```
RunDevelopment, line 328: `walk_sorted` doesn't really tell you what the method does. Maybe something like `list_all_files_sorted` would be better.

Also, `path` is a very generic name as well. Do we expect this to be a directory or can it be a file as well?

backend/src/nodes/nodes/image/image_file_iterator.py
```diff
@@ -84,23 +84,7 @@ async def run(self, directory: str, context: IteratorContext) -> None:
...
-        ):
-            await context.progress.suspend()
-
-            dirs.sort(key=alphanumeric_sort)
-            for name in sorted(files, key=alphanumeric_sort):
-                filepath = os.path.join(root, name)
-                _base, ext = os.path.splitext(filepath)
-                if ext.lower() in supported_filetypes:
-                    just_image_files.append(filepath)
+        just_image_files: List[str] = walk_sorted(directory, supported_filetypes)
```
RunDevelopment, line 87: We should probably also use this for the Pytorch model iterator as well. It's kinda weird that it's different in that regard.

RunDevelopment (changes requested):

backend/src/nodes/nodes/image/image_file_iterator.py
joeyballentine: True, tbh I forgot that one existed

theflyingzamboni:
> I also don't like that we zip the two lists by order. This is really brittle and means that any change to the way we sort paths is potentially a breaking change. That being said, we also have to do it by order, so I guess this is a problem we can't do anything about.

How would changing sort method be a breaking change? The change would apply to both lists, therefore resorting them the same way. If someone isn't using equivalent filenames/directory tree structure in the two folders, it would only ever work by coincidence anyway.

RunDevelopment:
> How else would you want to sort files other than by an order defined by their filenames?

By grouping. In C#, you can do this: `A.Concat(B).GroupBy(path => keyFn(path)).Where(group => group.Count() == 2)`. This would give you an iterator of "pairs". The nice thing here is that you can get arbitrary pairing using the key function.

Of course, this isn't really possible in chainner in the foreseeable future, and even it was, it'd be harder to use.

> If someone isn't using equivalent filenames/directory tree structure in the two folders, it would only ever work by coincidence anyway.

Isn't this one of the main use cases for this node? Let's ignore directory structures for now and just focus on filenames. On Discord, I've seen people wanting to compare a directory of source images with their upscaled counterparts. The upscaled images all ended in `-upscale` or the model name (e.g. `-4x-Ultrasharp`) IIRC. So this would not be a supported use case?

Also, if it only "works by coincidence", then why do we allow it at all? We are basically asking people to shoot themselves in the foot.

joeyballentine:
> Also, if it only "works by coincidence", then why do we allow it at all? We are basically asking people to shoot themselves in the foot.

This was in response to what you were saying. He meant there's a specific way this will be used where it will be useful, and if you aren't using it that way it's only going to work by coincidence. For the case you describe it will work fine. 

> The upscaled images all ended in -upscale or the model name (e.g. -4x-Ultrasharp) IIRC. So this would not be a supported use case?

If every file ends with the same suffix then it'll get sorted the same either way

joeyballentine:
The whole point of this node is to fulfill a specific want from multiple people. This is exactly how it would work if it was with rewritten iterators as well btw, since that would do implicit zips the exact same way as this

RunDevelopment:
> This is exactly how it would work if it was with rewritten iterators as well btw, since that would do implicit zips the exact same way as this

Good point.

Argh, I still kinda hate it, but whatever. Good enough :)

RunDevelopment (approved):

PR #1162, opened 2022-11-02
Implement handling for converting fp16 ONNX to NCNN

theflyingzamboni:
As discussed on discord. Main features are:
1. PyTorch to NCNN will now always use fp32 when converting to ONNX intermediately. 
2. Converting raw data in tensors to floats now handles fp16.
3. If fp16 is used in ONNX model where not supported in NCNN, data type converted to fp32.
4. Fp mode is now selectable for Convert to ONNX (though fp16 will error if pytorch fp16 mode not on/supported).

### DIFF
--- a/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
+++ b/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
@@ -48,7 +48,7 @@ class ConvertTorchToNCNNNode(NodeBase):
             )
 
         # Set final to False so intermediate conversion to ONNX is always fp32
-        onnx_model = ConvertTorchToONNXNode().run(model, final=False)[0]
+        onnx_model = ConvertTorchToONNXNode().run(model)[0]
         ncnn_model, fp_mode = ConvertOnnxToNcnnNode().run(onnx_model, is_fp16)
 
         return ncnn_model, fp_mode
--- a/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
+++ b/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
@@ -48,7 +48,7 @@ class ConvertTorchToNCNNNode(NodeBase):
             )
 
         # Set final to False so intermediate conversion to ONNX is always fp32
-        onnx_model = ConvertTorchToONNXNode().run(model, False)
+        onnx_model = ConvertTorchToONNXNode().run(model, final=False)
         ncnn_model, fp_mode = ConvertOnnxToNcnnNode().run(onnx_model, is_fp16)
 
         return ncnn_model, fp_mode
--- a/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
+++ b/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
@@ -48,7 +48,7 @@ class ConvertTorchToNCNNNode(NodeBase):
             )
 
         # Set final to False so intermediate conversion to ONNX is always fp32
-        onnx_model = ConvertTorchToONNXNode().run(model, final=False)
+        onnx_model = ConvertTorchToONNXNode().run(model, final=False)[0]
         ncnn_model, fp_mode = ConvertOnnxToNcnnNode().run(onnx_model, is_fp16)
 
         return ncnn_model, fp_mode
--- a/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
+++ b/backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
@@ -12,7 +12,7 @@ from ...utils.torch_types import PyTorchModel
 from .convert_to_onnx import ConvertTorchToONNXNode
 
 try:
-    from ..onnx.convert_to_ncnn import ConvertOnnxToNcnnNode
+    from ..onnx.convert_to_ncnn import ConvertOnnxToNcnnNode, FP_MODE_32
 except:
     ConvertOnnxToNcnnNode = None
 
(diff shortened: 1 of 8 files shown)


backend/src/nodes/utils/ncnn_param_schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "AbsVal": {
-    "weightOrder": []
+    "weightOrder": {}
```
RunDevelopment, line 3: Out of curiosity, where do the changes to this schema come from?

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -19,17 +20,37 @@
...
+        ]
 
         self.category = PyTorchCategory
         self.name = "Convert To ONNX"
         self.icon = "ONNX"
         self.sub = "Utility"
 
-    def run(self, model: PyTorchModel) -> OnnxModel:
+    def run(
+        self, model: PyTorchModel, is_fp16: int = 0, final: bool = True
```
RunDevelopment, line 46: > `final: bool = True`

No. The parameters of the `run` method as supposed to mirror the inputs of the node (except in the case of some iterator helper nodes). 

I also don't get why this parameter is even necessary. Couldn't you just us the existing `is_fp16`?

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -64,4 +85,6 @@ def run(self, model: PyTorchModel) -> OnnxModel:
             f.seek(0)
             onnx_model_bytes = f.read()
 
-        return OnnxModel(onnx_model_bytes)
+        fp_mode = "fp16" if fp16 else "fp32"
```
RunDevelopment, line 88: Why does `fp16` and not `should_use_fp16` determine `fp_mode`? Right now, the model could be fp32 with `fp_mode == "fp16"`.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -19,17 +20,37 @@
...
+            OnnxFpDropdown(),
+        ]
+        self.outputs = [
+            OnnxModelOutput(label="ONNX Model"),
+            TextOutput(
+                "FP Mode",
+                """match Input1 {
+                    FpMode::fp32 => "fp32",
+                    FpMode::fp16 => "fp16",
+                }""",
```
RunDevelopment, line 36: It feels like we should have a Navi function in chainner scope that does this string conversion. Right now, we already have 2 copies of this and I don't want to add a third.

Just something like this:
```
def FpMode::toString(mode: FpMode) {
  match mode {
    FpMode::fp32 => "fp32",
    FpMode::fp16 => "fp16",
  }
}

// call
FpMode::toString(Input1)
```

RunDevelopment (changes requested):
As always, I'll assume that the actual NCNN/ONNX code is correct. I have no idea what they do anyway.

backend/src/nodes/utils/ncnn_param_schema.json
theflyingzamboni: What do you mean, where do they come from? I just needed weightOrder to be a dict instead of a list so I could associate the weights with their permitted data types. This schema is just something we made based off the NCNN operators documentation.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
theflyingzamboni: I don't think it is now. I put it there before I decided to add the fp32/fp16 dropdown, so I needed a way to indicate that it should not just follow fp16 mode in settings.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
theflyingzamboni: I don't think it could because the assert statement up top would throw an error, but you are correct it should be `should_use_fp16` anyway

src/common/types/chainner-scope.ts
```diff
@@ -102,6 +102,12 @@ def BorderType::getOutputChannels(type: BorderType, channels: uint) {
         _ => channels
     }
 }
+def FpMode::toString(mode: FpMode) {
```
RunDevelopment, line 105: Could you also use this function in the 2 other places? Just search for uses of `FpMode::fp32` in our code base and you'll find them.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -19,17 +20,29 @@
...
+            TextOutput("FP Mode", "FpMode::toString(Input1)"),
+        ]
 
         self.category = PyTorchCategory
         self.name = "Convert To ONNX"
         self.icon = "ONNX"
         self.sub = "Utility"
 
-    def run(self, model: PyTorchModel) -> OnnxModel:
+    def run(self, model: PyTorchModel, is_fp16: int = 0) -> Tuple[OnnxModel, str]:
```
RunDevelopment, line 39: > `= 0`

Default values don't make sense for input parameters because the input itself determines the default. Please remove this and just pass 0 (or better, a constant, e.g. `FP_MODE_32`) in `convert_to_ncnn.py`.

backend/src/nodes/nodes/pytorch/convert_to_ncnn.py
```diff
@@ -46,7 +46,9 @@ def run(self, model: PyTorchModel, is_fp16: int) -> Any:
                 and therefore requires the ONNX dependency to be installed. Please install ONNX through the dependency \
                 manager to use this node."
             )
-        onnx_model = ConvertTorchToONNXNode().run(model)
+
+        # Set final to False so intermediate conversion to ONNX is always fp32
```
RunDevelopment, line 50: Please update this comment.

RunDevelopment (changes requested):

RunDevelopment (approved):

RunDevelopment (approved):

PR #1163, opened 2022-11-02
Added handle to Write Output Frame dir input

theflyingzamboni:
Output Video Directory input now has handle.


backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -76,7 +76,7 @@ def __init__(self):
         self.description = ""
         self.inputs = [
             ImageInput("Frame"),
-            DirectoryInput("Output Video Directory"),
+            DirectoryInput("Output Video Directory", True),
```
joeyballentine, line 79: Can you use the named param version of this?

joeyballentine (approved):

PR #1170, opened 2022-11-05
Video Iterator: Throw proper errors & use default for frames

joeyballentine:
Apparently videos can not only be encoded with the incorrect number of frames, they also can just not have the frame info at all. So now I have it using an absurdly high number if it doesn't exist

### DIFF
--- a/backend/src/nodes/nodes/image/video_frame_iterator.py
+++ b/backend/src/nodes/nodes/image/video_frame_iterator.py
@@ -210,12 +210,20 @@ class SimpleVideoFrameIteratorNode(IteratorNodeBase):
         if video_stream is None:
             raise Exception("No video stream found in file")
 
-        width = int(video_stream["width"])
-        height = int(video_stream["height"])
-        fps = int(video_stream["r_frame_rate"].split("/")[0]) / int(
-            video_stream["r_frame_rate"].split("/")[1]
-        )
-        frame_count = int(video_stream["nb_frames"])
+        width = video_stream.get("width", None)
+        if width is None:
+            raise Exception("No width found in video stream")
+        width = int(width)
+        height = video_stream.get("height", None)
+        if height is None:
+            raise Exception("No height found in video stream")
+        height = int(height)
+        fps = video_stream.get("r_frame_rate", None)
+        if fps is None:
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -210,12 +210,20 @@ async def run(self, path: str, context: IteratorContext) -> None:
...
+        height = video_stream.get("height", None)
+        if height is None:
+            raise Exception("No height found in video stream")
+        height = int(height)
+        fps = video_stream.get("r_frame_rate", None)
+        if fps is None:
+            raise Exception("No fps found in video stream")
+        fps = int(fps.split("/")[0]) / int(fps.split("/")[1])
+        frame_count = video_stream.get("nb_frames", 999999999)
+        frame_count = int(frame_count)
```
RunDevelopment, line 226: Is there some way to know the length of the video (e.g. in seconds)? If so, then we could calculate the number of frames by doing `length * fps`.

joeyballentine: I'll check, that's a good idea

joeyballentine: We do get a `duration`, i believe its in seconds

joeyballentine:
Just in case we ever need to come back to it: If we find a case where `duration` isn't in the info, but `duration_ts` is, we can multiply the `duration_ts` by the evaluated `timecode` (split by `/`, divide) to get the actual duration. 

Also, if we end up finding cases where width and height aren't there, there's two other fields called coded_width and coded_height we could maybe use as a backup (if those are present)

ALSO: If `r_frame_rate` isn't available, we could theoretically use `avg_frame_rate`, it would just end up potentially being inaccurate.

But, I won't worry about these things until we find cases in the wild.

RunDevelopment (approved):

RunDevelopment:
One thing I would note regarding the code is that we might want to move the code for individual aspects (width, height, fps, frame count) into separate functions. Right now, it's all just there together. It's not that bad right now, but we should do that when implementing your ideas.

PR #1171, opened 2022-11-05
Fix Swin2SR loading for specific models

joeyballentine:


backend/src/nodes/nodes/pytorch/upscale_image.py
```diff
@@ -6,8 +6,6 @@
 import numpy as np
 from sanic.log import logger
 
-from backend.src.nodes.properties.inputs.pytorch_inputs import SrModelInput
```
RunDevelopment, line 9: This feels like something a linter should have been able to catch. Is there nothing like a "no duplicate imports" rule?

RunDevelopment (approved):

backend/src/nodes/nodes/pytorch/upscale_image.py
RunDevelopment: Just looked through Pylint rules, and they don't have it.

joeyballentine: Possibly, we should definitely set that up, or at least have some check to make sure we aren't importing things this way. Upscale Image was unable to be imported cuz of it

PR #1172, opened 2022-11-05
Fix HAT not being included as a valid SR model type

joeyballentine:
This bug was caused by merging the torch type PR


RunDevelopment (approved):

RunDevelopment:
Kinda annoying that this could be a problem. Maybe we could find a way to derive the `PytorchSrModel` type from the types tuple? This stuff is possible in TS, so maybe it's also possible in Python.

joeyballentine:
That would be awesome. I tried doing that before and it kept complaining, but maybe there's a way.

PR #1177, opened 2022-11-06
Use Navi's `literal` for simpler type creation

RunDevelopment:


joeyballentine:
Can you do this for the pytorch one as well? This actually is great because it'll be really easy to make a more dynamic "tag output" that those both can use

RunDevelopment:
All other outputs already use `literal`.

joeyballentine:
Ah, i didn't realize.

joeyballentine (approved):

PR #1181, opened 2022-11-06
Normal Map Generator

### LINKED ISSUE #1160: How about some Sobel Normal Map generator node?
I would see it useful, especially with being able to blend few of those with different radius. Even better if limited color range could be selected as input for different nodes.

RunDevelopment:
Resolves #1160.

This PR adds a node for generating normal maps from images using Sobel or Scharr.


joeyballentine:
I'm unable to connect RGBA images to the handle, yet it doesn't give me a type restriction tag of RGB/Gray -- I thought something was broken for a sec

RunDevelopment:
> I thought something was broken for a sec

Yeah, we don't communicate that some nodes allow grayscale or RGB but not RGBA images right now.

joeyballentine:
I feel like for things on this we might wanna allow RGBA and just auto convert. In fact, i feel like we should do that for most nodes. People using iterators (where we can't type check) and having things error when one random image is RGBA instead of RGB probably isn't the best UX

RunDevelopment:
The problem is: How do you convert from RGBA to RGB? A lot of image formats set transparent pixel to transparent black, so you can't assume that transparent pixels have any useful information.

joeyballentine:
It's probably best to determine that based on the node. Like in this case, it would probably be whatever impacts the normal map the least (so it's just flat there)

RunDevelopment:
That's not really possible. The node includes a blur, so "whatever impacts the normal map the least" is not a simple local property we can decide by only looking at the pixel itself. That's why I don't want to deal with this complexity.

Since this issue with RGBA doesn't affect this node alone, I say we make an issue for this and proceed without it for this PR.

joeyballentine:
@RunDevelopment If someone has an image with alpha they want to make a normal map for, they'll have to do something with it. Most likely they'll just pop off the alpha layer and only use the RGB data, which like you said could be noisy. Since it's pretty difficult to replace alpha with pure color with chaiNNer at the moment, I feel like we should probably mask the alpha out of the normal map so the normal data is all flat in the alpha areas

RunDevelopment:
> I feel like we should probably mask the alpha out of the normal map so the normal data is all flat in the alpha areas

And how would you do that? We can't just say `final = mix(normal, flat, alpha)`, because (1) the node includes blurring and sharpening and (2) the Sobel kernel is 3x3. Both (1) and (2) mean that this simple approach will cause artifacts around transparent edges.

joeyballentine:
Good point

joeyballentine (approved):

PR #1182, opened 2022-11-06
Fix split transparency with gray images

joeyballentine:
This bug was caused when we made all gray images 3 channel, since the gray fix here was expecting two channels

### DIFF
--- a/backend/src/nodes/nodes/image_channel/transparency_split.py
+++ b/backend/src/nodes/nodes/image_channel/transparency_split.py
@@ -10,6 +10,7 @@ from ...node_factory import NodeFactory
 from ...properties.inputs import ImageInput
 from ...properties.outputs import ImageOutput
 from ...properties import expression
+from ...utils.image_utils import as_target_channels
 
 
 @NodeFactory.register("chainner:image:split_transparency")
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image_channel/transparency_split.py
```diff
@@ -39,14 +39,19 @@ def __init__(self):
...
             img = np.tile(np.expand_dims(img, axis=2), (1, 1, min(4, 3)))
         # Pad with solid alpha channel if needed (i.e three channel image)
-        elif img.shape[2] == 3:
+        if img.shape[2] == 3:
             logger.debug("Expanding image channels")
             img = np.dstack((img, np.full(img.shape[:-1], 1.0)))
+        # Convert from gray to RGB then pad with solid alpha channel
+        if img.shape[2] == 1:
+            logger.debug("Expanding image channels")
+            img = np.dstack((img, img, img, np.full(img.shape[:-1], 1.0)))
```
RunDevelopment, line 54: Why not just do `img = as_target_channels(img, 4)`?

joeyballentine: Because I forgot that existed

RunDevelopment (approved):

PR #1185, opened 2022-11-07
Add more available inputs to text append and text pattern

joeyballentine:
image

We're gonna have to do something about the type system lol

image

### DIFF
--- a/backend/src/nodes/utils/utils.py
+++ b/backend/src/nodes/utils/utils.py
@@ -23,6 +23,35 @@ MAX_VALUES_BY_DTYPE = {
 }
 NUMBERS = re.compile(r"(\d+)")
 
+ALPHABET = [
+    "A",
+    "B",
+    "C",
+    "D",
+    "E",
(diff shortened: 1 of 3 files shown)


backend/src/nodes/utils/utils.py
```diff
@@ -23,6 +23,35 @@
...
+    "R",
+    "S",
+    "T",
+    "U",
+    "V",
+    "W",
+    "X",
+    "Y",
+    "Z",
+]
```
RunDevelopment, line 53:
```suggestion
ALPHABET = [*"ABCDEFGHIJKLMNOPQRSTUVWXYZ"]
```

RunDevelopment (commented):
Do we really need 26 inputs? I would have said 10 or 12 are enough...

> We're gonna have to do something about the type system lol

Yeah, but this kind of thing would require tuples/arrays which are hard to implement.

RunDevelopment (approved):

PR #1190, opened 2022-11-08
Exhaustive deps for custom hooks

RunDevelopment:
We have a few custom hooks that also take deps arrays. Unfortunately, those deps arrays cannot be verified by ESLint. To solve this, I changed all custom hooks to either assume memoized callbacks or be of a form that ESLint can verify.

Changes:
- `useAsyncEffect` now takes a function that creates options as its first argument. This makes it so that ESLint can verify the deps array.
- `useIpcRendererListener` now assumes memoized callbacks.
- `useBackendEventSourceListener` now assumes memoized callbacks.
- `useBatchedCallback` now assumes memoized callbacks.
- Removed `useAsyncInterval`. The function wasn't easily convertible so that ESLint could understand it.
- Turned off ESLint's `consistent-return` rule. TS makes the kinds of errors this rule is supposed to prevent impossible.

With this PR, we can be reasonably certain that all deps arrays in our code base are correct.

### DIFF
--- a/src/renderer/contexts/ExecutionContext.tsx
+++ b/src/renderer/contexts/ExecutionContext.tsx
@@ -1,4 +1,4 @@
-import { memo, useEffect, useMemo, useRef, useState } from 'react';
+import { memo, useCallback, useEffect, useRef, useState } from 'react';
 import { useHotkeys } from 'react-hotkeys-hook';
 import { Edge, Node, useReactFlow } from 'reactflow';
 import { createContext, useContext, useContextSelector } from 'use-context-selector';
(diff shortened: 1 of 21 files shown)


src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -227,65 +227,45 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
-        eventSource,
-        'execution-error',
-        (data) => {
-            if (data) {
-                sendAlert(AlertType.ERROR, null, getExecutionErrorMessage(data, schemata));
-                unAnimate();
-                setStatus(ExecutionStatus.READY);
-            }
-        },
-        [setStatus, unAnimate, schemata, {}]
```
joeyballentine, line 255: Are you 100% sure you didn't cause a whole bunch of bugs by getting rid of these {} s?

joeyballentine (approved):

PR #1191, opened 2022-11-08
Better Save Changes popup

### LINKED ISSUE #1100: Quitting with unsaved changes warning popup is inconsistent with all other applications
## Description
The warning dialog about unsaved changes is inconsistent with all other such dialogs in all other applications. Key differences:

1. Hidden negative question ("Do not save?" instead of "Save?")
2. Different button order ("Don't save / Save" instead of "Save / Don't save / Cancel quitting")
3. Different default button ("Don't save" instead of "Save")
4. No Cancel button
5. File name isn't mentioned
6. Unnecessarily long text

Consistency is very important here because clicking the wrong button can lead to unexpected data loss. A lot of people don't read warning messages, especially when they're long, and click out of the decades old habit.

Furthermore, misclicking on the close button doesn't offer an option of canceling quitting. Considering chaiNNer doesn't load fast, restarting can be annoying. This is also important when custom action is needed, for example saving changes into a different file.

The dialog shoud be:

**Do you want to save changes to {chainName}?** *Yes / No / Cancel* (*Save / Don't save / Cancel*, if possible)

### Dialog of chaiNNer

image

**Do not save?** *Yes / No*

### Traditional approach

image

**Save?** *Yes / No / Cancel*

### Modern approach

image

**Save?** *Yes / No / Cancel*

### Advanced approach

image

**Save?** *Yes / No / Cancel*

Note that in all cases, quitting isn't even mentioned. The important part of the question is to save or not.

## Official design guidelines

**Microsoft:** https://learn.microsoft.com/en-us/windows/win32/uxguide/mess-confirm
**Apple:** https://developer.apple.com/design/human-interface-guidelines/components/presentation/alerts/

## Information:
- Chainner version: 0.13.0
- OS: Windows 10

RunDevelopment:
Fixes #1100.

I added a new `saveUnsavedChanges` function will save any unsaved changes. When closing the app with unsaved changes, the main process will now send an IPC message to the renderer to save, after which the renderer closes itself.

I also had to make some changes to alerts to consistently select the "Save" button as the default, while making "Cancel" the least destructive ref.

Alert:
image

System:
image

### DIFF
--- a/src/common/safeIpc.ts
+++ b/src/common/safeIpc.ts
@@ -83,6 +83,8 @@ export interface SendChannels {
     'show-collected-information': SendChannelInfo<[info: Record<string, unknown>]>;
     'disable-menu': SendChannelInfo;
     'enable-menu': SendChannelInfo;
+    'save-before-exit': SendChannelInfo;
+    'exit-after-save': SendChannelInfo;
 
     // history
     'history-undo': SendChannelInfo;
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -707,18 +707,37 @@ const createWindow = lazy(async () => {
     ipcMain.on('update-has-unsaved-changes', (_, value) => {
         hasUnsavedChanges = value;
     });
+    let forceExit = false;
+    ipcMain.on('exit-after-save', () => {
+        forceExit = true;
+        mainWindow.close();
+    });
 
     mainWindow.on('close', (event) => {
+        if (forceExit) {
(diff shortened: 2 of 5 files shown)


joeyballentine (dismissed):
Lgtm, just fix the conflicts

joeyballentine (approved):
Vielen Dank

RunDevelopment:
Gern geschehen!

PR #1215, opened 2022-11-10
General file drag'n'drop

### LINKED ISSUE #1196: Drag and drop model files with automatic creation of a node with the model
If I now try to drag a file (for example a pth) I get an warning:
изображение
And I would like to get a node with a specified model.

RunDevelopment:
Resolves #1196.

This was actually really easy to implement. Basically, every file input can now declare itself as the primary input for its file types. When a file is dragged into chainner, we go through all primary file inputs and create the node of the first matching one.

The NCNN group changed because only one input is set when the user drops a file. So I changed the group to fix that.

I also removed the `TorchFileInput`. It wasn't used.

### DIFF
--- a/backend/src/nodes/nodes/image/load_image.py
+++ b/backend/src/nodes/nodes/image/load_image.py
@@ -22,7 +22,7 @@ class ImReadNode(NodeBase):
     def __init__(self):
         super().__init__()
         self.description = "Load image from specified file."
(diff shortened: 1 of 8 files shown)


joeyballentine (approved):

joeyballentine:
that's a lot more elegant than I thought it would be. awesome 👍

PR #1225, opened 2022-11-12
Update React-flow & (potentially) optimize chain animation

joeyballentine:


src/renderer/components/inputs/InputContainer.tsx
```diff
@@ -123,7 +123,7 @@ export const HandleWrapper = memo(
                             position: 'absolute',
                             top: '50%',
                             left: '50%',
-                            height: '30px',
+                            height: '45px',
```
RunDevelopment, line 126: What's this about?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -370,6 +370,14 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
         [createNode, wrapper, reactFlowInstance, schemata, sendAlert, setEdges, setNodes]
     );
 
+    // TODO: I want to get this to work at some point but for now it needs to not exist
+    // const onEdgeUpdate = useCallback(
```
RunDevelopment, line 374: I don't really like commented-out code, but we can keep it if you expect to integrate it relatively quickly.

src/renderer/components/CustomEdge/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+        stroke-width: 8px;
+    }
+
+    &.running {
+        animation: 'dashdraw-chain 0.5s linear infinite';
+        opacity: 100%;
+    }
+
+    &.dot {
+        stroke: var(--chain-hole-color) !important;
```
RunDevelopment, line 58: Since you are using colors, could you add `@use 'colors.scss';`?

src/renderer/components/CustomEdge/CustomEdge.tsx
```diff
@@ -92,67 +92,42 @@ export const CustomEdge = memo(
                 onMouseOver={() => hoverTimeout()}
             >
                 <path
-                    className="edge-chain-links"
+                    className={useMemo(
```
RunDevelopment, line 95: Why did you memo this? You're just creating a string, it's not expensive. In fact, I even suspect that the overhead of the memo + the function + deps array is more expensive than the string interpolation...

src/renderer/components/CustomEdge/CustomEdge.tsx
```diff
@@ -92,67 +92,42 @@ export const CustomEdge = memo(
...
-                    }}
+                    stroke={currentColor}
                 />
                 <path
-                    className="edge-chain"
+                    className={useMemo(
+                        () =>
+                            `edge-chain ${isHovered ? 'hovered' : ''} ${
+                                showRunning && animateChain ? 'running' : ''
+                            }`,
```
RunDevelopment, line 112: It might be worth to extract the `${isHovered ? 'hovered' : ''} ${showRunning && animateChain ? 'running' : ''}` part into a variable.

RunDevelopment (changes requested):
Just a few minor things.

Also, why did you name the new stylesheet `styles.scss`? Very generic. I thought the common standard was to have folder structures like this:
```
dir/
  ComponentFoo.tsx
  ComponentFoo.scss
  ComponentWithoutStyles.tsx
```

src/renderer/components/ReactFlowBox.tsx
joeyballentine: I just don't wanna have to dig through the RF discord again to find this snippet

src/renderer/components/inputs/InputContainer.tsx
joeyballentine: I had been working on different change and forgot to revert this

src/renderer/components/CustomEdge/styles.scss
joeyballentine: Is that needed? This works as is

src/renderer/components/CustomEdge/CustomEdge.tsx
joeyballentine: It at least appeared to improve performance... Can't really explain it but dragging nodes around was laggy before i did this

src/renderer/components/ReactFlowBox.tsx
RunDevelopment: Okay, then let's keep it.

src/renderer/components/CustomEdge/styles.scss
RunDevelopment: Yes, it works because global.scss imports color.scss and global.scss happens to be present. However, this is global reasoning. There's no way to tell that a variable called `--chain-hole-color` exists just by looking at this file. This not only makes it more nebulous where this variable comes from, it will also stop working if we remove the color.scss import in global.scss (e.g. because we moved out most rules into dedicated files and don't need the import anymore).

src/renderer/components/CustomEdge/CustomEdge.tsx
RunDevelopment: That's strange cause it really shouldn't....

src/renderer/components/inputs/InputContainer.tsx
joeyballentine: Actually I didn't forget to revert this, this is actually consistent with the other handle now https://github.com/chaiNNer-org/chaiNNer/blob/12e219c2fa504fe0bda7c7d4926e127ca85b524e/src/renderer/components/outputs/OutputContainer.tsx#L107

src/renderer/components/CustomEdge/CustomEdge.tsx
joeyballentine: it apparently doesnt anymore, so might have been a coincidence

src/renderer/components/inputs/InputContainer.tsx
joeyballentine: gonna resolve this because I think of it as a bugfix

RunDevelopment (approved):

PR #1229, opened 2022-11-13
Add experimental Apple MPS support

### LINKED ISSUE #1218: Mac M1 arm MPS support
**Motivation**
Currently generating on Mac M1 is very slow due to using only CPU. However the pytorch-nightly supports running on MPS (aka Mac M1 "GPU") which, while used for example in Stable Diffusion is very fast. I have not found any info that this is currently being considered or done here. It could theoretically be as simple asi just switching the device in pytorch to the descriptor 'mps.' Or there might be some slight adjustments needed (some remaining methods in pytorch dont support mps so you need to convert data to cpu and back - but its minority of those and its possible they are not used in Chainner).

In the current state the Chainner is not usable for upscaling on the newest MacBook Pro M1 with 32 GB ram (upscaling simple image takes like 10 minutes) which is sad, so it should be added.

joeyballentine:
I don't know if this works, but it should. We'll need to get some apple folks to test the nightly

Resolves #1218

### DIFF
--- a/backend/src/nodes/nodes/pytorch/convert_to_onnx.py
+++ b/backend/src/nodes/nodes/pytorch/convert_to_onnx.py
@@ -45,14 +45,14 @@ class ConvertTorchToONNXNode(NodeBase):
             ), "PyTorch fp16 mode must be supported and turned on in settings to convert model as fp16."
 
         model = model.eval()
-        model = model.to(torch.device(exec_options.device))
+        model = model.to(torch.device(exec_options.full_device))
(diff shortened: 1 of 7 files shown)


RunDevelopment (approved):

sascha1337:
MPS fam here, testing nightly now

joeyballentine:
Hi @sascha1337. Keep in mind if you are testing the nightly you will need to use system python and install the latest PyTorch version, as chainner installs an older version that doesn't support MPS

PR #1230, opened 2022-11-13
Add installing gif for windows

joeyballentine:
image


package.json
```diff
@@ -222,4 +223,4 @@
     "uuid": "^8.3.2",
     "yargs": "^17.5.1"
   }
-}
+}
```
RunDevelopment, line 226: Reeeeeeeeeee

joeyballentine: maybe we should just add this file to a .eslintignore or something

RunDevelopment: ESLint doesn't do anything for JSON files, at least, we haven't configured it to lint JSON files and it won't do so by default. Your IDE is just configured to remove trailing new lines without leaving the last new line intact.

PR #1231, opened 2022-11-14
Connect node on node-on-edge collision detect

joeyballentine:
https://i.gyazo.com/1678e412e348a9c90caebaee05b392e5.mp4


0x4E69676874466F78:
Interesting solution. But it can cause problems.
You should then add the ability to easily rip out such a node (for example via SHIFT/ALT/CTRL) with auto-connection of neighbors.

joeyballentine:
@0x4E69676874466F78 we already discussed this in our private development chat, and the plan is to make it so this only happens when you hold a key down. Also i plan on having it visually show you that it will do this, therefore making it more obvious what's going on. And I'm working on getting more accurate (but still optimized) collision detection that better follows the curve of the edge

package.json
```diff
@@ -196,14 +197,17 @@
...
     "@fontsource/open-sans": "^4.5.10",
     "@react-nano/use-event-source": "^0.12.0",
+    "bezier-js": "^6.1.0",
     "cross-fetch": "^3.1.5",
     "decompress": "^4.2.1",
     "electron-log": "^4.4.8",
     "electron-squirrel-startup": "^1.0.0",
     "fast-deep-equal": "^3.1.3",
     "framer-motion": "^6.3.6",
+    "install": "^0.13.0",
```
joeyballentine, line 207: This is a reminder to myself to uninstall these, i must have done "npm install npm install" lol

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -297,25 +326,35 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
-                ) {
-                    return false;
+                    ];
+                    const curve = new Bezier(coords);
+                    const curveIntersectsLeft =
+                        curve.lineIntersects({
+                            p1: nodeBounds.TL,
+                            p2: nodeBounds.BL,
+                        }).length > 0;
+                    const curveIntersectsRight =
```
joeyballentine, line 336: Another note to myself: i can just do early returns here to avoid doing unnecessary calculations. There's probably also a way i can check which side is closest to the curve and check that first.

joeyballentine: Another idea: we can store the center point of the curve (from the edge data) and check if the center is within the node bounds. It would definitely mean collision and would be much faster to check.

RunDevelopment:
1. Multiple edges are highlighted:
    image
1. When dragging and dropping the Spritesheet iterator from the node selector, I got this: (doesn't happen on main)
    image
    image
1. If I hover over an edge and then let go of ALT, the node retains its outline color.
    image
1. Visual glitch around handles.
    image

RunDevelopment:
5. Something constantly updates and consumes CPU even I don't interact with Chainner and even when it's minimized. (doesn't happen on main)
   image
   image

joeyballentine:
The only remaining issue is, I caused an overlap issue with edges & handles by fixing that gap. I'm not sure what the best way to deal with that is
image

joeyballentine:
Ok, fixed that issue. It was caused by how we were ordering zIndexes when selected, so I had to add a few extra checks and modifications to that method.

joeyballentine:
There's still a slight issue with this though, which is that now if you select an edge that is connected to a node that is covering the edge, the edge wont go in front of it. But I think that's such an edge case it probably doesn't matter.

theflyingzamboni:
Could we maybe change the collision detection area to the input handle with a buffer around it? Right now, the collision detection is kind of funky when the node intersects multiple valid edges, and tiny movements in mouse position will cause it to switch between them, and not always in easily predictable ways. A more localized target area would reduce the chances of this.

Also, this would resolve issues with nodes that have multiple image inputs, like Blend Images or Average Color Fix. Currently, it always snaps to Base Layer, but putting the targeting on the input handle would allow either. Apparently none of the channel ones work with this snapping, I assume that's by intention?

joeyballentine:
> Could we maybe change the collision detection area to the input handle with a buffer around it? 

How would you differentiate between which handle you want it to connect to? handles are really small and really close top each other, so this just seems like a worse version of the problem you speak of.

> Also, this would resolve issues with nodes that have multiple image inputs, like Blend Images or Average Color Fix. Currently, it always snaps to Base Layer, but putting the targeting on the input handle would allow either. Apparently none of the channel ones work with this snapping, I assume that's by intention?

I have no idea what you mean by this

theflyingzamboni:
> How would you differentiate between which handle you want it to connect to? handles are really small and really close to each other, so this just seems like a worse version of the problem you speak of.

The smaller area is why it's better. If the target area is the orphan node's input handle, it reduces the chances that the target area will be overlapping multiple edges at the same time, so there are fewer cases where it has to decide between two edges.

https://user-images.githubusercontent.com/55669985/203388444-b9600376-2f5f-4808-b92a-5676cbda8c51.mp4

For the second part, I mean that if you for example drag an orphan Blend Images node onto an edge, it will automatically connect to the Base Layer input. If the orphan node target area were the input handle, it would be possible to choose which input the edge snaps to.

If you meant the second part, nodes in the Channels category do not snap at all. I was assuming that's intentional.

RunDevelopment:
I found a bug. In this chain, I can't insert a H&S node in the marked spot.

image

src/common/common-types.ts
```diff
@@ -209,9 +211,19 @@ export interface NodeData {
...
     readonly complete?: boolean;
+    sourceX?: number;
+    sourceY?: number;
+    targetX?: number;
+    targetY?: number;
+    type?: NonNeverType;
+    edgePath?: string;
+    edgeCenterX?: number;
+    edgeCenterY?: number;
+    colliding?: boolean;
```
RunDevelopment, line 226: I am against this change for node and edge data. I can already see a Ctrl+Z completely breaking this.

- `NodeData.collidingEdge` is just like `hoveredNode` and can be a simple `useState` in global. Same for `EdgeData.colliding`.
- `type` can be derived from `typeState` and doesn't need to be stored.
- `{source,target}{X,Y}` are already props of `Edge<T>`.
- `edgeCenter{X,Y}` and `edgePath` can be calculated on the fly and don't need to be stored.

joeyballentine: What about Edith edgePath? We'd need to recalculate that

joeyballentine: Also take a look at this: https://discord.com/channels/771389069270712320/859774873500778517/1041561117471805490

joeyballentine: > I can already see a Ctrl+Z completely breaking this

When you Ctrl+Z and an edge changes it would update this

RunDevelopment: > What about edgePath? We'd need to recalculate that

Yeah, just recalculate and cache it.

joeyballentine: If we're gonna recalculate it then I'm going to just copy how it gets each number from the RF code so we don't have to parse it from the string it gives. Could lead to desync in the future but should be a bit more efficient i guess

RunDevelopment (commented):
Performance is bad. The cause is React (just look at a perf profile, you won't see any of our code in there), so we'll have to do something to fix it.

I'll also make a PR with suggestions.

src/renderer/components/outputs/OutputContainer.tsx
```diff
@@ -95,6 +95,7 @@ export const OutputContainer = memo(
                 <HStack h="full">
                     {children}
                     <Center
+                        id={stringifySourceHandle({ nodeId: id, outputId })}
```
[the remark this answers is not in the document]
joeyballentine: Ah, thats left over from when i was trying to get the handle positions by id lol

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +147,69 @@ const updateZIndexes = (
         }
     }
 
+    const nodesToAdjust: { node: Node<NodeData>; zIndex: number }[] = [];
```
RunDevelopment, line 150: Could you please explain all this zIndex logic? I don't get what this is supposed to do.

RunDevelopment (commented):
Almost done now, I think.

I have another question from the type side of things. Right now, we are using the `getFirstPossible{Input,Output}` methods to find inputs/outputs to connect to. However, those methods were intended to be used on nodes with no inputs/outputs whatsoever. It is possible to create invalid connections with node collisions because of this.

The easiest example is probably Change Colorspace:
image

Inserting the Change Colorspace node will obviously result in an invalid change, but `getFirstPossible{Input,Output}` will allow this. Right now, collision allow any connection where you could *possibly* configure the inserted node to produce a valid chain, but it doesn't consider how the node is currently configured.

So my question is: Is this intended behavior?

RunDevelopment:
It is not possible to insert and Add Normals node like this via collisions if the input image is an RGBA image.

image

joeyballentine:
image

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -216,8 +286,200 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
+                    const edgeType = typeState.functions.get(e.source)?.outputs.get(outputId);
+                    if (!edgeType) {
+                        return EMPTY_ARRAY;
+                    }
+                    const { inputId } = parseTargetHandle(e.targetHandle);
+                    const targetNode = reactFlowInstance.getNode(e.target);
+                    if (!targetNode) {
+                        return EMPTY_ARRAY;
+                    }
+                    const targetEdgeType = functionDefinitions
```
RunDevelopment, line 346: Why didn't you use typeState?

joeyballentine: I don't want to check based on the current type of the input, I want to check what the type is without any connections. Otherwise, we'd have the same problem as in the screenshot you sent (we talked about this)

Unless there's something in the typestate that can do this that I'm unaware of?

RunDevelopment: TypeState functions have a reference to the type definition, so you don't need to call `reactFlowInstance.getNode`. You can do `typeState.functions.get(e.target)?.definition.inputDefaults.get(inputId)`.

joeyballentine: Oh nice, I'll do that then

joeyballentine:
@RunDevelopment i think this is good enough now. The only problem left as far as I know is the lag, which can't really be fixed

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +148,69 @@ const updateZIndexes = (
...
 
+    const nodesToAdjust: { node: Node<NodeData>; zIndex: number }[] = [];
     // set the zIndex of all edges
     for (const e of edges) {
         let zIndex = Math.max(
             byId.get(e.source)?.zIndex ?? STARTING_Z_INDEX,
             byId.get(e.target)?.zIndex ?? STARTING_Z_INDEX
         );
 
+        const connectedNodes = nodes.filter((n) => n.id === e.source || n.id === e.target);
```
RunDevelopment, line 159: When the user moves nodes, this code runs every frame, so it needs to be fast. This is why I made sure the whole zIndex code runs in O(n) and is as fast as I could make it.

While I don't want you to benchmark everything (not that it would even be that useful), please write your code so that it's still O(n) (here n being the number of nodes + edges). 

`byId.get` could be used in this specific line of code to reduce the time complexity from O(n) to O(1), so please use it like the code above already does.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +148,69 @@ const updateZIndexes = (
...
+
         if (e.selected && zIndex < MIN_SELECTED_INDEX) {
             zIndex += SELECTED_ADD;
+            // If an edge is selected, make the connected nodes the same zIndex
+            connectedNodes.forEach((n) => {
+                if (n.type !== 'iterator') {
+                    // We need to wait until after the loop to adjust the zIndex
+                    nodesToAdjust.push({ node: n, zIndex });
```
RunDevelopment, line 167: I cannot see this being correct. Nodes can have multiple edges connected to them, so nodes might end up multiple times in the `nodesToAdjust` array with different zIndex values each.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +148,69 @@ const updateZIndexes = (
...
+                }
+            });
+        } else if (connectedNodes.some((n) => n.selected)) {
+            // If an edge is connectec to a selected node, we need to make the other node it is connected to the same zIndex
+            connectedNodes.forEach((n) => {
```
RunDevelopment, line 172: I don't understand this and why you only update the zIndexes of nodes that are selected or that are connected to a selected edge. Why do you not update the zIndexes of both nodes connected to this edge?

src/renderer/components/ReactFlowBox.tsx
joeyballentine: This code only runs on node drag stop, doesnt it? That means once you stop dragging, not every frame.

src/renderer/components/ReactFlowBox.tsx
joeyballentine: The entirety of this code was wack-a-mole esque "solve one problem and three more appear" so everything i did here was because changing the z index of one thing caused it to be over something else that needed to be raised, etc. If you were to try this yourself i bet you'd end up with the same kind of logic for raising up what is connected.

src/renderer/components/ReactFlowBox.tsx
RunDevelopment: No, on move.

joeyballentine: Ah, yup

joeyballentine:
The last bit with the iterators for updating the child nodes doesn't work right. Will continue tomorrow.

joeyballentine:
Ok I think I finally got everything @RunDevelopment

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -87,6 +103,11 @@ const updateZIndexes = (
             if (n.selected && !selectedIterators.has(n.parentNode)) {
                 relatedIterators.add(n.parentNode);
             }
+            // We need this for later
+            iteratorChildNodesMap.set(n.parentNode, [
+                ...(iteratorChildNodesMap.get(n.parentNode) ?? []),
+                n.id,
+            ]);
```
RunDevelopment, line 110: Remember the total O(n) time complexity? For a chain with one iterator and n child nodes, this is O(n^2).
```suggestion
            let childNodes = iteratorChildNodesMap.get(n.parentNode);
            if (childNodes === undefine) {
                iteratorChildNodesMap.set(n.parentNode, childNodes = []);
            }
            childNodes.push(n.id);
```

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +156,119 @@ const updateZIndexes = (
...
+                    nodesToAdjust.set(targetNode.id, zIndex);
+                }
+            }
+            // Save this for later
+            const edgeParent = sourceNode.parentNode ?? targetNode.parentNode;
+            if (edgeParent) {
+                iteratorChildEdgesMap.set(edgeParent, [
+                    ...(iteratorChildEdgesMap.get(edgeParent) ?? []),
+                    e.id,
+                ]);
```
RunDevelopment, line 200: Same here. Please push onto the existing list. You might want to make a little helper function `mapPush<K, T>(map: Map<K, T[]>, key: K, item: T): void`.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +156,119 @@ const updateZIndexes = (
...
 
+    const nodesToAdjust = new Map<string, number>();
     // set the zIndex of all edges
     for (const e of edges) {
+        edgesById.set(e.id, e);
         let zIndex = Math.max(
-            byId.get(e.source)?.zIndex ?? STARTING_Z_INDEX,
-            byId.get(e.target)?.zIndex ?? STARTING_Z_INDEX
+            nodesById.get(e.source)?.zIndex ?? STARTING_Z_INDEX,
+            nodesById.get(e.target)?.zIndex ?? STARTING_Z_INDEX
```
RunDevelopment, line 165: Could you please use the below variables here too?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +156,119 @@ const updateZIndexes = (
...
-            zIndex += SELECTED_ADD;
+        const sourceNode = nodesById.get(e.source);
+        const targetNode = nodesById.get(e.target);
+
+        if (sourceNode && targetNode) {
+            if (e.selected && zIndex < MIN_SELECTED_INDEX) {
+                zIndex += SELECTED_ADD;
+                // If an edge is selected, make the connected nodes the same zIndex so that the handle stays above
+                if (sourceNode.type !== 'iterator') {
+                    // We need to wait until after the loop to adjust the zIndex
```
RunDevelopment, line 176: Please don't just copy this comment everywhere. Maybe have one comment near `nodesToAdjust` explain what it does.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -135,19 +156,119 @@ const updateZIndexes = (
...
 
         e.zIndex = zIndex - 1;
     }
+
+    // Adjust the zIndex of the nodes that were connected to selected edges
+    // The reason we don't do this in the loop above is because we don't want to modify as we detect)
+    for (const [nodeId, zIndex] of nodesToAdjust.entries()) {
+        const node = nodesById.get(nodeId);
+        if (node) {
+            node.zIndex = zIndex;
```
RunDevelopment, line 212: I still don't see this being correct. What if one node is included with multiple different zIndexes in `nodesToAdjust`?

RunDevelopment (changes requested):
I still don't like the new zIndex stuff. It's complex and the fact that you essentially derived the algorithm through trail and error doesn't give me much confidence in its correctness. 

So what I would like you to do is to write down *why* and *how* you do this. The `updateZIndexes` has a long comment explaining how zIndexes are derived. Please adjust this documentation.

RunDevelopment (approved):

PR #1232, opened 2022-11-15
Move update check to after main window opens

joeyballentine:


src/main/main.ts
```diff
@@ -82,33 +82,39 @@ process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true';
...
-            });
-            if (buttonResult.response === 0) {
-                await shell.openExternal(latest.releaseUrl);
-                app.exit();
-            }
-        })
-        .catch((reason) => log.error(reason));
-}
+const checkForUpdate = () => {
+    if (app.isPackaged && checkUpdateOnStartup) {
```
RunDevelopment, line 86: The name `checkForUpdate` does not imply that it only checks on start up. Could you move the `checkUpdateOnStartup` check to where the function is called?

RunDevelopment (approved):

PR #1242, opened 2022-11-19
Fix video formats list

### LINKED ISSUE #1239: Add MKV Support to the Video Frame Iterator node (MKV Input, Video Length error)
**Description**
MKV Input is not listed as a supported input format, but there's also issues with video length. 

Here's an example of the lack of MKV Input:
image

See this message regarding the error: https://discord.com/channels/930865462852591648/930875396277280788/1043129615645954088

When forcing MKV as an input file, it generates the error above.

joeyballentine:
Closes #1239

Note that the second part of that issue is already fixed (but not released yet)


RunDevelopment (approved):
I'm just gonna assume that this makes sense. I don't know much about different video formats.

PR #1243, opened 2022-11-19
Viewport export

Sryvkver:
Added the functionality to export the current node viewport. [(disord suggestion)](https://discord.com/channels/930865462852591648/930865780663398451/1042090143911268353)
Currently the main node viewport is used for the export. This means the viewport will change its position and scaling during the export, but will revert back afterwards.
I couldn't get it to work with a second, hidden viewport.

Preview of the export button
Screenshot 2022-11-19 122252

Example export
chaiNNer-Viewport

Export padding settings
Screenshot 2022-11-19 123105

### DIFF
--- a/src/renderer/components/SettingsModal.tsx
+++ b/src/renderer/components/SettingsModal.tsx
@@ -149,7 +149,8 @@ const Dropdown = memo(
 );
 
 const AppearanceSettings = memo(() => {
-    const { useSnapToGrid, useIsDarkMode, useAnimateChain, useViewportExportPadding } = useContext(SettingsContext);
+    const { useSnapToGrid, useIsDarkMode, useAnimateChain, useViewportExportPadding } =
+        useContext(SettingsContext);
 
     const [isDarkMode, setIsDarkMode] = useIsDarkMode;
     const [animateChain, setAnimateChain] = useAnimateChain;
--- a/src/renderer/components/SettingsModal.tsx
+++ b/src/renderer/components/SettingsModal.tsx
@@ -149,10 +149,11 @@ const Dropdown = memo(
 );
 
 const AppearanceSettings = memo(() => {
-    const { useSnapToGrid, useIsDarkMode, useAnimateChain } = useContext(SettingsContext);
+    const { useSnapToGrid, useIsDarkMode, useAnimateChain, useViewportExportPadding } = useContext(SettingsContext);
 
     const [isDarkMode, setIsDarkMode] = useIsDarkMode;
     const [animateChain, setAnimateChain] = useAnimateChain;
+    const [viewportExportPadding, setViewportExportPadding] = useViewportExportPadding;
 
     const [isSnapToGrid, setIsSnapToGrid, snapToGridAmount, setSnapToGridAmount] = useSnapToGrid;
 
@@ -198,6 +199,25 @@ const AppearanceSettings = memo(() => {
                     min={1}
                     value={Number(snapToGridAmount || 1)}
                     onChange={(number: string) => setSnapToGridAmount(Number(number || 1))}
+                >
+                    <NumberInputField />
+                    <NumberInputStepper>
+                        <NumberIncrementStepper />
+                        <NumberDecrementStepper />
+                    </NumberInputStepper>
+                </NumberInput>
+            </SettingsItem>
+
+            <SettingsItem
+                description="The amount of padding for the viewport PNG export."
+                title="Viewport PNG export padding"
+            >
+                <NumberInput
+                    defaultValue={viewportExportPadding || 20}
+                    max={100}
+                    min={0}
+                    value={Number(viewportExportPadding || 0)}
+                    onChange={(number: string) => setViewportExportPadding(Number(number || 0))}
                 >
                     <NumberInputField />
                     <NumberInputStepper>
--- a/src/renderer/components/SettingsModal.tsx
+++ b/src/renderer/components/SettingsModal.tsx
@@ -214,11 +214,16 @@ const AppearanceSettings = memo(() => {
                 title="Viewport PNG export padding"
             >
                 <NumberInput
-                    defaultValue={viewportExportPadding || 20}
                     max={100}
                     min={0}
-                    value={Number(viewportExportPadding || 0)}
-                    onChange={(number: string) => setViewportExportPadding(Number(number || 0))}
+                    value={viewportExportPadding}
+                    onChange={(number: string) => {
+                        const value = Number(number);
+
+                        if (!Number.isNaN(value)) {
+                            setViewportExportPadding(value);
+                        }
+                    }}
                 >
                     <NumberInputField />
                     <NumberInputStepper>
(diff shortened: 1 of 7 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1136,6 +1139,98 @@ export const GlobalProvider = memo(
...
+                        ) {
+                            return false;
+                        }
+
+                        return true;
+                    },
+                })
+                    .then((dataUrl: string) => {
+                        currentReactFlowInstance.setViewport(oldViewport);
+                        downloadImage(dataUrl, 'chaiNNer-Viewport.png');
```
joeyballentine, line 1221: Thanks for this!
What's your opinion on putting the chain name + the current timestamp in the file name?

src/renderer/components/SettingsModal.tsx
```diff
@@ -206,6 +208,25 @@ const AppearanceSettings = memo(() => {
                     </NumberInputStepper>
                 </NumberInput>
             </SettingsItem>
+
+            <SettingsItem
+                description="The amount of padding for the viewport PNG export."
+                title="Viewport PNG export padding"
+            >
+                <NumberInput
+                    defaultValue={viewportExportPadding || 20}
```
RunDevelopment, line 217: According to the type, `viewportExportPadding` is a number, so what does this `|| 20` mean?

Also, why is the default value dependent on the current value? Did you mean for the default to simply be 20?

src/renderer/components/SettingsModal.tsx
```diff
@@ -206,6 +208,25 @@ const AppearanceSettings = memo(() => {
...
+                    max={100}
+                    min={0}
+                    value={Number(viewportExportPadding || 0)}
```
RunDevelopment, line 220: `viewportExportPadding` is a number already.
```suggestion
                    value={viewportExportPadding}
```

src/renderer/components/SettingsModal.tsx
```diff
@@ -206,6 +208,25 @@ const AppearanceSettings = memo(() => {
...
+                    onChange={(number: string) => setViewportExportPadding(Number(number || 0))}
```
RunDevelopment, line 221: This will set the value to NaN if the current input is not a valid number.

(My suggestion still needs some code formatting.)
```suggestion
                    onChange={(number: string) => {
                       const value = Number(number)
                       if (!Number.isNaN(value)) {
                           setViewportExportPadding(value)
                       }
                    }}
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1136,6 +1139,98 @@ export const GlobalProvider = memo(
...
+            );
+
+            currentReactFlowInstance.setViewport({
+                x: paddedBoundingBox.x * -1 * exportZoom,
+                y: paddedBoundingBox.y * -1 * exportZoom,
+                zoom: exportZoom,
+            });
+
+            const currentFlowWrapper = reactFlowWrapper.current;
+            if (!(currentFlowWrapper instanceof HTMLElement)) return;
```
RunDevelopment, line 1194: Could you move this check all the way up to where `if (!reactFlowWrapper.current) return;` is? That would allow us to remove the `if (!reactFlowWrapper.current) return;`.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1136,6 +1139,98 @@ export const GlobalProvider = memo(
 
         const [connectingFrom, setConnectingFrom] = useState<OnConnectStartParams | null>(null);
 
+        const getNodesBoundingBox = useCallback(() => {
```
RunDevelopment, line 1142: This function will return `{x:Infinity, y:Infinity, width:-Infinity, height:-Infinity}` for any empty chain.

Could you please add a check that returns `undefined` if there are no nodes? `exportViewportScreenshot` could then check for `undefined` and return early.

On that note: Maybe we should *also* disable the export button if the chain is empty?

RunDevelopment (changes requested):
Thank you for the PR @Sryvkver!

Very good work overall, I just have a few suggestions and question.

src/renderer/contexts/GlobalNodeState.tsx
Sryvkver: Good idea, will add that

src/renderer/components/SettingsModal.tsx
Sryvkver: Well i actually just copied the setup from the snapToGridAmount settings above. Which uses the following:
`defaultValue={snapToGridAmount || 1}`

But I just tested it without all it, and it works as expected.
Will change that.

src/renderer/contexts/GlobalNodeState.tsx
Sryvkver: Good catch, completely forgot to test that.

Have added the check and disabled the button when there are no nodes.

src/renderer/components/SettingsModal.tsx
joeyballentine: We may want to look into why we even have that for snapToGridAmount... That's probably a holdover from before we used typescript and I was just trying to make sure nothing errored.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1136,6 +1139,115 @@ export const GlobalProvider = memo(
...
+                            return false;
+                        }
+
+                        return true;
+                    },
+                })
+                    .then((dataUrl: string) => {
+                        currentReactFlowInstance.setViewport(oldViewport);
+
+                        const currentNodename = savePath ? parse(savePath).name : 'Untitled';
```
joeyballentine, line 1225: Minor nit but this should probably be called `currentChainName`

Sryvkver: Whoopsie...
Variable names are hard sometimes

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1136,6 +1139,115 @@ export const GlobalProvider = memo(
...
+
+            return {
+                x: minX,
+                y: minY,
+                width: maxX - minX,
+                height: maxY - minY,
+            };
+        }, [getNodes]);
+
+        const downloadImage = (dataUrl: string, fileName: string) => {
```
RunDevelopment, line 1160: Where will the image be downloaded to? The user's download folder?

Sryvkver: Currently it opens the save to dialog with a default location of the users download folder.

joeyballentine: I think that's good

RunDevelopment: Same. Sounds good.

RunDevelopment (approved):

PR #1246, opened 2022-11-20
Fix face upscaling in video iterator

joeyballentine:
Without this, using face upscaling in a video iterator makes the output inverted and overflows. I assume the output was a 0-255 range float image, which we wouldn't catch with the enforced normalization. Anyway, this fixes it.


RunDevelopment:
> which we wouldn't catch with the enforced normalization

Maybe we should add a check to image output that ensure that all images are normalized? This would find those bugs directly at the source.

joeyballentine:
We already have the enforce on the inputs, the issue is just that the initial check we do checks if the image is int and assumesv floats are normalized

RunDevelopment (approved):

PR #1248, opened 2022-11-20
Refactor header component

joeyballentine:
It's always bothered me that this wasn't using a grid. Also, I fixed the fact that you could resize the window to be so small that it would cause the header to overflow off the page.


RunDevelopment (commented):
The code itself looks good.

src/main/main.ts
```diff
@@ -663,8 +663,8 @@ const createWindow = lazy(async () => {
         width: lastWindowSize?.width ?? 1280,
         height: lastWindowSize?.height ?? 720,
         backgroundColor: '#1A202C',
-        minWidth: 720,
-        minHeight: 640,
+        minWidth: 840,
```
RunDevelopment, line 666: The min width was already quite large before. Is this really necessary?

E.g. my laptop has 1920x1080px screen with 125% DPI scaling, so an app window will be 768px (adjusted for DPI scaling) wide when snapped to the left/right side of the screen. With this change, chainner will be annoyingly larger than half.

joeyballentine: I suppose I could lower it back, but keep in mind things start overflowing off the screen when it becomes that small. If we ever add anything more to the header this could start becoming a big issue. I'm not really familiar with ways to resolve that issue. I guess we could hide things that don't matter when it becomes smaller?

RunDevelopment: Yes, basically we need responsive design.

joeyballentine:
Since you said the code looks good, gonna merge this so I can preemptively fix the conflict on #1247

PR #1249, opened 2022-11-20
Added `normalizeAccessKeys: true`

RunDevelopment:
This should fix [the bug reported here](https://discord.com/channels/930865462852591648/930865725135011851/1044022936232800376).


RunDevelopment:
Small rant: Why is feature off by default?! I thought Electron was cross-platform, but no! Here you have to enable cross-platform support manually. Who thought that this was a good idea?

joeyballentine (approved):

joeyballentine:
Yeah that's dumb

PR #1250, opened 2022-11-21
Use restricted unpickler

joeyballentine:
This is something I just copied wholesale from my ESRGAN-Bot code. Basically, all it does is prevent arbitrary code execution from loading pickles (which .pth files are). Pickles are notoriously unsafe and this just restricts it to what torch needs. I completely forgot that I hadn't added this to chaiNNer until now.

### DIFF
--- a/backend/src/nodes/nodes/pytorch/load_model.py
+++ b/backend/src/nodes/nodes/pytorch/load_model.py
@@ -15,6 +15,7 @@ from ...utils.exec_options import get_execution_options
 from ...utils.torch_types import PyTorchModel
 from ...utils.pytorch_model_loading import load_state_dict
 from ...utils.pytorch_utils import to_pytorch_execution_options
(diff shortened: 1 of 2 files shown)


RunDevelopment:
GFPGAN has 3 `torch.load(...)`s in their arch code. Should those also be restricted?

---

> The pickle module is not secure. Only unpickle data you trust. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling.

How the fuck is ***the default*** for loading Torch models to allow arbitrary code execution?!

joeyballentine:
@RunDevelopment i don't think those are used as those would require having the sytelgan2 weights which we do not include. I'm pretty sure those are just leftover from wherever they copied the network from.

RunDevelopment:
Alright, then let's say they're safe for now.

RunDevelopment (approved):

PR #1251, opened 2022-11-21
Add Noise Node

### LINKED ISSUE #1087: Add noise node
> A node to add grain/noise to images would be great

[Discord](https://discord.com/channels/930865462852591648/930865780663398451/1029202962163187732)

joeyballentine:
image
Closes #1087

### DIFF
--- a/backend/src/nodes/nodes/image_filter/add_noise.py
+++ b/backend/src/nodes/nodes/image_filter/add_noise.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from typing import Literal
 
 from math import ceil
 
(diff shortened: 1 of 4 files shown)


RunDevelopment:
This node breaks determinism. Right now, chaiNNer assumes that nodes (except for IO load operations) are deterministic, which it uses to determine whether broadcast data is still fresh. Any form of future caching must also assume determinism.

So yeah, a non-determinist node is a no-go. (I am so not looking forward to SD...)

To fix this, we just need a seeded RNG, which [numpy provides](https://numpy.org/doc/stable/reference/random/generator.html). The problem is: which seed? I can think of 3 ways to choose a seed:
1. Constant seed (e.g. 123). This is simple, but also means that all images of the same shape always get the exact same noise.
2. Image "hash" seed. The idea is to "hash" the image and use that as the seed. I put hash in quotation because we don't actually need to hash the whole image. We can just throw the image shape and some sparsely sampled pixels into a hasher and that should give a good-enough hash.
3. Let the user input a seed. At its simplest, this could be just a number input. To get a better UX, we use a group to (1) initialize it to a random value and (2) provide an easy way to get a new random seed (e.g. in the form of a "Reseed" button).

I prefer option 3 because it's simple and lets the user change the randomness added to their image.

joeyballentine:
I like option 3. That group can be reused for any other non-deterministic node that requires a seed as well. Does that have to be done for this PR though?

joeyballentine:
For now, I've done option 1. I think option 3 can be done in a separate PR, and this is good for now.

RunDevelopment (approved):

PR #1252, opened 2022-11-21
Ctrl+D Duplication

### LINKED ISSUE #1149: Add Ctrl+D for duplicating
Self explanatory

joeyballentine:
Closes #1149

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -126,7 +126,7 @@ interface Global {
     ) => readonly [Readonly<Size> | undefined, (size: Readonly<Size>) => void];
     removeNodeById: (id: string) => void;
     removeEdgeById: (id: string) => void;
-    duplicateNodes: (nodeIds: string[], edgeIds: string[]) => void;
+    duplicateNodes: (nodeIds: string[]) => void;
     toggleNodeLock: (id: string) => void;
     clearNode: (id: string) => void;
     setIteratorSize: (id: string, size: IteratorSize) => void;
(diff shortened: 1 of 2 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -1067,18 +1067,21 @@ export const GlobalProvider = memo(
...
                     return [...setSelected(nodes, false), ...newNodes];
                 });
 
                 changeEdges((edges) => {
                     const newEdge = copyEdges(
                         edges.filter((e) => {
-                            return nodesToCopy.has(e.target) && nodesToCopy.has(e.source);
+                            return (
+                                (nodesToCopy.has(e.target) && nodesToCopy.has(e.source)) ||
+                                edgeIds.includes(e.id)
```
RunDevelopment, line 1083: Question: Users are able to select arbitrary edges. What will this do when the user selects a single edge and nothing else? What will the result of the user copying that be?

More general: How is copying edges supposed to behave?

joeyballentine: When i tested it, nothing happened (as expected) but i suppose it's possible it was still duplicating it. I didn't see any errors in the console though

RunDevelopment: Try this:
1. Create a chain with a lest one edge.
2. Select one edge and nothing else.
3. <kbd>Ctrl+D</kbd>
4. You will see that the edge you previously selected was deselected. Select it again by clicking on it.
5. Press <kbd>DEL</kbd>
6. You will see that the edge will be deselected.

What actually happened is that we duplicated the edge and then removed the duplicate again.

joeyballentine: Ah. I'll fix that then

RunDevelopment (approved):

PR #1253, opened 2022-11-22
Hide categories when empty due to search

joeyballentine:
image


src/renderer/components/NodeSelectorPanel/NodeSelectorPanel.tsx
```diff
@@ -219,6 +219,10 @@ export const NodeSelector = memo(() => {
                                                 ? getSubcategories(categoryNodes)
                                                 : null;
 
+                                            if (!categoryNodes && !categoryIsMissingNodes) {
```
RunDevelopment, line 222: Could you move this check above `const subcategoryMap = ...´? I think the check should be closer to the variables it checks.

RunDevelopment (approved):

PR #1254, opened 2022-11-22
Added High Boost Filter sharpen node

theflyingzamboni:
Sharpening method [mentioned in Discord](https://discord.com/channels/930865462852591648/930865780663398451/1043536769603350599).

Has two different kernel modes.


backend/src/nodes/nodes/image_filter/sharpen.py
```diff
@@ -21,7 +21,7 @@ def __init__(self):
         ]
         self.outputs = [ImageOutput(image_type="Input0")]
         self.category = ImageFilterCategory
-        self.name = "Sharpen"
+        self.name = "Sharpen (Unsharp)"
```
joeyballentine, line 24: Unless it's too big to fit in the width of the node selector thingy, i would use "unsharp mask" for it.

theflyingzamboni: That's how I had it originally, but it ended up leaving a gap to the right of the NumberInput it was so large.

joeyballentine: What about having separate subcategories for blur and sharpen so this can just be called "Unsharp Mask" and the new one can just be called "Sharpen"

backend/src/nodes/nodes/image_filter/high_boost_sharpen.py
```diff
@@ -0,0 +1,43 @@
...
+        self.sub = "Blur/Sharpen"
+
+    def run(
+        self,
+        img: np.ndarray,
+        kernel_type: int,
+    ) -> np.ndarray:
+        if kernel_type == KernelType.STRONG:
+            # 8-neighbor kernel
+            kernel = np.array([[-1, -1, -1], [-1, 9, -1], [-1, -1, -1]])
```
RunDevelopment, line 36: Could we get a sharpen amount here too? The formula for how to generate a kernel from a specific amount can be found [here for NORMAL](https://en.wikipedia.org/wiki/Unsharp_masking#Implementation). The idea is the same for STRONG: just replace the [[0,1,0],[1,1,1],[0,1,0]]/5 kernel in the formula with a [[1,1,1],[1,1,1],[1,1,1]]/9 kernel.

RunDevelopment (changes requested):

backend/src/nodes/nodes/image_filter/sharpen.py
theflyingzamboni: Sounds good, mostly. I'd rather be more specific naming the new one though. Calling it Sharpen when we currently have Unsharp Mask called Sharpen would be confusing.

joeyballentine: Tbh the way you have it now is fine

src/common/types/chainner-scope.ts
```diff
@@ -93,6 +93,7 @@ enum PaddingAlignment { Start, End, Center }
 enum ResizeCondition { Both, Upscale, Downscale }
 enum RotateSizeChange { Crop, Expand }
 enum SideSelection { Width, Height, Shorter, Longer }
+enum KernelType { Normal, Strong }
```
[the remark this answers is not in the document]
theflyingzamboni: Because your type system still confuses me, and it seemed like it fit with the others. 😆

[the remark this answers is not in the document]
RunDevelopment: Then just leave it as is for now.

I guess I'll make it my next task to make something that will automatically generate you the type for a dropdown.

backend/src/nodes/nodes/image_filter/high_boost_sharpen.py
theflyingzamboni: That makes a lot more sense than the example I'd found, which weakened the effect at higher values lol.

joeyballentine (approved):

backend/src/nodes/nodes/image_filter/high_boost_sharpen.py
```diff
@@ -0,0 +1,50 @@
...
+        if amount == 0:
+            return img
+
+        delta = np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]])
+        if kernel_type == KernelType.STRONG:
+            # 8-neighbor kernel
+            kernel = delta - np.array([[1, 1, 1], [1, 1, 1], [1, 1, 1]]) / 9
+        else:
+            # 4-neighbor kernel
+            kernel = np.array([[0, -0.2, 0], [-0.2, 0.8, -0.2], [0, -0.2, 0]])
```
RunDevelopment, line 45: Could we make this consistent with the above kernel?

```suggestion
            kernel = delta - np.array([[0, 1, 0], [1, 1, 1], [0, 1, 0]]) / 5
```

This should make it easier to understand how we got those numbers.

theflyingzamboni: Sure. I was just optimizing by pre-calculating. Just couldn't do it with the other because the decimals are infinite.

joeyballentine: This is another case where it would be nice to have some sort of python compiler that can optimize that stuff at compile time

joeyballentine (approved):

backend/src/nodes/nodes/image_filter/high_boost_sharpen.py
RunDevelopment: "Optimizing"

While it's true that we are wasting cycles calculating constants as runtime, we only save ~100 cycles tops while the actual convolution will take millions of cycles for small images (assuming fast memory allocations and efficient algorithms). It's just not worth the less understandable code. As they say: "premature optimization is the root of all evil."

RunDevelopment (approved):

PR #1259, opened 2022-11-23
NCNN Optimizer

theflyingzamboni:
Translation of the ncnnoptimize tool to work with chaiNNer's NcnnModel class. Should produce identical results to using convertmodel.com with optimize onnx and optimize ncnn both checked. Currently runs during onnx to ncnn conversion (so converted models are automatically optimized) and Save Model (in case someone needs an existing ncnn model optimized).

For testing, it would be helpful if people tried it both with and without onnx optimization first. To test without onnx optimization, comment out these lines in onnx convert_to_ncnn.py, and change opt_model on line 42 to model_proto. This should produce the same results as running the C++ ncnnoptimize exe or convertmodel.com with just optimize ncnn checked. Some of the weights in the bin will differ by one because of rounding differences. This is normal.

This PR also fixes a couple conversion bugs encountered while testing other things, as well as expanding the param schema.


backend/src/nodes/nodes/ncnn/save_model.py
```diff
@@ -34,6 +35,8 @@ def run(self, model: NcnnModelWrapper, directory: str, name: str) -> None:
         full_bin_path = os.path.join(directory, full_bin)
         full_param_path = os.path.join(directory, full_param)
 
+        model.model = NcnnOptimizer(model.model).optimize()
```
RunDevelopment, line 38: 1. Why does the Save Model node do the optimization?
2. `NcnnOptimizer` mutates the model, and mutating cached values isn't aokay.

backend/src/nodes/utils/ncnn_optimizer.py
```diff
@@ -0,0 +1,1209 @@
...
+        self.__eliminate_reshape_after_global_pooling()
+        self.__eliminate_reshape_before_binaryop()
+
+        self.__replace_convolution_with_innerproduct_after_global_pooling()
+        self.__replace_convolution_with_innerproduct_after_innerproduct()
+
+        self.__eliminate_flatten_after_innerproduct()
+        self.__eliminate_orphaned_memorydata()
+
+        return self.model
```
RunDevelopment, line 1209: Not a fan of this API. This makes it seem like you are returning a new optimized model, but that's not what's going on.

backend/src/nodes/utils/onnx_to_ncnn.py
```diff
@@ -3005,46 +3018,46 @@ def convert(self, is_fp16: bool = False, include_mem_data: bool = True):
...
                     layer.add_param(4, -234)
                 else:
                     if pads.size == 1:
-                        layer.add_param(4, pads[0])
+                        layer.add_param(4, int(pads[0]))
                     elif pads.size == 2:
-                        layer.add_param(4, pads[1])
-                        layer.add_param(14, pads[0])
+                        layer.add_param(4, int(pads[1]))
+                        layer.add_param(14, int(pads[0]))
```
RunDevelopment, line 3041: Lots of casting going on here. Why is that?

src/renderer/components/outputs/NcnnModelOutput.tsx
```diff
@@ -77,14 +77,6 @@ export const NcnnModelOutput = memo(
                             maxW={60}
                             spacing={2}
                         >
-                            <WrapItem>
-                                <Tag
-                                    bgColor={tagColor}
-                                    textColor={fontColor}
-                                >
-                                    {current.scale}x
```
RunDevelopment, line 85: Why did you remove this?

RunDevelopment (changes requested):
I found a few minor things.

Also, what is this `ncnnoptimize.cpp` file doing in there?

backend/src/nodes/nodes/ncnn/save_model.py
theflyingzamboni: The description says why I did it. If it's not there, there's no way to optimize an an existing ncnn model with chaiNNer. I'd rather not make a separate node for that (which would be the other possibility), since it should just be part of conversion anyway. There's no reason to use an unoptimized model.

theflyingzamboni: As for 2, do you want it to return a new model object instead?

backend/src/nodes/utils/ncnn_optimizer.py
theflyingzamboni: Same question as before: should it copy the model before modifying it?

backend/src/nodes/utils/onnx_to_ncnn.py
theflyingzamboni: Not having it gave me a bug when I added the checked cast in the optimizer iirc. I think it was that some of the values were np.int64 instead of int, which is now strictly required. Prior to checked_cast it didn't matter, because both types convert to a string the same.

RunDevelopment: > it was that some of the values were np.int64 instead of an int

Argh. Guess we have no choice then.

backend/src/nodes/nodes/ncnn/save_model.py
RunDevelopment: My main problem with optimizing on save is that can cause problems. You just added ~2k lines of code for the optimizer and co. I would expect there to be 1 or 2 bugs in there somewhere. This can cause a situation where a user loads a model, it upscales fine, saves the model, and now the model doesn't work anymore.

Also, if there's no reason to use an unoptimized model, then shouldn't we optimize on load instead? If we optimize on load, we could also do stuff like caching the optimized model on disk, which could help to reduce the performance impact. Also, chainner already "preloads" models to show tags, so Load Model taking a little longer doesn't count towards total execution time.

Also also, how load does the optimization usually take and how much time/memory does it save? Just curious.

backend/src/nodes/utils/ncnn_optimizer.py
RunDevelopment: (Let's talk about cache mutation and API here.)

> should it copy the model before modifying it?

Yes and no. In the case of Save Model, a copy is necessary for correctness, but in ONNX->NCNN the copy is unnecessary since we just created that model.

Maybe we could split the API into `def optimize(self) -> None` and `def optimized(self) -> NcnnModel`. This should make it clearer which method mutates the underlying model and which one doesn't (I hope). Maybe we could even call it `optimized_copy`?

backend/src/nodes/nodes/ncnn/save_model.py
theflyingzamboni: Optimization is a processing time-saver rather than a memory saver. So putting it on load and forcing it to run every single time a model is used is counterproductive. Save at least is something only used occasionally, and not as part of the normal upscaling process.

I do see your point about possible breaking though. It's at least something that can be fixed, rather than the constant time penalty of putting it in load.

As for how long it takes, depends on the model size and the optimization operations required. When I timed what is probably one of the more intensive ones (a lot of weight recalculations), it was about 0.22s

backend/src/nodes/utils/ncnn_optimizer.py
theflyingzamboni: I'm going to hold off on this until we definitely settle on when it will be used. I don't particularly like the idea, but if we do decide to put it in Load, then I'd think mutating the model would be fine.

backend/src/nodes/nodes/ncnn/save_model.py
joeyballentine: What if we added an optimization toggle on the save node? Same with ONNX too

RunDevelopment: > rather than the permanent time penalty of putting it in load.

My whole point is that there is no time penalty for putting it in load. Load Model is executed once when the chain is loaded and its result is cached. We pay for it once before the user is even executing the chain and then never again.

> it was about 0.22s

Then that's not worth caching on disk.

So performing the optimization cost 0.22s at most, how much upscaling time did it save?

theflyingzamboni: > So performing the optimization cost 0.22s at most, how much upscaling time did it save?

Don't know for that model, it wasn't an SR model and can't run in chaiNNer (scale computes as 0, so it errors). Which is actually something that needs to be fixed, because it actually crashes chaiNNer.

theflyingzamboni: @RunDevelopment For ESRGAN/RealESRGAN, time saved on upscaling a single image seems about equal to time spent optimizing on load, so it seems net neutral on a single run. 

I did make a mistake on time calculation though. I remembered that the 0.22s was just the time for performing the fuse_x_batchnorm operation, not the full conversion. Testing in chaiNNer on ESRGAN, it adds between 0.2-0.25s to loading an unoptimized model and 0.1-0.15s to an already optimized one. I'm leaning toward being swayed, though I still dislike adding that time to loading. I'm actually more swayed by thinking how much of a pain it would be to communicate that people should optimize old conversions. Putting it in Load would just do it for the user without them even needing to be aware of it. 

The time cost is not *too* bad now, but if we ever can get larger models like SwinIR working in ncnn, the extra load time could be more of an annoyance. We could always revisit the issue if something like that happens.

RunDevelopment: > if we ever can get larger models like SwinIR working in ncnn, the extra load time could be more of an annoyance.

I honestly think that any increase in load time <1sec is acceptable when part of Load Model. Whether a model takes a second longer to load or not isn't as important when the user needs a while to select the model in the first place (be it via an open file dialog or via dropping a file).

This could become a problem in the future when users can load many models via iterators, but we can think about a solution for this when we get to it.

theflyingzamboni: I was thinking it might be possible that a model the size of SwinIR would take longer than 1s, but given how responsive the NCNN team is, that may never be a concern for us lol

backend/src/nodes/utils/ncnn_optimizer.py
theflyingzamboni: So it should be fine if it always mutates the cached model now, yes? Since the optimized model is now always passed out of Load and Convert?

RunDevelopment: Not really. The API of this function is still confusing.

theflyingzamboni: When I asked if it should always copy before though, you said it was unnecessary because the model had just been created, which is the same for Load Model.

theflyingzamboni: I changed the Optimizer so it now returns None.

RunDevelopment (approved):

PR #1263, opened 2022-11-24
Added torch.LongStorage to safe_list

### LINKED ISSUE #1262: Can't seem to load any SwinIR or HAT models
I can't seem to load any SwinIR or HAT models in the newest version that was just released. But I downgraded to `v0.15.0` and I am able to load the models again.

Tested on Windows 11.

theflyingzamboni:
Fixes #1262

### DIFF
--- a/backend/src/nodes/utils/unpickler.py
+++ b/backend/src/nodes/utils/unpickler.py
@@ -5,7 +5,10 @@ from types import SimpleNamespace
 safe_list = {
     ("collections", "OrderedDict"),
     ("torch._utils", "_rebuild_tensor_v2"),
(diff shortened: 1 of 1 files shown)


theflyingzamboni:
> Should we add more storage types just to future proof it? If there are any others

There are lots. Up to you. I can run down the list and add them if you want.

joeyballentine:
I'd say add half to be safe

theflyingzamboni:
Added Int and BFloat16 as well. These seem like the most likely ones we could encounter.

joeyballentine (approved):

PR #1266, opened 2022-11-24
Added support for reading DDS via texconv

### LINKED ISSUE #141: Many DDS DXT10 formats are not supported
The underlying image library we use does not support many DDS DXT10 formats. 

This includes:
- ATI1 (legacy)
- ATI2 (legacy)
- BC1 (all variants)
- BC7 (all variants)

RunDevelopment:
Fixes #141.

This PR uses [texconv](https://github.com/microsoft/DirectXTex/wiki/Texconv) to read DDS files by first converting them to PNG and then reading that PNG. The PNG will always be deleted after reading it. I tested multiple DDS formats, and it seems to work reliably. In Load Image itself, texconv is only used as a last resort. OpenCV and PIL will be used first to read any DDS files. Texconv will only be used if OpenCV and PIL fail. 

Performance seems to be okay too. I tried loading a large DDS file and a PNG version of it. I was not able to see any noticeable difference in execution time when using a Load Image node.

Texconv itself is included as an executable in the backend folder. I also included its license text.

Errors are okay-ish, Windows text encoding isn't great...
image

---

~~The only remaining question is: In which directory should the temporary PNGs be saved to? 
I picked the texconv directory for now, but that's not a great choice.~~


joeyballentine (approved):
LGTM

iwubcode:
@RunDevelopment  - I noticed this says reading, are there plans to support _writing_ to DDS formats?

RunDevelopment:
There weren't any until now :) #1286

PR #1269, opened 2022-11-24
Refactored `hoveredNode`

RunDevelopment:
Changes:
- `hoveredNode` cannot be null.
- Added a `hoveredNodeRef` for use in `createNode`.
- `createNode` and `createConnection` are now part of the nonvolatile context. This should improve performance.


RunDevelopment:
> I'm pretty sure without it that drag-n-dropped nodes would still go inside of the iterators they're placed on top of

They don't.

> we'd just need to extract out that code and call it from both onNodeDragStop and onDrop

Not sure if that will be easy, but it's also not the point of this PR. That should be its own PR IMO.

joeyballentine:
> Not sure if that will be easy, but it's also not the point of this PR. That should be its own PR IMO.

Understandable.

PR #1278, opened 2022-11-27
NCNN input image dims getter

theflyingzamboni:
Per @RunDevelopment's request in discord. Returns height, width, channels for an NCNN model's input. 0 is default for each and means that there is no specific required dimension. Static method of NcnnModelWrapper, called just like get_broadcast_data().

### DIFF
--- a/backend/src/nodes/utils/ncnn_model.py
+++ b/backend/src/nodes/utils/ncnn_model.py
@@ -787,3 +787,12 @@ class NcnnModelWrapper:
         in_nc = weight_data_size // nf // kernel_w // kernel_h
 
         return nf, in_nc
+
+    @staticmethod
+    def get_input_dims(model: NcnnModel) -> Tuple[int, int, int]:
+        input_layer = model.layers[0]
+        w = checked_cast(int, input_layer.params[0].value)
+        h = checked_cast(int, input_layer.params[1].value)
+        c = checked_cast(int, input_layer.params[2].value)
+
+        return h, w, c


RunDevelopment:
Wait? Don't models always require a certain number of channels (`in_nc`)?

theflyingzamboni:
> Wait? Don't models always require a certain number of channels (`in_nc`)?

They do. Sorry, I meant that just height and width being 0 meant no shape was specified. Tbh, you could probably just check one of them. I don't *think* it would be possible to set only one required dimension in an Input layer?

Anyway, I just had the function return all three dims, because you never know. This isn't where we get in_nc though, as most of the time the Input layer doesn't specify a shape. in_nc is being calculated from the first Convolution layer, since that layer always contains in_nc.

RunDevelopment (approved):

RunDevelopment (changes requested):
Actually, could we change the API so that it returns `None` instead of `0` if there is no requirement?

theflyingzamboni:
Lol sure. I actually had it that way originally, then decided to keep it to returning ints and let any check be done outside the method. Guess I should have left it like that. :laughing:

RunDevelopment:
That's not really what I meant. I expected the function to return `Tuple[int | None, int | None, int | None]`. In my mind, width, height, and channel restrictions are all independently optional. I.e. there can a width restriction but no channel restriction. Is that not the case?

theflyingzamboni:
I can't say with certainty, but I don't *think* they will ever be independent? NCNN documentation being what it is though, I don't know for sure. I tried putting Input shape params into a model, but it actually had no impact on its ability to function lol, so now I'm questioning whether this is something we even need for NCNN.

RunDevelopment:
bruh

But we *know* we need it for ONNX, right?

joeyballentine:
we definitely need it for onnx. models converted without a dynamic shape wont work without it

RunDevelopment:
Alright, then let's just do ONNX for now.

We can reopen this PR once we know that we need this for NCNN as well.

theflyingzamboni:
> bruh

Now you get it. Joke's on us for assuming that because NCNN *can* have a defined input shape that it would actually do anything with it.

PR #1279, opened 2022-11-27
Make FFMPEG download failure just a warning

### LINKED ISSUE #1195: Unable to install integrated Ffmpeg
chaiNNer-windows-x64-0.14.2
```
[Window Title]
Unable to install integrated Ffmpeg

[Content]
Chainner was unable to install FFMPEG. Please ensure that your computer is connected to the internet and that chainner has access to the network.

[Exit]
```
Why doesn't it even want to start?

joeyballentine:
Fixes #1195

### DIFF
--- a/src/common/common-types.ts
+++ b/src/common/common-types.ts
@@ -228,8 +228,8 @@ export interface PythonInfo {
 }
 
 export interface FfmpegInfo {
-    readonly ffmpeg: string;
-    readonly ffprobe: string;
+    readonly ffmpeg?: string;
+    readonly ffprobe?: string;
 }
 
 export type FileSaveResult = FileSaveSuccess | FileSaveCanceled;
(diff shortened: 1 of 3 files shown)


src/common/common-types.ts
```diff
@@ -228,8 +228,8 @@ export interface PythonInfo {
 }
 
 export interface FfmpegInfo {
-    readonly ffmpeg: string;
-    readonly ffprobe: string;
+    readonly ffmpeg?: string;
+    readonly ffprobe?: string;
 }
```
RunDevelopment, line 233: I want to slightly change the type. In the code, it's actually either (1) we know both or (2) we know neither. That's a lot stronger than "either might or might not be known".

```ts
export type FfmpegInfo = {
    readonly ffmpeg: string;
    readonly ffprobe: string;
} | {
    readonly ffmpeg: undefined;
    readonly ffprobe: undefined;
}
```

RunDevelopment (approved):

PR #1283, opened 2022-11-27
Fix spawning chosen python path

joeyballentine:
Fixes https://github.com/chaiNNer-org/chaiNNer-nightly/issues/2

I was able to recreate this issue on windows, and can confirm this does fix it.

What I did:
- Normalize paths, just in case
- Use spawn instead of exec so it can be properly escaped

### DIFF
--- a/src/main/childProc.ts
+++ b/src/main/childProc.ts
@@ -1,0 +1,25 @@
+import { spawn } from 'child_process';
+
+export const promisifiedSpawn = async (command: string, args: string[]): Promise<string> => {
+    return new Promise((resolve, reject) => {
+        const proc = spawn(command, args);
+        let stdout = '';
+        let stderr = '';
+        proc.stdout.on('data', (data) => {
+            stdout += data;
+        });
+        proc.stderr.on('data', (data) => {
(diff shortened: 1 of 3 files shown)


RunDevelopment (approved):

limnibal:
Latest-nightly 2022.11.30 gets stuck on the splash screen when I set system python and corresponding exe (in my dml folder). Im on Win10. 
Logs:
[main.log](https://github.com/chaiNNer-org/chaiNNer/files/10135750/main.log)

joeyballentine:
Ok so this is a different issue. We've seen this pop up before -- it's an issue with python loading the SSL package. Are you using a VPN or something?

limnibal:
I don't use any VPN or special Firewall/Antivirus setup on that machine, just standard Win10pro and the Microsoft Defender.

joeyballentine:
There wasn't an issue for it, just someone mentioned it in discord. I don't think it was ever resolved though. https://discord.com/channels/930865462852591648/930875396277280788/1029102995633352715

Feel free to open a new issue though, because it's unrelated to the path issue

PR #1285, opened 2022-11-27
Add "Create Color" nodes

### LINKED ISSUE #475: Add nodes that generate images with constant color
These nodes should come in 3 flavors: grayscale (single-channel), RGB, and RGBA.

joeyballentine:
Closes #475 

image


RunDevelopment (approved):
I kinda don't like the approach, but it's good enough for now.

I always imagined that we'd make colors values (like numbers or strings) and that we'd just have 3 nodes to create color values and then 1 node to make those into an image. Color values can be useful for e.g. fill colors (e.g. Create Border) or as blend values (Blend node).

PR #1287, opened 2022-11-27
Add TensorRT caching

### LINKED ISSUE #959: TensorRT: save conversion model into caching file
TensorRT works for me with no problem! Thank you very much.
At the beginning, it compiles the model for a long time (some models over 230 seconds), but then it works quickly (slightly faster than CUDA).
Is conversion caching possible? so that after restarting each time you do not have to wait.

joeyballentine:
Add an option to cache TensorRT conversions done by onnx (it's a built in tensorrt execution provider setting) which is good since they take so long to convert. No clue how they determine a hash or anything like that and I can't get tensorrt to work for me anymore so I have no clue if this even works, but I don't see why it wouldn't (it at least gets through all the parts I changed fine)

Closes #959

### DIFF
--- a/src/renderer/components/SettingsModal.tsx
+++ b/src/renderer/components/SettingsModal.tsx
@@ -33,6 +33,8 @@ import {
     VStack,
     useDisclosure,
 } from '@chakra-ui/react';
+import log from 'electron-log';
+import { readdir,unlink } from 'fs/promises';
 import path from 'path';
 import {
     PropsWithChildren,
@@ -46,7 +48,7 @@ import {
 import { BsFillPencilFill, BsPaletteFill } from 'react-icons/bs';
 import { FaPython, FaTools } from 'react-icons/fa';
 import { useContext } from 'use-context-selector';
-import { hasTensorRt } from '../../common/env';
+import { getOnnxTensorRtCacheLocation, hasTensorRt } from '../../common/env';
 import { ipcRenderer } from '../../common/safeIpc';
 import { BackendContext } from '../contexts/BackendContext';
 import { SettingsContext } from '../contexts/SettingsContext';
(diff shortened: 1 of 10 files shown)


src/renderer/components/SettingsModal.tsx
```diff
@@ -580,6 +586,49 @@ const PythonSettings = memo(() => {
...
+                                                            });
+                                                        }
+                                                    })
+                                                    .catch((err) => {
+                                                        log.error(err);
+                                                    });
+                                            })
+                                            .catch((err) => {
+                                                log.error(err);
+                                            });
```
RunDevelopment, line 625: Let's make this less nested.

```suggestion
                                            .then((appDataPath: string) => {
                                                const onnxTensorRtCacheLocation =
                                                    getOnnxTensorRtCacheLocation(appDataPath);
                                                return readdir(onnxTensorRtCacheLocation)
                                            })
                                            .then((files) => {
                                                for (const file of files) {
                                                    unlink(
                                                        path.join(
                                                            onnxTensorRtCacheLocation,
                                                            file
                                                        )
                                                    ).catch((error) => {
                                                        log.error(error);
                                                    });
                                                }
                                            })
                                            .catch((err) => {
                                                log.error(err);
                                            });
```

joeyballentine: that doesnt work because the second .then needs to have `onnxTensorRtCacheLocation` available for the path joining

RunDevelopment: Then make it 
```ts
                                            .then(async (appDataPath: string) => {
                                                const onnxTensorRtCacheLocation =
                                                    getOnnxTensorRtCacheLocation(appDataPath);
                                                const files = await readdir(onnxTensorRtCacheLocation);
                                                for (const file of files) {
                                                    unlink(
                                                        path.join(
                                                            onnxTensorRtCacheLocation,
                                                            file
                                                        )
                                                    ).catch((error) => {
                                                        log.error(error);
                                                    });
                                                }
                                            })
                                            .catch((err) => {
                                                log.error(err);
                                            });
```

RunDevelopment (approved):

Issue #835, opened 2022-08-26
Actions with selected nodes via the context menu


0x4E69676874466F78:
At the moment I can't mass disable and enable nodes.
The context menu does not open if more than 1 node is selected.

RunDevelopment:
In general, it would be nice to have a context menu for mass-selected nodes.

PR #1289, opened 2022-11-28
Multi-select context menu

joeyballentine:
Partially implements #835

You still can't mass enable/disable or lock nodes though. I'm not sure how to accomplish that with the way we currently have the hooks set up (for individual changes). 

Also, this is only for the actual shift+drag multi-select, not the "hold ctrl down" multi-select

image

### DIFF
--- a/src/renderer/contexts/GlobalNodeState.tsx
+++ b/src/renderer/contexts/GlobalNodeState.tsx
@@ -124,11 +124,11 @@ interface Global {
         inputId: InputId,
         inputSize: InputSize | undefined
     ) => readonly [Readonly<Size> | undefined, (size: Readonly<Size>) => void];
-    removeNodeById: (id: string) => void;
+    removeNodesById: (ids: string[]) => void;
     removeEdgeById: (id: string) => void;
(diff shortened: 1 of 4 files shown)


src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -124,11 +124,11 @@ interface Global {
         inputId: InputId,
         inputSize: InputSize | undefined
     ) => readonly [Readonly<Size> | undefined, (size: Readonly<Size>) => void];
-    removeNodeById: (id: string) => void;
+    removeNodesById: (ids: string[]) => void;
```
RunDevelopment, line 127: Whenever you declare a function takes a mutable array, I expect that function to modify the array in some way. However, that is not the case for pretty much all array parameters added by this PR. Could you please make them readonly?

RunDevelopment (commented):
Just one thing.

Also, I'm not a fan of the `ids.forEach` in `clearNodes` (seems inefficient), but it's probably okay for now.

RunDevelopment (approved):

Issue #1233, opened 2022-11-15
Resizable View Image node


RunDevelopment:
[As requested on Discord](https://discord.com/channels/930865462852591648/930865780663398451/1041878176697159731), add a way to resize the View Image node.

joeyballentine:
In order for this to be useful you're going to need to be ok with sending larger images to the frontend. I'd even argue right now we send too small images, and quite frankly previewing it only useful if you see the actual image and not a resized version

joeyballentine:
turns out this actually duplicates #782

Issue #782, opened 2022-08-20
Custom size node (for View Image)


0x4E69676874466F78:
I want to see the view image node much larger than the other nodes, but there is no way to stretch it right now.
Using the zoom all the time is inconvenient.

joeyballentine:
I'm not sure how I didn't see this issue before now. It is definitely possible for us to add resizability to View Image nodes. However, the preview we display isn't actually the full image, it is a downscaled version of the image (if it is over a certain size). This was to prevent it from taking a long time to encode and send the image data to the frontend. So, you would not get much utility out of being able to expand the View Image node unless we increased the maximum resolution the images get resized to.

0x4E69676874466F78:
Is it possible to keep images in shared memory accessible between python and electron?
This would allow not to copy images once again.
But in general, need just set size preview enough to fit the new size node.

joeyballentine:
> Is it possible to keep images in shared memory accessible between python and electron?
> This would allow not to copy images once again.
> But in general, need just set size preview enough to fit the new size node.

Even if you could share memory, the image would still need to be encoded in order for the frontend to display it at all. In the backend it's just a numpy array, which the frontend is not able to do anything with. It needs to be encoded to PNG to be able to be displayed at all.

As for how sharing memory would work, I have no idea if that's at all possible but I could look into it I guess. Would at least be more efficient to not have to send the data over a rest request.

PR #1290, opened 2022-11-28
Increase maximum image preview resolution

joeyballentine:
It's really not any slower than 512 tbh

Figured it was a good idea to do this at least to make the existing preview a bit more useful before fully implementing #782 / #1233

### DIFF
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -96,18 +96,22 @@ class LargeImageOutput(ImageOutput):
         img = value
         h, w, c = get_h_w_c(img)
 
+        previews = []
+
         # Encode for multiple scales. Use the preceding scale to save time encoding the smaller sizes.
-        base64_2048, np_2048 = preview_encode(img, 2048)
-        base64_1024, np_1024 = preview_encode(np_2048, 1024)
-        base64_512, np_512 = preview_encode(np_1024, 512)
-        base64_256, _ = preview_encode(np_512, 256)
-        del np_2048, np_1024, np_512
+        last_encoded = img
+        for size in [2048, 1024, 512, 256]:
+            if h >= size or w >= size:
+                base64, last_encoded = preview_encode(last_encoded, size)
+                le_h, le_w, _ = get_h_w_c(last_encoded)
+                previews.insert(0, {"size": max(le_h, le_w), "url": base64})
+
+        if h < 256 and w < 256:
+            base64, _ = preview_encode(img, 256)
+            previews.insert(0, {"size": max(h, w), "url": base64})
 
         return {
-            "256": base64_256,
-            "512": base64_512,
-            "1024": base64_1024,
-            "2048": base64_2048,
+            "previews": previews,
             "height": h,
             "width": w,
             "channels": c,
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -96,10 +96,18 @@ class LargeImageOutput(ImageOutput):
         img = value
         h, w, c = get_h_w_c(img)
 
-        base64_img = preview_encode(img, 2048)
+        # Encode for multiple scales. Use the preceding scale to save time encoding the smaller sizes.
+        base64_2048, np_2048 = preview_encode(img, 2048)
+        base64_1024, np_1024 = preview_encode(np_2048, 1024)
+        base64_512, np_512 = preview_encode(np_1024, 512)
+        base64_256, _ = preview_encode(np_512, 256)
+        del np_2048, np_1024, np_512
 
         return {
-            "image": base64_img,
+            "256": base64_256,
+            "512": base64_512,
+            "1024": base64_1024,
+            "2048": base64_2048,
             "height": h,
             "width": w,
             "channels": c,
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -96,7 +96,7 @@ class LargeImageOutput(ImageOutput):
         img = value
         h, w, c = get_h_w_c(img)
 
-        base64_img = preview_encode(img, 512)
+        base64_img = preview_encode(img, 2048)
 
         return {
             "image": base64_img,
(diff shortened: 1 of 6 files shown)


RunDevelopment:
We chose 512x512 as the maximum size because that ensured that each image is at most 1 MB in size (1.33 MB in base64). With 2048x2084 it's 16 MB (21.3 MB in base64). Are we really sure that we want the frontend to deal with 20 MB data URL strings?

Also, the reason we made previews small and didn't use the original image in Load Image (back when the preview was still tied to the file input) is that large image previews lag the UI. #116

joeyballentine:
Like i said in #1233 you're going to need to be ok with this otherwise resizing the view image node will be useless. It already doesn't even use up all the resolution it can, and resizing it to make it bigger will only make that more obvious.

RunDevelopment:
Lagging the UI for everyone for a feature that isn't going to be used by everyone is not an okay solution.

How about this: (pseudo) mip maps. We'll broadcast multiple preview images at 128, 512 and, 2084 pixels in size. The frontend then dynamically picks one of those depending on whether a node is in view and the current zoom level. So the frontend will only show high-res images if necessary. Of course, we don't need to send the large versions if the image isn't actually that large.

joeyballentine:
That sounds like a good idea

joeyballentine:
This actually worked really well IMO

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -96,10 +96,18 @@ def get_broadcast_data(self, value: np.ndarray):
         img = value
         h, w, c = get_h_w_c(img)
 
-        base64_img = preview_encode(img, 512)
+        # Encode for multiple scales. Use the preceding scale to save time encoding the smaller sizes.
+        base64_2048, np_2048 = preview_encode(img, 2048)
+        base64_1024, np_1024 = preview_encode(np_2048, 1024)
```
RunDevelopment, line 101: Using the downscaled image from the previous iteration is a really clever trick. I like that a lot.

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -96,10 +96,18 @@ def get_broadcast_data(self, value: np.ndarray):
...
+        base64_512, np_512 = preview_encode(np_1024, 512)
+        base64_256, _ = preview_encode(np_512, 256)
+        del np_2048, np_1024, np_512
```
RunDevelopment, line 104: Why do you do this?

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -96,10 +96,18 @@ def get_broadcast_data(self, value: np.ndarray):
         img = value
         h, w, c = get_h_w_c(img)
 
-        base64_img = preview_encode(img, 512)
+        # Encode for multiple scales. Use the preceding scale to save time encoding the smaller sizes.
+        base64_2048, np_2048 = preview_encode(img, 2048)
```
RunDevelopment, line 100: I don't like that we PNG encode small images multiple times. Basically, if the image is smaller than the next smaller preview size, then we don't need that preview (e.g. if the image is 500px, then we don't need >=1024px).

Of course, variably many previews doesn't work with the hard-coded preview sizes, so we should probably replace them with a `preview: { size: number, url: string }[]` prop.

src/renderer/components/outputs/LargeImageOutput.tsx
```diff
@@ -10,12 +10,22 @@ import { GlobalContext, GlobalVolatileContext } from '../../contexts/GlobalNodeS
...
+    '512': string;
+    '1024': string;
+    '2048': string;
     width: number;
     height: number;
     channels: number;
 }
 
+const pickImage = (last: LargeImageBroadcastData, zoom: number) => {
+    if (zoom < 1) return last['256'];
```
RunDevelopment, line 23: This method of picking makes various assumptions:
1. It assumes no DPI scaling/zooming via Electron.
2. It assumes that the preview is (roughly) 200px (CSS units).

What we are really interested in to pick the right preview is this question: How many *screen* pixels is the preview container currently?

So we want `realSize = cssSize * zoom * dpiScaling * electronZoom`. In our case, `cssSize` is the 200px of the preview container. I think the browser already has a method to give you `dpiScaling * electronZoom`, so you can use that. 

It's quite easy to pick the right preview image with `realSize` and it also makes it easy to pick the right interpolation method (`imageRendering: pixelated/auto`).

backend/src/nodes/properties/outputs/numpy_outputs.py
RunDevelopment: Also, since `preview_encode` treats `target_size` as more of a suggestion, `size: number` should be the actual size of the (returned) preview image.

RunDevelopment (changes requested):

backend/src/nodes/properties/outputs/numpy_outputs.py
joeyballentine: Just in case... I don't really know how pythons garbage collection works

backend/src/nodes/properties/outputs/numpy_outputs.py
joeyballentine: How will the frontend decide which to show then? Index?

RunDevelopment: It knows the size of each preview. That's enough information to pick one.

backend/src/nodes/properties/outputs/numpy_outputs.py
RunDevelopment: Unless you [`gc.collect` afterwards](https://stackoverflow.com/questions/1316767/how-can-i-explicitly-free-memory-in-python), it won't help much. It's like we need that those relatively small images anyway. GC should be smart enough to clean those up in a timely manner.

backend/src/nodes/properties/outputs/numpy_outputs.py
```diff
@@ -96,10 +96,22 @@ def get_broadcast_data(self, value: np.ndarray):
         img = value
         h, w, c = get_h_w_c(img)
 
-        base64_img = preview_encode(img, 512)
+        previews = []
+
+        # Encode for multiple scales. Use the preceding scale to save time encoding the smaller sizes.
+        last_encoded = img
+        for size in [2048, 1024, 512, 256]:
+            if h >= size or w >= size:
```
RunDevelopment, line 104: This condition isn't correct. Suppose we have an image 700x700px. This image will only get 512px and 256px previews, but no 700px preview.

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #1291, opened 2022-11-28
Add translation hooks to more files

joeyballentine:


src/common/locales/en/translation.json
```diff
@@ -22,7 +25,30 @@
...
+  "outputs": {
+    "largeImage": {
+      "imageNotAvailable": "Image not available."
+    },
+    "ncnn": {
+      "modelNotAvailable": "Model data not available."
+    },
+    "pytorch": {
+      "modelNotAvailable": "Model data not available."
+    }
```
RunDevelopment, line 50: I wonder, should those 2 (and ONNX when added) be unified?

joeyballentine: Probably

RunDevelopment (approved):

PR #1292, opened 2022-11-28
Add Surface Blur Node

### LINKED ISSUE #1180: Surface blur
Add a Surface blur node. Kinda like PS' or Paint.net's Surface blur filter.

https://en.wikipedia.org/wiki/Bilateral_filter

joeyballentine:
Closes #1180

I may have done something wrong, because the sigma values don't seem to do anything, but I also don't really know what they're supposed to do

### DIFF
--- a/backend/src/nodes/nodes/image_filter/blur_bilateral.py
+++ b/backend/src/nodes/nodes/image_filter/blur_bilateral.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import cv2
+import numpy as np
+
+from . import category as ImageFilterCategory
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput
+from ...properties.outputs import ImageOutput
+from ...utils.utils import get_h_w_c
+
+
+@NodeFactory.register("chainner:image:bilateral_blur")
+class BlurNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply surface/bilateral blur to an image."
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Radius", controls_step=1),
+            NumberInput("Color Sigma", controls_step=1, default=75),
+            NumberInput("Space Sigma", controls_step=1, default=75),
+        ]
+        self.outputs = [ImageOutput(image_type="Input0")]
+        self.category = ImageFilterCategory
+        self.name = "Surface Blur"
+        self.icon = "MdBlurOn"
+        self.sub = "Blur"
+
+    def run(
+        self,
+        img: np.ndarray,
+        radius: int,
+        sigma_color: int,
+        sigma_space: int,
+    ) -> np.ndarray:
+
+        if radius == 0:
+            return img
+
+        _, _, c = get_h_w_c(img)
+        if c == 4:
+            rgb = img[:, :, :3]
+            alpha = img[:, :, 3]
+            rgb = cv2.bilateralFilter(
+                rgb,
+                radius,
+                sigma_color,
+                sigma_space,
+                borderType=cv2.BORDER_REFLECT_101,
+            )
+            alpha = cv2.bilateralFilter(
+                alpha,
+                radius,
+                sigma_color,
+                sigma_space,
+                borderType=cv2.BORDER_REFLECT_101,
+            )
+            result = np.dstack((rgb, alpha))
+            return np.clip(result, 0, 1)
+
+        # Bilateral filter with reflected padding
+        return np.clip(
+            cv2.bilateralFilter(
+                img, radius, sigma_color, sigma_space, borderType=cv2.BORDER_REFLECT_101
+            ),
+            0,
+            1,
+        )


RunDevelopment:
Alright, this is gonna be a bit of a rant about `cv.bilateralFilter`: This function utter garbage for images. Just look at this:

| Info | Image |
| :-- | --- |
| Original | image |
| gauss blur <br> r=2 | image |
| bilateralFilter <br> r=6 color=1.0 space=1.0 | image |
| bilateralFilter <br> r=6 color=0.5 space=0.5 | image |
| bilateralFilter <br> r=6 color=0.1 space=0.1 | image |
| bilateralFilter <br> r=6 color=0.05 space=0.05 | image |
| bilateralFilter <br> r=6 color=0.01 space=0.01 | image |
| bilateralFilter <br> r=6 color=0 space=0 | image |
| **Paint.net surface blur** <br> r=6 threshold=15 | image |

(Values are given directly to `cv.bilateralFilter`, so you need to multiply by 100 to get the sigma inputs for the node.)

Compared to Paint.net's surface blur, `cv.bilateralFilter` produces images with blurry edges and not enough blurring on surfaces. At high sigmas, it's even comparable to a gauss blur. Not even the steep edges of the dark weapons on the light background in the bottom of the image are preserved.

I probably just don't understand `cv.bilateralFilter` enough to use it correctly, but as it is, it's not usable.

joeyballentine:
Turns out, this actually works really well -- in the 0-255 range.

joeyballentine:
One thing I just thought of is I'm not actually sure what the cap of the sigmas is, I assumed 100, but it might be higher or just infinite

backend/src/nodes/nodes/image_filter/blur_bilateral.py
```diff
@@ -0,0 +1,68 @@
...
+@NodeFactory.register("chainner:image:bilateral_blur")
+class BlurNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply surface/bilateral blur to an image."
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Radius", controls_step=1),
+            NumberInput("Color Sigma", controls_step=1, default=75, maximum=100),
+            NumberInput("Space Sigma", controls_step=1, default=75, maximum=100),
```
theflyingzamboni, line 23: There isn't actually a maximum for these values, and the documentation even mentions large values as >150. There's definitely a point around 300-400 where it starts approximating Gaussian too much to be very useful, but I think we should let the max be higher than 100. Maybe the 250-300 range.

backend/src/nodes/nodes/image_filter/blur_bilateral.py
```diff
@@ -0,0 +1,68 @@
...
+        img: np.ndarray,
+        radius: int,
+        sigma_color: int,
+        sigma_space: int,
+    ) -> np.ndarray:
+
+        if radius == 0:
+            return img
+
+        img = (img * 255.0).astype(np.uint8)
```
theflyingzamboni, line 42: Okay, after some research and testing, I believe this is unnecessary. To get it to work with 0-1, you have to divide *only* sigmaColor by 255 to normalize it (but not sigmaSpace, it would seem). Unless I'm tired and messed up my testing, this should give equivalent results without destroying data in the image.

backend/src/nodes/nodes/image_filter/blur_bilateral.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@NodeFactory.register("chainner:image:bilateral_blur")
+class BlurNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply surface/bilateral blur to an image."
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Radius", controls_step=1),
```
theflyingzamboni, line 21: According to the docs, this is diameter, not radius.

joeyballentine: That would make sense why I had to use 12 to get similar results to paint.net's 6

joeyballentine: Should I just multiply the radius by 2 then? or should I change this to be diameter?

theflyingzamboni (dismissed):

RunDevelopment:
Alright, this works way better. I still found some things:

1. If Color Sigma=0, it seems to be ignored.
2. Please make r=6, color=25, and space=25 the default values. Especially for complex nodes, the starting values should be something that works and users can experiment with.
3. Could you make the sigmas log-scale sliders between 0 and 1000? This makes them easier to use and 1000 should be big enough.

joeyballentine:
Works for me

joeyballentine:
> If Color Sigma=0, it seems to be ignored.

what should I do about this? As far as I'm concerned, that's just how this is supposed to work....

RunDevelopment:
> As far as I'm concerned, that's just how this is supposed to work....

Is it? From what I see, this is a bug in opencv.

I mainly say this because it breaks continuity. Looks at this:

Color=100 (0.390625)
image

Color=10 (0.0390625)
image

Color=1 (0.00390625)
image

Color=0 (0.0)
image

Compare this with 
Color=256 (1.0)
image

It seems like color=0 gets interpreted as "value missing" or something, and so it uses a default of 1.0. [This behavior is not documented](https://docs.opencv.org/3.4/d4/d86/group__imgproc__filter.html#ga9d7064d478c95d60003cf839430737ed). I guess, I'll open an OpenCV issue for this.

From what I see, color=0 should result in the image being returned as is, so let's use that as a fix for now.

backend/src/nodes/nodes/image_filter/blur_bilateral.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def run(
+        self,
+        img: np.ndarray,
+        diameter: int,
+        sigma_color: int,
+        sigma_space: int,
+    ) -> np.ndarray:
+
+        if diameter == 0 or sigma_space == 0:
```
RunDevelopment, line 53: The 0==1 thing applies to both sigmas. [More info](https://github.com/opencv/opencv/issues/22886#issuecomment-1331383727).

```suggestion
        if diameter == 0 or sigma_color == 0 or sigma_space == 0:
```

RunDevelopment (approved):

PR #1293, opened 2022-11-28
Added amount input for Unsharp Mask

RunDevelopment:
This PR adds an amount slider to the Unsharp Mask node. Turns out, the blur radius (which we previously incorrectly called "amount") is only half of unsharp mask. [Gimp docs explain this quite nicely](https://docs.gimp.org/2.10/en/gimp-filter-unsharp-mask.html).

I made amount a log scale, since small amount values are more important. I also adjusted the scale of High Boost for consistency.

image


joeyballentine (approved):

RunDevelopment:
About Gimp's threshold: We could implement that too. Unsharp mask essentially is just `image + (image - blurred) * amount`. So a threshold can be implemented by modifying the difference `image - blurred`.

What do you think?

PR #1296, opened 2022-11-28
Onnx getter for input shape and better error message for models

RunDevelopment:
Changes:
- Added function to determine the input shape of an Onnx model.
- Added error message for models with input size requirements that are not met.

I'll add the tiling and padding stuff for exact-size models some day else. For now, an error will do.

### DIFF
--- a/backend/src/nodes/nodes/onnx/upscale_image.py
+++ b/backend/src/nodes/nodes/onnx/upscale_image.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from typing import Tuple, Literal
 import numpy as np
 import onnxruntime as ort
 from sanic.log import logger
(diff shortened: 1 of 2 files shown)


joeyballentine:
I feel like @theflyingzamboni should review this one

backend/src/nodes/nodes/onnx/upscale_image.py
```diff
@@ -10,14 +11,36 @@
...
-from ...utils.auto_split_tiles import parse_tile_size_input, TileSize
+from ...utils.auto_split_tiles import parse_tile_size_input, TileSize, NO_TILING
 from ...utils.onnx_auto_split import onnx_auto_split
 from ...utils.onnx_model import OnnxModel
 from ...utils.onnx_session import get_onnx_session
 from ...utils.utils import get_h_w_c, convenient_upscale
 from ...utils.exec_options import get_execution_options
 
 
+OnnxInputShape = Literal["BCHW", "BHWC"]
```
theflyingzamboni, line 22: Can we figure out whether the dim orders are HW or WH? We're doing HW here, but when we define input/output in the Convert to ONNX node, we have:
```
dynamic_axes = {
    "input": {0: "batch_size", 2: "width", 3: "height"},
    "output": {0: "batch_size", 2: "width", 3: "height"},
}
```

I have found people reference it both ways, so I'm currently not sure which is technically correct. It *probably* won't matter in terms of functioning, but it would be good to at least be internally consistent.

theflyingzamboni (commented):
Logic seems correct and produced expected results, so other than the minor comment, it seems good.

backend/src/nodes/nodes/onnx/upscale_image.py
joeyballentine: ONNX docs use `b, c, h, w`

example: https://onnx.ai/onnx/operators/onnx__DepthToSpace.html

So I think our exporter is wrong

RunDevelopment: I was using the same logic that the upscaling process uses. If you follow the `change_shape` parameter into `onnx_auto_split`, you'll find that `onnx_auto_split` assume either `BCHW` or `BHWC`. So if WH is also possible, we have some fixing to do.

RunDevelopment: So the PR is gtg @theflyingzamboni?

theflyingzamboni (dismissed):

backend/src/nodes/nodes/onnx/upscale_image.py
RunDevelopment: I don't even know what this export you guys were talking about is lol

joeyballentine: The PyTorch -> onnx node

theflyingzamboni (approved):

joeyballentine (approved):

PR #1299, opened 2022-11-29
Inform user of why their attempted connection is invalid

joeyballentine:
The type ones are super ugly right now. I'm not sure what to do about that, so @RunDevelopment will need to weigh in on that.

image
image
image


RunDevelopment:
The infinite loop warning doesn't show up for self-referential connections.

image

joeyballentine:
image

PR #1302, opened 2022-11-29
Improved settings modal appearance

joeyballentine:
- Darker color (now shows scrollbar)
- Fixed scrolling so tabs float
- Fixed padding

image


RunDevelopment:
I like the new darker color, but I don't like that settings is now inconsistent with all other popups. Could you also change the bg color of the dependency manager modal and alerts?

joeyballentine:
I was going to do that in separate PRs

RunDevelopment (approved):

PR #1303, opened 2022-11-29
Improved errors for invalid connection tooltip

RunDevelopment:
I just extracted out the whole type mismatch logic from `checkNodeValidity` and reused in `isValidConnect`.

image

image

image


joeyballentine (approved):
Perfect

RunDevelopment:
Note: Generated type errors still need to be improved, but that can be its own PR.

image

PR #1307, opened 2022-11-29
Removed PyTorch scale tag in output

RunDevelopment:
We already removed this for NCNN, so I removed for PyTorch as well.
image


joeyballentine:
Is there any way we could also move the model arch and the arch type tags to the type tag area as well? Especially since showing the arch type as an input type restriction would be nice

RunDevelopment:
You mean SR vs Face, right?

joeyballentine:
Yes

RunDevelopment:
image

Maybe we should rename the model input of Upscale Face to just "Model"? The "Face SR" is redundant now.

joeyballentine:
Yeah that sounds good to me

joeyballentine (approved):

PR #1309, opened 2022-11-29
Fix "final newline" config 

RunDevelopment:
@joeyballentine your last translation PR removed the last empty line that is auto generated. This is really annoying cause it changes that file every time I start chainner.


joeyballentine:
Yeah I noticed that. Same issue as the package.json I think, since i manually edited the file. Annoying. What setting do i need to change for that again?

joeyballentine:
And if it's a vscode setting is there a way you could add that to the vscode settings file in the repo?

RunDevelopment:
image

I'll add it to our project settings.

RunDevelopment:
Okay, I updated the workspace settings to always add a final new line.

I also set it to always trim trailing spaces because we had some of those in some of zamboni's PRs.

joeyballentine (approved):

PR #1314, opened 2022-11-30
Improve alert box design

joeyballentine:
- Use a darker color
- Added an icon for each alert type.

image

### DIFF
--- a/src/renderer/contexts/AlertBoxContext.tsx
+++ b/src/renderer/contexts/AlertBoxContext.tsx
@@ -67,7 +67,7 @@ const EMPTY_MESSAGE: InternalMessage = {
 
 const ALERT_FOCUS_ID = 'alert-focus-button';
 
-const pickAlertIcon = (type: AlertType) => {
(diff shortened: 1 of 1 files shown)


src/renderer/contexts/AlertBoxContext.tsx
```diff
@@ -67,6 +67,49 @@ const EMPTY_MESSAGE: InternalMessage = {
 
 const ALERT_FOCUS_ID = 'alert-focus-button';
 
+const pickAlertIcon = (type: AlertType) => {
```
RunDevelopment, line 70: The return type of this function is incorrect. It's `JSX.Element | undefined` right now, but you likely intended it to be `JSX.Element`. This is because just calling `assertNever(type)` won't communicate the TS that this path never returns.

To fix this, please add a return type annotation to this function and make it `return assertNever(type)`.
```suggestion
const pickAlertIcon = (type: AlertType): JSX.Element => {
```

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #1342, opened 2022-12-04
Improved performance of previews

RunDevelopment:
Changes:
- Use PNG or JPEG as the preview encoding format. As discussed on discord, PNG encoding is around 10x faster than WebP and JPEG is around 2x faster than PNG. Right now, PNG will always be used for RGBA images and the largest preview.
- The maximum preview size is now picked optimally. One of the reasons for the performance problem was that the image was encoded as a preview at fully size twice if it was `p_size <= image_size <= p_size * 1.2` where `p_size` is that size of the largest preview. This won't happen anymore.


joeyballentine (approved):
This is so much faster, wow

PR #1343, opened 2022-12-04
Show current overall progress on taskbar

joeyballentine:
cant take a screenshot because for some reason windows decided that the taskbar is more important than everything else and wont hide it (for anything, even fullscreen videos)

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -264,7 +264,7 @@ class Executor:
             ProgressController() if not parent_executor else parent_executor.progress
         )
 
-        self.completed_node_ids = []
+        self.completed_node_ids = set()
 
         self.loop: asyncio.AbstractEventLoop = loop
         self.queue: EventQueue = queue
@@ -408,8 +408,7 @@ class Executor:
 
         async def send_broadcast():
             data = await self.loop.run_in_executor(self.pool, compute_broadcast_data)
-            if not node_id in self.completed_node_ids:
-                self.completed_node_ids.append(node_id)
+            self.completed_node_ids.add(node_id)
             await self.queue.put(
                 {
                     "event": "node-finish",
@@ -429,8 +428,7 @@ class Executor:
             # broadcasts are done is parallel, so don't wait
             self.__broadcast_tasks.append(self.loop.create_task(send_broadcast()))
         else:
-            if not node_id in self.completed_node_ids:
-                self.completed_node_ids.append(node_id)
+            self.completed_node_ids.add(node_id)
             await self.queue.put(
                 {
                     "event": "node-finish",
@@ -450,8 +448,7 @@ class Executor:
         if not node_id in finished:
             finished.append(node_id)
 
-        if not node_id in self.completed_node_ids:
-            self.completed_node_ids.append(node_id)
+        self.completed_node_ids.add(node_id)
 
         total_length = len(self.chain.nodes.keys())
         progress_percent = len(self.completed_node_ids) / total_length
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -264,6 +264,8 @@ class Executor:
             ProgressController() if not parent_executor else parent_executor.progress
         )
 
+        self.completed_node_ids = []
+
         self.loop: asyncio.AbstractEventLoop = loop
         self.queue: EventQueue = queue
         self.pool: ThreadPoolExecutor = pool
@@ -406,6 +408,8 @@ class Executor:
 
         async def send_broadcast():
             data = await self.loop.run_in_executor(self.pool, compute_broadcast_data)
+            if not node_id in self.completed_node_ids:
+                self.completed_node_ids.append(node_id)
             await self.queue.put(
                 {
                     "event": "node-finish",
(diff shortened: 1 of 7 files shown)


RunDevelopment:
Idk how, but you broke something pretty badly. I get this when I load an image. Chainner closes after that.

image

```
[1] 23:44:23.434 > Unhandled Exception TypeError: Error processing argument at index 0, conversion failure from null
[1]     at IpcMainImpl.<anonymous> (C:\Users\micha\Git\chaiNNer\.webpack\main\index.js:5932:112)
[1]     at IpcMainImpl.emit (node:events:527:28)
[1]     at EventEmitter.<anonymous> (node:electron/js2c/browser_init:165:11237)
[1]     at EventEmitter.emit (node:events:527:28)
```

joeyballentine:
wtf

src/renderer/hooks/useBackendEventSource.ts
```diff
@@ -28,6 +28,7 @@ export interface BackendEventMap {
         nodeId: string;
         executionTime?: number | null;
         data?: OutputData | null;
+        progressPercent?: number;
```
RunDevelopment, line 31: There's the problem. Wrong type. Python's None gets jsonified to null.
```suggestion
        progressPercent?: number | null;
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -209,6 +209,16 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
 
     const [isBackendKilled, setIsBackendKilled] = useState(false);
 
+    const [percentComplete, setPercentComplete] = useState<number | undefined>(undefined);
+
+    useEffect(() => {
+        if (percentComplete === undefined) {
+            ipcRenderer.send('set-progress-bar', -1);
```
RunDevelopment, line 216: You spread the fact that -1 is interpreted as "none" all the way into the frontend. Could you change the API of `set-progress-bar` to take a `number | null`? This would make for a cleaner API and the handler can be implemented as a nice oneliner: `mainWindow.setProgressBar(progress ?? -1)`.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -209,6 +209,16 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
 
     const [isBackendKilled, setIsBackendKilled] = useState(false);
 
+    const [percentComplete, setPercentComplete] = useState<number | undefined>(undefined);
```
RunDevelopment, line 212: I don't understand why this useState and useEffect exists. The state is never read outside the effect, so why not just call `ipcRenderer.send('set-progress-bar', value)` directly in `updateNodeFinish`'s callback?

RunDevelopment (changes requested):

src/renderer/contexts/ExecutionContext.tsx
joeyballentine: I plan on making a future PR that shows this percentage on the frontend as well. This state will be exported out of the context.

RunDevelopment:
Just noticed this. Stopping the current execution should reset the progress.

backend/src/process.py
```diff
@@ -264,6 +264,8 @@ def __init__(
             ProgressController() if not parent_executor else parent_executor.progress
         )
 
+        self.completed_node_ids = []
```
RunDevelopment, line 267: Please make this a set. You only use `completed_node_ids` to add ids not already in `completed_node_ids` and to get its length. Both are O(1) operations for sets.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -209,11 +209,22 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+
+    useEffect(() => {
+        if (percentComplete === undefined) {
+            ipcRenderer.send('set-progress-bar', null);
+        } else {
+            ipcRenderer.send('set-progress-bar', percentComplete);
+        }
```
RunDevelopment, line 219:
```suggestion
        ipcRenderer.send('set-progress-bar', percentComplete ?? null);
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -249,6 +260,13 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                         inputHash,
                         data ?? undefined
                     );
+                    if (progressPercent !== undefined && progressPercent !== null) {
```
RunDevelopment, line 263: Small trick.

```suggestion
                    if (progressPercent != null) {
```

backend/src/process.py
```diff
@@ -440,13 +448,19 @@ def __create_node_finish(self, node_id: NodeId) -> Event:
         if not node_id in finished:
             finished.append(node_id)
 
+        self.completed_node_ids.add(node_id)
+
+        total_length = len(self.chain.nodes.keys())
```
RunDevelopment, line 453: I don't think we need the `.keys()`. Same for the other 2 occurrences.

```suggestion
        total_length = len(self.chain.nodes)
```

PR #1344, opened 2022-12-04
Handle electron second-instance events

joeyballentine:


RunDevelopment (approved):
I'm just gonna assume that this works, cause I don't want to make an installation to test this.

PR #1346, opened 2022-12-05
Use custom font (Supreme)

joeyballentine:
Was looking at fonts and really liked [this one](www.fontshare.com/fonts/supreme). I think it might be a bit more narrow which gives us more space for node headings and whatnot.

Drafting this PR because 1) I want feedback on font choice, and 2) I already see a specific new issue (node timer) which I will have to address, and there may be more. Will just need to go through the whole UI and test.

image


RunDevelopment:
Most people have to zoom out when looking at large chains, so font rendering at small font sizes is very important. It seems like Supreme is worse at this than the one we are currently using.

Example: current font vs supreme
image
image

joeyballentine:
Honestly, I don't think that's much of an issue personally

PR #1348, opened 2022-12-05
Automatically restart backend on deps changes

RunDevelopment:
Changes:
- The deps manager will automatically restart the backend after (un)installing dependencies.
- Removed "Restart chaiNNer" button from deps manager.
- Everything related to types is not allowed to assume that the schema id of a node is valid.
- While testing, I found that the backend event source is not a good indicator to find out whether the backend is still alive. When restarting the backend, the event source *sometimes* errors (as in, its current state is "error"), but it actually still works just fine. So I demoted the event source error alert to a log warning.

### DIFF
--- a/src/common/safeIpc.ts
+++ b/src/common/safeIpc.ts
@@ -45,7 +45,6 @@ export interface InvokeChannels {
     >;
     'get-cli-open': ChannelInfo<FileOpenResult<ParsedSaveData> | undefined>;
     'owns-backend': ChannelInfo<boolean>;
-    'kill-backend': ChannelInfo<void>;
     'restart-backend': ChannelInfo<void>;
     'relaunch-application': ChannelInfo<void>;
     'quit-application': ChannelInfo<void>;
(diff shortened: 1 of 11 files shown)


RunDevelopment:
Invalid nodes are automatically removed now, and when the current chain has unsaved changes, we will display another warning.

I also slightly changed alerts so that these "nested" alerts aren't detected as overlapping.

joeyballentine:
Seems like when it removes the nodes it doesn't immediately remove the connected edges. I got two of these warnings 
image

RunDevelopment:
I remove nodes via `removeNodesById`, so removing them likely isn't the problem. Invalid nodes don't have handles, and since they are only removed after schemata changes have been detected, RF will render one frame where the invalid nodes + edges are still present.

I don't think that there is anything I can do about that.

joeyballentine:
ok, thats fine

joeyballentine (approved):

PR #1351, opened 2022-12-05
Fixed preview resizing not being alpha-corrected

### LINKED ISSUE #1321: Preview resizing is not alpha-corrected
When the frontend switches between the original image and a resized preview, noticeable black outlines can occur. The issue seems to be that OpenCV's resize (which is used by `preview_encode`) does not properly account for the alpha channel and resizes each channel independently instead. PIL's resize (which is used by our resize nodes) does not have this issue.

In the following example, I used a different checkerboard color to make it more obvious.
image
image

RunDevelopment:
Fixes #1321.

I needed to move `preview_encode` out of `image_util.py` because `pil_util.py` (which is where `resize` comes from) depends on that file. So the diff is kinda bad... This is the fully diff of `preview_encode`

```diff
-        img = cv2.resize(img, (int(w / f), int(h / f)), interpolation=cv2.INTER_AREA)
+        t = (int(w / f), int(h / f))
+        if c == 4:
+            # https://github.com/chaiNNer-org/chaiNNer/issues/1321
+            img = resize(img, t, InterpolationMethod.BOX)
+        else:
+            img = cv2.resize(img, t, interpolation=cv2.INTER_AREA)
```

### DIFF
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -1,9 +1,11 @@
-from typing import Optional
+import base64
+from typing import Optional, Tuple
 import numpy as np
+import cv2
 
-from ...utils.image_utils import preview_encode
 from ...utils.utils import get_h_w_c
 from ...utils.format import format_image_with_channels
+from ...utils.pil_utils import resize, InterpolationMethod
 from .base_output import BaseOutput, OutputKind
 from .. import expression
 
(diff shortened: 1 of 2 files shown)


RunDevelopment:
> Is it faster to use cv2 for <4 channel images?

I would suspect that it is because we don't need to convert the image, but I haven't measured it. Will do.

RunDevelopment:
Yup, cv is quite a bit faster. Downsampling an 8k RGB image to 2k took cv 40ms and Pil 600ms. So roughly a 10x difference.

RunDevelopment:
Nope.

joeyballentine (approved):

PR #1356, opened 2022-12-05
Added support for saving DDS files

### LINKED ISSUE #1286: Save DDS images
Add a DDS format option to Save Image.

This will have to come with multiple conditional inputs:
- Format: There are many DDS formats, so the user has to pick one.
- Dithering: Whether to use dithering.
- BC7 has a setting to adjust quality vs time it takes to save the image.

RunDevelopment:
This fixes #1286 by adding support for a few commonly used DDS formats. I only added a few for now, but adding more is easy since we just need to extend `DdsFormatDropdown`.

TODO:

- [x] Defaults. A lot of dropdowns have suboptimal defaults because chainner is hard-coded to use the first value as the default. (#1357)
- [x] Conditional inputs. I need support for nested groups to only show the DDS format specific options when necessary. (#1360)

### DIFF
--- a/backend/src/nodes/nodes/image/save_image.py
+++ b/backend/src/nodes/nodes/image/save_image.py
@@ -104,7 +104,7 @@ class ImWriteNode(NodeBase):
 
         os.makedirs(base_directory, exist_ok=True)
 
-        # DDS files are handles separately
+        # DDS files are handled separately
         if extension == "dds":
             save_as_dds(
                 full_path,


backend/src/nodes/nodes/image/save_image.py
```diff
@@ -80,6 +106,21 @@ def run(
         img = (np.clip(img, 0, 1) * 255).round().astype("uint8")
 
         os.makedirs(base_directory, exist_ok=True)
+
+        # DDS files are handles separately
```
joeyballentine, line 110:
```suggestion
        # DDS files are handled separately
```

joeyballentine (approved):

PR #1280, opened 2022-11-27
Possible pytorch-DML support

joeyballentine:
Should theoretically resolve #174, but requires system python and manually setting up the package. No clue if it actually works though. Someone who wants to set up python 3.8 will need to try it

PR #1359, opened 2022-12-06
theoretically fix for non-cuda device mapping

joeyballentine:
This might fix the error mentioned in the log at the bottom of #1280. I think the error originates from us giving it "dml:0" even though dml probably cannot be mapped to a specific device. Therefore, I changed this check to only add the :{number} in the case of cuda.

### DIFF
--- a/backend/src/nodes/utils/exec_options.py
+++ b/backend/src/nodes/utils/exec_options.py
@@ -38,7 +38,7 @@ class ExecutionOptions:
 
     @property
     def full_device(self) -> str:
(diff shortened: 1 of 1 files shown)


limnibal:
In Alpha v0.16.0 i still get a error message when trying to load pytorch models. 
System Win10pro, using miniconda and dml
Logs: 
[main.log](https://github.com/chaiNNer-org/chaiNNer/files/10188958/main.log)

FNsi:
Tbh I think it's much easier to use dml in onnx runtime...

PR #1361, opened 2022-12-06
Use same pyright version as pylance

RunDevelopment:


joeyballentine:
Can you please set it to ignore the block file?

joeyballentine (approved):

PR #1365, opened 2022-12-06
Fixed stuck progress bar

RunDevelopment:
This fix consistent of 3 parts:
1. The `percentComplete` state is now the source of truth and the only thing sending to `'set-progress-bar'`.
2. When the current status is "not running", we will (1) reset progress (like before) and (2) never display progress.
3. I added an invisible element to ensure that changes to `status` and `percentComplete` cause DOM changes, which are a requirement for effects to run.


joeyballentine (approved):

joeyballentine:
Oops

PR #1366, opened 2022-12-06
Refactored convenient_upscale

theflyingzamboni:
- Added handling for models where in_nc != out_nc. These models will now work properly and output images with c == out_nc. Output for models where in_nc == out_nc will continue to output images with c == input image c.
- Refactored convenient_upscale to divide it into preprocessing, upscaling, and postprocessing to try and make it a little clearer.


backend/src/nodes/nodes/onnx/upscale_image.py
```diff
@@ -41,6 +41,19 @@ def get_input_shape(
...
+    session: ort.InferenceSession,
+) -> Tuple[OnnxInputShape, int, int | None, int | None]:
+    """
+    Returns the output shape, output channels, output width (optional), and output height (optional).
+    """
+    shape = session.get_outputs()[0].shape
+    if isinstance(shape[1], int) and shape[1] <= 4:
+        return "BCHW", shape[1], as_int(shape[3]), as_int(shape[2])
+    else:
+        return "BHWC", shape[3], as_int(shape[2]), as_int(shape[1])
```
RunDevelopment, line 54: Maybe we could move this logic into a `parse_onnx_shape` function?

RunDevelopment (commented):
Could you explain the general approach for handling models with out_nc != in_nc?

theflyingzamboni:
> Could you explain the general approach for handling models with out_nc != in_nc?

For models where out_nc != in_nc, we can't rely on image channels in == image channels out, because these models are meant to output specific types of images (e.g. a 1 -> 3 model is for colorizing, so outputting a grayscale image makes no sense).

For preprocessing, these models aren't really any different than the usual ones, since they number of image input channels still needs to match the model input nc. The only difference is for images with alpha; for example, feeding an RGBA image to a RGB -> GRAY model. When feeding an RGBA image to something like an RGB -> RGB model, we need to retain some extra information to restore the alpha channel. When in_nc != out_nc, we can discard that information because we aren't adding an alpha channel back to the output.

Similarly, since the postprocessing section is for altering the number of output image channels to match the number of input image channels, I skipped it for these models (so the number of output image channels is whatever the model outputs).

RunDevelopment:
Alright, so I skimmed through the code and I don't like the approach you've taken at all. You've basically taken the old code, split it into 3 parts, and then inserted the `in_nc != out_nc` logic into these 3 parts. This resulted in complex and tightly coupled code.

But all of this interwoven `in_nc != out_nc` logic could be removed if `convenient_upscale` started with `if model_in_nc != model_out_nc: return upscale(to_target_channels(img, model_in_nc))`. Of course, this isn't a magic trick that solves everything, but it makes the two different behaviors of the function so much clearer. As I see it, this `if` is currently spread out over 3 functions.

Now for `restore_input_channels`: This variable and associated logic can be completely removed by adding `if inimg_c == model_in_nc: return upscale(img)`. This also makes it very clear that all the complex processing is only done if the image and model channels don't match.

theflyingzamboni:
@RunDevelopment So it's definitely not there yet (because I only really thought about it after I was done), but part of the reason I want to split it into 3 distinct steps is that I'm trying to get it to a place where it can provide a framework for supporting architectures like RealCUGAN that require data pre-processing and post-processing to function.

RunDevelopment:
Will this additional pre- and postprocessing even make sense for general images? Right now, convient_upscale takes images with any number of channels, so the architecture-specific preprocessing also has to work on images with any number of channels. Otherwise, why shouldn't we put the pre- and postprocessing into the passed in `upscale` method itself?

RunDevelopment:
Alright, let's take a few steps back. What is the purpose of `convenient_upscale`? Well, most upscalers only accept images with a certain number of channels, and this function's purpose is to make images with any number of channels compatible with a given upscaler. Consequently, the number of channels of the image passed to `upscale` is **the only guarantee** this function makes.

However, you want `convenient_upscale` to make additional guarantees that are enforced by some user-provided functions.

IMO this is fundamentally the wrong approach. Instead of adding more responsibilities into one function, each function should only do one thing. Adding pre- and post-processing responsibilities just make `convenient_upscale` and more complex, when the same functionality can also be achieved with function composition.

As an example: To upscale with NCNN, we need to (1) adjust the channels of the image, (2) tile the image, and (3) convert the image to uint8. Of course, all these steps then have to be reversed to get the final upscaled image. This is really easy to do with function composition. Each function essentially looks like this:

```py
def transform_image(image, upscale):
    prepared_image = preprocessing(image)
    upscaled_image = upscale(prepared_image)
    return postprocessing(image)
```

And we then chain the processing tasks via function composition:
```py
upscaled_image = convenient_upscale(
    image,
    lambda i: auto_split(
        i, 
        lambda: i ncnn.upscale(i) # uint8/float conversions are done inside this function
    )
)
```

Of course, python sucks, so function composition is very verbose in this language. To get around this, we make give these composed functions a name. E.g.:
```py
def ncnn_auto_split(image):
    return auto_split(image, lambda i: ncnn.scale())

# I am simplifying, but that is all the <arch>_auto_split functions are in essence
```

Getting to your example:
How would we do the normalization for quant models? With function composition of course!
```py
def quant_normalize(image, upscale):
    norm_image = image * 2 - 1
    upscaled_image = upscale(norm_image)
    return (upscaled_image + 1) / 2

upscaled_image = convenient_upscale(
    image,
    lambda i: auto_split(
        i, 
        lambda i: quant_normalize(
            i,
            lambda: i quant.upscale(i)
        )
    )
)
```

Again, python sucks, so you would give these composed function names.

All these function compositions essentially form a pipeline that gradually processes the image, with each individual function only doing one thing. And to change the pipeline (e.g. to add or remove steps), you only need to change how these functions are composed.

---

Side note: Having written all that, maybe we should move to a different pattern than function composition. By factoring out the image parameter, we can get an even nicer pattern that is a bit easier to compose in python:

```py
Upscaler = Callable[[Image], Image]

def quant_normalize(inner_upscale: Upscaler) -> Upscaler:
    def upscale(image: Image) -> Image:
        norm_image = image * 2 - 1
        upscaled_image = upscale(norm_image)
        return (upscaled_image + 1) / 2
    return upscale

def auto_split(inner_upscale: Upscaler, tiler, other_params) -> Upscaler:
    ...
def convenient_upscaler(inner_upscale: Upscaler, model_channels) -> Upscaler:
    ...

# usage
u: Upscaler = quant.upscale
u = quant_normalize(u)
u = auto_split(u, ...)
u = convenient_upscaler(u, ...)

u(image)
```

backend/src/nodes/utils/utils.py
```diff
@@ -187,88 +187,126 @@ def nptensor2np(
...
+) -> Tuple[np.ndarray, np.ndarray | int | None]:
+    """Adjust number of image channels if it does not equal number of channels
+    required by the model."""
     c = get_h_w_c(img)[2]
 
-    # Transparency hack (white/black background difference alpha)
-    if c == 4 and input_channels in (1, 3):
-        # Ignore single-color alpha
+    imgout = img
+    supplemental_output = None
```
RunDevelopment, line 198: You might want to call this "alpha" everywhere, cause that's what it is.

RunDevelopment (commented):
Even without looking at `to_target_channels` and `from_target_channels` too closely, I know that they are incorrect right now. This isn't your fault, the code was buggy before the refactor, it's just that those bugs are easy to see now.

Why are they incorrect? Since `convenient_upscale` already handles the trivial `model_in_nc != model_out_nc` and `inimg_c == model_in_nc` cases, we know that `{to,from}_target_channels` have to do some work, they have to change the channels of the input image in some way. However, there are multiple ways in which both functions can just return the input image as is right now.

To fix this, please return the results right away instead of assigning them to a variable with a default value and then returning that variable. This allows pyright to do simply verify that the functions always return a value. Example:

```py
def from_target_channels(
    img: np.ndarray, alpha: np.ndarray | None, inimg_c: int
) -> np.ndarray:
    """Adjust number of output image channels to match number of
    input image channels."""
    outimg_c = get_h_w_c(img)[2]

    if (inimg_c == 4 and outimg_c in (1, 3)) or (inimg_c == 3 and outimg_c == 1):
        if img.ndim == 2:
            return np.tile(np.expand_dims(img, axis=2), (1, 1, min(inimg_c, 3)))
        elif outimg_c == 1:
            return np.tile(img, (1, 1, min(inimg_c, 3)))

        if alpha is not None:
            return np.dstack((img, alpha))
    elif (inimg_c in (1, 3) and outimg_c == 4) or (inimg_c == 1 and outimg_c == 3):
        img = img[:, :, :3]
        if inimg_c == 1:
            return cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)

    ... # well, that's the bug
```

I would probably replace the `...` with raising an error. It's better to just fail instead of returning invalid results.

backend/src/nodes/utils/utils.py
theflyingzamboni: It isn't alpha everywhere though. When using difference alpha, it's returning a modified copy of the original image which is then upscaled and used to calculate alpha later in the code.

theflyingzamboni:
> However, there are multiple ways in which both functions can just return the input image as is right now.

What is an example of an input that would return the image unmodified?

backend/src/nodes/utils/utils.py
RunDevelopment: Right, we do this white/black background trick. Good point.

RunDevelopment:
For example, give `from_target_channels` an RGB image, no alpha, and ask it to output an RGBA image. Sure, assuming that `convenient_upscale` and `to_target_channels` are implemented correctly, this will never happen, but that's global reasoning. If the correctness of `from_target_channels` depends entirely on the 2 other functions, then `from_target_channels` is tightly coupled with them. The situation is even worse for `to_target_channels` because that function will just do the wrong thing if the image or model doesn't have 1/3/4 channels.

> I believe we enforce this on numpy inputs

It isn't.

---

Frankly, I question whether splitting convenient upscale into pre- and post-processing even makes sense. Just having a general channel converter function and making the alpha logic into `convenient_upscale` makes the whole thing a lot simpler to reason about:

<details>

```py
def to_target_channels(img: np.ndarray, target: int) -> np.ndarray:
    """Adjusts the given image to have `target` number of channels."""
    c = get_h_w_c(img)[2]

    if c == target:
        return img

    if c == 1:
        if target == 3:
            return cv2.cvtColor(img, cv2.COLOR_GRAY2BGR)
        if target == 4:
            return cv2.cvtColor(img, cv2.COLOR_GRAY2BGRA)

    if c == 3:
        if target == 1:
            return cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
        if target == 4:
            return cv2.cvtColor(img, cv2.COLOR_BGR2BGRA)

    if c == 4:
        if target == 1:
            return cv2.cvtColor(img, cv2.COLOR_BGRA2GRAY)
        if target == 3:
            return cv2.cvtColor(img, cv2.COLOR_BGRA2BGR)

    raise ValueError(f"Unable to convert {c} channel image to {target} channel image")


def clipped(upscale: Callable[[np.ndarray], np.ndarray]):
    return lambda i: np.clip(upscale(i), 0, 1)


def with_black_and_white_backgrounds(img: np.ndarray):
    c = get_h_w_c(img)[2]
    assert c == 4

    black = np.copy(img[:, :, :3])
    white = np.copy(img[:, :, :3])
    for c in range(3):
        black[:, :, c] *= img[:, :, 3]
        white[:, :, c] = (white[:, :, c] - 1) * img[:, :, 3] + 1

    return black, white


def convenient_upscale(
    img: np.ndarray,
    model_in_nc: int,
    model_out_nc: int,
    upscale: Callable[[np.ndarray], np.ndarray],
) -> np.ndarray:
    """
    Upscales the given image in an intuitive/convenient way.

    This method guarantees that the `upscale` function will be called with an image with
    `model_in_nc` number of channels.

    Additionally, guarantees that the number of channels of the output image will match
    that of the input image in cases where `model_in_nc` == `model_out_nc`, and match
    `model_out_nc` otherwise.
    """
    img_c = get_h_w_c(img)[2]

    unclipped_upscale = upscale
    upscale = clipped(upscale)

    if model_in_nc != model_out_nc:
        return upscale(to_target_channels(img, model_in_nc))

    model_c = model_in_nc

    if img_c == model_c:
        return upscale(img)

    if img_c == 4:
        # Ignore alpha if single-color or not being replaced
        unique = np.unique(img[:, :, 3])
        if len(unique) == 1:
            rgb = convenient_upscale(img[:, :, :3], model_c, model_c, unclipped_upscale)
            unique_alpha = np.full(rgb.shape[:-1], unique[0], np.float32)
            return np.dstack((rgb, unique_alpha))

        # Transparency hack (white/black background difference alpha)
        black, white = with_black_and_white_backgrounds(img)
        black_up = convenient_upscale(black, model_c, model_c, unclipped_upscale)
        white_up = convenient_upscale(white, model_c, model_c, unclipped_upscale)

        alpha = 1 - np.mean(white_up - black_up, axis=2)
        return np.dstack((black_up, alpha))

    return to_target_channels(upscale(to_target_channels(img, model_c)), img_c)
```

</details>

theflyingzamboni:
@RunDevelopment This code example is incorrect and will return 2-channel images if inimg_c is 4 and outimg_c is 1. The function shouldn't return after converting 1 channel to 3.

```def from_target_channels(
    img: np.ndarray, alpha: np.ndarray | None, inimg_c: int
) -> np.ndarray:
    """Adjust number of output image channels to match number of
    input image channels."""
    outimg_c = get_h_w_c(img)[2]

    if (inimg_c == 4 and outimg_c in (1, 3)) or (inimg_c == 3 and outimg_c == 1):
        if img.ndim == 2:
            return np.tile(np.expand_dims(img, axis=2), (1, 1, min(inimg_c, 3)))
        elif outimg_c == 1:
            return np.tile(img, (1, 1, min(inimg_c, 3)))

        if alpha is not None:
            return np.dstack((img, alpha))
    elif (inimg_c in (1, 3) and outimg_c == 4) or (inimg_c == 1 and outimg_c == 3):
        img = img[:, :, :3]
        if inimg_c == 1:
            return cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)

    ... # well, that's the bug
```

theflyingzamboni:
> It isn't.

Is there a reason why not?

RunDevelopment:
> Is there a reason why not?

No idea. Are there models out there that upscale 2-channel or >4-channel images? I could image 2-channel models for normal maps, but I don't know very much about which models are out there.

But I also want to point out the `convenient_input` should rely on the validation of ImageInput because that's global reasoning again.

RunDevelopment:
> Also, recursion seems kind of excessive for this?

We can also use `rgb = to_target_channels(upscale(to_target_channels(img[:, :, :3], model_c)), 3)` (similar for the 2 other occurrences). Actually, that's kinda nice because it makes it really explicit that `rgb` is indeed a 3-channel image.

theflyingzamboni:
> No idea. Are there models out there that upscale 2-channel or >4-channel images? I could imagine 2-channel models for normal maps, but I don't know very much about which models are out there.

I have never seen anything except a 1-, 3-, or 4-channel model. I don't know that anyone has even bothered with losses that would handle a 2-channel image when training. @joeyballentine?

joeyballentine:
2 channel models are possible of course but yeah, unlikely since most losses just work with 3. But that doesn't mean there could never be a 2 channel model for normal maps or something, if someone wanted to they could make it (like how i trained a few 4 channel ones)

theflyingzamboni:
I implemented the latest suggestion. There was a bug in it in that it could return 2D images instead of always 3D, so I fixed that.

RunDevelopment:
> There was a bug in it in that it could return 2D images instead of always 3D, so I fixed that.

Why is that a bug? I thought `convenvient_upscale` only guaranteed the number of channels?

theflyingzamboni:
> Why is that a bug? I thought `convenvient_upscale` only guaranteed the number of channels?

It also has to (and did before) guarantee dimensions because pytorch needs `ndim == 3`.

RunDevelopment:
Some thoughts on this:

1. If this must be guaranteed and others are relying on that guarantee, then it should be documented.
2. Instead of making `to_target_channel` more complex, I think it's better to wrap `upscale` into a function that enforces the guarantee (like `clipped`). (Let's call that wrapper function `ensure_3d` for later.)
3. Why should `convenvient_upscale` guarantee this? This need seems very specific to pytorch and if pytorch can't deal with 2d images, then we should fix our pytorch code IMO. Also, if we had an `ensure_3d` function, then pytorch could just use that.

theflyingzamboni:
I disagree with points 2 and 3. Making sure that the input image has a proper shape to be inputted into the upscaler is the whole purpose of convenient_upscale, and the shape of the input is part of that. Additionally, the fact that the cv2 color conversion methods take a 3D array and output a 2D array when converting to gray is an implementation detail that the to_target_channels method should be correcting for to yield the input needed for upscaling.

(Also, I was partially incorrect, ONNX needs 3D input as well. Only NCNN doesn't seem to care.)

theflyingzamboni:
Actually, it does need to come out of to_target_channels to a separate function. The way you refactored it, it no longer ensures that an image is 3D for `in_img_c == model_in_nc`. I still believe this is part of the job of convenient_upscale though, so I will move it there.

theflyingzamboni:
So I made as_3d/ensure_3d methods and wrapped upscale in convenient upscale like with clipped. Wanted to put the former in image_utils, but it would have been a circular import. Oh well.

theflyingzamboni:
> We use 2d images to represent grayscale images everywhere, it's their natural representation. PyTorch is the weird one here for not supporting that. If we ever want to use PyTorch upscaling without convenient_upscale (as I have suggested doing in the past), then we have to account for this strange restriction again.

As I amended above, ONNX requires 3D arrays as well, so NCNN is the odd one out among our upscalers in not requiring it.

As for the 2D enforcement, I recall this came about because you opted to use np.concatenate in a function instead of dstack, which errors when receiving a 3D array and 2D array. Aside from calling concatenate sometimes (which could probably be replaced), is there actually a reason to prefer 2D? Unless there's a situations where we have functions that genuinely need grayscale to be 2D, I'd prefer we enforce 3D instead, so that all of our numpy arrays have consistent ndims. Right now, grayscale is the odd one out.

theflyingzamboni:
If you're truly dead set that grayscale images must be 2D and to_target_channels should not return consistent shape, as_3d could also be enforced in np2nptensor, as this is where the transpose call for both ONNX and PyTorch is. This would make more sense than putting it in the upscale functions themselves.

theflyingzamboni:
> Yeah, convenient_upscale should probably be in its own file TBH. It's not exactly a general util method. I originally wanted to use as_target_channels (with some modifications) instead of making a new to_target_channels functions, but I couldn't even import it.

I was thinking about this as well, especially now that it has additional methods it depends on. I'll do that.

theflyingzamboni:
@RunDevelopment While I'm at it, instead of having `to_target_channels` and `as_target_channels`, should I just get rid of the former and modify `as_target_channels` to support channel narrowing operations? I had forgotten we had that method, and the two together seem somewhat redundant.

RunDevelopment:
> should I just get rid of the former and modify `as_target_channels` to support channel narrowing operations?

Yes, I think that would be good. But add narrowing operations behind an optional parameter (e.g. `narrowing=False`). Current callers of `as_target_channels` assume that the function errors on narrowing operations (since that is the current specified behavior).

Regarding the name: I think `as*` is better than `to*` here. `to*` is commonly used in conversions that yield a new object, but our function returns the image as is if the number of channels match.

> as_3d could also be enforced in np2nptensor

If that is the code that errors for 2d images, then yes, I believe that is the code that should be fixed.

---

On the subject of `np2nptensor`: Maybe we should restructure `src/nodes/util`? This folder contains the implementations of dozens of nodes and seems almost completely unorganized...

---

Also, what are `np2nptensor` and ``np2tensor``? They seem to do the same things but tensors and numpy.

theflyingzamboni:
> On the subject of np2nptensor: Maybe we should restructure src/nodes/util? This folder contains the implementations of dozens of nodes and seems almost completely unorganized...

I agree, I think that would be a very good idea at this point. After this PR is merged, that is. There are way too many different things grouped together, and it's a pain to hunt for files. Certainly the ncnn/onnx/pytorch stuff could easily have their own util folders.

> Also, what are np2nptensor and np2tensor? They seem to do the same things but tensors and numpy.

No idea. @joeyballentine?

joeyballentine:
Np2tensor converts a numpy array to a PyTorch tensor. This includes putting it in the right data range and shape and order and whatnot.

When i implemented onnx stuff, it needed the same things, but obviously not with pytorch. So it's basically the same exact function but does all that shape/range/etc stuff with numpy

theflyingzamboni:
Sorry for the diff, things had to move for imports to work right.

backend/src/nodes/utils/noise_utils.py
```diff
@@ -18,6 +20,7 @@ def __add_noises(
         img = img[:, :, :3]
 
     noises = noise_gen(h, w)
+    logger.info(noises[0].dtype)
```
RunDevelopment, line 23: Is that intentional?

backend/src/nodes/utils/image_utils.py
```diff
@@ -199,31 +199,54 @@ def as_2d_grayscale(img: np.ndarray) -> np.ndarray:
...
-
-    if target_channels == 3:
-        if c == 1:
-            return np.dstack((img, img, img))
-
-    if target_channels == 4:
-        return convert_to_BGRA(img, c)
-
-    assert False, "Unable to convert image"
+    if narrowing is False:
```
RunDevelopment, line 226: I mean, `is` technically works, but should that really be used over `==` for bools? This is still my first python project, so I'm not really aware of best practices of the language, but seems a bit strange to me.

backend/src/nodes/utils/noise_utils.py
theflyingzamboni: Ugh, no. I thought I got all those.

backend/src/nodes/utils/image_utils.py
RunDevelopment:
```py
>>> isinstance(True, int)
True
```

You just gave me another good reason to hate python. I never would have expected this. Not even JS is that stupid. Like, I know why they did it, bools in C and all, but come on.

RunDevelopment (approved):

joeyballentine (approved):

PR #1369, opened 2022-12-07
Support node-search context menu in iterators

### LINKED ISSUE #1353: Right click in *iterator space not works
Can't create nodes by right click popup because it not works.

joeyballentine:
Closes #1353

### DIFF
--- a/src/renderer/hooks/usePaneNodeSearchMenu.tsx
+++ b/src/renderer/hooks/usePaneNodeSearchMenu.tsx
@@ -324,7 +324,8 @@ interface Position {
 }
 
 export const usePaneNodeSearchMenu = (
-    wrapperRef: React.RefObject<HTMLDivElement>
+    wrapperRef: React.RefObject<HTMLDivElement>,
+    parent?: string
 ): UsePaneNodeSearchMenuValue => {
     const typeState = useContextSelector(GlobalVolatileContext, (c) => c.typeState);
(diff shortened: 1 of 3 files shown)


src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -379,7 +380,7 @@ export const usePaneNodeSearchMenu = (
                     },
                     nodeType: schema.nodeType,
                 },
-                stoppedOnIterator || undefined
+                stoppedOnIterator || parent || undefined
```
RunDevelopment, line 383: The `|| undefined` is unnecessary, I think.

joeyballentine: Since it was there before I'd rather not remove that unless you're sure

RunDevelopment: I am sure. The reason it was there in the first place is that `stoppedOnIterator` allows `null` (for some reason), but the parameter expected `string | undefined`.

RunDevelopment: Also, with `null` and `undefined` in general: TypeScript is really good at tracking these. If TS says it's fine, then it most likely is.

joeyballentine, line 383:
```suggestion
                stoppedOnIterator || parent
```

RunDevelopment (approved):

PR #1377, opened 2022-12-09
Added checkbox variants of dropdowns

RunDevelopment:
image


src/renderer/global.scss
```diff
@@ -148,3 +148,42 @@ body {
 .chainner-handle:has(+ .with-label) {
     margin-top: 1rem;
 }
+
+.chainner-node-checkbox > .chakra-checkbox__control {
```
joeyballentine, line 152: would this be better placed in a folder with the actual checkbox component?

src/renderer/components/inputs/DropDownInput.tsx
```diff
@@ -1,55 +1,41 @@
...
-    // reset invalid values to default
-    useEffect(() => {
-        if (value !== undefined && options.every((o) => o.value !== value)) {
-            setValue(def);
-        }
-    }, [value, setValue, options, def]);
-
-    let selection = options.findIndex((o) => o.value === value);
-    if (selection === -1) selection = 0;
+    if (options.length === 2 && options[0].option === 'Yes') {
```
joeyballentine, line 15: Could you make this an exhaustive check for both Yes and No, and also include checks for things like true/false?

Also, I'm thinking maybe it should actually be defined in the backend as a checkbox input that uses a boolean rather than a dropdown. Would make more sense IMO, and then you wouldn't need this check at all.

RunDevelopment: Yes, the current yes/no detection is only a temporary solution. Thanks for pointing that out, I almost forgot to change it.

I was thinking about something like a `preferredStyle` property for dropdowns. This would allow dropdowns to communicate how they are would like to displayed. The frontend can then choose to display them differently (e.g. in the style of a checkbox).

RunDevelopment:
Latest design (just for completeness)
image

joeyballentine (approved):

PR #1384, opened 2022-12-11
Convert values to numbers for number and slider inputs

### LINKED ISSUE #1364: Crash when opening an old chain
[This chain](https://github.com/chaiNNer-org/chaiNNer/files/10168916/big.ol.test.zip) produces this error when attempting to open it:

```
SliderInput.tsx?ce2e:102 Uncaught TypeError: val.toFixed is not a function
    at eval (SliderInput.tsx?ce2e:102:69)
    at eval (SliderInput.tsx?ce2e:182:39)
    at renderWithHooks (react-dom.development.js?61bb:16175:1)
    at updateFunctionComponent (react-dom.development.js?61bb:20387:1)
    at updateSimpleMemoComponent (react-dom.development.js?61bb:20222:1)
    at updateMemoComponent (react-dom.development.js?61bb:20081:1)
    at beginWork (react-dom.development.js?61bb:22502:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:4161:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:4210:1)
    at invokeGuardedCallback (react-dom.development.js?61bb:4274:1)
```

RunDevelopment:
Fixes #1364 by converting the saved values of all number and slider inputs to numbers. If the saved value is not a value number, it will be removed so the frontend can reset it to its default value. However, invalid values *shouldn't* be possible.

The big chain now loads without problems.
image


joeyballentine:
Is there no way to do this with the inputs themselves rather than migrations? Like when it gets its initial value cast it to a number?

RunDevelopment:
The problem isn't that inputs are bugged, but that save files have invalid values. I don't think that inputs should have to validate their values. It's just not their responsibility. (TBH, what dropdown does is already a bit too much IMO.)

joeyballentine (approved):

PR #1396, opened 2022-12-13
Reorganize utils folder

theflyingzamboni:
As discussed on Discord. Most file now moved into `impl` or one of its subfolders.

image


backend/src/nodes/nodes/onnx/upscale_image.py
```diff
@@ -10,12 +10,12 @@
...
-from ...utils.convenient_upscale import convenient_upscale
-from ...utils.auto_split import ExactTileSize
-from ...utils.onnx_auto_split import onnx_auto_split
-from ...utils.onnx_model import OnnxModel
-from ...utils.onnx_session import get_onnx_session
+from ...impl.upscale.auto_split_tiles import parse_tile_size_input, TileSize
+from ...impl.upscale.convenient_upscale import convenient_upscale
+from ...impl.upscale.auto_split import ExactTileSize
+from ...impl.onnx.onnx_auto_split import onnx_auto_split
+from ...impl.onnx.onnx_model import OnnxModel
```
RunDevelopment, line 17: I don't think we need the `onnx` prefix anymore. Same for ncnn and PyTorch files. The classes and functions within those files should keep any prefixes, though.
```suggestion
from ...impl.onnx.model import OnnxModel
```

backend/src/nodes/impl/pytorch/torch_types.py
```diff
@@ -6,9 +6,9 @@
 from .architecture.SwinIR import SwinIR
 from .architecture.Swin2SR import Swin2SR
 from .architecture.HAT import HAT
-from .architecture.face_archs.gfpganv1_clean_arch import GFPGANv1Clean
-from .architecture.face_archs.restoreformer_arch import RestoreFormer
-from .architecture.face_archs.codeformer import CodeFormer
+from .architecture.face_arches.gfpganv1_clean_arch import GFPGANv1Clean
```
RunDevelopment, line 9: Maybe we could just call it `face`? We already know that it's about architectures.

```suggestion
from .architecture.face.gfpganv1_clean_arch import GFPGANv1Clean
```

RunDevelopment (commented):
2 suggestions, gtg otherwise.

theflyingzamboni:
I just realized I forgot to remove `_util` from various files too.

RunDevelopment (approved):

PR #1397, opened 2022-12-14
More impl refactoring

RunDevelopment:
My goal is to eventually get rid of `image_util` and `pil_util`, so here are a few changes toward that goal.

Changes:
- Renamed `blend_modes.py` to `blend.py` and moved the `blend_images` into it.
- Added `caption.py` for `add_caption`.
- Added `image_formats.py` for image formats.
- Moved `normalize_normals` into `impl/normals/util.py`.

### DIFF
--- a/backend/src/nodes/impl/blend.py
+++ b/backend/src/nodes/impl/blend.py
@@ -1,6 +1,9 @@
 import cv2
 import numpy as np
 
+from ..utils.utils import get_h_w_c
+from .image_utils import as_target_channels
+
 
 class BlendModes:
     """Blending mode constants"""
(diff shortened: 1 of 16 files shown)


theflyingzamboni:
One thing I was thinking about when reorganizing was whether a couple of methods (like `get_h_w_c`) should move from `utils/utils.py` to 'impl/image_utils.py`.

joeyballentine (approved):

RunDevelopment:
> One thing I was thinking about when reorganizing was whether a couple of methods (like `get_h_w_c`) should move from `utils/utils.py` to `impl/image_utils.py`.

Same. Although, I would have `get_h_w_c` as a util util cause it feels like that too me.

PR #1404, opened 2022-12-16
Shorter directory type tags

RunDevelopment:
I noticed that directory type tags are now causing Load Image nodes to become wider because of the more compact design. I tried to fix this by making the tag 2 chars shorter. 

While this does help, we really need a general solution for type tags causing nodes to widen, but I don't have any...


joeyballentine:
Yeah, I'm not really sure what to do about that to be honest

joeyballentine (approved):

PR #1413, opened 2022-12-19
Make FillColor and RotateSizeChange enums

RunDevelopment:


backend/src/nodes/impl/pil_utils.py
```diff
@@ -64,5 +69,10 @@ def rotate(
     interpolation = INTERPOLATION_METHODS_MAP[interpolation]
 
     pimg = Image.fromarray((img * 255).astype("uint8"))
-    pimg = pimg.rotate(angle, interpolation, expand, fillcolor=fill_color)  # type: ignore
+    pimg = pimg.rotate(
+        angle,
+        interpolation,  # type: ignore
```
joeyballentine, line 74: Why the type ignore?

RunDevelopment: Cause PIL types are wrong-ish. Basically, their constants are typed as int (not literal ints), but their functions expect a special subtype of int.

joeyballentine (approved):

PR #1437, opened 2022-12-26
New iterator system specification

RunDevelopment:
This PR adds a specification for the new iterator system. The system and its behavior are quite complex, so I wanted to write all of it down before even attempting to implement it.

The specification shouldn't only guide our implementation efforts but also be helpful when writing documentation for this feature.


joeyballentine:
Does this PR still need to be open? I know it documents things that we still want to accomplish one day, but 1) does this need to be a PR, and 2) does this need to be in our repo at all?

PR #1442, opened 2022-12-29
Added Image Convolve Node

JumpyJacko:
I added this because I thought it would be nice to have a generalised convolve node for other image processing tasks like the difference of gaussians.

### DIFF
--- a/backend/src/nodes/nodes/image_utility/convolve.py
+++ b/backend/src/nodes/nodes/image_utility/convolve.py
@@ -9,13 +9,12 @@ from ...properties.inputs import ImageInput, NumberInput, TextInput
 from ...properties.outputs import ImageOutput
 from ...properties import expression
 
+
 @NodeFactory.register("chainner:image:image_convolve")
 class ImageConvolveNode(NodeBase):
     def __init__(self):
         super().__init__()
-        self.description = (
-            "Convolves input image with input kernel (kernel values separated by commas)"
-        )
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
         self.inputs = [
             ImageInput(),
             TextInput("Kernel String"),
@@ -30,7 +29,7 @@ class ImageConvolveNode(NodeBase):
         self.name = "Convolve"
         self.icon = "MdAutoFixHigh"
         self.sub = "Miscellaneous"
-    
+
     def run(
         self,
         img: np.ndarray,
@@ -38,15 +37,15 @@ class ImageConvolveNode(NodeBase):
         kernel_dim: int,
         padding: int,
         strides: int,
-        ) -> np.ndarray:
-        
+    ) -> np.ndarray:
+
         kernel = np.array([float(d) for d in kernel_in.split(",")])
-        kernel = kernel.reshape(kernel_dim,kernel_dim)
-        
+        kernel = kernel.reshape(kernel_dim, kernel_dim)
+
         # Grayscale image if it is not already
         if len(img.shape) != 2:
             img = img[:, :, 0]
-        
+
         # Thanks Samrat Sahoo on Medium for the convolution code
         # Cross Correlation
         kernel = np.flipud(np.fliplr(kernel))
@@ -64,8 +63,12 @@ class ImageConvolveNode(NodeBase):
 
         # Apply Equal Padding to All Sides
         if padding != 0:
-            imagePadded = np.zeros((img.shape[0] + padding*2, img.shape[1] + padding*2))
-            imagePadded[int(padding):int(-1 * padding), int(padding):int(-1 * padding)] = img
+            imagePadded = np.zeros(
+                (img.shape[0] + padding * 2, img.shape[1] + padding * 2)
+            )
+            imagePadded[
+                int(padding) : int(-1 * padding), int(padding) : int(-1 * padding)
+            ] = img
         else:
             imagePadded = img
 
@@ -83,7 +86,10 @@ class ImageConvolveNode(NodeBase):
                     try:
                         # Only Convolve if x has moved by the specified Strides
                         if x % strides == 0:
-                            output[x, y] = (kernel * imagePadded[x: x + xKernShape, y: y + yKernShape]).sum()
+                            output[x, y] = (
+                                kernel
+                                * imagePadded[x : x + xKernShape, y : y + yKernShape]
+                            ).sum()
                     except:
                         break
 
--- a/backend/src/nodes/nodes/image_utility/convolve.py
+++ b/backend/src/nodes/nodes/image_utility/convolve.py
@@ -21,7 +21,7 @@ class ImageConvolveNode(NodeBase):
             TextInput("Kernel String"),
             NumberInput("Kernel Width/Height", minimum=0, default=3),
             NumberInput("Padding", minimum=0, default=0),
-            NumberInput("Strides", minimum=0, default=1),
+            NumberInput("Strides", minimum=1, default=1),
         ]
         self.outputs = [
             ImageOutput(
--- a/backend/src/nodes/nodes/image_utility/convolve.py
+++ b/backend/src/nodes/nodes/image_utility/convolve.py
@@ -16,10 +16,9 @@ class ImageConvolveNode(NodeBase):
         super().__init__()
         self.description = "Convolves input image with input kernel"
         self.inputs = [
-            ImageInput("Grayscale Image", channels=1),
+            ImageInput("Image"),
             NoteTextAreaInput("Kernel String"),
             NumberInput("Padding", minimum=0, default=0),
-            NumberInput("Strides", minimum=1, default=1),
         ]
         self.outputs = [
             ImageOutput(
@@ -30,15 +29,12 @@ class ImageConvolveNode(NodeBase):
                     let kernel = Input1;
 
                     let padding = Input2;
-                    let stride = Input3;
 
                     Image {
-                        width: int & floor((w + padding * 2 - 1) / stride + 1),
-                        height: int & floor((h + padding * 2 - 1) / stride + 1),
-                        channels: 1,
+                        width: int & floor((w + padding * 2 - 1) + 1),
+                        height: int & floor((h + padding * 2 - 1)+ 1),
                     }
                 """,
-                channels=1,
             )
         ]
         self.category = ImageUtilityCategory
@@ -51,35 +47,24 @@ class ImageConvolveNode(NodeBase):
         img: np.ndarray,
         kernel_in: str,
         padding: int,
-        strides: int,
     ) -> np.ndarray:
 
         kernel = np.stack([l.split() for l in kernel_in.splitlines()], axis=0).astype(
             float
         )
 
-        # Thanks Samrat Sahoo on Medium for the convolution code
         kernel = np.flipud(np.fliplr(kernel))
 
-        xKernShape = kernel.shape[0]
-        yKernShape = kernel.shape[1]
-        xImgShape = img.shape[0]
-        yImgShape = img.shape[1]
+        img = cv2.copyMakeBorder(
+            img,
+            top=padding,
+            left=padding,
+            right=padding,
+            bottom=padding,
+            borderType=cv2.BORDER_CONSTANT,
+            value=0,
+        )
 
-        xOutput = int(((xImgShape - xKernShape + 2 * padding) / strides) + 1)
-        yOutput = int(((yImgShape - yKernShape + 2 * padding) / strides) + 1)
-        output = np.zeros((xOutput, yOutput))
-
-        if padding != 0:
-            imagePadded = np.zeros(
-                (img.shape[0] + padding * 2, img.shape[1] + padding * 2)
-            )
-            imagePadded[
-                int(padding) : int(-1 * padding), int(padding) : int(-1 * padding)
-            ] = img
-        else:
-            imagePadded = img
-
-        output = cv2.filter2D(imagePadded, -1, kernel)
+        output = cv2.filter2D(img, -1, kernel)
 
         return output
--- a/backend/src/nodes/nodes/image_utility/convolve.py
+++ b/backend/src/nodes/nodes/image_utility/convolve.py
@@ -31,8 +31,8 @@ class ImageConvolveNode(NodeBase):
                     let padding = Input2;
 
                     Image {
-                        width: int & floor((w + padding * 2 - 1) + 1),
-                        height: int & floor((h + padding * 2 - 1)+ 1),
+                        width: w + padding * 2,
+                        height: h + padding * 2,
                     }
                 """,
             )
--- a/backend/src/nodes/nodes/image_utility/convolve.py
+++ b/backend/src/nodes/nodes/image_utility/convolve.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+import numpy as np
+
+from . import category as ImageUtilityCategory
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput, TextInput
+from ...properties.outputs import ImageOutput
+from ...properties import expression
+
+@NodeFactory.register("chainner:image:image_convolve")
+class ImageConvolveNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = (
+            "Convolves input image with input kernel (kernel values separated by commas)"
+        )
+        self.inputs = [
+            ImageInput(),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
+            NumberInput("Padding", minimum=0, default=0),
+            NumberInput("Strides", minimum=0, default=1),
+        ]
+        self.outputs = [
+            ImageOutput(image_type=expression.Image(size_as="Input0"), channels=1)
+        ]
+        self.category = ImageUtilityCategory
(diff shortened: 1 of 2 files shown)


backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,87 @@
...
+class ImageConvolveNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
+        self.inputs = [
+            ImageInput(),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
+            NumberInput("Padding", minimum=0, default=0),
+            NumberInput("Strides", minimum=0, default=1),
```
RunDevelopment, line 23: What is the purpose of stride? I don't see how it is useful.

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,87 @@
...
+@NodeFactory.register("chainner:image:image_convolve")
+class ImageConvolveNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
+        self.inputs = [
+            ImageInput(),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
+            NumberInput("Padding", minimum=0, default=0),
```
RunDevelopment, line 22: Is this parameter really necessary? I mean, we already have a node to do this (Create Border).

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,87 @@
...
+        padding: int,
+        strides: int,
+    ) -> np.ndarray:
+
+        kernel = np.array([float(d) for d in kernel_in.split(",")])
+        kernel = kernel.reshape(kernel_dim, kernel_dim)
+
+        # Grayscale image if it is not already
+        if len(img.shape) != 2:
+            img = img[:, :, 0]
```
RunDevelopment, line 47: I don't think that this is a good idea. Chainner stores color images as BGR internally, so you just get the B channel here.

Maybe we could just require grayscale images using [the `channels` parameter](https://github.com/chaiNNer-org/chaiNNer/blob/3fae92c05ffc94be5e9b84c0abcffa084bd6cff6/backend/src/nodes/properties/inputs/numpy_inputs.py#L24)? We could also use `as_target_channels(img, 1, narrowing=True)` to convert the image to grayscale in a way that is consistent with the rest of chainner.
Not sure what option is best here.

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,87 @@
...
+                int(padding) : int(-1 * padding), int(padding) : int(-1 * padding)
+            ] = img
+        else:
+            imagePadded = img
+
+        for y in range(img.shape[1]):
+            if y > img.shape[1] - yKernShape:
+                break
+            if y % strides == 0:
+                for x in range(img.shape[0]):
```
RunDevelopment, line 75: Implementing the convolution yourself isn't ideal for one reasons: Python is slow as heck. Chainner is used on images that with millions of pixels, and pixel-wise operations in Python code are just too slow for that. (There are also more efficient algorithms for convolution based on DTF, but they only really matter for larger kernel sizes starting at around 11x11.)

Please use [`cv2.filter2D`](https://docs.opencv.org/3.4/d4/d86/group__imgproc__filter.html#ga27c049795ce870216ddfb366086b5a04) instead. This function does the convolution for you and is implemented in highly optimized C code. This function is really fast (even for ridiculously large kernel sizes). 
There are good examples for its usage online, but we also use this function throughout chainner, so you can also find a few examples in our code base.

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,87 @@
...
+
+@NodeFactory.register("chainner:image:image_convolve")
+class ImageConvolveNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
+        self.inputs = [
+            ImageInput(),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
```
RunDevelopment, line 21: Regarding the kernel format: Maybe we could use a format that encodes the kernel dimensions as well as its weights? With a text area inputs, we could use a format like this:
```
1   0 -1
2.5 0 -2.5
1   0 -1
```
Newlines are used as row separators and spaces are used as column separators.

This format should be rather intuitive for new users, and it fully describes the kernel. The latter is very useful, because it will us to make a specialized kernel input in the future (e.g. as a popover as suggested on discord).

RunDevelopment (commented):
I really like the idea of adding a convolve node. This will make advanced users less reliant on the nodes we implement because they can just implement some image operations themselves.

Taking the idea even further, we could even add more kernel operations (e.g. transpose, slice, sum) in future PRs (not that you *have to* make those PRs @JumpyJacko, no worries 😄) and make the Math node compatible with kernels.

backend/src/nodes/nodes/image_utility/convolve.py
joeyballentine: It's purpose is that it's a way to shrink the image down

backend/src/nodes/nodes/image_utility/convolve.py
joeyballentine: Padding is something generally used at the same step as convolution. While normally I'd say let the other node do it, since this is standard for convolution i say just keep it.

joeyballentine: For reference for this and stride, check this out https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html

backend/src/nodes/nodes/image_utility/convolve.py
joeyballentine: I'm fine with this so long as we have a node that can create kernels without having to format things like this (or i guess using the popover would be fine like you said)

Generally though i think I'd prefer using separate nodes for these kinds of things though, that way other numbers can be connected to the node in order to create them

backend/src/nodes/nodes/image_utility/convolve.py
RunDevelopment: Oh, in that case, I'm fine with both padding and stride.

backend/src/nodes/nodes/image_utility/convolve.py
RunDevelopment: Yes, this string representation was mostly meant as an internal string format for kernels that is also usable right now. Eventually, we should have a dedicated kernel input that creates these strings.

>  other numbers can be connected to the node in order to create them

Do you really want to have an e.g. 3x3 kernel node with 9 number inputs? I would have said that we treat kernels as literals and then support math operations on them, so people can create more complex kernels from simpler ones. Well, I guess we don't have to iron out the specifics right now.

backend/src/nodes/nodes/image_utility/convolve.py
JumpyJacko: Oh, that's a great point, I forgot about the other nodes as I was first testing the convolve in a Python notebook. I think it'll be best to require grayscale images in this case, I'll implement it now.

backend/src/nodes/nodes/image_utility/convolve.py
JumpyJacko: Oh man, I can't believe I missed this function. I did realise that this 2D convolution was extremely slow on large images but couldn't find a solution somehow. I'll implement this as well.

backend/src/nodes/nodes/image_utility/convolve.py
JumpyJacko: Regarding the current string representation, there doesn't seem to be a way to have multiline string input. I have made a version of the kernel input to support this multiline input though which looks like this:
```python
kernel = np.stack([l.split() for l in kernel_in.splitlines()], axis=0).astype(float)
```
and works for inputs that look like this:
```yaml
integer only input:
1 0 -1
2 0 -2
1 0 -1

irregular input:
1   0  -1
2.5  0   -2.5
1    0  -1
```
But does not work in chaiNNer due to the lack of support for multiline text input. I'm not sure what to do from here.

RunDevelopment: > there doesn't seem to be a way to have multiline string input

We have [NoteTextAreaInput](https://github.com/chaiNNer-org/chaiNNer/blob/3fae92c05ffc94be5e9b84c0abcffa084bd6cff6/backend/src/nodes/properties/inputs/generic_inputs.py#L225). This input is only used in the Note nodes right now, so we might encounter a few bugs since it hasn't really been tested a lot. But this is actually a good thing, because we can then fix those bugs in later PRs. For now, let's just use that input, bugy or not, and fix anything that comes up later.

RunDevelopment:
2 things I want to talk about:

1. `cv2.filter2D` isn't quite a convolution, because it handles border pixels as well, so input size is always output size. Do we want to keep this behavior, or do we want to crop filter2D's output to get a clean convolution?
2. Since we went for grayscale image input -> grayscale image output, maybe we could generalize this to number of input channels == number of output channels. filter2D does the convolution for all channels anyway, so we could use that. Not sure whether this channel-wise convolution is easy to understand for users though.

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,70 @@
...
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
+        self.inputs = [
+            ImageInput("Grayscale Image", channels=1),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
+            NumberInput("Padding", minimum=0, default=0),
+            NumberInput("Strides", minimum=0, default=1),
+        ]
+        self.outputs = [
+            ImageOutput(image_type=expression.Image(size_as="Input0"), channels=1)
```
RunDevelopment, line 27: We have to account for padding and stride here to compute the correct output size. Assuming padding and stride as explained [here](https://adeshpande3.github.io/A-Beginner's-Guide-To-Understanding-Convolutional-Neural-Networks-Part-2/), the output type has to be this:

```
let w = Input0.width;
let h = Input0.height;

let kernel = Input1;

let padding = Input2;
let stride = Input3;

Image {
	width: int & floor((w + padding * 2 - 1) / stride + 1),
	height: int & floor((h + padding * 2 - 1) / stride + 1),
	channels: 1,
}
```

Sorry for the complex type expression. We have a whole custom language for [chainner's type system](https://github.com/chaiNNer-org/Navi/blob/main/navi.md) called Navi. This might be a bit overwhelming, so I'll keep it short. `let foo = value;` are "variables" (you can't reassign them), `InputX` are the types of the inputs starting with `Input0`, and the last `Image { ... }` is an image type with the specified dimensions.

If we choose to crop filter2D's output, we'll have to change the formula a bit. I'll help again then.

```suggestion
            ImageOutput(
                image_type="""
                    let w = Input0.width;
                    let h = Input0.height;

                    let kernel = Input1;

                    let padding = Input2;
                    let stride = Input3;

                    Image {
                        width: int & floor((w + padding * 2 - 1) / stride + 1),
                        height: int & floor((h + padding * 2 - 1) / stride + 1),
                        channels: 1,
                    }
                """,
                channels=1
            )
```

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,70 @@
...
+class ImageConvolveNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Convolves input image with input kernel (kernel values separated by commas)"
+        self.inputs = [
+            ImageInput("Grayscale Image", channels=1),
+            TextInput("Kernel String"),
+            NumberInput("Kernel Width/Height", minimum=0, default=3),
+            NumberInput("Padding", minimum=0, default=0),
+            NumberInput("Strides", minimum=0, default=1),
```
RunDevelopment, line 24: The minimum has to be 1.

```suggestion
            NumberInput("Strides", minimum=1, default=1),
```

RunDevelopment (changes requested):

JumpyJacko:
> 1. `cv2.filter2D` isn't quite a convolution, because it handles border pixels as well, so input size is always output size. Do we want to keep this behavior, or do we want to crop filter2D's output to get a clean convolution?

I'm not too sure what to do here, I think someone other than me should decide whether we want clean convolution or use OpenCV's filter2d which says that,
> "The function does actually compute correlation, not the convolution" - [OpenCV Docs](https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#ga27c049795ce870216ddfb366086b5a04)

And I'm not well versed in convolution as the most I know is from 3Blue1Brown's videos on linear algebra and convolutions (😅).

> 2. Since we went for grayscale image input -> grayscale image output, maybe we could generalize this to number of input channels == number of output channels. filter2D does the convolution for all channels anyway, so we could use that. Not sure whether this channel-wise convolution is easy to understand for users though.

I currently do not know how I would do this but I'll look around to see because this seems like a good idea instead of just having _only_ grayscale image convolution input -> output.

RunDevelopment:
> I think someone other than me should decide whether we want clean convolution or use OpenCV's filter2d which says that

@joeyballentine @theflyingzamboni 

(Another idea would be to have a dropdown that lets the user choose whether they want convolution or correlation. This might be confusing though.)

> the most I know is from 3Blue1Brown's videos on linear algebra and convolutions

Same lol

> I currently do not know how I would do this

`cv2.filter2D` already does channel-wise convolution: "`dst`: output image of the same size and the same number of channels as `src`." You just have to adjust your padding code to handle 3d images. You can use `cv2.copyMakeBorder` for that like this:
```py
cv2.copyMakeBorder(
        img,
        top=padding,
        left=padding,
        right=padding,
        bottom=padding,
        borderType=cv2.BORDER_CONSTANT,
        value=0,
    )
```

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,86 @@
...
+        self.name = "Convolve"
+        self.icon = "MdAutoFixHigh"
+        self.sub = "Miscellaneous"
+
+    def run(
+        self,
+        img: np.ndarray,
+        kernel_in: str,
+        padding: int,
+        strides: int,
```
RunDevelopment, line 55: Stride still has to be implemented. This can be done like this `output[::stride, ::stride, ...]` after the convolution.

That being said, do we really need stride? It's essentially just nearest neighbor downsampling. 
Also @joeyballentine

joeyballentine: I suppose we don't

JumpyJacko: I just removed the strides so it should be fine now.

JumpyJacko:
> I think someone other than me should decide whether we want clean convolution or use OpenCV's filter2d which says that

I think it might actually already be doing convolution (except for the handling border pixels part) because I realised that there's this bit of code which flips the kernel.
```python
kernel = np.flipud(np.fliplr(kernel)) # might be able to just do np.transpose(kernel)
```
I read in [this article](https://towardsdatascience.com/convolution-vs-correlation-af868b6b4fb5) that the difference with _correlation_ and _convolution_ is that the kernel for a correlation (what filter2d is doing right now) is
```
1 2 3
4 5 6
7 8 9
```
and by rotating 180 degrees (or transposing?) it into this
```
9 8 7
6 5 4
3 2 1
```
makes the correlation instead do a convolution. I might be wrong and we still have to decide what to do with the border pixels.

RunDevelopment:
You're right. Convolution is just correlation with the kernel rotated by 180°. Well, that makes things easier :)

Since you already do a 180° rotation before filter2D, it should cancel out the rotation done by filter2D's correlation.

>  (or transposing?)

Transposing is not the same as rotating. Transposition mirrors the kernel along the diagonal, so it's a 90° rotation + a flip.

> we still have to decide what to do with the border pixels

I think we could just leave it as is for now. Convolution producing an image with the same size as the input image is very useful when combining the convolved result with the input image.

backend/src/nodes/nodes/image_utility/convolve.py
```diff
@@ -0,0 +1,70 @@
...
+                    let w = Input0.width;
+                    let h = Input0.height;
+
+                    let kernel = Input1;
+
+                    let padding = Input2;
+
+                    Image {
+                        width: int & floor((w + padding * 2 - 1) + 1),
+                        height: int & floor((h + padding * 2 - 1)+ 1),
```
RunDevelopment, line 35: Since stride is gone, this can be simplified.

```suggestion
                        width: w + padding * 2,
                        height: h + padding * 2,
```

RunDevelopment (approved):
LGTM;

Thank you @JumpyJacko!

theflyingzamboni:
I agree on having input dims == output dims. I think even a lot of people advanced enough to use a node like this are likely just copying kernels they've seen somewhere else without knowing the fine details of convolution, and would be expecting the image to remain the same size.

joeyballentine:
LGTM, thanks for this

PR #1444, opened 2022-12-30
Added seed input and group

RunDevelopment:
I finally got around to adding a "Seed" input for Add Noise. I did this via a new group that just adds a little "Random seed" button to a number input. Adding the button was pretty tricky though. In the end, I did it via an optional prop in `SchemaInput`. Not an elegant solution, but it works.

image

### DIFF
--- a/backend/src/nodes/nodes/image_filter/add_noise.py
+++ b/backend/src/nodes/nodes/image_filter/add_noise.py
@@ -4,9 +4,9 @@ from enum import Enum
 import numpy as np
 
 from . import category as ImageFilterCategory
-from ...node_base import NodeBase
+from ...node_base import NodeBase, group
 from ...node_factory import NodeFactory
-from ...properties.inputs import ImageInput, SliderInput, EnumInput
+from ...properties.inputs import ImageInput, SliderInput, EnumInput, NumberInput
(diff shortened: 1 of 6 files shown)


joeyballentine:
I think the new seed icon would be better off as a shuffle icon than as a dice. I couldn't tell what it was at first

RunDevelopment:
What does a new seed icon look like? Like this
image
image

or like this
image

RunDevelopment:
I used a die because the button generates a random seed.

joeyballentine:
My suggestion was to use a shuffle icon, not a seed icon lol. Shuffle icons are commonly used for buttons that change some kind of random value, i think

RunDevelopment:
Shuffle icons are typically used when you, well, shuffle something, changing its order. This has nothing to do with generating a random seed. I would never expect to get a random number from pressing this icon:
image

I used dice because I saw it in Automatic1111's web UI.

That being said, I also see that the minimalist die icon might be hard to understand when you don't know what it does. 

Maybe as inspiration, Bitwarden uses this icon for generating new passwords:
image

RunDevelopment:
I think the refresh icon kinda works
image

joeyballentine (approved):

PR #1451, opened 2023-01-01
Make nodes 1px thinner

RunDevelopment:
This PR changes the minimum width of nodes to be 1px less. This makes their width divisible by 16, which makes nodes fit perfectly into the default 16x16 (snap to) grid.

I implemented this by moving the min width property into the element with the node's border. The border width is 0.5px, so all nodes are 1px thinner now.

image

image


joeyballentine:
It looks like you somehow made the handles a bit squished by doing this

RunDevelopment:
No, they are perfect circles in CSS. This might be another rendering bug like I recently had with checkboxes...
image

joeyballentine (approved):

PR #1458, opened 2023-01-07
Added Number and Text nodes

RunDevelopment:
This adds 2 nodes that simply output their given inputs. These nodes are very useful for defining constants used by multiple nodes across the chain.

I recently talked to someone that used Text Padding to define text constants, so there really has to be a solution.


joeyballentine:
Can't believe it took this long to add this, I've wanted this for so long but never actually did it

joeyballentine (approved):

PR #1468, opened 2023-01-10
Setup progress refactor

RunDevelopment:
Firstly, nomenclature:
- The setup process will refer to the process of starting chainner. This includes starting the backed and installing any deps that it might need.
- An owned backend is a backend process that was spawned by the main process of the app.
- A borrowed backend is a backend process the frontend connects to in --no-backend mode.

Changes:
- Removed unused `get-ffmpeg` ICP channel.
- Replaced all IPC channels going to the splash screen with a single `splash-setup-progress` channel.
- Added a new user-interface-independent way of dealing with user interaction. During the setup process, we might need to tell the user something via a dialog. This just isn't possible in CLI mode, so I had to find a user interaction model that work for both GUI and CLI. My solution to this are interrupts. The basic idea is that an interrupt... interrupts the program and only yields back control after the user has interacted with it. 
    Right now, there are 2 types of interrupts: critical error and warning. Critical error will always close the program after being interacted with, while warning will do nothing. More types may be added in the future. Other options for interaction can also be specified.
    Note that interrupts are a declarative format. The one requesting an interrupt is not the one actually performing the interrupt. This allows interrupts to work with both GUI and CLI. To switch from GUI dialogs to CLI prompts, we simply have to change the interrupt listener.
- Added a new progress API for the setup progress. This makes it easy to change how progress is displayed. The setup progress no longer requires the splash screen for displaying progress. The splash screen is now simply a listener to the progress emitted by the setup process. This makes it easy to swap out the splash screen for something else.
- The backend new has a `/python-info` JSON API call. This returns the python info of the current backend server. This python info is now used for borrowed backends.
- Most of the setup process in now in `main/backend/setup.ts`.
- There are 2 types of backend processes now: owned and borrowed. Owned processes are sub processes spawned by the app, while borrowed processes are independent processes simply connected (--no-backend). Both types have make up the `BackendProcess` interface, which makes it easy to abstract over how the backend was created.
- The backend new has a `/python-info` JSON API call. This returns the python info of the current backend server. This python info is now used for borrowed backends. This means that the correct python executable will be used for borrowed backend, which allows you to install/update deps from them with chainner's deps manager.
- Chainner will no longer check deps integrated python when started with `--no-backend`. It will simply connect to the borrowed backend and that's it.

Overall, I'm really happy with how this turned out. The new unified `BackendProcess` API made it easy to extract out the whole setup process into its own file, which made `main/main.ts` a lot simpler. The new progress API is also really simple and easy to use. The whole architecture around `main/main.ts` and the setup process feels a lot cleaner now.


src/renderer/splash.tsx
```diff
@@ -1,104 +1,81 @@
...
 import { ChaiNNerLogo } from './components/chaiNNerLogo';
 import { theme } from './splashTheme';
 
 const Splash = memo(() => {
     const { t } = useTranslation();
 
-    const [status, setStatus] = useState(t('splash.loading', 'Loading...'));
+    const getStatusText = useCallback(
+        (stage: SetupStage): string => {
+            switch (stage) {
```
joeyballentine, line 17: Instead of this gigantic switch, couldn't we just send the text data with the event?

joeyballentine: I see why you did it this way (so you could use the translation hook), but I think you can just use i18next for that in the main process

joeyballentine: I guess that makes sense, I'm just thinking about it as the frontend shouldn't (imo) care about how the backend is implemented. With this (and the current code) if i had a backend that did different steps, the frontend wouldn't be able to display that

RunDevelopment: > I'm just thinking about it as the frontend shouldn't (imo) care about how the backend is implemented.

Yeah, agreed. I'll move the switch into `main/`.

> if i had a backend that did different steps, the frontend wouldn't be able to display that

The backend also wouldn't compile because both the frontend and backend communicate to each other via a strictly type-checked API. If you want to add a new stage, TS will ensure that you add it everywhere.

joeyballentine (approved):

PR #1472, opened 2023-01-11
Added Apply LUT node

RunDevelopment:
image

### DIFF
--- a/backend/src/nodes/nodes/image_utility/lut.py
+++ b/backend/src/nodes/nodes/image_utility/lut.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 import numpy as np
-import cv2
 
 from . import category as ImageUtilityCategory
 from ...node_base import NodeBase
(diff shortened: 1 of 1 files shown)


RunDevelopment:
@theflyingzamboni I heard that `np.take` is way faster than indexing a LUT like I did here, but I could get this to work. Do you happen to know this?

RunDevelopment:
I change the implementation to always quantize the grayscale image to the size of the LUT. This allows users to achieve color quantization effects that would be hard to achieve otherwise.

image

If users want chainner to smoothly interpolate between LUT values, then they can do that by resizing the LUT themselves using our Resize nodes.

joeyballentine (approved):

joeyballentine:
LGTM but I'd like to wait until Zamboni responds to your question

theflyingzamboni:
> @theflyingzamboni I heard that `np.take` is way faster than indexing a LUT like I did here, but I couldn't get it to work. Do you happen to know this?

I've never come across that function before. I could try messing around with it and see if I can figure out how it works.

theflyingzamboni:
> Also, we should definitely add a node to create LUTs

Create RGB + Shift 😛

joeyballentine:
> > Also, we should definitely add a node to create LUTs
> 
> Create RGB + Shift 😛

256 of them?? Lmao

theflyingzamboni:
@RunDevelopment Okay, I got `np.take` working, and it is about twice as fast. The issue is LUT dimensionality. The way it is right now, the LUT has dimensions (1, w, c), and for `np.take` to work, it needs to be (w, c).

RunDevelopment:
> so `return np.take(lut, img, axis=0)`.

Doesn't work.

image

```
  File "C:\Users\micha\Git\chaiNNer\backend\src\process.py", line 349, in __process
    raw_output, execution_time = await self.loop.run_in_executor(
  File "C:\Python38\lib\concurrent\futures\thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\micha\Git\chaiNNer\backend\src\process.py", line 223, in wrapper
    result = supplier()
  File "C:\Users\micha\Git\chaiNNer\backend\src\nodes\nodes\image_utility\lut.py", line 68, in run
    return np.take(lut, img, axis=0)
  File "<__array_function__ internals>", line 5, in take
  File "C:\Python38\lib\site-packages\numpy\core\fromnumeric.py", line 190, in take
    return _wrapfunc(a, 'take', indices, axis=axis, out=out, mode=mode)
  File "C:\Python38\lib\site-packages\numpy\core\fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
IndexError: index 72 is out of bounds for axis 0 with size 1
```

RunDevelopment:
Argh, you know what? Let's just not use `np.take` right now. If the node is too slow, we can try to make it faster later.

theflyingzamboni:
@RunDevelopment You didn't correct the dimensions of the LUT.

> The way it is right now, the LUT has dimensions (1, w, c), and for np.take to work, it needs to be (w, c).

PR #1473, opened 2023-01-12
Update version of opencv-python to 4.7.0.68

joeyballentine:
We can finally support macos under 10.15 lol, assuming everything else works


RunDevelopment (approved):
Well, let's hope that it works.

PR #1481, opened 2023-01-13
Switched Apply LUT to np.take for speed-up

theflyingzamboni:


joeyballentine (approved):

RunDevelopment (approved):
Thank you @theflyingzamboni! It's about twice as fast in my tests.

PR #1484, opened 2023-01-14
Make zip builds more portable (python & ffmpeg)

joeyballentine:
This PR does three things:
- Moves the forge config out of package.json and into its own config file
- Uses a build hook to write a `portable` file to the zip, if the build is a zip
- Checks for the existence of this file when checking for python/ffmpeg, and if the file exists will use the current executable directory rather than appdata.

Note: Settings still save to appdata. I will accomplish that next.

### DIFF
--- a/src/main/backend/setup.ts
+++ b/src/main/backend/setup.ts
@@ -1,4 +1,3 @@
-import { app } from 'electron';
 import log from 'electron-log';
 import { t } from 'i18next';
 import path from 'path';
@@ -317,9 +316,6 @@ export const setupBackend = async (
     getRootDir: () => Promise<string>
 ): Promise<BackendProcess> => {
     token.submitProgress({ totalProgress: 0 });
-
-    const currentExecutableDir = path.dirname(app.getPath('exe'));
-    log.info(`currentExecutableDir setupBackend: ${currentExecutableDir}`);
 
     const backend = getArguments().noBackend
         ? await setupBorrowedBackend(token, 8000)
--- a/src/main/backend/setup.ts
+++ b/src/main/backend/setup.ts
@@ -12,7 +12,6 @@ import {
 import { runPipInstall, runPipList } from '../../common/pip';
 import { CriticalError } from '../../common/ui/error';
 import { ProgressToken } from '../../common/ui/progress';
-import { checkFileExists } from '../../common/util';
 import { versionGt } from '../../common/version';
 import { getArguments } from '../arguments';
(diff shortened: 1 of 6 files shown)


src/main/backend/setup.ts
```diff
@@ -267,17 +269,21 @@ const setupOwnedBackend = async (
     });
     const port = await getValidPort();
 
+    const currentExecutableDir = path.dirname(app.getAppPath());
+    const isPortable = await checkFileExists(path.join(currentExecutableDir, 'portable'));
+    const rootDir = isPortable ? currentExecutableDir : app.getPath('userData');
```
RunDevelopment, line 274: I think the logic for getting the root dir should be moved out of `backend/setup.ts` completely. It could be passed in via `getRootDir: () => Promise<string>` parameter, like `hasNvidia`.

RunDevelopment (changes requested):
The empty `portable` file seems like a hack to me, but it's fine for now, I guess.

joeyballentine:
> The empty `portable` file seems like a hack to me, but it's fine for now, I guess.

That's quite literally the only way i can possibly think of to do it

RunDevelopment:
I would probably make it a setting we don't show in the UI, but I honestly think that we should rethink how we store settings anyway. The whole "I'm shadowing built-in `localStorage`" doesn't smell like a best practice to me either.

So let's keep the `portable` file for now and think about settings as a totality later.

src/main/backend/setup.ts
```diff
@@ -315,13 +313,23 @@ export const setupBackend = async (
...
     useSystemPython: boolean,
     systemPythonLocation: string | undefined | null,
-    hasNvidia: () => Promise<boolean>
+    hasNvidia: () => Promise<boolean>,
+    getRootDir: () => Promise<string>
 ): Promise<BackendProcess> => {
     token.submitProgress({ totalProgress: 0 });
 
+    const currentExecutableDir = path.dirname(app.getPath('exe'));
+    log.info(`currentExecutableDir setupBackend: ${currentExecutableDir}`);
```
RunDevelopment, line 322: You might want to use the variable in `platform.ts`.

RunDevelopment (approved):
One comment. GTG otherwise. Feel free to merge after resolving the comment.

src/main/backend/setup.ts
joeyballentine: oops i forgot to remove that log

joeyballentine: both those lines were from debugging

PR #1485, opened 2023-01-14
Fix splash screen progress bars

joeyballentine:
the python and ffmpeg progress bars were being divided by 100 despite being already in 0-1 range


RunDevelopment:
AFAIK, python and ffmpeg progress was reported in 0-100 (as indicated by the variable name "percentage"), but the splash screen's [Progress component](https://chakra-ui.com/docs/components/progress) expects 0-100.

Closed in favor of #1486.

PR #1487, opened 2023-01-15
Make settings portable for zips

joeyballentine:
Unfortunately, I needed to make a synchronous version of the check. This makes me wonder if I should use that sync check for everything since I could get it once at the top of main.ts and pass it through to the other stuff from the previous PR.

Alternatively, we could switch to Node 18 and use top-level-await (though we'd need to wait for electron to update first)

### DIFF
--- a/src/main/backend/setup.ts
+++ b/src/main/backend/setup.ts
@@ -260,15 +260,13 @@ const setupOwnedBackend = async (
     useSystemPython: boolean,
     systemPythonLocation: string | undefined | null,
     hasNvidia: () => Promise<boolean>,
-    getRootDir: () => Promise<string>
+    rootDir: string
 ): Promise<OwnedBackendProcess> => {
     token.submitProgress({
         status: t('splash.checkingPort', 'Checking for available port...'),
         totalProgress: 0.1,
     });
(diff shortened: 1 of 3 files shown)


RunDevelopment:
> if I should use that sync check for everything

Yes. We already pay for the FS lookup in the sync version. No need to pay for it again later.

I would also suggest wrapping the sync versions with `lazy`. That way you can call them over and over again with perf cost.

---

There is one last use of `app.getPath('userData')` that you should look at: It's for the `'get-appdata'` IPC channel. Could you please verify that those still work as intended?

joeyballentine:
> Could you please verify that those still work as intended?

How would that have been affected by this?

joeyballentine:
Wait nvm I see what you mean, the frontend could be doing something that expects these things to be in the right spots

joeyballentine:
That's used for the tensorrt cache location and opening the logs folder, all of which I'll make separate PRs for to change their locations

RunDevelopment:
The non-sync functions aren't used anymore. Do you think we should remove them?

RunDevelopment (approved):

PR #1489, opened 2023-01-15
Added basic CLI mode

RunDevelopment:
This PR adds a very basic CLI mode. It can just about run a single complete chain. This PR does not include more fancy feature such as progress display and input overrides. Those will come later.

Changes:
- Moved several functions and functionalities from `renderer/` into `common/`.
- Moved everything that opens a window into `main/gui/`. This neatly groups everything that needs access to a GUI into on folder.
- Changed argument parsing. Argument parsing will now be pretty much the first thing chainner does, no matter CLI or GUI mode. I also documented the CLI. Here's what `chainner.exe --help` (or `npx electron-forge start -- --help`) emits:
    ```
    chainner [file]
    
    Open the chaiNNer GUI
    
    Commands:
      chainner open [file]  Open the chaiNNer GUI                          [default]
      chainner run <file>   Run the given chain in the command line without opening
                            the GUI
    
    Positionals:
      file  An optional chain to open. This should be a .chn file           [string]
    
    Options:
      --help     Show help                                                 [boolean]
      --version  Show version number                                       [boolean]
    ```
    Note that the `--no-backend` option is hidden here, because it's an internal option.
- CLI arguments now have to be passed as function parameters, no more `getArguments()`. This was necessary because arguments could be from a different command as far the types are concerned. While a bit more cumbersome, this also encourages us to write abstractions instead of using CLI args directly, which is good.
- Added `settingStorage`. No more `localStorage` overshadowing.
- I also took this change to better structure how IPC handles and other event listener are registered for the main window.


joeyballentine:
Code looks fine. Fix the merge conflicts and we'll be good to go

joeyballentine (approved):

PR #1499, opened 2023-01-23
add random number node

Zarxrax:
A node which generates a random integer within a user-defined range. A seed parameter can also be specified. A value of -1 for the seed will generate a completely random number every time, while a non-negative number will allow the random numbers to be reproduced on a later run. In that case, it can be useful to pass the frame index from an iterator into the seed value.

### DIFF
--- a/backend/src/nodes/nodes/utility/random_number.py
+++ b/backend/src/nodes/nodes/utility/random_number.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+from random import seed, randint
+from typing import Union
+
+from . import category as UtilityCategory
+
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import NumberInput
+from ...properties.outputs import NumberOutput
+
+
+@NodeFactory.register("chainner:utility:random_number")
+class RandomNumberNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Generate a random integer."
+        self.inputs = [
+            NumberInput(
+                "Minimum Value",
+                minimum=None,
+                maximum=None,
+                precision=100,
+                controls_step=1,
+            ),
+            NumberInput(
+                "Maximum Value",
+                minimum=None,
+                maximum=None,
+                precision=100,
+                controls_step=1,
+            ),
+            NumberInput(
+                "Seed",
+                minimum=-1,
+                maximum=None,
+                default=-1,
+                precision=100,
+                controls_step=1,
+            )
+        ]
+        self.outputs = [
+            NumberOutput("Result", output_type="number")
+        ]
+
+        self.category = UtilityCategory
+        self.name = "Random Number"
+        self.icon = "MdCalculate"
+        self.sub = "Math"
+
+    def run(self, a: int, b: int, c: int) -> int:
+        if c != -1:  #if seed value is -1, don't use a seed
+            seed(c)
+        result = randint(a,b)
+        return result
--- a/backend/src/nodes/nodes/utility/random_number.py
+++ b/backend/src/nodes/nodes/utility/random_number.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
-from random import seed, randint
-from typing import Union
-
+from random import Random, randint
 from . import category as UtilityCategory
 
-from ...node_base import NodeBase
+from ...node_base import NodeBase, group
 from ...node_factory import NodeFactory
 from ...properties.inputs import NumberInput
 from ...properties.outputs import NumberOutput
@@ -31,8 +29,9 @@ class RandomNumberNode(NodeBase):
             ),
         ]
         self.outputs = [
-            NumberOutput("Result", output_type="int & max(.., Input0) & min(.., Input1)")
-
+            NumberOutput(
+                "Result", output_type="int & max(.., Input0) & min(.., Input1)"
+            )
         ]
 
         self.category = UtilityCategory
@@ -44,4 +43,4 @@ class RandomNumberNode(NodeBase):
         if c == -1:  # random seed
             return randint(a, b)
         else:
-            return random.Random(c).randint(a, b)
+            return Random(c).randint(a, b)
--- a/backend/src/nodes/nodes/utility/random_number.py
+++ b/backend/src/nodes/nodes/utility/random_number.py
@@ -20,8 +20,6 @@ class RandomNumberNode(NodeBase):
                 "Minimum Value",
                 minimum=None,
                 maximum=None,
-                precision=100,
-                controls_step=1,
             ),
             NumberInput(
                 "Maximum Value",
--- a/backend/src/nodes/nodes/utility/random_number.py
+++ b/backend/src/nodes/nodes/utility/random_number.py
@@ -26,14 +26,9 @@ class RandomNumberNode(NodeBase):
                 minimum=None,
                 maximum=None,
             ),
-            NumberInput(
-                "Seed",
-                minimum=-1,
-                maximum=None,
-                default=-1,
-                precision=100,
-                controls_step=1,
-            )
+            group("seed")(
+                NumberInput("Seed", minimum=-1, maximum=None, default=-1),
+            ),
         ]
         self.outputs = [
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/utility/random_number.py
```diff
@@ -0,0 +1,55 @@
...
+        self.category = UtilityCategory
+        self.name = "Random Number"
+        self.icon = "MdCalculate"
+        self.sub = "Math"
+
+    def run(self, a: int, b: int, c: int) -> int:
+        if c != -1:  #if seed value is -1, don't use a seed
+            seed(c)
+        result = randint(a,b)
+        return result
```
RunDevelopment, line 55:
```suggestion
        if c == -1:  # random seed
            return randint(a, b)
        else:
            return random.Random(c).randint(a, b)
```

backend/src/nodes/nodes/utility/random_number.py
```diff
@@ -0,0 +1,55 @@
...
+                "Seed",
+                minimum=-1,
+                maximum=None,
+                default=-1,
+                precision=100,
+                controls_step=1,
+            )
+        ]
+        self.outputs = [
+            NumberOutput("Result", output_type="number")
```
RunDevelopment, line 43: We actually know quite a bit about the output type. It's not just any number, we know that it's an integer within the range given by the above inputs.

Don't worry about the below expression. It's a [Navi type expression](https://github.com/chaiNNer-org/Navi/blob/main/navi.md), but you can also treat it as a black box for now :)
```suggestion
            NumberOutput("Result", output_type="int & max(.., Input0) & min(.., Input1)")

```

RunDevelopment (changes requested):
Thank you for the PR @Zarxrax!

I left you a few suggestions to make the node better.

Also, could you please fix the formatting? We use black for formatting, so you just run the command `black ./backend`.

Zarxrax:
@RunDevelopment, thank you for the assistance. After some further testing, I have realized that this node is not functioning the way that I intended.
When used in an iterator and without specifying a seed, I am getting the same result for each iteration of the iterator. If I do not use an iterator, I will get a different number every time I run the chain. Does the run() function not get called for every iteration?

joeyballentine:
I believe that's due to the chain optimization it does. I think that means we need to build in some kind of exception to not cache this node's result

RunDevelopment:
Nodes are expected to be deterministic. Without this assumption, we wouldn't be able to cache the outputs of nodes and do computations efficiently.

So with a random seed, your node will create one random number and that random number will be cached and reused when the node is referenced again. This caching also happens inside iterators, unless the random number node is connected to an iterated value. E.g. if you connect the frame index or file index to the seed input, you'll get a new random number every time.

Because of this caching, I think it would be best to remove random seed for now. This will solve the issue by making the node deterministic. 

While not a perfect substitute, you can use the random seed button to generate new random seeds outside iterators instead.

Zarxrax:
So I believe I have come up with a solution to get around the caching, while also making it a bit more flexible and also more intuitive to the end user. I have added an additional input for the frame index, separate from the seed value. It now seems to work more like I originally intended.

RunDevelopment:
Your idea is great, but I don't like the execution. It currently has a few problems:
1. "Index from Iterator" is a number input users can also just manually assign numbers to

backend/src/nodes/nodes/utility/random_number.py
```diff
@@ -0,0 +1,54 @@
...
+        ]
+
+        self.category = UtilityCategory
+        self.name = "Random Number"
+        self.icon = "MdCalculate"
+        self.sub = "Math"
+
+    def run(self, minval: int, maxval: int, seedval: int, frameval: int) -> int:
+        seed((frameval + 1) * (seedval + 1))
+        return randint(minval, maxval)
```
RunDevelopment, line 54: Instead of setting the global seed (which is a side effect), please create a new RNG instead.
```suggestion
        return random.Random((frameval + 1) * (seedval + 1)).randint(minval, maxval)

```

backend/src/nodes/nodes/utility/random_number.py
```diff
@@ -0,0 +1,54 @@
...
+                    "Seed",
+                    minimum=0,
+                    maximum=None,
+                ),
+            ),
+            NumberInput(
+                "Index from Iterator",
+                minimum=0,
+                maximum=None,
+            ),
```
RunDevelopment, line 39: Using a number input here isn't ideal. Users aren't supposed to enter numbers into this input, they are supposed to connect it to something. We can enforce this by making this an optional generic input.

Please note that this means that `frameval` can now be `None`. Please make sure to denotes this in the type hint.

```suggestion
            BaseInput(input_type="uint", label="Index from Iterator", kind="generic").make_optional(),
```

RunDevelopment (commented):
Your idea is great. Using the iterator index as a source of entropy to derive a unique seed is a perfect fit for chainner's deterministic execution model.

However, we can improve this a little:

backend/src/nodes/nodes/utility/random_number.py
Zarxrax: > Using a number input here isn't ideal. Users aren't supposed to enter numbers into this input, they are supposed to connect it to something. We can enforce this by making this an optional generic input.
> 
> Please note that this means that `frameval` can now be `None`. Please make sure to denotes this in the type hint.

Thank you for the suggestions. I have updated it per your recommendation, and in the case of the 'None' for the frameval I change it to 0.

The only other thing I wonder now is would this node be better categorized under the "Value" subheading rather than "Math"? I just now noticed the "Number" node and thought perhaps it would be better to group them together.

joeyballentine:
I'm curious why that input is needed when you can just connect the index to the seed input? what do we actually gain from that?

Zarxrax:
> I'm curious why that input is needed when you can just connect the index to the seed input? what do we actually gain from that?

So, if you just connect the index directly to the seed input, you essentially just get a fixed sequence of numbers every time. That's not particularly random anymore, right? This makes it impossible to do things like have multiple random numbers in the same chain. For example, what if you wanted to get a random x,y position value? You have 2 random number nodes connected to the frame index, and they are both giving you the same values on every frame. Also, the user may just not like the particular sequence of numbers that they obtain, so being able to change the seed independently gives them the ability to find a sequence that works better for their needs.
Also, I think its not terribly intuitive to just know that you would need to connect the index into the seed in order to get a different random number on every frame. There would probably be a lot of people asking why they keep getting the same number on every frame.

RunDevelopment:
@joeyballentine One way you can think about the additional input is to see it as a source of entropy. Essentially, the node implements not only the random number generation but also a seed derivation.

@Zarxrax On that note: Maybe it would make sense to move the seed derivation into its own node? This would allow us to reuse it for other nodes as well (e.g. the SD nodes from #1500 or the Add noise node). 

The "Derive Seed" node would have a seed number input plus a few generic inputs to connect sources of entropy to (e.g. the iteration index). The node would then output a seed (basically a random number). Users can then connect the derived seed to the seed input of whatever node needs a seed. 

Moving the seed derivation into a separate node would also justify using a more complex but better function to combine the given sources of entropy with the initial seed. The easiest way would probably be to just use a hash function like SHA. This would also free us up to support more types of data than just integers (e.g. text). 

That being said, I won't make this a requirement for this PR. We can still add this in followup PRs. It would be nice to know what you think about it though :)

backend/src/nodes/nodes/utility/random_number.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
+from typing import Union
+from random import randint, Random
```
RunDevelopment, line 3: Lint fails because `randint` is an unused import.
```suggestion
from random import Random

```

Zarxrax:
@RunDevelopment I'm not sure if I have any particular opinion on that idea at the moment. It sounds like it might make things more complex, but at the same time provide more power and flexibility, so I guess that's a trade off.

RunDevelopment (approved):

joeyballentine:
I like the derive seed idea. We should definitely do that

PR #1500, opened 2023-01-23
Stable Diffusion via REST

adodge:
Hi!  Thanks for this project.  It's exactly the sort of thing I was looking for.  I made some nodes to hook into Automatic1111.

To get it running, you have to set up the service from here:  https://github.com/AUTOMATIC1111/stable-diffusion-webui and run it with `./webui.sh --api` to expose the REST API.  Here are some examples of it working:

Screenshot from 2023-01-23 11-29-23
Screenshot from 2023-01-23 11-44-06


joeyballentine:
This is really cool! Thanks!

I'm not too sure how i feel about relying on a separate server to run this stuff though. It makes sense in that it's a lot easier to not have to integrate everything directly, but i feel like it has the potential to cause lots of issues and incompatibility given enough time. For example if automatic's SD repo updates and makes breaking changes, people would need to use an old version for this to work properly.

This would however be the perfect use case for a plugin once we have the plugin system implemented, but I'll defer to @rundevelopment on this one to see what he thinks.

joeyballentine:
Skimmed through the code. I like that you put this in a separate REST category to make it clear that it's interacting with external things and not anything built into chaiNNer.

If you don't have SD running, do these nodes still show up? What happens in that case?

adodge:
Thanks for taking a look!  Yes, all good points.

The nodes do show up, but obviously they report an error if you try to run them.  I could have it try to connect to the API when the nodes are imported.

joeyballentine:
We're currently talking about this, and we feel the benefits of adding this most likely outweigh any drawbacks. Are you in our Discord server? I think it would be easier to discuss this with you there, if you don't mind

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+
+@NodeFactory.register("chainner:rest:sd_img2img")
+class Img2Img(AsyncNodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Modify an image using an external Stable Diffusion service"
+        self.inputs = [
+            ImageInput(),
+            TextInput("Prompt", default="an astronaut riding a horse"),
+            TextInput("Negative Prompt", default=""),
```
RunDevelopment, line 32: This default value is invalid. Text inputs enforce a min length of 1 by default. If you want any input to be optional, use its `make_optional()` method. E.g. `TextInput("Negative Prompt").make_optional()`. If the user doesn't assign a value to an optional input, then the `run` of the node will get `None` for this parameter (be sure to denote this in the type hint of the parameter).

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+            NumberInput(
+                "Denoising Strength",
+                minimum=0,
+                default=0.75,
+                maximum=1,
+                controls_step=0.1,
+                precision=4,
+            ),
```
RunDevelopment, line 40: Generally, sliders are nicer for users because they visualize not only the current value but also the relation of the current value to the minimum and maximum.

image
```suggestion
            SliderInput(
                "Denoising Strength",
                minimum=0,
                default=0.75,
                maximum=1,
                slider_step=0.01,
                controls_step=0.1,
                precision=4,
            ),
```

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+            NumberInput("Seed", minimum=0, default=42, maximum=4294967296),
```
RunDevelopment, line 41: We have a group (=wrappers around inputs that add functionality) for seeds. It adds a button that when pressed generates a random 6-digit number.

image

(The `group` function is imported from `node_base`.)

```suggestion
            group("seed")(
                NumberInput("Seed", minimum=0, default=42, maximum=4294967296)
            ),
```

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+            NumberInput("Steps", minimum=1, default=20, maximum=150),
+            EnumInput(SamplerName, default_value=SamplerName.EULER),
+            NumberInput(
+                "CFG Scale",
```
RunDevelopment, line 45: Could Steps and CFG Scale also be slider inputs? The maximum values seem kinda weird, so I don't know if it makes sense.

Also, we also support sliders with log scales. So if you need that, we have it :)

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+                minimum=1,
+                default=7,
+                maximum=20,
+                precision=4,
+                controls_step=0.1,
+            ),
+            NumberInput("Width", minimum=64, default=512, maximum=2048),
+            NumberInput("Height", minimum=64, default=512, maximum=2048),
```
RunDevelopment, line 53: These are the output width and height of the image below, right? Does the REST service guarantee that it will return images with example those dimensions?

Can these be any number between 64 and 2048 or are only powers of 2 allowed? I remember playing with an SD upscaler, and it only correctly support multiples of 64.

backend/src/nodes/nodes/rest/img2img.py
```diff
@@ -0,0 +1,99 @@
...
+            TextInput("Model Checkpoint Override", default=""),
```
RunDevelopment, line 54: This should probably be an optional input too.

backend/src/nodes/impl/rest.py
```diff
@@ -0,0 +1,95 @@
...
+async def get_async(url) -> Dict:
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url) as response:
+            return await response.json()
+
+
+async def post_async(url, json_data: Dict) -> Dict:
+    async with aiohttp.ClientSession() as session:
+        async with session.post(url, json=json_data) as response:
+            return await response.json()
```
RunDevelopment, line 41: Do these methods have to be async? The only reason you had to add `AsyncNodeBase` and everything around it is because of these methods.

We run nodes in a separate thread pool, so it should be fine to use a sync blocking API, I believe. I'm not a python expert, so I could be wrong here.

backend/src/nodes/impl/rest.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+STABLE_DIFFUSION_OPTIONS = asyncio.run(get_async(STABLE_DIFFUSION_OPTIONS_URL))
```
RunDevelopment, line 44: This doesn't seem to be used anywhere. What's its purpose?

RunDevelopment (changes requested):
Thank you for this @adodge! Great work!

I left you a few comments with questions and suggestions.

backend/src/nodes/nodes/rest/__init__.py
```diff
@@ -0,0 +1,8 @@
+from ...category import Category
+
+category = Category(
+    name="REST",
+    description="Interact with external components via REST",
```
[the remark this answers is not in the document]
joeyballentine: "Stable Diffusion (External)" maybe?

backend/src/nodes/impl/rest.py
```diff
@@ -0,0 +1,95 @@
...
+import numpy as np
+import os
+from PIL import Image
+from enum import Enum
+from typing import Dict, Union
+
+from .image_utils import normalize
+from ..utils.utils import get_h_w_c
+
+STABLE_DIFFUSION_HOST = os.environ.get("STABLE_DIFFUSION_HOST", "127.0.0.1")
```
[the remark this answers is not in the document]
adodge: Good point.  How about `external_stable_diffusion.py`?

backend/src/nodes/nodes/rest/img2img.py
adodge: Ah!  Okay, thank you.

backend/src/nodes/nodes/rest/img2img.py
adodge: Cool!  That is a lot nicer

backend/src/nodes/nodes/rest/img2img.py
adodge: They are sliders in the Automatic1111 UI, so it would make sense here too.  The upper bounds are, I think, arbitrary but reasonable, and the same as what are in the other UI.

backend/src/nodes/nodes/rest/img2img.py
adodge: Yes.  I don't know if it explicitly guarantees the output size, but it does seem to either produce output of the given size or crash.  It looks like it works with any integer values, though I believe models work best on the dimensions they were trained on.  (Usually 512x512 or 768x768.)

In the Automatic UI, these are sliders from 64 to 2048.  I used another UI that had a dropdown menu for these with various canonical values.

Imitating the UI decisions of the thing we're calling seems like a good baseline.

backend/src/nodes/impl/rest.py
adodge: I'll investigate...  It seemed necessary at the time, but maybe I didn't have the full context.

backend/src/nodes/impl/rest.py
adodge: Currently it's purpose is to make a call to the API at import time that will fail if it's not available, causing the nodes to be hidden.  I'll make a comment.

backend/src/nodes/impl/rest.py
adodge: Oh yeah.  Absolutely unnecessary.  Works fine with `requests` instead of `aiohttp`.

backend/src/nodes/nodes/rest/__init__.py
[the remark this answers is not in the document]
adodge: I do like the idea of an external category with the specific service as a subcategory.  We could call it "Automatic1111" instead of "Stable Diffusion", just to make that relationship more explicit.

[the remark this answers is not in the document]
joeyballentine: Or maybe "Automatic1111 Stable Diffusion"? I feel like SD should be referenced somewhere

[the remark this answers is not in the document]
adodge: How about: category "Stable Diffusion (External)" subcategory "Automatic1111"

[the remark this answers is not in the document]
adodge: I guess a question is whether there's a connection between categories and dependencies.  Right now, it looks like it's 1-to-1.  Installing PyTorch unlocks the entire PyTorch category.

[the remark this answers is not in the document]
joeyballentine: Yeah I'd say go with that. Makes enough sense.

adodge:
Implemented the feedback given.  Here are some new screenshots:

image
image

joeyballentine:
The results of this all still get saved by automatic's repo right? Like there's no chance someone would generate something cool but forget to save it and never be able to get it back?

adodge:
Weirdly it doesn't save, and I don't see an option in the API to do so.  Would be good to have an easy way to do something similar on the ChaiNNer side.  Maybe a node that outputs the current time as text, that could be used with the save node.

There's also the problem of saving metadata.  The API does give a JSON blob with metadata, so that could be an output from the SD node, and we could have a "Save Text" node.

adodge:
Here's a possible implementation of the saving:

image
image

It's kind of a lot of spaghetti to do something basic.  Maybe there could be one node that does it all.

requirements.txt
```diff
@@ -17,6 +17,7 @@ pylint
 pyright==1.1.280
 pytest
 pywin32 ; sys_platform=="win32"
+requests
```
RunDevelopment, line 20: We also need to declare this dependency in `dependencies.ts` (the file chainner uses to manage python deps inside the GUI). 

@joeyballentine Should this be a required dep or an optional dep? `requests` a pretty commonly used (and small) library, so I wouldn't mind making it a required dep.

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
```diff
@@ -0,0 +1,102 @@
...
+            SliderInput("Width", minimum=64, default=512, maximum=2048),
+            SliderInput("Height", minimum=64, default=512, maximum=2048),
+            TextInput("Model Checkpoint Override").make_optional(),
+        ]
+        self.outputs = [
+            LargeImageOutput(),
+        ]
+
+        self.category = ExternalStableDiffusionCategory
+        self.name = "Image-to-Image"
```
RunDevelopment, line 66: Could we call this "Image To Image"? We don't use kebab case anywhere else. Same for text to image.

backend/src/nodes/impl/external_stable_diffusion.py
```diff
@@ -0,0 +1,96 @@
...
+
+def post(url, json_data: Dict) -> Dict:
+    response = requests.post(url, json=json_data)
+    return response.json()
+
+
+# Call the API at import time
+# If this fails (because the API isn't up) it will raise an exception and the
+# nodes importing this file won't be registered
+STABLE_DIFFUSION_OPTIONS = get(STABLE_DIFFUSION_OPTIONS_URL)
```
RunDevelopment, line 45: I wonder if this is a good idea. When I initially checked out this PR, the SD nodes didn't even show up because it couldn't connect to the options URL for whatever reason. I'm also not sure whether connecting to this URL once is a reliable indicator for the availability of the service. The nodes have to handle network errors anyway, since the device might disconnect for the network (laptop).

Hiding nodes based on factors outside the user's machine also presents usability issues. Imagine a user creating a chain with SD yesterday. Today the SD service has maintenance and so it refuses connections. Suddenly, the chain the user create but yesterday fails to load as chainner doesn't recognize the SD nodes anymore.

Yeah, I think having an error explaining that the service is unavailable at the moment would be preferable.

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
```diff
@@ -0,0 +1,102 @@
...
+                maximum=20,
+                controls_step=0.1,
+                precision=1,
+            ),
+            SliderInput("Width", minimum=64, default=512, maximum=2048),
+            SliderInput("Height", minimum=64, default=512, maximum=2048),
+            TextInput("Model Checkpoint Override").make_optional(),
+        ]
+        self.outputs = [
+            LargeImageOutput(),
```
RunDevelopment, line 62: `LargeImageOutput` implicitly have side effects (the image is shown to the user). Please use a regular `ImageOutput`. Same for text to image.

requirements.txt
joeyballentine: I'm fine with it being required

backend/src/nodes/impl/external_stable_diffusion.py
joeyballentine: It's not outside the user's machine though, this is a local server it's connecting to.

And my biggest thing was just not confusing people. We hide pytorch nodes if pytorch isn't installed, i thought this should be similar.

RunDevelopment: Oh, you're right. I didn't even look at the host address and just assumed it was some remote server.

adodge:
Made a few more changes in response to feedback:

- Added requests library to requirements.ts
- Removed kebab case in favor of "Text to Image"
- Changed LargeImageOutput to ImageOutput

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
```diff
@@ -0,0 +1,102 @@
...
+                maximum=20,
+                controls_step=0.1,
+                precision=1,
+            ),
+            SliderInput("Width", minimum=64, default=512, maximum=2048),
+            SliderInput("Height", minimum=64, default=512, maximum=2048),
+            TextInput("Model Checkpoint Override").make_optional(),
+        ]
+        self.outputs = [
+            ImageOutput(),
```
RunDevelopment, line 62: We talked about the Width and Height inputs before and how the REST service seems to guarantee that the returned images have those dimensions. If we actually guarantee that the returned images have those dimensions (e.g. by checking and throwing an error if they aren't the right size), then we can denote this in the output type here.

Chainner has a custom type system that allows us to verify chains and to predict output types. E.g. when you see chainner knowing the exact size of all output images in [this screenshot](https://github.com/chaiNNer-org/chaiNNer/blob/main/src/public/screenshot.png), that's the type system at work. 

Assuming that the Width and Height inputs are correct, we can denotes in the `ImageOutput` like this:
```py
ImageOutput(image_type="Image { width: Input8, height: Input9 }")
```

`Input8` and `Input9` are the Width and Height inputs. All inputs are assigned an id based on their position (starting with 0) and those to inputs have the ids 8 and 9 respectively.

But we can potentially do even better. Do we also know the number of channels of the returned image? E.g. if given a grayscale, RGB, or RGBA image, what type of image will be returned? If we do, then we can denote this in the type as well.

RunDevelopment (commented):
One last thing for the Img2Img and Tex2Img nodes.

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
adodge: Very cool!

It looks like `Image to Image` will always return an RGB image of the size you specify, even if it's given a greyscale or RGBA input, or an input with a different size.  `Text to Image` is the same.

I added these notes to the ImageOutputs.

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
```diff
@@ -0,0 +1,104 @@
...
+                controls_step=0.1,
+                precision=1,
+            ),
+            SliderInput("Width", minimum=64, default=512, maximum=2048).with_id(8),
+            SliderInput("Height", minimum=64, default=512, maximum=2048).with_id(9),
+            TextInput("Model Checkpoint Override").make_optional(),
+        ]
+        self.outputs = [
+            ImageOutput(
+                image_type="Image {width: Input8, height: Input9, channels: 3}"
```
RunDevelopment, line 63: Using `channels: x` inside the type expression does work, but we have something better. `ImageOutput` also has a [`channels` parameter](https://github.com/chaiNNer-org/chaiNNer/blob/main/backend/src/nodes/properties/outputs/numpy_outputs.py#L42). This parameter not only adds the channel number to the type, it also enforces it at runtime. So if the API returns something that is not an RGB image, the user will see an error saying that the node is implemented incorrectly. This is important because chainner uses type information to verify the correctness of chains, so the type information has to be correct.

Same for text to image.

```suggestion
                image_type="Image { width: Input8, height: Input9 }",
                channels=3,
```

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
```diff
@@ -0,0 +1,104 @@
...
+            "width": width,
+            "height": height,
+            "override_settings": {},
+        }
+        if sd_model_checkpoint:
+            request_data["override_settings"][
+                "sd_model_checkpoint"
+            ] = sd_model_checkpoint
+        response = post(url=STABLE_DIFFUSION_IMG2IMG_URL, json_data=request_data)
+        return decode_base64_image(response["images"][0])
```
RunDevelopment, line 104: As with channels, we should also have a runtime check here to verify the size of the returned image.

Same for text to image.
```suggestion
        result = decode_base64_image(response["images"][0])
        h, w, _ = get_h_w_c(result)
        assert (w, h) == (width, height), f"Expected the returned image to be {width}x{height}px but found {w}x{h}px instead"
        return result
```

RunDevelopment (changes requested):
Sorry about this. Last last last change request.

backend/src/nodes/nodes/external_stable_diffusion/img2img.py
adodge: Cool!  Yes, makes sense.

> Sorry about this.

No worries!  I appreciate you taking the time to review it.

adodge: Actually, this has raised a weird issue that I missed.  The API does not in fact always return an image of the size you request.  I thought I'd tested this, but I guess I happened to only try multiple of 8, because actually it seems to round down.

If you ask for 513x513, it will return 512x512.  If you ask for 511x511 it will return 504x504.

I added `control_step` and `slider_step` arguments to the SliderInputs, but it doesn't stop someone from manually setting an invalid size.

https://github.com/chaiNNer-org/chaiNNer/blob/059bb95a1fef73b30935a25dd8624dcc360fa6cf/backend/src/nodes/nodes/external_stable_diffusion/img2img.py#L60-L67

For now I added this runtime error, but I think it'd be better if we could predict the output size instead.  Is it possible, I wonder, to specify this "round down" behavior in the image type of the ImageOutput?  I get lost when I try to find out where things are happening in the typescript code.

https://github.com/chaiNNer-org/chaiNNer/blob/059bb95a1fef73b30935a25dd8624dcc360fa6cf/backend/src/nodes/nodes/external_stable_diffusion/img2img.py#L101-L104

RunDevelopment: > For now I added this runtime error, but I think it'd be better if we could predict the output size instead. Is it possible, I wonder, to specify this "round down" behavior in the image type of the ImageOutput? I get lost when I try to find out where things are happening in the typescript code.

Navi docs are [here](https://github.com/chaiNNer-org/Navi/blob/main/navi.md#built-in-functions), but it seems like I forgot to document the `floor` and `ceil` functions. I'll fix that soon.
And yes, there is a lot of TS code for the type system. It's a whole little programming language :)

Anyway, yes, we can adjust the output type to do rounding to get rid of the runtime error. Since we have a `floor` function, we can do `int & floor(InputX / 8) * 8` to round down to the nearest multiple of 8.

If you want to, you can even extract this into a function:
```
def nearest_valid(n: number) = int & floor(n / 8) * 8;
Image { 
    width: nearest_valid(InputX),
    height: nearest_valid(InputY)
}
```

adodge: Oh, awesome!  I didn't realize there was documentation!  Thank you.  I've put that in and it seems to be working.

RunDevelopment (approved):

adodge:
@joeyballentine Oops.  Done!

joeyballentine (approved):
Awesome. Thanks again for this

PR #1502, opened 2023-01-25
Dithering Node

adodge:
Applies a Bayer ordered dithering filter.  (Maybe a little off-mission for an upscaling-oriented project, but a neat and maybe useful effect.)

image
ford_dithered2


backend/src/nodes/nodes/image_adjustment/dither.py
```diff
@@ -0,0 +1,58 @@
...
+@NodeFactory.register("chainner:image:dither")
+class DitherNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply a Bayer ordered dithering algorithm."
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Colors per channel", minimum=2, default=4),
+            DropDownInput(
+                input_type=get_number_type(0, 16, 0),
```
RunDevelopment, line 24: That function wasn't meant to be exported by the module. Sorry about that.

And this input type isn't meant to be a number type. It's meant to be a distinct type that is incompatible with any other type. But this distinct type has to be declared in `chainner-scope.ts`. Frankly, I wouldn't recommend doing it that way. I would recommend using an `EnumInput` + `option_labels` ([example](https://github.com/chaiNNer-org/chaiNNer/blob/4f4caaca5f3f49769e53620617eb92fff91d0e90/backend/src/nodes/nodes/image_dimension/resize_to_side.py#L106)) because `EnumInput` automatically handles typing for you.

theflyingzamboni:
Dithering is actually something I've been wanting to add at some point, so I don't consider it off-mission. I've seen multiple people who want to redither images after upscaling them.

To that end, would you be willing to expand the scope of this node? Ideally, I'd like a dithering node to have Bayer, Floyd-Steinberg, and Riemersma options (as can be found in ImageMagick), but even just FS would be a good addition. 

Nodes allow for inputs to be displayed/hidden depending on the value of a dropdown input, so if you have a Method dropdown, you could display only the inputs needed for each type, if they differ.

joeyballentine:
This makes me think we should also have a way of extracting color palettes, so you'd be able to dither with specific colors? That might be out of the scope of this PR though

backend/src/nodes/impl/dithering.py
```diff
@@ -0,0 +1,220 @@
...
+                148,
+                85,
+            ],
+        ]
+    ),
+}
+
+
+# Copied from onnx/np_tensor_utils.py
+MAX_VALUES_BY_DTYPE = {
```
joeyballentine, line 150: This should probably be put into a common helper file that can be imported in both places, rather than being copy/pasted

backend/src/nodes/nodes/image_adjustment/dither.py
```diff
@@ -0,0 +1,136 @@
...
+
+UNIFORM_DITHER_ALGORITHM_LABELS = {
+    UniformDitherAlgorithm.NONE: "No dithering",
+    UniformDitherAlgorithm.ORDERED: "Ordered Dithering",
+    UniformDitherAlgorithm.DIFFUSION: "Error Diffusion",
+}
+
+
+@NodeFactory.register("chainner:image:dither")
+class OitherNode(NodeBase):
```
RunDevelopment, line 32: Oither?

backend/src/nodes/impl/dithering.py
```diff
@@ -0,0 +1,427 @@
...
+    return apply_to_all_channels(one_channel_ordered_dither,
+                                 image, threshold_map=threshold_map, num_colors=num_colors)
+
+
+def error_diffusion_dither(image: np.ndarray, error_diffusion_map: ErrorDiffusionMap, num_colors: int) -> np.ndarray:
+    return apply_to_all_channels(one_channel_error_diffusion,
+                                 image, num_colors=num_colors, error_diffusion_map=error_diffusion_map)
+
+
+# TODO handle palettes
```
RunDevelopment, line 423: That's a lot of code, and it seems like more will come. Maybe we could make one file per dithering method?

backend/src/nodes/nodes/image_adjustment/dither.py
adodge: Oops.  Wow what an improbable typo.

backend/src/nodes/impl/dithering.py
adodge: Yes, good idea.  I'll break it up

adodge:
Okay, I think this is ready for review.  It supports paletted and uniform quantization with error diffusion (e.g. Floyd-Steinberg) and Riemersma algorithms, and uniform quantization with ordered dithering (e.g. Bayer).  It also supports quantization without dithering, and [some other kinds of error diffusion](https://tannerhelland.com/2012/12/28/dithering-eleven-algorithms-source-code.html), mostly because they were easy to add once I'd done Floyd-Steinberg.

There's also a node for extracting a color palette from an image, using either k-means or median cut methods.

Here's a chain demoing the different options:

[dither_demo.chn.txt](https://github.com/chaiNNer-org/chaiNNer/files/10529623/dither_demo.chn.txt)

Screenshot from 2023-01-29 03-08-54

ford_dither_demo

(I expect that the methods that don't look good in this example probably have other applications where they do better.)

backend/src/nodes/nodes/image_channel/palette_extraction.py
```diff
@@ -0,0 +1,81 @@
...
+                },
+            )(
+                NumberInput(
+                    "Palette Size",
+                    minimum=2,
+                    default=8,
+                ).with_id(2),
+            ),
+        ]
+        self.outputs = [ImageOutput(image_type=expression.Image(channels_as="Input0"))]
```
RunDevelopment, line 64: I think we can choose a better output type here. Firstly, we know that the returned palette will have a height of 1px. We can't know the width for the "all" method, but for the others it should the palette size, right?
```suggestion
        self.outputs = [
            ImageOutput(
                image_type=expression.Image(
                    width="""
                        match Input1 {
                            PaletteExtractionMethod::All => int(1..),
                            _ => Input2
                        }
                    """, 
                    height=1, 
                    channels_as="Input0"
                )
            ),
        ]
```

backend/src/nodes/nodes/image_filter/dither_palette.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@NodeFactory.register("chainner:image:palette_dither")
+class PaletteDitherNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply one of a variety of dithering algorithms using colors from a given palette. (Only the top row of pixels (y=0) of the palette will be used.)"
+        self.inputs = [
+            ImageInput(),
+            ImageInput(label="LUT", image_type=expression.Image(channels_as="Input0")),
```
RunDevelopment, line 39: This should probably be called "Palette" instead.

backend/src/nodes/impl/dithering/constants.py
```diff
@@ -0,0 +1,209 @@
...
+                int,
+                """
+         0  8  2 10
+        12  4 14  6
+         3 11  1  9
+        15  7 13  5
+        """.strip().split(),
+            )
+        )
+    ).reshape((4, 4)),
```
RunDevelopment, line 166: You did this to get around black messing up the formatting, right? I think a better solution might be to either turn off formatting (`# fmt: off`) for those lines or to use [some of the tricks in this post](https://stackoverflow.com/questions/58584413/black-formatter-ignore-specific-multi-line-code)

RunDevelopment (commented):
Awesome work! Thank you @adodge!

I just have a few minor suggestions.

backend/src/nodes/nodes/image_channel/palette_extraction.py
adodge: Nice!

backend/src/nodes/nodes/image_filter/dither_palette.py
adodge: Okay.  I was trying to have some consistency with the "Apply LUT" node, but I guess a LUT and a Palette are actually different things, even if they're both 1 row images.

backend/src/nodes/impl/dithering/constants.py
adodge: Nice.  Yes, this looks a lot better.

backend/src/nodes/nodes/image_filter/dither_palette.py
RunDevelopment: They are both look up tables, but I think that have slightly different purposes. The LUTs for Apply LUT are essentially gradient maps, so the order of colors matters. But with palettes, the order of colors doesn't matter, right?

adodge: Yeah, that's a good distinction.

RunDevelopment:
@adodge Could you please fix the CI (our linter don't like some of your code)? We can merge this once the CI passes.

adodge:
@RunDevelopment Should be fixed now!

adodge:
So, there are still formatting complaints, but they're coming from files that were changed in the `main` branch.

joeyballentine:
Yeah i think something got screwed up when i merged that last PR. I'll fix that

backend/src/nodes/nodes/image_channel/palette_extraction.py
```diff
@@ -0,0 +1,106 @@
...
+                ).with_id(2),
+            ),
+        ]
+        self.outputs = [
+            ImageOutput(
+                image_type=expression.Image(
+                    width="""
+                match Input1 {
+                    PaletteExtractionMethod::All => int(1..),
+                    _ => Input2
```
RunDevelopment, line 72: The palette size fix you recently pushed invalidates this type. If the images has, let's say, 3 distinct colors and the user set a palette size of 4, then your fix will return a 3x1px image, right?

adodge: It should still be 4x1, since I'm padding it:

https://github.com/chaiNNer-org/chaiNNer/blob/eb457466cc68a7d6777ba9a10b3d976ca452ae54/backend/src/nodes/nodes/image_channel/palette_extraction.py#L97-L99

adodge: This test suggests it's working as intended:  

extracting 10 colors from a 8x1 image:  image

10x1 output:  image

adodge: Sorry to sneak that in after it's already reviewed.  I'll open another PR next time.

RunDevelopment: > It should still be 4x1, since I'm padding it

Ah, I didn't see that part. My bad.

> Sorry to sneak that in after it's already reviewed. I'll open another PR next time.

No, don't worry about it. GH automatically dismisses my review when you push new commits, so I would have reviewed them anyway.

adodge:
Looks like `main` is back in order.  Can we rerun the linter here to make sure?  (It passes when I merge and run it locally.)

joeyballentine:
If you merge main to this branch and push it, it'll re-run. If I just re-run it here it will still fail since it doesn't have the fixes from main

RunDevelopment:
> If you merge main to this branch and push it, it'll re-run.

Not until someone with write access approves of them.

adodge:
Okay...  apparently I was on pylint 2.15.10 not 2.16.1, so I wasn't seeing this failure locally.  Now I am 95% sure it will pass in CI.

backend/src/nodes/nodes/image_channel/palette_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+            )(
+                NumberInput(
+                    "Palette Size",
+                    minimum=2,
+                    default=8,
+                ).with_id(2),
+            ),
+        ]
+        self.outputs = [
+            ImageOutput(
```
joeyballentine, line 67: Can you give this output a label of "Palette"?

adodge: Of course!

joeyballentine:
LGTM but just wanna make sure @RunDevelopment is fine with those ignored types

RunDevelopment:
Yeah, that's fine. Pyright can't handle numpy types for some reason. So all you can do is to `# type: ignore` yourself out of this mess.

RunDevelopment (approved):

PR #1506, opened 2023-01-25
Added PixTransform guided upsampling

RunDevelopment:
Code comes from this repo: https://github.com/prs-eth/PixTransform

There is still a lot to do, but the node is in a functioning state. The next steps will be to add support for:
- ~~Non-square images,~~
- ~~Tiling (this takes a lot of VRAM for images >256px),~~
- ~~RGB images, and~~
- ~~User configurable params (e.g. iterations).~~


RunDevelopment:
Okay, this is pretty much done now, but I still need to test everything. I also want to investigate batch-size-based splitting.

RunDevelopment:
Since PixTransform uses a lot of features that aren't used extensively in other nodes, I ran into a few bugs and fixed them. So this PR includes the following bug fixes unrelated to PixTransform:

- Overlap is now capped at 1/4th the smallest dimensions for exact splitting. Without this cap, processing very small images is impossible. E.g. it wouldn't be possible to upscale a 32x32px image with this fix.
- Fixed image segmentation for exact splitting. That method was honestly just broken (infinite loop type of broken), and we never knew because existing nodes only used relatively trivial splitting which didn't trigger the broken parts.
- Writing into the result buffer image (the one constructed from upscaled tiles) didn't work for grayscale images if they had a different number of dimensions (so e.g. the result image is a 3D grayscale image and the current tile is a 2D grayscale image).

RunDevelopment:
Alright, PixTransform works now. It's really slow, but it works.

image

joeyballentine:
Are you sure we should call the node "Pix Transform"? I would think "Guided Upscale" would be more appropriate, and in the future if we add more than just pix transform (other architectures or whatever) we can add a dropdown for which arch it uses

RunDevelopment:
Agreed. I rebranded it to "Guided Upscale".

joeyballentine (approved):

PR #1516, opened 2023-01-29
Lens Blur node

harisreedhar:
Add lens blur node
Screenshot from 2023-01-29 13-24-52

### DIFF
--- a/backend/src/nodes/nodes/image_filter/blur_lens.py
+++ b/backend/src/nodes/nodes/image_filter/blur_lens.py
@@ -1,0 +1,150 @@
+from __future__ import annotations
+
+import math
+import numpy as np
+from scipy import signal
+from functools import reduce
+
+from . import category as ImageFilterCategory
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput, SliderInput
+from ...properties.outputs import ImageOutput
+
+# lens blur adapted from: https://github.com/NatLee/Blur-Generator
+
+kernel_scales = [1.4,1.2,1.2,1.2,1.2,1.2]
+
+kernel_params = [
+                [[0.862325, 1.624835, 0.767583, 1.862321]],
+
+                [[0.886528, 5.268909, 0.411259, -0.548794],
+                [1.960518, 1.558213, 0.513282, 4.56111]],
+
+                [[2.17649, 5.043495, 1.621035, -2.105439],
+                [1.019306, 9.027613, -0.28086, -0.162882],
+                [2.81511, 1.597273, -0.366471, 10.300301]],
+
+                [[4.338459, 1.553635, -5.767909, 46.164397],
+                [3.839993, 4.693183, 9.795391, -15.227561],
+                [2.791880, 8.178137, -3.048324, 0.302959],
+                [1.342190, 12.328289, 0.010001, 0.244650]],
+
+                [[4.892608, 1.685979, -22.356787, 85.91246],
+                [4.71187, 4.998496, 35.918936, -28.875618],
+                [4.052795, 8.244168, -13.212253, -1.578428],
+                [2.929212, 11.900859, 0.507991, 1.816328],
+                [1.512961, 16.116382, 0.138051, -0.01]],
+
+                [[5.143778, 2.079813, -82.326596, 111.231024],
+                [5.612426, 6.153387, 113.878661, 58.004879],
+                [5.982921, 9.802895, 39.479083, -162.028887],
+                [6.505167, 11.059237, -71.286026, 95.027069],
+                [3.869579, 14.81052, 1.405746, -3.704914],
+                [2.201904, 19.032909, -0.152784, -0.107988]]]
+
+def get_parameters(component_count = 2):
+    parameter_index = max(0, min(component_count - 1, len(kernel_params)))
+    parameter_dictionaries = [dict(zip(['a','b','A','B'], b)) for b in kernel_params[parameter_index]]
+    return (parameter_dictionaries, kernel_scales[parameter_index])
+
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image_filter/blur_lens.py
```diff
@@ -0,0 +1,150 @@
...
+        self.outputs = [ImageOutput(image_type="Input0")]
+        self.category = ImageFilterCategory
+        self.name = "Lens Blur"
+        self.icon = "MdBlurOn"
+        self.sub = "Blur"
+
+    def run(
+        self,
+        img: np.ndarray,
+        in_radius: int,
```
RunDevelopment, line 141: This is a float because of `precision=1`.

```suggestion
        in_radius: float,
```

backend/src/nodes/nodes/image_filter/blur_lens.py
```diff
@@ -0,0 +1,150 @@
...
+                controls_step=1,
+            ),
+            NumberInput(
+                "Exposure Gamma",
+                minimum=0.01,
+                maximum=100,
+                default=5,
+                precision=4,
+                controls_step=0.1,
+            ),
```
RunDevelopment, line 130: It *might* be a better user experience to use a log-scale slider here. Emphasis on might, I'm not sure. Please test it out and choose what you think feels better.

```suggestion
            SliderInput(
                "Exposure Gamma",
                minimum=0.01,
                maximum=100,
                default=5,
                precision=4,
                controls_step=0.1,
                scale="log",
            ),
```

backend/src/nodes/nodes/image_filter/blur_lens.py
```diff
@@ -0,0 +1,150 @@
...
+def weighted_sum(kernel, params):
+    return np.add(kernel.real * params['A'], kernel.imag * params['B'])
+
+def multiply_kernel(kernel):
+    kernel_size = kernel.shape[1]
+    a = np.repeat(kernel, kernel_size, 0)
+    b = np.repeat(kernel.transpose(), kernel_size, 1)
+    return np.multiply(a,b)
+
+def lens_blur(img, radius=3, components=5, exposure_gamma=5):
```
RunDevelopment, line 80: Please add type hints to all parameters on all function.

backend/src/nodes/nodes/image_filter/blur_lens.py
```diff
@@ -0,0 +1,150 @@
...
+from scipy import signal
+from functools import reduce
+
+from . import category as ImageFilterCategory
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput, SliderInput
+from ...properties.outputs import ImageOutput
+
+# lens blur adapted from: https://github.com/NatLee/Blur-Generator
```
RunDevelopment, line 14: The license of that project requires us to credit the author, so let's add a little note.

```suggestion
# Lens blur adapted from: https://github.com/NatLee/Blur-Generator
# MIT License; Copyright (c) 2022 Nat Lee
```

backend/src/nodes/nodes/image_filter/blur_lens.py
```diff
@@ -0,0 +1,150 @@
...
+    return np.add(kernel.real * params['A'], kernel.imag * params['B'])
+
+def multiply_kernel(kernel):
+    kernel_size = kernel.shape[1]
+    a = np.repeat(kernel, kernel_size, 0)
+    b = np.repeat(kernel.transpose(), kernel_size, 1)
+    return np.multiply(a,b)
+
+def lens_blur(img, radius=3, components=5, exposure_gamma=5):
+    img = np.ascontiguousarray(img.transpose(2,0,1), dtype=np.float32)
```
RunDevelopment, line 81: Chainner uses 2D np arrays to represent grayscale images. I'm not sure whether this transpose call will work for that. If it doesn't, we have a `as_3d` function that converts a 2d grayscale image into a single-channel 3d image.

RunDevelopment (changes requested):
Thank you for the PR @harisreedhar!

I looked through code and have a few suggestions.
Also, please format the new code with [black](https://github.com/psf/black).

backend/src/nodes/nodes/image_filter/blur_lens.py
harisreedhar: Radius is int so removed precision

backend/src/nodes/nodes/image_filter/blur_lens.py
harisreedhar: Done

RunDevelopment:
@joeyballentine Does opencv or numpy include scipy?

joeyballentine:
@RunDevelopment I don't think so, so that means we would need to add that as another dependency. It's actually required by facexlib so anyone who has the pytorch deps installed already has it.

harisreedhar:
> @RunDevelopment I don't think so, so that means we would need to add that as another dependency. It's actually required by facexlib so anyone who has the pytorch deps installed already has it.

``scipy.signal.convolve2d`` can be replaced with this function. So no need to add extra dependency.
```python
def conv2d(a, f):
    s = f.shape + tuple(np.subtract(a.shape, f.shape) + 1)
    strd = np.lib.stride_tricks.as_strided
    subM = strd(a, shape = s, strides = a.strides * 2)
    return np.einsum('ij,ijkl->kl', f, subM)
```

RunDevelopment:
Maybe we could use `cv2.filter2D`? That one also does a convolution.

RunDevelopment:
> It is faster than current implementation

Holy, it's waaay faster. Nice! My little test example dropped from 51s to 4.3s, so it's over 10x faster.

RunDevelopment (approved):

PR #1519, opened 2023-01-29
NCNN Model File Iterator

joeyballentine:


backend/src/nodes/nodes/ncnn/model_file_iterator.py
```diff
@@ -0,0 +1,101 @@
...
+
+        model_path_node_id = context.get_helper(NCNN_ITERATOR_NODE_ID).id
+
+        just_param_files: List[str] = list_all_files_sorted(directory, [".param"])
+        just_bin_files: List[str] = list_all_files_sorted(directory, [".bin"])
+
+        if len(just_param_files) != len(just_bin_files):
+            raise ValueError(
+                "The number of param files and bin files are not the same. Please check your directory."
+            )
```
RunDevelopment, line 91: Checking just the length here is not enough IMO. I think that we should only pair .bin and .param files with the same name (well, relative path really).

RunDevelopment (approved):

PR #1521, opened 2023-01-30
Force a backend restart when killing the backend takes over 2.5s

joeyballentine:


src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -328,8 +328,21 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
-            const response = await backend.kill();
-            if (response.type === 'error') {
+            // Try to kill the current executor
+            // If it doesn't respond within 1 second, force restart it
+            const backendKillPromise = backend.kill();
+            const timeoutPromise = new Promise<string>((resolve) => {
+                setTimeout(() => {
+                    resolve('timeout');
+                }, 1000);
+            });
```
RunDevelopment, line 338: Please use the `delay` function and promise chaining. This not only makes the code simpler, it also results in better types. The `typeof response !== 'string'` below should be unnecessary now.

```suggestion
            const timeoutPromise = delay(1000).then(() => 'timeout');
```

RunDevelopment (commented):
I think 1sec might be a bit too harsh. Maybe we could give it 5sec? Restarting the backend also takes some time, and we also lose its cache (e.g. Load Image nodes), so the next run will take longer.

src/renderer/contexts/ExecutionContext.tsx
theflyingzamboni: Maybe make it adjustable in settings? 5sec feels too long to me, since the point is to stop an upscale instead of waiting for it to complete.

joeyballentine: Honestly, even just the 1 second feels a bit too long. When i was playing around with it yesterday, since the backend takes time to restart it really feels like nothing happens for quite a while. Maybe we need a modal with a spinner to pop up while it's stopping or something, then we can increase it to like 2.5s

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -326,9 +327,48 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+            const backendKillPromise = backend.kill();
+            const timeoutPromise = delay(KILL_TIMEOUT).then(() => ({
+                type: 'timeout',
+                exception: '',
+            }));
+            const response = await Promise.race([backendKillPromise, timeoutPromise]);
+            if (response.type === 'timeout') {
+                log.info('Executor did not respond to kill request. Forcing restart.');
+                // Force restart the backend
+                await ipcRenderer.invoke('restart-backend');
```
RunDevelopment, line 369: Please use `restart` from BackendContext.

src/common/Backend.ts
```diff
@@ -93,17 +93,22 @@ export interface BackendError {
 export class Backend {
     readonly port: number;
 
+    abortController: AbortController;
```
RunDevelopment, line 96: This should probably be private and readonly.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -326,8 +331,36 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+    const KILL_TIMEOUT = 2500;
+    useEffect(() => {
+        statusRef.current = status;
+        if (status === ExecutionStatus.KILLING) {
+            setTimeout(() => {
+                if (statusRef.current === ExecutionStatus.KILLING) {
+                    log.info('Executor did not actually kill the backend. Forcing restart.');
+
+                    ipcRenderer
+                        .invoke('restart-backend')
```
RunDevelopment, line 345: Could you please use the `restart` method from `BackendContext`? That method would probably be a better place to put the `backend.abort();` as well.

joeyballentine: Whoops, accidentally reverted that when i copied the old code back

src/common/Backend.ts
joeyballentine: I can't make it readonly because I make a new abortcontroller each run

joeyballentine: Which I'm pretty sure I have to, or it'll just always remain in an aborted state

RunDevelopment: Yeah, you're right. It can't be readonly.

src/common/Backend.ts
```diff
@@ -123,6 +128,7 @@ export class Backend {
     run(
         data: BackendRunRequest
     ): Promise<BackendSuccessResponse | BackendExceptionResponse | BackendAlreadyRunningResponse> {
+        this.abortController = new AbortController();
```
RunDevelopment, line 131: Why does `run` create a new AC every time? Like, this won't even work correctly if you have concurrent `run`s, which is not ensured here. Also, if I understand AC correctly, then all requests after `abort` will fail until we call `run` again.

Why don't make it the responsibility of `abort` to create a new AC? It used up the current one, so it should replace it with a new one.

src/renderer/components/Header/ExecutionButtons.tsx
```diff
@@ -71,10 +71,9 @@ export const ExecutionButtons = memo(() => {
                 <IconButton
                     aria-label={t('header.stopButton', 'Stop button')}
                     colorScheme="red"
-                    disabled={
-                        !(status === ExecutionStatus.RUNNING || status === ExecutionStatus.PAUSED)
-                    }
+                    disabled={![ExecutionStatus.RUNNING, ExecutionStatus.PAUSED].includes(status)}
                     icon={<IoStop />}
+                    isLoading={[ExecutionStatus.KILLING].includes(status)}
```
RunDevelopment, line 76:
```suggestion
                    isLoading={ExecutionStatus.KILLING === status}
```

src/renderer/contexts/BackendContext.tsx
```diff
@@ -100,6 +100,7 @@ export const BackendProvider = memo(
                     try {
                         // eslint-disable-next-line no-await-in-loop
                         await ipcRenderer.invoke('restart-backend');
+                        backend.abort();
```
RunDevelopment, line 103: I wonder whether we should abort all current requests before restarting. The timeline of 'restart-backend' is as follows:

```
-> call 'restart-backend'
    - stop python process
    - start new python process
<- 'restart-backend' yields back (await resolves)
    - new python process starts running our code (sanic starts the server)
```

I know this because I ran into this in CLI mode. Just because the backend process is open, doesn't mean it's ready.

So right now, we have a race condition. If the server starts up fast enough, we might not abort the old requests in time.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -266,10 +268,14 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                 });
             }
         } catch (err: unknown) {
-            sendAlert({
-                type: AlertType.ERROR,
-                message: `An unexpected error occurred: ${String(err)}`,
-            });
+            if (err instanceof DOMException && err.name === 'AbortError') {
+                // We could maybe send a success message here
```
RunDevelopment, line 272: If we aborted successfully, then the user will see the control buttons in the UI change. No need for any messages here.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -326,8 +332,34 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
         }
     }, [backend, sendAlert]);
 
+    const statusRef = useRef(status);
+
+    const KILL_TIMEOUT = 2500;
+    useEffect(() => {
```
RunDevelopment, line 338: I don't like this `useEffect`. Not only do we know that `useEffect` is kinda bugged in contexts, the effect might also be executed multiple times (e.g. if `restart` or `sendAlert` changes), which is not protected against. I think that this should be part of `kill` instead.

src/common/Backend.ts
joeyballentine: > if you have concurrent runs

That's not even possible at the moment... But yeah, I'll move it to after the abort

src/renderer/components/Header/ExecutionButtons.tsx
joeyballentine: whoops, leftover from when i had _another_ status in there as well

src/renderer/contexts/BackendContext.tsx
joeyballentine: IIRC it didn't work right the other way around, but I can try it

src/common/Backend.ts
RunDevelopment: > That's not even possible at the moment...

There is nothing *here* that guarantees that. You're using global reasoning. Like, in general, assuming that the whole application works correctly for a small piece of code to work isn't exactly best practice...

src/renderer/contexts/BackendContext.tsx
joeyballentine: works fine

src/renderer/contexts/ExecutionContext.tsx
joeyballentine: I still need to keep the useEffect for setting the ref, but otherwise I moved everything else to be part of kill

RunDevelopment: Why not use the original approach of `Promise.race`? This would eliminate the need for the ref.

joeyballentine: Because that doesn't work how you think it does. Calling `/kill` doesn't wait until it's killed. It returns immediately. So, we have to check whether or not the actual execution is over, which this timeout and status ref does.

RunDevelopment: You're right, `/kill` doesn't work like that. I wonder if it should though. I would expect that no executor is running after `/kill` responds with "Successfully killed execution."

It wouldn't even be difficult to change `/kill`. After `ctx.executor.kill()`, just do `while ctx.executor: await sleep(1ms)` and now we wait until the executor is actually killed. Spinning is a simple solution, although it's slightly incorrect. A new `/run` request could come in after the current `/run` as killed but before the `await sleep(1ms)` yields back. That's not really a concern right now though.

RunDevelopment (approved):

PR #1522, opened 2023-01-30
Update React Flow

joeyballentine:
This gets us two cool things:
- Dragging a node to the edge now scrolls
- Dragging a connection near a handle now snaps the edge to the handle

However, it appears this broke the pane context menu, and I'm not sure why. So, I'm opening this as a draft


joeyballentine:
Context menu on this is still broken

joeyballentine:
Fixed the issue. There doesn't seem to be anything else wrong so I think this is good. Thanks to Moritz for figuring that out for me because I never would have figured out it was the onPaneClick callback. He ultimately doesn't know why that changed, but it isn't necessary anymore anyway, so it doesn't really matter.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -447,7 +450,9 @@ export const usePaneNodeSearchMenu = (
     ));
 
     const onConnectStart = useCallback(
-        (event: React.MouseEvent, handle: OnConnectStartParams) => {
+        (event: React.MouseEvent | React.TouchEvent, handle: OnConnectStartParams) => {
+            // eslint-disable-next-line no-param-reassign
+            event = event as React.MouseEvent;
```
RunDevelopment, line 455: I think it would be better to support touch events instead of assuming that they'll never happen (can they happen?). In theory, we could just average all touchpoints and use that.

joeyballentine: Sure i could try that. I just assumed touch would never be use for this (especially since I'm pretty sure we don't have any of the necessary react flow stuff set up that would allow touch dragging

RunDevelopment: Hmmm, then maybe we (read you) should the RF guys about this?

joeyballentine: Tbh I think handling touch stuff should be in a different PR. I don't see a reason to not merge this just because suddenly we wanna support touch devices

RunDevelopment: Yeah, you're right.

RunDevelopment (approved):

PR #1528, opened 2023-02-04
Run black formatting + fix pylint linting

joeyballentine:
Black and pylint updated, which is why our CI has been failing. This fixes everything that is now wrong (and hopefully the pylint stuff doesn't break anything else)


RunDevelopment:
> Black and pylint updated

Then we should probably use fixed versions in our requirements.txt.

joeyballentine:
Would you rather have that than this?

RunDevelopment:
Why not both?

RunDevelopment (approved):

PR #1532, opened 2023-02-06
Color Levels node

harisreedhar:
Color levels node with similar functionality like gimp or photoshop.
chaiNNer-Untitled-2023-2-6_10-35

### DIFF
--- a/backend/src/nodes/nodes/image_adjustment/color_levels.py
+++ b/backend/src/nodes/nodes/image_adjustment/color_levels.py
@@ -25,7 +25,7 @@ class ColorLevelsNode(NodeBase):
         super().__init__()
         self.description = "Adjust color levels"
         self.inputs = [
-            ImageInput(),
+            ImageInput(channels=[1, 3, 4]),
             BoolInput("Red", default=True),
             BoolInput("Green", default=True),
             BoolInput("Blue", default=True),
@@ -94,7 +94,6 @@ class ColorLevelsNode(NodeBase):
         """Adjust color levels of image"""
 
         _, _, c = get_h_w_c(img)
-        c = min(c, 4)
 
         if c == 1:
             img = as_3d(img)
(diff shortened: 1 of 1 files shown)


joeyballentine:
Very cool, thanks! Code looks fine to me, I'll try it out tomorrow.

backend/src/nodes/nodes/image_adjustment/color_levels.py
```diff
@@ -0,0 +1,125 @@
...
+        """Adjust color levels of image"""
+
+        _, _, c = get_h_w_c(img)
+        c = min(c, 4)
+
+        if c == 1:
+            img = as_3d(img)
+            red, green, blue = True, True, True
+
+        img = (img * 255).astype("uint8")
```
RunDevelopment, line 103: Is the conversion to 8 bit really necessary? Some people use chainner to process 16 bit images.

harisreedhar: Removed 8 bit conversion. It was unnecessary

backend/src/nodes/nodes/image_adjustment/color_levels.py
```diff
@@ -0,0 +1,123 @@
...
+
+    This code was adapted from a Stack-Overflow answer by Iperov,
+    can found at: https://stackoverflow.com/a/60339950
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.description = "Adjust color levels"
+        self.inputs = [
+            ImageInput(),
```
RunDevelopment, line 28: This code assumes grayscale, RGB, or RGBA images. Chainner doesn't guarantee that. We have no way to create e.g. 2 channel images right now, but we might add this in the future. 
But there is an easy way to let the `ImageInput` enforce the number of channels of the input image. This will provide the guarantees you need. You can even remove the `c = min(c, 4)` since it's no longer necessary.

```suggestion
            ImageInput(channels=[1, 3, 4]),
```

RunDevelopment (changes requested):
One last nit. Otherwise gtg.

PR #1533, opened 2023-02-06
Rembg Node

### LINKED ISSUE #818: Onboard rembg for image background removal
**Motivation**
I want to be able to remove image backgrounds as part of my processing.

**Description**
[The rembg python module](https://github.com/danielgatis/rembg) uses a pytorch model to remove image backgrounds. It's not perfect but it's really damn good. I would like to have an option to use it during processing

theflyingzamboni:
Implements rembg (https://github.com/danielgatis/rembg) as a node in chaiNNer, for use with the models in the repo (https://github.com/danielgatis/rembg/releases/tag/v0.0.0).

Closes #818


backend/src/nodes/impl/rembg/pymatting/estimate_alpha_cf.py
```diff
@@ -0,0 +1,76 @@
...
+from .ichol import ichol
+from .cg import cg
+import numpy as np
+
+
+def estimate_alpha_cf(
+    image, trimap, preconditioner=None, laplacian_kwargs={}, cg_kwargs={}
+):
+    """
+    Estimate alpha from an input image and an input trimap using Closed-Form Alpha Matting as proposed by :cite:`levin2007closed`.
```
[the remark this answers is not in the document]
theflyingzamboni: The licenses are already in the folders, and there are copyright notes in the __init__.py files.

backend/src/nodes/impl/rembg/pymatting/cf_laplacian.py
```diff
@@ -0,0 +1,176 @@
+import numpy as np
+import scipy.sparse
+from numba import njit
```
RunDevelopment, line 3: This looks like another dependency.

backend/src/nodes/impl/rembg/session_factory.py
```diff
@@ -0,0 +1,19 @@
...
+
+import onnxruntime as ort
+
+from ...impl.onnx.utils import get_input_shape
+from .session_base import BaseSession
+from .session_cloth import ClothSession
+from .session_simple import SimpleSession
+
+
+def new_session(session: ort.InferenceSession) -> BaseSession:
```
RunDevelopment, line 11: Please explain what sessions are here.

backend/src/nodes/impl/onnx/model.py
```diff
@@ -1,5 +1,69 @@
...
+
+OnnxRemBgModels = (OnnxU2Net, OnnxU2NetCloth)
+OnnxRemBgModel = Union[OnnxU2Net, OnnxU2NetCloth]
+
+
+OnnxModels = (OnnxGeneric, *OnnxRemBgModels)
+OnnxModel = Union[OnnxGeneric, OnnxRemBgModel]
+
+
+def isRemBgModel(model_as_bytes: bytes) -> bool:
```
RunDevelopment, line 48: Python naming conventions, please.

backend/src/nodes/nodes/onnx/load_model.py
```diff
@@ -24,7 +24,7 @@ def __init__(self):
         )
         self.inputs = [OnnxFileInput(primary_input=True)]
         self.outputs = [
-            OnnxModelOutput(),
+            OnnxModelOutput(kind="onnx"),
```
RunDevelopment, line 27: Why does this not broadcast?

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -79,4 +79,4 @@ def run(self, model: PyTorchSRModel, is_fp16: int) -> Tuple[OnnxModel, str]:
 
         fp_mode = "fp16" if should_use_fp16 else "fp32"
 
-        return OnnxModel(onnx_model_bytes), fp_mode
+        return load_onnx_model(onnx_model_bytes), fp_mode
```
RunDevelopment, line 82: Since the PyTorch model is guaranteed to be an SR model, shouldn't the ONNX model be one as well? If so, then the ONNX output of this node should reflect that.

(I added the comment here, because GH wouldn't allow me to near `self.outputs`.)

src/common/types/chainner-scope.ts
```diff
@@ -106,6 +113,18 @@ def convenientUpscale(model: PyTorchModel | NcnnNetwork | OnnxModel, image: Imag
     }
 }
 
+def removeBackground(model: OnnxRemBgModel, image:Image) {
```
RunDevelopment, line 116:
```suggestion
def removeBackground(model: OnnxRemBgModel, image: Image) {
```

backend/src/nodes/impl/rembg/pymatting/cf_laplacian.py
theflyingzamboni: Ah crap, that's what I get for forgetting to test with `npm start`.

backend/src/nodes/impl/onnx/model.py
theflyingzamboni: We should change that for pytorch too, because I just copied the format of all that from the pytorch model code we have.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
theflyingzamboni: What does PyTorch have to do with ONNX?

backend/src/nodes/impl/rembg/session_factory.py
theflyingzamboni: From what I can tell, what rembg does is wrap the onnx runtime session in a custom session class. There are a couple of different classes, depending on the model. They do some extra pre- and post-processing on the input/output data.

backend/src/nodes/nodes/onnx/load_model.py
theflyingzamboni: Because when I was trying to figure out how to do the generic output + tag, I thought should_broadcast was related to the model preview layout, so I'd tried removing it. Once I got it working, I didn't add it back in. Everything works without it, including the pre-connection validation, so I honestly don't really know what it's meant for? Could you explain that, because it doesn't seem like it's doing what I thought it did here.

backend/src/nodes/impl/rembg/session_factory.py
RunDevelopment: Wdym "from what I can tell?" This function references one of our functions, so how can it be from rembg?

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
RunDevelopment: From what I understand, a pytorch SR model could never be a ONNX rembg model.

theflyingzamboni: To expand on that, we've always allowed ONNX and NCNN to load any model so people aren't restricted by what we add in. So we don't guarantee SR models for anything. The only difference here is that for the Remove Background node, we need to be able to guarantee a valid model.

backend/src/nodes/nodes/onnx/load_model.py
RunDevelopment: This shouldn't work. Without broadcasting, the frontend cannot receive the information necessary to narrow down types.

backend/src/nodes/impl/rembg/session_factory.py
theflyingzamboni: I thought you were asking about the Session classes in general.

Here's the original version: https://github.com/danielgatis/rembg/blob/main/rembg/session_factory.py

session_factory.py exists to acquire the models, determine what kind of session wrapper that model needs, and set of the InferenceSession. I modified it heavily to work within chaiNNer.

1. We don't want to determine the models by filename, so I replaced that check with one that uses one of our functions to reference a unique characteristic of the cloth model vs the others.
2.  We aren't downloading the models, so I removed all that code.
3. We already have our own method of creating and caching InferenceSessions, so I passed it in instead of creating a new one.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
RunDevelopment: > So we don't guarantee SR models for anything.

We guarantee SR models for PyTorch.

theflyingzamboni: I'm talking about ONNX, not PyTorch.

backend/src/nodes/nodes/onnx/load_model.py
theflyingzamboni: ¯\\\_(ツ)_/¯

Not sure what to tell you. I don't have that in, and you can see it's working.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
RunDevelopment: But *I am* talking about PyTorch. This is the PyTorch to ONNX node.

theflyingzamboni: Ohhhh. I didn't see that. I thought you were commenting on the Load Model node.

theflyingzamboni: Hmm. I kind of don't want to make a specific SR model class there? Since the supported PyTorch SR models are a subset of possible ONNX SR models that a user could load. It would be a little strange to categorize some as Generic and some as SR. From the user end, it doesn't really matter whether the convert node guarantees something as "not rembg," but if I were to do that, I'd rather just specify it as Generic.

RunDevelopment: > If I were to do restrict the output type, I think I'd rather just specify it as Generic.
That's what I wanted to suggest.

backend/src/nodes/nodes/onnx/load_model.py
RunDevelopment: Now, I know why. You implemented  `get_broadcast_data` incorrectly. The output is supposed to decide itself whether it broadcasts by returning either `None` or a value. [Look at the NCNN implementation](https://github.com/chaiNNer-org/chaiNNer/blob/a635cf2a111ddd7382de8017f921ad5d71aa7807/backend/src/nodes/properties/outputs/ncnn_outputs.py#L18). So the ONNX output always broadcasts because it completely ignores `should_broadcast`, which is incorrect.

backend/src/nodes/impl/onnx/model.py
```diff
@@ -1,5 +1,69 @@
...
+        self.bytes: bytes = model_as_bytes
+        self.arch = "u2net"
+        self.sub_type = "RemBg"
+
+
+class OnnxU2NetCloth:
+    def __init__(self, model_as_bytes: bytes):
+        self.bytes: bytes = model_as_bytes
+        self.arch = "u2net_cloth"
+        self.sub_type = "RemBg"
```
RunDevelopment, line 37: What's the motivation for making those 2 separate classes? Couldn't this just be one `OnnxRemBg` class?

src/common/types/chainner-scope.ts
```diff
@@ -106,6 +113,18 @@ def convenientUpscale(model: PyTorchModel | NcnnNetwork | OnnxModel, image: Imag
...
 }
 
+def removeBackground(model: OnnxRemBgModel, image: Image) {
+    Image {
+        width: image.width,
+        height: if model.arch == "u2net_cloth" {
+            image.height * 3
+        } else {
+            image.height
+        },
```
RunDevelopment, line 123: I don't like this. Why does the type has to know about arch-specific things? Instead, I think we should give the type scales for width and height. This moves the arch-specific knowledge into the backend.

backend/src/nodes/impl/onnx/model.py
theflyingzamboni: Kind of similar to why we have different classes for different PyTorch arches. I split them because I needed to distinguish them for output typing, but they are also distinct in that they require entirely different session wrappers.

src/common/types/chainner-scope.ts
theflyingzamboni: You mean add a `output_height_scale` attribute to the backend classes?

backend/src/nodes/impl/onnx/model.py
RunDevelopment: > I split them because I needed to distinguish them for output typing, but they are also distinct in that they require entirely different session wrappers

These are 2 classes are only used here. Everything else only uses `OnnxRemBgModel`, `OnnxGeneric`, and `OnnxModel`.

So what's the point of having 2 classes for rembg models?

src/common/types/chainner-scope.ts
RunDevelopment: Yes, but also to the type. Right now, the knowledge of how u2net_cloth models work is spread between the backend and this function. So I am suggesting to fix that by given the type `scaleWidth: uint` and `scaleHeight: uint` fields. Yes, this means that we need to track those values in the backend and broadcast them over.

backend/src/nodes/impl/onnx/model.py
theflyingzamboni: Because I needed to distinguish the u2net arch from the u2net_cloth arch, because the removeBackground type function uses those to type the output.

backend/src/nodes/nodes/pytorch/convert_to_onnx.py
```diff
@@ -27,7 +27,12 @@ def __init__(self):
             OnnxFpDropdown(),
         ]
         self.outputs = [
-            OnnxModelOutput(label="ONNX Model"),
+            OnnxModelOutput(
+                model_type="OnnxGenericModel",
+                label="ONNX Model",
+                kind="onnx",
+                should_broadcast=True,
```
RunDevelopment, line 34: Only starting nodes should broadcast types.

src/common/types/chainner-scope.ts
```diff
@@ -64,10 +64,23 @@ struct NcnnNetwork {
...
-    scale: int(1..),
-    inputChannels: int(1..),
-    outputChannels: int(1..),
+    arch: string,
+    subType: string,
+    scaleHeight: int(1..) | null,
+    scaleWidth: int(1..) | null,
 }
+let OnnxRemBgModel = OnnxModel {
+    arch: "u2net",
```
RunDevelopment, line 73: Is arch even useful anymore?

theflyingzamboni: Maybe not? I was just copying the way PyTorch does this.

src/common/types/chainner-scope.ts
```diff
@@ -64,10 +64,23 @@ struct NcnnNetwork {
...
+    subType: "RemBg",
+    scaleHeight: 1 | 3,
+    scaleWidth: 1,
+};
+let OnnxGenericModel = OnnxModel {
+    arch: invStrSet(OnnxRemBgModel.arch),
+    subType: "Generic",
+    scaleHeight: null,
+    scaleWidth: null,
```
RunDevelopment, line 82: That's not really how types are supposed to be used. We don't need `null` to represent that we don't know the scale, because we just omit the field (all fields are optional).

Here's how I would do it:
```
struct OnnxModel {
    arch: string,
    subType: string,
    scaleWidth: int(1..),
    scaleHeight: int(1..),
}
let OnnxRemBgModel = OnnxModel {
    arch: "u2net",
    subType: "RemBg",
};
let OnnxGenericModel = OnnxModel {
    arch: invStrSet(OnnxRemBgModel.arch),
    subType: "Generic",
}
```

I mean, generic models do have a scale, we just don't have a way of knowing it yet.

src/common/types/chainner-scope.ts
theflyingzamboni: This may possibly be used in the future if we ever get ONNX broadcasting fully. I'm not sure.

src/common/types/chainner-scope.ts
theflyingzamboni: Why remove the specific scale definitions from OnnxRemBgModel?

src/common/types/chainner-scope.ts
theflyingzamboni: Eh, I can just remove it for now and add it back if it's needed.

src/common/types/chainner-scope.ts
RunDevelopment: Because (1) they will be "added back in" when the model is loaded and gets its output type overridden and (2) because it's an implementation detail (if there is a model with a height scale of 2, we'd have to change the type again).

RunDevelopment (approved):

PR #1538, opened 2023-02-08
Add Convert Normal Maps node

joeyballentine:
Thanks, Mark from Nvidia

includes dx <--> ogl conversion and conversion to octahedral

image


backend/src/nodes/impl/normals/nvidia/LightspeedOctahedralConverter.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# To then load these into RTX Remix, you can convert it to a DDS file using
+#   https://developer.nvidia.com/nvidia-texture-tools-exporter
+#   Use BC5 compression, and the flag --no-mip-gamma-correct
+class LightspeedOctahedralConverter:
+    # Convert DirectX style normal maps (green is down)
+    @staticmethod
+    def convert_dx_to_octahedral(image):
+        image = LightspeedOctahedralConverter._check_for_spherical_normals(image)
+        normals = LightspeedOctahedralConverter._pixels_to_normals(image)
```
RunDevelopment, line 46: These 2 lines don't work, kinda. The problem is that they assume that the blue channel stores the z component of the normal as -1..1 mapping to 0..1 (`b = (z+1)/2`) (or 0..255 in the original implementation). This is somewhat uncommon. Since most game engines only support hemispherical normal maps (where z is between 0 and 1), those game engines don't encode negative z value (`b = abs(z)`). As we talked about before, z can be computed on the fly, which is why many game engines use the b channel to store other material information in there. So we can't assume anything about the b channel.

To fix this, we just have to ignore the blue channel and compute z on the fly. This is exactly what my `gr_to_xyz` util method does.

backend/src/nodes/impl/normals/nvidia/LightspeedOctahedralConverter.py
```diff
@@ -0,0 +1,100 @@
...
+    @staticmethod
+    def _pixels_to_normals(image):
+        image = image[:, :, 0:3]
+        image = image * 2.0 - 1.0
+        return image / np.linalg.norm(image, axis=2)[:, :, np.newaxis]
+
+    @staticmethod
+    def _octahedrals_to_pixels(octahedrals):
+        image = octahedrals + (0.5 / 255)
+        return np.pad(image, ((0, 0), (0, 0), (0, 1)), mode="constant")  # type: ignore
```
RunDevelopment, line 83: The +0.5 is a trick to get correct rounding by flooring. This can go because we don't convert to 8bit anymore.
```suggestion
        return np.pad(octahedrals, ((0, 0), (0, 0), (0, 1)), mode="constant")  # type: ignore
```

backend/src/nodes/impl/normals/nvidia/LightspeedOctahedralConverter.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @staticmethod
+    def ogl_to_dx(image):
+        # flip the g channel to convert to DX style
+        image[:, :, (1)] = 1 - image[:, :, (1)]
+        return image
```
RunDevelopment, line 89: I don't like the fact that we write to the image and then return it. Someone (e.g. me) looking at just the function signature could assume that this is a pure function.

backend/src/nodes/impl/upscale/convenient_upscale.py
```diff
@@ -68,7 +68,7 @@ def convenient_upscale(
             upscale(as_target_channels(white, model_in_nc, True)), 3, True
         )
 
-        alpha = 1 - np.mean(white_up - black_up, axis=2)
+        alpha = 1 - np.mean(white_up - black_up, axis=2)  # type: ignore
```
RunDevelopment, line 71: Why?

joeyballentine: It started failing for me, idk why

joeyballentine: All of a sudden pyright has a problem with it

joeyballentine: guess it only is an issue for me locally

backend/src/nodes/nodes/image_filter/normal_convert.py
```diff
@@ -0,0 +1,78 @@
...
+    ) -> np.ndarray:
+        if from_type == to_type:
+            return img
+
+        normal = np.stack(gr_to_xyz(img), axis=2)
+
+        if (from_type == NormalMapType.DIRECTX and to_type == NormalMapType.OPENGL) or (
+            from_type == NormalMapType.OPENGL and to_type == NormalMapType.DIRECTX
+        ):
+            normal = LightspeedOctahedralConverter.ogl_to_dx(normal)
```
RunDevelopment, line 73: This is incorrect. `ogl_to_dx` expects an RGB image and not XYZ, so the conversion doesn't work.

backend/src/nodes/nodes/image_filter/normal_convert.py
```diff
@@ -0,0 +1,78 @@
...
+        if from_type == NormalMapType.DIRECTX and to_type == NormalMapType.OCTAHEDRAL:
+            normal = LightspeedOctahedralConverter.convert_dx_to_octahedral(normal)
```
RunDevelopment, line 75: This is also incorrect. The `convert_*_to_octahedral` function return an RGB image, not normals.

backend/src/nodes/nodes/image_filter/normal_convert.py
```diff
@@ -0,0 +1,78 @@
...
+        self.sub = "Normal Map"
+
+    def run(
+        self,
+        img: np.ndarray,
+        from_type: NormalMapType,
+        to_type: NormalMapType,
+    ) -> np.ndarray:
+        if from_type == to_type:
+            return img
```
RunDevelopment, line 66: `img` may be an RGBA image, so we can't just return it as is.

RunDevelopment (commented):
I think we need to further specify the behavior of this node. Right now, it is not clear at all what the contents of the B channel is for each To value. 

I would recommend the following: 
- For DirectX and OpenGL, the B channel is calculated z component as defined by `xyz_to_bgr`.
- For Octahedral, the B channel is zero.
It's important to note that those are guarantees. So even if no actual conversion takes place (e.g. OpenGL -> OpenGL), those guarantees to have to be met, so no `if from_type == to_type: return img`.

backend/src/nodes/nodes/image_filter/normal_convert.py
joeyballentine: Is an XYZ image not an RGB image? I don't understand

backend/src/nodes/nodes/image_filter/normal_convert.py
```diff
@@ -0,0 +1,78 @@
...
+from ...properties import expression
+from ...impl.image_utils import NormalMapType
+from ...impl.normals.nvidia.LightspeedOctahedralConverter import (
+    LightspeedOctahedralConverter,
+)
+from ...impl.normals.util import gr_to_xyz, xyz_to_bgr
+
+
+@NodeFactory.register("chainner:image:convert_normal_map")
+class SpecularToMetal(NodeBase):
```
joeyballentine, line 19:
```suggestion
class ConvertNormalMap(NodeBase):
```

backend/src/nodes/nodes/image_filter/normal_convert.py
joeyballentine: I see now, thanks

RunDevelopment (approved):
image

PR #1539, opened 2023-02-08
rearrange inputs on upscale nodes so images come first

Zarxrax:


backend/src/nodes/nodes/pytorch/upscale_face.py
```diff
@@ -31,12 +31,12 @@ def __init__(self):
         super().__init__()
         self.description = "Uses face-detection to upscales and restore face(s) in an image using a PyTorch Face Super-Resolution model. Right now supports GFPGAN, RestoreFormer, and CodeFormer."
         self.inputs = [
-            FaceModelInput("Model"),
-            ImageInput(),
-            ImageInput("Upscaled Background").make_optional(),
+            ImageInput().with_id(1),
+            ImageInput("Upscaled Background").with_id(2).make_optional(),
+            FaceModelInput("Model").with_id(0),
```
RunDevelopment, line 36: I think Image, Model, BG Image would be better. This makes it more consistent with the other nodes. The optional bg image is essentially just an upscale-specific input (like tile size), which come after the model input in other nodes.

joeyballentine: That makes sense

Zarxrax: ok, sounds fine. Changed.

RunDevelopment (approved):

PR #1540, opened 2023-02-08
Stable Diffusion Outpainting

adodge:
I don't have a good sense of how to configure this node to get good results, but I'm pretty sure if it's possible in the GUI, it should be possible with this node.

image

image


joeyballentine:
Awesome. Man, these nodes are gigantic lol

adodge:
> Man, these nodes are gigantic lol

They sure are!  I'm not sure what can be done about it, right now.  I thought about having a "basic" node that has minimal inputs and just uses the defaults for the others, but I don't know if anyone would ever use it.  Fiddling with all the settings is sort of part of the SD workflow.

joeyballentine:
Yeah i think it's an unavoidable issue. It's far better to have all the options than not.

backend/src/nodes/nodes/external_stable_diffusion/outpainting.py
```diff
@@ -0,0 +1,271 @@
...
+                    maximum=1,
+                    precision=2,
+                    slider_step=0.01,
+                    controls_step=0.01,
+                ),
+            ),
+        ]
+
+        # target_w = math.ceil((init_img.width + left + right) / 64) * 64
+        # target_h = math.ceil((init_img.height + up + down) / 64) * 64
```
joeyballentine, line 147: what are these for? are they still necessary?

adodge: Ah, nope!  Note to myself.  I've removed it.

RunDevelopment (approved):

PR #1544, opened 2023-02-09
Gradients Node

adodge:
A node that creates a variety of gradients.

image


joeyballentine:
This has made me realize how badly we need a color picker input lol

theflyingzamboni:
Before we add a gradient node, we should really finally implement a proper color picker (as a separate PR prior to this one). We need this for multiple other nodes, and this node is just too much without it. My suggestion remains to implement a color picker as a Popover (https://chakra-ui.com/docs/components/popover) that activates when selecting the singular color input.

Perhaps we could pick from and use one of these pre-designed React color pickers: https://casesandberg.github.io/react-color/

As a side note, the color and position could be implemented as a group, allowing people to add additional color inputs (like how Text Append works).

joeyballentine:
Yeah I was thinking the exact same thing

adodge:
How about a version of this that doesn't offer any color options, and just produces a gradient from [0,255] grayscale with whatever shape options?  The rest could be done with a LUT if needed.

joeyballentine:
That definitely could work

adodge:
Okay.  Here's what that looks like:

image

theflyingzamboni:
That could work as a stopgap, but I think that's all that should be seen as. With a proper color picker and groups, I think we can manage something much fuller featured.

adodge:
We could have a node for turning greys to a color gradient, with specialized color picker widgets.  It would be useful for the output of this and also other images.  As inspiration: [Blender's color ramp node](https://docs.blender.org/manual/en/latest/render/shader_nodes/converter/color_ramp.html)

theflyingzamboni:
Something color ramp-like would be cool, but I'm not sure how we could implement it in the most chaiNNer-esque way. It would be nice to make it so users could input a colors via connection so we could have a Color Input node that could be connected to other nodes. Time for a crappy mockup (I'm so sorry):

node mockup


The colors swatches would be clickable to bring up the color picker. In an ideal world, the preview would have the movable squares like blender or GIMP as well so that their positions could be adjusted manually or by number input. There would also be dimensions and a Gradient Style dropdown like in the initial version (I was too lazy to add it to the mockup).

joeyballentine (approved):
I think this looks good, and even though the overall way of creating colored gradients is limited right now i think this is a great starting point. Anyone disagree?

theflyingzamboni:
Could we at least add some directionality controls to horizontal, vertical, and diagonal? With that, I think it would be an okay starting step until we can get the fundamentals necessary for a more advanced gradient node implemented.

adodge:
How about this?  I added a "reverse" input to all the styles, and redid diagonal to have an angle and width control.

image

theflyingzamboni:
Sounds good for now. 👍

joeyballentine (approved):

PR #1545, opened 2023-02-10
Internal Stable Diffusion

adodge:
I just found out about this `sdkit` library.  https://github.com/easydiffusion/sdkit  Looks like we could use it to load and run SD models in-process.  Here's a quick proof of concept.

image

### DIFF
--- a/backend/src/nodes/impl/stable_diffusion/stable_diffusion.py
+++ b/backend/src/nodes/impl/stable_diffusion/stable_diffusion.py
@@ -1,0 +1,145 @@
+import hashlib
+from enum import Enum
+from typing import List, Optional
+
+import cv2
+import numpy as np
+import sdkit
+import torch
+from PIL import Image
+from sdkit.generate import generate_images
+from sdkit.models import load_model
+
+from nodes.utils.utils import get_h_w_c
+
+
+class InternalSamplerName(Enum):
+    DDIM = "ddim"
+    PLMS = "plms"
+    HEUN = "heun"
+    EULER = "euler"
+    EULER_A = "euler_a"
+    DPM2 = "dpm2"
+    DPM2_A = "dpm2_a"
+    LMS = "lms"
+    dpm_solver_stability = "dpm_solver_stability"
+    DPMpp_2S_A = "dpmpp_2s_a"
+    DPMpp_2M = "dpmpp_2m"
+    DPMpp_SDE = "dpmpp_sde"
+    DPM_FAST = "dpm_fast"
+    DPM_A = "dpm_adaptive"
+
+
+def nearest_valid_size(width, height):
+    return (width // 64) * 64, (height // 64) * 64
+
+
+class StableDiffusion:
+    def __init__(self, model):
+        self.model = model
+
+        hsh = hashlib.sha256()
+        hsh.update(self.__class__.__name__.encode("utf-8"))
+        state_dict = model.state_dict()
+        for key in sorted(state_dict.keys()):
+            hsh.update(key.encode("utf-8"))
+            hsh.update(state_dict[key].detach().cpu().numpy().tobytes())
+
+        self.cache_hash: bytes = hsh.digest()
+
+    @classmethod
+    def from_file(cls, path: str):
+        context = sdkit.Context()
+        # VRAM usage level "high" disables some clever hooks that load and
+        # unload layers from VRAM during processing.  Saves VRAM, but causes
+        # some weird errors when we try to move the loaded models out of the
+        # sdkit.Context().  (Maybe we don't have to do that though...)
+        context.vram_usage_level = "high"
+        context.model_paths["stable-diffusion"] = path
+        load_model(context, "stable-diffusion")
+        return cls(context.models["stable-diffusion"])
+
+    @staticmethod
+    def _array_to_image(image_nparray: np.ndarray) -> Image:
+        image_nparray = (np.clip(image_nparray, 0, 1) * 255).round().astype("uint8")
+        _, _, c = get_h_w_c(image_nparray)
+        if c == 1:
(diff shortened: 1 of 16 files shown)


backend/src/nodes/impl/stable_diffusion/stable_diffusion.py
```diff
@@ -0,0 +1,33 @@
...
+        load_model(self.context, "stable-diffusion")
+
+        images: List[Image] = generate_images(context=self.context, prompt=prompt)
+        image = np.array(images[0])
+
+        unload_model(self.context, "stable-diffusion")
+
+        # XXX TODO Would be very nice to not load and unload the model completely every time. Need to figure out a
+        #  way to load_model an sdkit.Context to cpu, pass that around, and send it to the gpu when we want to run.
+        #  like ONNX or PyTorch
```
joeyballentine, line 31: Maybe what might be a good thing to attempt with this is actually just putting all the necessary code into chainner itself. We tend to do this with anything we can do we can 
1) cut down on the number of dependencies needed 
2) modify things as necessary 

For example, this library installs a bunch of unnecessary stuff like the realesrgan and gfpgan packages. We could just remove everything pertaining to that since we don't need those. 

And of course part of that could be implementing what this comment says (if it's not possible with what's available)

joeyballentine: The other thing that would be nice is being able to parse the state dict to determine what version of the model it is, if the arch changed between versions. I do this with all the other models. Only thing is, i have no idea if that's possible with .ckpt files and it's part of why i gave up initially on implementing SD :/

joeyballentine:
Overall, this is probably a great idea. i had no idea this library existed

backend/src/nodes/impl/stable_diffusion/stable_diffusion.py
adodge: Would it make sense to fork `sdkit` and make the changes we need to that?  If we just copy code around, it'll be a much more manual process to include upstream improvements.

joeyballentine: Yeah, in this case that would make more sense. Good idea

backend/src/nodes/impl/ncnn/model.py
```diff
@@ -704,7 +704,7 @@ def bin(self) -> bytes:
 
 class NcnnModelWrapper:
     def __init__(self, model: NcnnModel) -> None:
-        self.model: NcnnModel = model
+        self.model: NcannModel = model
```
joeyballentine, line 707: Uhhhh....

joeyballentine (commented):
How does this work now? You removed the dependency so how does it know to use the fork?

backend/src/nodes/impl/ncnn/model.py
adodge: Uhh, not sure how that happened.

adodge: > How does this work now? You removed the dependency so how does it know to use the fork?

It doesn't...  I've just installed it manually for testing.  I'm not sure the best way to do this.  I think the dependency manager code needs to change a bit to accommodate installing from a github link.  Or maybe it's better to use a git submodule.

joeyballentine: I think I'd rather you just publish it to pypi and we install like normal. I forked NCNN and upload it to pypi under ncnn_vulkan, for example, so its not unprecedented

Ryu1845:
Will this support image to image?

joeyballentine:
@Ryu1845 in case you don't know, the latest nightly already supports img2img (as well as other SD features) via automatic1111's webui API. This PR is basically just a proof of concept for having an internal implementation, which if we go down the route of fully implementing, will also have img2img.

Ryu1845:
Oh, thanks. sorry I didn't look close enough :sweat_smile:

joeyballentine:
@Ryu1845 and just to be clear, you can get the nightlies from here: https://github.com/chaiNNer-org/chaiNNer-nightly

Not sure when it will be officially released, but hopefully soon.

ExoPhantasm:
This is an interesting development, but is this the right direction to go having to rely on what seems to be a fairly stripped back library? It is lacking a lot of features that are accessible via automatic api such as embeddings, LORA, alternate samplers for img2img, prompt blending, weighting (?) and likely more.

Perhaps if a native SD is required then it should use [diffusers](https://github.com/huggingface/diffusers) as technically it should be possible to implement any community pipeline as a node, this covers almost all features needed and even some that aren't even in popular UIs yet, such as [SAG](https://github.com/KU-CVLAB/Self-Attention-Guidance).

adodge:
> This is an interesting development, but is this the right direction to go having to rely on what seems to be a fairly stripped back library? It is lacking a lot of features that are accessible via automatic api such as embeddings, LORA, alternate samplers for img2img, prompt blending, weighting (?) and likely more.
> 
> Perhaps if a native SD is required then it should use [diffusers](https://github.com/huggingface/diffusers) as technically it should be possible to implement any community pipeline as a node, this covers almost all features needed and even some that aren't even in popular UIs yet, such as [SAG](https://github.com/KU-CVLAB/Self-Attention-Guidance).

I've been looking at it.  It's a good reference implementation.

adodge:
Not that we want to follow them uncritically, but it's worth noting that the major implementations of SD that people are using don't use the `diffusers` library.  Most implementations seem to be based ultimately on this repo:  https://github.com/Stability-AI/stablediffusion

adodge:
I'm currently looking at adapting this codebase:  https://github.com/comfyanonymous/ComfyUI, which does support all the hip new SD things.  (And incidentally has a copy of the Stability-AI code buried in it.)

adodge:
Superseded by #1602

PR #1546, opened 2023-02-10
Inpaint node

harisreedhar:
Read more about [opencv inpaint](https://docs.opencv.org/3.4/df/d3d/tutorial_py_inpainting.html)
chaiNNer-Inpaint-2023-2-8_20-49

### DIFF
--- a/backend/src/nodes/nodes/image_utility/inpaint.py
+++ b/backend/src/nodes/nodes/image_utility/inpaint.py
@@ -7,6 +7,7 @@ import numpy as np
 from . import category as ImageUtilityCategory
 from ...node_base import NodeBase
 from ...node_factory import NodeFactory
+from ...properties import expression
 from ...properties.inputs import ImageInput, NumberInput, EnumInput
 from ...properties.outputs import ImageOutput
 
@@ -20,7 +21,7 @@ class InpaintAlgorithm(Enum):
 class InpaintNode(NodeBase):
     def __init__(self):
         super().__init__()
-        self.description = "Inpaint Image"
+        self.description = "Inpaint an image with given mask."
         self.inputs = [
             ImageInput(channels=[1, 3]),
             ImageInput(label="Mask", channels=1),
@@ -39,7 +40,17 @@ class InpaintNode(NodeBase):
                 controls_step=1,
             ),
         ]
-        self.outputs = [ImageOutput(image_type="Input0")]
+        self.outputs = [
+            ImageOutput(
+                image_type=expression.Image(
+                    width="Input0.width & Input1.width",
+                    height="Input0.height & Input1.height",
+                    channels="Input0.channels",
+                )
+            ).with_never_reason(
+                "The given image and mask must have the same resolution."
+            )
+        ]
         self.category = ImageUtilityCategory
         self.name = "Inpaint"
         self.icon = "MdOutlineAutoFixHigh"
@@ -52,8 +63,7 @@ class InpaintNode(NodeBase):
         inpaint_method: InpaintAlgorithm,
         radius: float,
     ) -> np.ndarray:
-        """Inpaint can be used to remove unwanted elements from an image,
-        such as dust and scratches in scanned photographs."""
+        """Inpaint an image"""
 
         assert (
             img.shape[:2] == mask.shape[:2]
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image_utility/inpaint.py
```diff
@@ -0,0 +1,66 @@
...
+            ),
+            NumberInput(
+                "Search Radius",
+                minimum=0,
+                default=1,
+                precision=1,
+                controls_step=1,
+            ),
+        ]
+        self.outputs = [ImageOutput(image_type="Input0")]
```
RunDevelopment, line 42: Time for some more type system tricks!

Firstly, since we know that the image and its mask must have the same size, we can use a type intersection to model this:
```
Image {
    width: Input0.width & Input1.width,
    height: Input0.height & Input1.height,
    channels: Input0.channels
}
```

This has a nice side effect: we know the size of the output image, even if we only know the size of one of the input images. E.g. if the size the main image is unknow, but the mask is 100x100, then we know that the node will either output a 100x100 image or error.

However, this creates an interesting case: If we know the sizes of both images and they are different, then we statically know the node will always error.
In terms of the type system, this is represented by the output type being `never`. `never` is the same concept as TypeScript's and Python's `never` type, it's the type that represents the empty set. When an output is `never`, we know that an error is going to occur, but we don't know that that error is. So outputs can optionally have a never reason. This a simple error message that will be shown to the user if an output is never.

With this little trick, we can statically verify that the 2 images have the same size using the type system. The [Add Normals node](https://github.com/chaiNNer-org/chaiNNer/blob/main/backend/src/nodes/nodes/image_filter/normal_addition.py#L42) is doing the same thing btw.

RunDevelopment (commented):
Great node idea. Thank you for adding this!

2 minor nits and then we can merge this :)

---

It's not gonna happen, but I wanted to mention it anyway: OpenCV also supports [FSR and ShiftMap](https://docs.opencv.org/4.x/de/daa/group__xphoto.html#gab4febba6be53e5fddc480b8cedf51eee) as part of xphoto, an unstable extension...

backend/src/nodes/nodes/image_utility/inpaint.py
harisreedhar: So, this helps to raise an error before python execution? Interesting! 
Thanks for detailed explanation.

harisreedhar:
> It's not gonna happen, but I wanted to mention it anyway: OpenCV also supports [FSR and ShiftMap](https://docs.opencv.org/4.x/de/daa/group__xphoto.html#gab4febba6be53e5fddc480b8cedf51eee) as part of xphoto, an unstable extension...

``opencv-contrib`` have tons of extra image processing tools. Any plans on including that to dependency?

RunDevelopment:
> `opencv-contrib` have tons of extra image processing tools. Any plans on including that to dependency?

I mean, it would be nice. It has tons of extra image processing functionality, so we could potentially implement some cool nodes with it. Since `opencv-contrib-python` is a strict superset of plain `opencv-python` that we currently use, we wouldn't have to change any of our existing python code either.

The main problem is how OpenCV is distributed. [Their README](https://github.com/opencv/opencv-python#installation-and-usage) explains it best, but I'll quickly summarize it here. OpenCV has 2 optional features: GUI and unstable. But instead of the optional features being add-on packages, they distribute 4 variants of OpenCV for all possible combination of features. So we get `opencv` (GUI, no unstable), `opencv-headless` (no GUI, no unstable), `opencv-contrib` (GUI, unstable), and `opencv-contrib-headless` (no GUI, unstable).

Right now, chainner works with any of those variants, which is great for compatibility with system python. If we go and require a variant with unstable features (`contrib`), then we will run into compatibility issues. We can resolve those issues in 2 ways:
1. We write nodes that use unstable features in a way that they "feature detect" what OpenCV variant is installed, and adjust accordingly.
2. We change the dependency managing code to always install a `contrib` variant.

No matter which way we go, we'll have to change how we install OpenCV. You can only install one of the variants, so we can't blindly install our preferred variant.

@joeyballentine Thoughts?

joeyballentine:
@RunDevelopment This actually made me realize that the way we currently do it could cause a conflict, since if they already have a different one of the 4 versions, we're installing another and creating an issue.

It actually would make the most sense for us to run the uninstall command for the other three, and then install the contrib version. It would ensure the most compatibility as well as give us those extra features

joeyballentine:
Is this PR good now?

RunDevelopment (approved):

PR #1552, opened 2023-02-12
Signed Distance Field Node

adodge:
A node that takes a monochrome image and returns a signed distance field, where the color of a pixel is related to the distance from that pixel to the nearest pixel of the other color, such that a threshold at 50% will restore the original image.

Can be used to create low-resolution textures that can be scaled up rendered at a higher resolution with fewer artifacts.  ([paper](https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf))

image

left: Original full size
middle: scaled down, scaled up, thresholded
right: distance transformed, scaled down, scaled up, thresholded

image

Can also be used to expand or contract a mask:

image

### DIFF
--- a/backend/src/nodes/nodes/image_filter/distance_transform.py
+++ b/backend/src/nodes/nodes/image_filter/distance_transform.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+import cv2
+import numpy as np
+
+from . import category as ImageFilterCategory
+from ...impl.dithering.common import dtype_to_uint8
+from ...impl.image_utils import as_3d
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput
+from ...properties.outputs import ImageOutput
+
+
+@NodeFactory.register("chainner:image:distance_transform")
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I wonder whether this node is too low-level. My worry right now is that nobody is going to use this awesome technique because they don't know how. I mean, I wouldn't know how without having seen this.

From what I see, the SDFs aren't that useful on their own (or maybe they are? idk). Is there a use case for just having the SDF of an image?

And maybe we could package SDFs differently. The primary use case of SDFs will here will be to enhance binary images, right? I was thinking that maybe we could have a node "Binary Upscale" that gives users the options to set an upscale scale (e.g. 4x), a smoothing level, and maybe something like border extend (to make the image's subject bolder or thinner since that is something SDF can easily do as well).

joeyballentine:
@RunDevelopment that sounds like the perfect use case for presets/node-functions that we don't have yet :)

RunDevelopment:
That is a very good point. Alright, I'm on board. We should create an issue for this though.

RunDevelopment (approved):

PR #1557, opened 2023-02-13
Node Output Caching

adodge:
I was messing around with longer pipelines and wanted to not constantly recompute earlier nodes, so I made this.  I don't know if this is something we want, or if this is the way we want to do it, but I'll submit it here for comment.

### DIFF
--- a/backend/src/nodes/node_cache.py
+++ b/backend/src/nodes/node_cache.py
@@ -4,7 +4,7 @@ import os
 import tempfile
 import time
 from enum import Enum
-from typing import List, Dict, Tuple, Optional
+from typing import List, Dict, Tuple, Optional, Union, Any
 
 import numpy as np
 from sanic.log import logger
@@ -77,9 +77,12 @@ class NodeOutputCache:
     def _enforce_limits():
         while True:
             total_bytes = sum([cache.size() for cache in CACHE_REGISTRY])
+            logger.debug(
+                f"Cache size: {total_bytes} ({100*total_bytes/CACHE_MAX_BYTES:0.1f}% of limit)"
+            )
             if total_bytes <= CACHE_MAX_BYTES:
                 return
-            logger.info("Dropping oldest cache key")
+            logger.debug("Dropping oldest cache key")
 
             oldest_keys = [
                 (cache, cache.oldest()) for cache in CACHE_REGISTRY if not cache.empty()
@@ -103,13 +106,16 @@ class NodeOutputCache:
         ]
 
     @staticmethod
-    def _output_to_list(output):
-        if not isinstance(output, (list, tuple)):
+    def _output_to_list(output) -> List:
+        if isinstance(output, list):
+            return output
+        elif isinstance(output, tuple):
+            return list(output)
+        else:
             return [output]
-        return output
 
     @staticmethod
-    def _list_to_output(output):
+    def _list_to_output(output: List):
         if len(output) == 1:
             return output[0]
         return output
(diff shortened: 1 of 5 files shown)


joeyballentine:
Oh wow, this is something that we wanted to do already, so this is great. I like the use of decorators for it as well. 

How does this clean up old caches when the inputs change?

adodge:
Nice!

> How does this clean up old caches when the inputs change?

Very lazily.  Cache entries have time that gets updated whenever they're accessed, and when a global max cache size (default 1 GiB) is reached, it drops the entries with the oldest access times until it's back under the limit.

joeyballentine:
I think we would be able to do something better than that. We know when inputs have changed on the frontend, and keep a hash stored (for frontend things). We could send that to the backend as a means of determining if it should run again. Or, when an input value changes, we could send a request to remove the cached data immediately. We actually already do that for starting nodes

RunDevelopment:
I agree with Joey. Although I am not sure how much the backend should rely on the hashes the frontend keeps. Technically, we could let the backend cache nodes aggressively the first time around, and then diff the last chain we ran with the one we want to run now. This eliminates the need for the backend to rely on the frontend for caching things correctly.

Implementing this will be quite challenging though. We will likely have to reimplement the way we currently do caching during a run.

> Or, when an input value changes, we could send a request to remove the cached data immediately.

I don't think that this is a good idea. The user might change values back to what they were before. Imagine accidentally deleting an edge somewhere at the start of the chain. Suddenly, the cache of pretty much the whole chain becomes invalidated.

joeyballentine:
Good point. In that case, maybe we could have a separate input hash system in the backend?

adodge:
Hmm...  So, do we want to invalidate the cache as soon as they change the node's inputs, or do we want to allow them to go back to an old input and have it (maybe) still be cached?  The way it's currently implemented, it should be doing the latter.

> maybe we could have a separate input hash system in the backend?

The PR does hash the node inputs to create the cache key.  It's probably different from what's happening on the frontend, (I didn't reference the typescript) but I think it's "correct" (the hash changes if and only if the node needs to be rerun.)

RunDevelopment:
> The PR does hash the node inputs to create the cache key.

To quickly explain how the frontend hashing system works: The frontend doesn't have access to the actual value of e.g. images, so it computes hashes by just looking at input values (strings and numbers). The basic algorithm is `hash(node) = sha256(node.type, node.inputs.map(i => i.isConnected ? hash(i.connectedNode) : i.value))`. If an input is connected to another node, we use the hash of that node instead of the input value.

This system works, because we assume that all nodes are pure and deterministic functions. I.e. we might not be able to know the sha256 of an upscaled image, but we do know the hashes of all values used to generate it. So we derive a hash not for the output image, but for the computation that creates that image. Since the result of the computation (in this case, the image) is uniquely determined by its input values, we can use the hash of the computation as the hash of the output.

This is why I am so vehemently against adding impure or non-deterministic nodes.

[Here is the actual implementation](https://github.com/chaiNNer-org/chaiNNer/blob/d1b275df584b393a0ab994c1ed3049e88d0d2bfc/src/renderer/hooks/useInputHashes.ts#L28) used by the frontend btw.

joeyballentine:
Also, the only reason I'm against keeping things around for longer than we need to is ram usage. If we cache to disk instead i wouldn't mind as much, but large images take up a ton of ram, so storing a cached history of nodes that produce large images is going to make chainner even more of a ram hog

adodge:
I've changed it to store cached numpy arrays in a temporary file.  This should make the RAM footprint negligible in most scenarios.  (A pathological case might be a pipeline involving lots of very small images)  I don't know if 1 GiB is a good default max size.  It's plenty as long as we're just talking about stable diffusion, because those images tend to be 10 MiB at the most.  I imagine we will want to make other slow-running nodes `@cached` as well.  An easy way of setting CACHE_MAX_BYTES in the UI would be good, but I haven't worked out how to do that yet.

joeyballentine (approved):

PR #1559, opened 2023-02-14
Add way to view external connections

joeyballentine:
Somewhat temporary solution

image
image

### DIFF
--- a/src/common/externalIntegrations.ts
+++ b/src/common/externalIntegrations.ts
@@ -1,0 +1,13 @@
+export interface Integration {
+    name: string;
+    url: string;
+    port: number;
+}
+
+export const externalIntegrations: Integration[] = [
+    {
+        name: 'AUTOMATIC1111/stable-diffusion-webui',
+        url: process.env.STABLE_DIFFUSION_HOST ?? '127.0.0.1',
+        port: Number(process.env.STABLE_DIFFUSION_PORT ?? 7860),
+    },
+];
--- a/src/renderer/contexts/DependencyContext.tsx
+++ b/src/renderer/contexts/DependencyContext.tsx
@@ -52,6 +52,11 @@ export interface DependencyContextValue {
     availableUpdates: number;
 }
 
+export interface ExternalIntegrationConnectionStatus {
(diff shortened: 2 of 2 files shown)


src/common/externalIntegrations.ts
```diff
@@ -0,0 +1,13 @@
...
+    name: string;
+    url: string;
+    port: number;
+}
+
+export const externalIntegrations: Integration[] = [
+    {
+        name: 'AUTOMATIC1111/stable-diffusion-webui',
+        url: process.env.STABLE_DIFFUSION_HOST ?? '127.0.0.1',
+        port: Number(process.env.STABLE_DIFFUSION_PORT ?? 7860),
```
RunDevelopment, line 11: You are assuming that the env variables for the electron frontend process and the python backend process are the same.

src/renderer/contexts/DependencyContext.tsx
```diff
@@ -383,6 +385,37 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
         availableUpdates,
     });
 
+    const [loadingExtInts, setLoadingExtInts] = useState(true);
+    const [externalIntegrationConnections, setExternalIntegrationConnections] = useState<
+        {
+            integration: Integration;
+            connected: boolean;
+        }[]
```
RunDevelopment, line 393: Please make the record (`{ ... }`) an interface.

RunDevelopment (changes requested):
Man, we really need that backend refactor so the backend can declare *and check* its own dependencies.

I don't like this, but I don't think that there is a better way to do it rn.

src/common/externalIntegrations.ts
joeyballentine: I am, because someone typically would start both at the same time with one executable. Also, it's not really assuming anything when we get to be the deciders of what is correct ;)

RunDevelopment (approved):
Still don't like `process.env`, but that problem will solve itself when we finally let the backend declare its own dependencies.

joeyballentine:
yeah its just a temporary solution

PR #1560, opened 2023-02-15
Add "Use Pip Directly" option in dependency manager

joeyballentine:
Also I made the view console button have text so its more obvious what it does.

This should hopefully prevent issues with people's VPNs and whatnot, as it will now default to just using pip directly. Progress bars (and thus our jank workaround for them) is now optional.

image

image

### DIFF
--- a/src/renderer/contexts/DependencyContext.tsx
+++ b/src/renderer/contexts/DependencyContext.tsx
@@ -468,6 +468,7 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
                                     <HStack>
                                         <Switch
                                             isChecked={usePipDirectly}
+                                            isDisabled={isRunningShell}
                                             onChange={() => {
                                                 setUsePipDirectly(!usePipDirectly);
                                             }}
--- a/src/renderer/contexts/DependencyContext.tsx
+++ b/src/renderer/contexts/DependencyContext.tsx
@@ -284,7 +284,7 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
     const refreshInstalledPackages = useCallback(() => setPipList(undefined), [setPipList]);
 
     const [isConsoleOpen, setIsConsoleOpen] = useState(false);
-    const [useProgressBars, setUseProgressBars] = useState(false);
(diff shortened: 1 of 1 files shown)


RunDevelopment:
> Progress bars (and thus our jank workaround for them) is now optional.

They aren't because we still use this method to install/update required dependencies. If this download method really is a problem, then changing this for optional dependencies won't do much when chainner won't even start because it can't install its required dependencies.

joeyballentine:
We don't use the one with progress to update or install the required ones. https://github.com/chaiNNer-org/chaiNNer/blob/bba5f44aaf5d6720592a5fdf28e7899a53b85aa2/src/main/backend/setup.ts#L231. it uses the regular pip command when you don't pass in an onProgress method. We've never had progress indicators for the startup deps.

RunDevelopment:
Oh, you're right. I thought we were doing there as well, but I was thinking of the download for integrated python and ffmpeg.

---

With progress bars being off by default and a warning/justification shown to the user, who is going to use progress bars anymore? Progress bars work for most of our users and offer a way better user experience. Downloading PyTorch will take multiple minutes, so not showing anything to indicate progress is far from ideal.

I would make the option to use pip directly a setting and enable progress bars by default. People with VPNs or other network issues can turn on the setting to use pip directly, but that will only be a small portion of our users.

joeyballentine:
yeah i guess that makes sense

src/renderer/contexts/DependencyContext.tsx
```diff
@@ -458,12 +464,38 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
...
-                                <IconButton
-                                    aria-label="Open Console View"
-                                    icon={<BsTerminalFill />}
-                                    size="sm"
-                                    onClick={() => setIsConsoleOpen(!isConsoleOpen)}
-                                />
+                                <HStack>
+                                    <HStack>
+                                        <Switch
+                                            isChecked={usePipDirectly}
```
RunDevelopment, line 470: We should disable the switch while (un)installing deps.

```suggestion
                                            isChecked={usePipDirectly}
                                            isDisabled={isRunningShell}
```

RunDevelopment (approved):

PR #1561, opened 2023-02-15
Make onnxoptimizer not required for m1 macs + use onnxruntime-silicon

joeyballentine:


src/common/dependencies.ts
```diff
@@ -21,6 +21,16 @@ export interface Dependency {
     description?: string;
 }
 
+const getOnnxRuntime = (canCuda: boolean) => {
```
RunDevelopment, line 24: Could we change this function to return a full PyPiPackage instead? Since onnxruntime-silicon doesn't seem to be official, I worry that the version of it and the official package diverge eventually.

Basically, I think the function should be like this:
```
if m1: return { package: 'onnxruntime-silicon', ... }
return { ...previousPackageCode }
```

This clearly shows the difference between the two.

Also, onnxruntime-silicon is 6MB :)

backend/src/nodes/nodes/onnx/convert_to_ncnn.py
```diff
@@ -36,8 +40,11 @@ def run(self, model: OnnxModel, is_fp16: int) -> Tuple[NcnnModelWrapper, str]:
         fp16 = bool(is_fp16)
 
         model_proto = onnx.load_model_from_string(model.bytes)
-        passes = onnxoptimizer.get_fuse_and_elimination_passes()
-        opt_model = onnxoptimizer.optimize(model_proto, passes)
+        if onnxoptimizer is not None:
+            passes = onnxoptimizer.get_fuse_and_elimination_passes()
+            opt_model = onnxoptimizer.optimize(model_proto, passes)
```
RunDevelopment, line 45: ONNX optimization always seems to be this. Maybe we could move this out into its own shared function to reduce the number of try-import-excepts down to one.

RunDevelopment (approved):

PR #1562, opened 2023-02-15
Lower timeout for SD check on startup

joeyballentine:
It was taking way too long to load, presumably because of the 10-minute timeout. I decided to make it really low. I'm not sure if it's too low though

### DIFF
--- a/backend/src/nodes/impl/external_stable_diffusion.py
+++ b/backend/src/nodes/impl/external_stable_diffusion.py
@@ -58,7 +58,7 @@ class ExternalServiceTimeout(Exception):
     pass
 
 
-def get(url, timeout=None) -> Dict:
+def get(url, timeout: float = STABLE_DIFFUSION_REQUEST_TIMEOUT) -> Dict:
     try:
         response = requests.get(
             url,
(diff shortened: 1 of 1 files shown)


backend/src/nodes/impl/external_stable_diffusion.py
```diff
@@ -58,9 +58,14 @@ class ExternalServiceTimeout(Exception):
     pass
 
 
-def get(url) -> Dict:
+def get(url, timeout=None) -> Dict:
```
RunDevelopment, line 61: Why not this?
```suggestion
def get(url, timeout: float = STABLE_DIFFUSION_REQUEST_TIMEOUT) -> Dict:
```

backend/src/nodes/impl/external_stable_diffusion.py
```diff
@@ -58,9 +58,14 @@ class ExternalServiceTimeout(Exception):
...
-def get(url) -> Dict:
+def get(url, timeout: float = STABLE_DIFFUSION_REQUEST_TIMEOUT) -> Dict:
     try:
-        response = requests.get(url, timeout=STABLE_DIFFUSION_REQUEST_TIMEOUT)
+        response = requests.get(
+            url,
+            timeout=timeout
+            if timeout is not None
+            else STABLE_DIFFUSION_REQUEST_TIMEOUT,
+        )
```
RunDevelopment, line 68: I find it worrying that pyright accepted this code. It really should have complained because it knows that `timeout` can never be `None`.

```suggestion
        response = requests.get(url, timeout=timeout)
```

PR #1563, opened 2023-02-15
use movflag=faststart to avoid corrupted unfinished videos

joeyballentine:


RunDevelopment (approved):
No idea what this does.

PR #1564, opened 2023-02-15
Add "Copy Audio" toggle for videos

### LINKED ISSUE #1558: Mux video audio back into video once processing is finished
Should be a toggle

joeyballentine:
Closes #1558

### DIFF
--- a/backend/src/nodes/nodes/image/video_frame_iterator.py
+++ b/backend/src/nodes/nodes/image/video_frame_iterator.py
@@ -267,14 +267,17 @@ class SimpleVideoFrameIteratorNode(IteratorNodeBase):
             out_path = writer["video_save_path"]
             if out_path is not None:
                 base, ext = os.path.splitext(out_path)
-                full_out_path = f"{base}_audio{ext}"
-                audio_stream = ffmpeg.input(path).audio
-                video_stream = ffmpeg.input(writer["video_save_path"])
-                # output_video = ffmpeg.concat(video_stream, audio_stream, v=1, a=1)
-                output_video = ffmpeg.output(
-                    audio_stream,
-                    video_stream,
-                    full_out_path,
-                    c="copy",
-                ).overwrite_output()
-                ffmpeg.run(output_video)
+                if "gif" not in ext.lower():
+                    full_out_path = f"{base}_audio{ext}"
+                    audio_stream = ffmpeg.input(path).audio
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -251,3 +256,28 @@ def before(index: int):
             )
 
         await context.run_while(frame_count, before, fail_fast=True)
+
+        ffmpeg_reader.stdout.close()
+        ffmpeg_reader.wait()
+        if writer["out"] is not None:
+            writer["out"].stdin.close()
+            writer["out"].wait()
```
joeyballentine, line 264: Btw, I think these might help fix the video corruption stuff, potentially. It was necessary to add otherwise the video stream might not be closed before we try to mux the audio

backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -251,3 +256,28 @@ def before(index: int):
...
+
+        if writer["copy_audio"]:
```
RunDevelopment, line 266: I just noticed. Why is `writer` an untyped dict and not something like a dataclass instance?

backend/src/nodes/nodes/image/video_frame_iterator.py
```diff
@@ -251,3 +256,28 @@ def before(index: int):
...
+            out_path = writer["video_save_path"]
+            if out_path is not None:
+                base, ext = os.path.splitext(out_path)
+                if "gif" not in ext.lower():
+                    full_out_path = f"{base}_audio{ext}"
+                    audio_stream = ffmpeg.input(path).audio
+                    video_stream = ffmpeg.input(writer["video_save_path"])
```
RunDevelopment, line 273:
```suggestion
                    video_stream = ffmpeg.input(out_path)
```

backend/src/nodes/nodes/image/video_frame_iterator.py
joeyballentine: It exists only to transfer data between the iterator and the helper. It's kinda shitty but it works

RunDevelopment: We should fix that after this PR.

RunDevelopment (approved):

PR #1565, opened 2023-02-15
Dilate/Erode node

harisreedhar:
Dilate - Erode operation
chaiNNer-Untitled-2023-2-15_11-43

morph shapes(Rectangle, Ellipse, Cross)
chaiNNer-Untitled-2023-2-15_11-45

### DIFF
--- a/backend/src/nodes/nodes/image_utility/dilate.py
+++ b/backend/src/nodes/nodes/image_utility/dilate.py
@@ -1,0 +1,69 @@
+from __future__ import annotations
+from enum import Enum
+
+import cv2
+import numpy as np
+
+from . import category as ImageUtilityCategory
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput, EnumInput
+from ...properties.outputs import ImageOutput
+
+
+class MorphShape(Enum):
+    RECTANGLE = cv2.MORPH_RECT
+    ELLIPSE = cv2.MORPH_ELLIPSE
+    CROSS = cv2.MORPH_CROSS
+
+
+@NodeFactory.register("chainner:image:dilate")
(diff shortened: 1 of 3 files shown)


RunDevelopment (changes requested):
Thank you for the PR @harisreedhar!

I wonder whether we should give the node a different name? I had to google what dilate and erode even mean in terms of image processing... 
They turned out to be a simple convolution with min/max. Maybe we could call the node "Min/Max" instead? Photoshop calls these filter "Minimum" and "Maximum" as well. Maybe we could even make 2 nodes for this, like PS. What do you think?

joeyballentine:
I wouldn't just call them "minimum" and "maximum". Maybe "min/max filter" to make it clear that it isn't just a math min/max

harisreedhar:
> Thank you for the PR @harisreedhar!
> 
> I wonder whether we should give the node a different name? I had to google what dilate and erode even mean in terms of image processing... 
> They turned out to be a simple convolution with min/max. Maybe we could call the node "Min/Max" instead? Photoshop calls these filter "Minimum" and "Maximum" as well. Maybe we could even make 2 nodes for this, like PS. What do you think?

Gimp & Blender uses the term ``dilate/erode``.
``Minimum/Maximum`` feels too much generalized🤔
What about ``shrink/grow``?

RunDevelopment:
> Gimp & Blender uses the term `dilate/erode`.

Oh, I didn't know that. In that case, let's keep the name.

joeyballentine:
Agreed

harisreedhar:
@RunDevelopment Should i need to make two separate nodes for each operation?

RunDevelopment:
Yes, I think that would be good.

RunDevelopment (approved):
Thank you @harisreedhar!

PR #1567, opened 2023-02-16
Fix `run dev` script

RunDevelopment:
As discussed on Discord. This PR replaces `localhost` with `127.0.0.1` to work around NodeJS 18 and removes the now unnecessary `cross-fetch` dependency.


joeyballentine (approved):
RunDevelopment runs to develop a fix for run dev

PR #1568, opened 2023-02-17
Improved `auto_split` for PixTransform

RunDevelopment:
To add tiling to #1506, I need 2 new features in `auto_split`:

1. Support for splitting in exact size mode. PixTransform requires square images, but we can largely choose the size of the square.
2. Region information. Since guided upsampling has a guide image, we need to be able to map the region of the current tile to its region in the guide image.

This PR adds those 2 features and makes a few additional changes:
- Added support for splitting in exact size mode. This is done in a simple loop where we repeatedly start from scratch when the tile size is changed.
- Added region information. `auto_split`'s upscale function parameter now takes a `Region` values. This is the location of the current tile within the given image.
- Added support for non-square tile sizes. By simplifying the `Tiler` interface, this just kinda happened. The underlying split implementations already supported non-square tile sizes internally, so this was easy to implement.
- Renamed a few module-scoped functions and variables to not export them.

I made this its own PR, because `auto_split` is very important, so I'd like us to test this PR in isolation.

### DIFF
--- a/backend/src/nodes/impl/ncnn/auto_split.py
+++ b/backend/src/nodes/impl/ncnn/auto_split.py
@@ -34,7 +34,7 @@ def ncnn_auto_split(
     staging_vkallocator,
     tiler: Tiler,
 ) -> np.ndarray:
-    def upscale(img: np.ndarray):
+    def upscale(img: np.ndarray, _):
         ex = net.create_extractor()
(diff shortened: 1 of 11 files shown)


joeyballentine:
Will test this soon

joeyballentine (approved):

PR #1569, opened 2023-02-17
PyTorch Inpaint Node

theflyingzamboni:
Adds a PyTorch inpainting node for models like LaMa (only one supported currently).


backend/src/nodes/impl/pytorch/model_loading.py
```diff
@@ -1,15 +1,16 @@
 from sanic.log import logger
 
+from .architecture.face.codeformer import CodeFormer
+from .architecture.face.gfpganv1_clean_arch import GFPGANv1Clean
+from .architecture.face.restoreformer_arch import RestoreFormer
+from .architecture.HAT import HAT
+from .architecture.LaMa import LaMa
 from .architecture.RRDB import RRDBNet as ESRGAN
```
RunDevelopment, line 8: Man, we need something that enforces sorted imports. What are you using? It seems to be quite good.

backend/src/nodes/nodes/pytorch/inpaint.py
```diff
@@ -0,0 +1,119 @@
...
+            ImageInput(label="Mask", channels=1),
+            InpaintModelInput(),
+        ]
+        self.outputs = [
+            ImageOutput(
+                image_type=expression.Image(
+                    width="Input0.width & Input1.width",
+                    height="Input0.height & Input1.height",
+                    channels="Input0.channels",
+                )
```
RunDevelopment, line 40: We know that input0 always is RGB, so the output is also RGB. It's better to use the `channels` parameter of the `ImageOutput` in that case, since it will check the type at runtime.

```suggestion
                image_type=expression.Image(
                    width="Input0.width & Input1.width",
                    height="Input0.height & Input1.height",
                ),
                channel=3,
```

backend/src/nodes/nodes/pytorch/inpaint.py
```diff
@@ -0,0 +1,119 @@
...
+            mask_tensor = np2tensor(mask, change_range=True)
+
+            d_img = None
+            d_mask = None
+            try:
+                d_img = img_tensor.to(device)
+                d_img = d_img.half() if use_fp16 else d_img.float()
+
+                d_mask = mask_tensor.to(device)
+                d_mask = (d_mask > 0) * 1
```
RunDevelopment, line 74: This is intended to binarize the mask, right? If so, could we use 0.5 as the threshold? If the mask was created by another AI, then it might contain a tiny bit of noise that would be amplified by a threshold of 0.

backend/src/nodes/impl/pytorch/model_loading.py
RunDevelopment: It's probably isort, right?

backend/src/nodes/nodes/pytorch/inpaint.py
theflyingzamboni: Seems like it. This is just pulled directly from the lama code. We could make it 0.5 if we want though, I guess.

backend/src/nodes/impl/pytorch/model_loading.py
theflyingzamboni: I think it's isort? It's whatever VSCode uses when I do right click > sort imports, and at some point the isort extension got installed. I think maybe the python extension started installing isort alongside itself?

backend/src/nodes/nodes/pytorch/inpaint.py
theflyingzamboni: lol I shouldn't have committed this suggestion directly. I didn't notice you had a typo. 😄I've been led astray!

RunDevelopment (approved):

PR #1574, opened 2023-02-18
Load state dicts from pt models

theflyingzamboni:
Also added link to big-lama.pt

### DIFF
--- a/backend/src/nodes/nodes/pytorch/load_model.py
+++ b/backend/src/nodes/nodes/pytorch/load_model.py
@@ -51,11 +51,18 @@ class LoadModelNode(NodeBase):
 
         try:
             logger.debug(f"Reading state dict from path: {path}")
-            state_dict = torch.load(
-                path,
-                map_location=torch.device(exec_options.full_device),
-                pickle_module=RestrictedUnpickle,  # type: ignore
-            )
+
+            if os.path.splitext(path)[1] == ".pt":
+                state_dict = torch.jit.load(  # type: ignore
+                    path, map_location=torch.device(exec_options.full_device)
+                ).state_dict()
+            else:
+                state_dict = torch.load(
+                    path,
+                    map_location=torch.device(exec_options.full_device),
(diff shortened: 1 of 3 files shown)


backend/src/nodes/nodes/pytorch/load_model.py
```diff
@@ -51,11 +52,18 @@ def run(self, path: str) -> Tuple[PyTorchModel, str, str]:
 
         try:
             logger.debug(f"Reading state dict from path: {path}")
-            state_dict = torch.load(
-                path,
-                map_location=torch.device(exec_options.full_device),
-                pickle_module=RestrictedUnpickle,  # type: ignore
-            )
+
+            if os.path.splitext(path)[1] == ".pt":
```
joeyballentine, line 56: is there a way we can check that it's a torchscript file without the extension? Technically there's nothing preventing anyone from saving a normal model as .pt or a JIT one as .pth, so a more robust check based on the file content would be much better imo

theflyingzamboni: I do not know. Seems kind of hard to do that without opening it first.

theflyingzamboni: I can think of a couple things. One is to attempt torch load, and then try jit load if it errors. The other is regex. TorchScript models have some extra files and folders that pth files don't have, including an extra pkl file (constants.pkl). Searching that string might be enough.

theflyingzamboni: Nevermind, loading a torchscript with torch load only fails if an operation isn't implemented, otherwise it's just a warning.

joeyballentine: This is fine then

theflyingzamboni: It will *probably* be fine. We can always revisit it if we have to, but I really don't know what to do other than read the file and regex it.

joeyballentine (approved):

PR #1575, opened 2023-02-19
Change color and icon of SD category + fix hardcoded accent colors

joeyballentine:


src/renderer/components/NodeSelectorPanel/RepresentativeNode.tsx
```diff
@@ -26,8 +27,11 @@ export const RepresentativeNode = memo(
         collapsed = false,
         createNodeFromSelector,
     }: RepresentativeNodeProps) => {
+        const { categories } = useContext(BackendContext);
+
         const bgColor = 'var(--selector-node-bg)';
-        const accentColor = getNodeAccentColor(category);
+        const currentCategory = categories.find((c) => c.name === category);
+        const accentColor = currentCategory?.color ?? '#CCCCCC';
```
RunDevelopment, line 34: The fallback color will be pretty much invisible in light mode.

RunDevelopment (commented):
Instead of duplicating `categories.find((c) => c.name === category)?.color ?? '#CCCCCC'` everywhere, could you extract it into a function `getNodeAccentColor(category: string | undefined) -> colorString` in `BackendContext`?

src/renderer/helpers/accentColors.ts
```diff
@@ -39,3 +40,7 @@ export const getTypeAccentColors = (inputType: Type): readonly [string, ...strin
     }
     return colors.length > 0 ? (colors as [string, ...string[]]) : defaultColorList;
 };
+
+export const getCategoryAccentColor = (categories: Category[], category: string) => {
```
RunDevelopment, line 44: We don't mutate the array.

```suggestion
export const getCategoryAccentColor = (categories: readonly Category[], category: string) => {
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -48,16 +48,16 @@ import { useThemeColor } from './useThemeColor';
 interface SchemaItemProps {
     schema: NodeSchema;
     isFavorite?: boolean;
+    accentColor?: string;
```
RunDevelopment, line 51: Why is this optional? We always pass in an accent color.

joeyballentine:
I'm just going to assume this is good now

PR #1582, opened 2023-02-20
Fix default size for textarea inputs

joeyballentine:
This ended up being a much easier fix than I thought. Turns out, the default value we were using was not the minimum value of the textarea, but a much larger value. I guess the default size when size was undefined would just be the minimum, but then the size would get set by the useEffect, so on refresh it was setting it to that default size we were setting, hence it getting randomly bigger.

So, I reduced the default size down and set a null condition for the default size.


theflyingzamboni:
It's working for me.

PR #1583, opened 2023-02-20
Switch SD text inputs to TextAreas

joeyballentine:
The only thing about this is the two textareas can have different widths, which ends up looking super weird. Not sure what to do about that

### DIFF
--- a/src/common/common-types.ts
+++ b/src/common/common-types.ts
@@ -85,6 +85,7 @@ export interface TextInput extends InputBase {
 export interface TextAreaInput extends InputBase {
     readonly kind: 'text';
     readonly resizable: boolean;
+    readonly def: string | null;
 }
 export interface NumberInput extends InputBase {
     readonly kind: 'number';
--- a/src/common/common-types.ts
+++ b/src/common/common-types.ts
@@ -82,7 +82,7 @@ export interface TextInput extends InputBase {
     readonly placeholder?: string | null;
     readonly def?: string | null;
 }
-export interface NoteTextAreaInput extends InputBase {
+export interface TextAreaInput extends InputBase {
     readonly kind: 'text';
     readonly resizable: boolean;
 }
@@ -118,7 +118,7 @@ export type Input =
     | FileInput
     | DirectoryInput
     | TextInput
-    | NoteTextAreaInput
+    | TextAreaInput
     | DropDownInput
     | SliderInput
     | NumberInput;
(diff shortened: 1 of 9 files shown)


joeyballentine:
Yeah i tried to do that but since the sizes aren't controlled, it didn't update until after refreshing

RunDevelopment:
Can we make them controlled?

joeyballentine:
I think so, but I'm not entirely sure what the performance impact of that would be. I can try it.

I'd preferably like to handle that with a different PR. I don't think it looking a bit jank is worth holding this back, since prompts can get pretty large and the limited size of regular text boxes definitely makes the experience worse for using these nodes

RunDevelopment (approved):

joeyballentine:
Fixed it

src/common/common-types.ts
```diff
@@ -82,7 +82,7 @@ export interface TextInput extends InputBase {
     readonly placeholder?: string | null;
     readonly def?: string | null;
 }
-export interface NoteTextAreaInput extends InputBase {
+export interface TextAreaInput extends InputBase {
     readonly kind: 'text';
     readonly resizable: boolean;
 }
```
RunDevelopment, line 88:
```suggestion
export interface TextAreaInput extends InputBase {
    readonly kind: 'text';
    readonly resizable: boolean;
    readonly def: string | null;
}
```

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -964,10 +964,17 @@ export const GlobalProvider = memo(
                 const setInputSize = (size: Readonly<Size>) => {
                     modifyNode(id, (old) => {
                         const nodeCopy = copyNode(old);
-                        nodeCopy.data.inputSize = {
+                        const newInputSize: Record<string, Size> = {
```
RunDevelopment, line 967: They may all be strings under the hood, but let's pretend that there's more structure to it.

```suggestion
                        const newInputSize: Record<InputId, Size> = {
```

joeyballentine: I'll commit that but I'm guessing it'll error. Maybe not though, since ultimately i did end up changing the way i did this after fixing that issue

RunDevelopment: Argh, you're right. Object.entries returns strings. Yeah, my suggestion was wrong, it has to be string. My ad.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -964,10 +964,17 @@ export const GlobalProvider = memo(
                 const setInputSize = (size: Readonly<Size>) => {
                     modifyNode(id, (old) => {
                         const nodeCopy = copyNode(old);
-                        nodeCopy.data.inputSize = {
+                        const newInputSize: Record<InputId, Size> = {
```
joeyballentine, line 967:
```suggestion
                        const newInputSize: Record<string, Size> = {
```

RunDevelopment (approved):

PR #1585, opened 2023-02-20
Value/Simplex Noise

adodge:
Node to generate noise textures.

image

### DIFF
--- a/backend/src/nodes/nodes/image/create_noise.py
+++ b/backend/src/nodes/nodes/image/create_noise.py
@@ -4,20 +4,20 @@ from enum import Enum
 
 import numpy as np
 
-from . import category as ImageUtilityCategory
-from ...impl.image_utils import cartesian_product
-from ...impl.noise_functions.simplex import SimplexNoise
-from ...impl.noise_functions.value import ValueNoise
-from ...node_base import NodeBase, group
-from ...node_factory import NodeFactory
-from ...properties import expression
-from ...properties.inputs import (
+from . import category as ImageCategory
+from nodes.impl.image_utils import cartesian_product
+from nodes.impl.noise_functions.simplex import SimplexNoise
+from nodes.impl.noise_functions.value import ValueNoise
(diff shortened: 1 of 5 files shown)


theflyingzamboni:
Quick note, we've recategorized all basic image-generating nodes into Image > Make Images.

backend/src/nodes/nodes/image/create_noise.py
```diff
@@ -0,0 +1,198 @@
...
+                {
+                    "enum": 6,
+                    "conditions": [
+                        FractalMethod.PINK.value,
+                        FractalMethod.PINK.value,
+                        FractalMethod.PINK.value,
+                        FractalMethod.PINK.value,
+                    ],
+                },
+            )(
```
RunDevelopment, line 72: I'd recommend using the new `conditional_group` function.
```suggestion
            conditional_group(enum=6, condition=FractalMethod.PINK.value)(
```

backend/src/nodes/nodes/image/create_noise.py
```diff
@@ -0,0 +1,198 @@
...
+                BoolInput("Increment Seed", default=True).with_id(13),
+            ),
+        ]
+        self.outputs = [
+            ImageOutput(
+                image_type=expression.Image(
+                    width="Input0",
+                    height="Input1",
+                    channels="1",
+                )
```
RunDevelopment, line 89: Using the channels parameter on `ImageOutput` is preferred.
```suggestion
                image_type=expression.Image(
                    width="Input0",
                    height="Input1",
                ),
                channels=1,
```

RunDevelopment:
About the CI: I recently added isort to organize imports. If you are using VSCode, just install the extension and hit Ctrl+S on the offending file.

backend/src/nodes/nodes/image/create_noise.py
adodge: Oh, cool!

joeyballentine (approved):

PR #1586, opened 2023-02-21
Add "tiling" option to text2img and img2img

adodge:
image

image


joeyballentine:
Is it just labeled as "tiling" in the actual webui? I kinda wonder if we should call it something else like "seamless edges (tiled textures)" or something, to differentiate it from the kind of tiling done during upscaling. But it's not really that important.

Neat feature 👍

adodge:
Good point.  That is what it's called in the webui, but it is actually a bad name.  I changed it to "Seamless Edges"

joeyballentine:
Is this also an option on img2img?

adodge:
It is, and it seems to work there too.

Without seamless textures:

image

With seamless textures:

image

joeyballentine (approved):

PR #1588, opened 2023-02-22
Auto-detect SD webui endpoint

adodge:
Evidently, it is possible to run `webui.sh` with TLS, so people may need to configure chaiNNer to use HTTPS to find it.  This adds that capability, via an environment variable.

### DIFF
--- a/backend/src/nodes/impl/external_stable_diffusion.py
+++ b/backend/src/nodes/impl/external_stable_diffusion.py
@@ -4,7 +4,7 @@ import base64
 import io
 import os
 from enum import Enum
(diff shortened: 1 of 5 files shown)


joeyballentine:
Should we instead just check over both http and https and use whichever one works?

adodge:
Oh, yeah.  That would make sense.  Could also try both ports 7860 (normal) and 7861 (`--nowebui`)

joeyballentine:
Yeah that sounds good to me

joeyballentine (approved):

PR #1602, opened 2023-02-26
Internal Stable Diffusion

adodge:
- Proof of concept and request for comment
- Using a fork of ComfyUI that's packaged as a library (https://github.com/adodge/ComfyLib)
  - Currently you need to manually install from git, but I'll put it in PyPi if this looks viable
- Need to work on the Navi input/output types and the renderer components
- Would be nice to cache checkpoints between runs.  loading the model can take 5-10 seconds
- Needs a higher-level node that does something like the exsting external "Text to Image"
- There are lots of other nodes in ComfyUI that could be ported over

image


adodge:
Did some work on the library.  Better VRAM memory management options.  Rearranged the node inputs to have the image and model as the first two inputs.

image
foo

Stuff to do before I mark this ready to review:

- [ ] more specific input/output types and frontend stuff to, e.g., show the resolution of latent images or version of models
- [ ] make monolithic "text to image" and "image to image" nodes
- [ ] make sure this works for v2.x stable diffusion models, or make sure it emits a useful error
- [ ] put the library on PyPi and link it in the dependency manager

adodge:
image

`text to image` and `image to image` nodes

adodge:
- [ ] Automatically round down sizes to the nearest valid values, like we do for the external SD

adodge:
image

Types for SD1/SD2 models, as well as latent image sizes.

backend/src/nodes/nodes/stable_diffusion/__init__.py
```diff
@@ -0,0 +1,9 @@
+from ...category import Category
+
+category = Category(
+    name="Stable Diffusion",
+    description="Nodes for using Stable Diffusion",
+    icon="PyTorch",
+    color="#DD6B20",
```
joeyballentine, line 7: IMO these should probably be a different icon & color (like the webui ones). While they are still PyTorch, I think it's separate enough where it makes sense to make it unique. The webui nodes I made purple, which coincidentally also happens to be the same color as [pytorch lightning](https://lightning.ai/docs/pytorch/stable/). Might at least be good to use the lightning logo for the load checkpoint node.

backend/src/nodes/nodes/stable_diffusion/clip_encode.py
```diff
@@ -0,0 +1,46 @@
...
+        self.category = StableDiffusionCategory
+        self.name = "CLIP Encode"
+        self.icon = "PyTorch"
+        self.sub = "Conditioning"
+
+    @torch.no_grad()
+    def run(self, clip: CLIPModel, prompt: Optional[str]) -> Conditioning:
+        prompt = prompt or ""
+        try:
+            clip.cuda()
```
joeyballentine, line 42: instead of doing `.cuda()`, could you do `.to(torch.device(exec_options.full_device))` like we do elsewhere? This not only ensures it goes on the right GPU but also allows things like DML to work (if someone has that manually set up)

backend/src/nodes/nodes/stable_diffusion/clip_encode.py
```diff
@@ -0,0 +1,46 @@
...
+            out = clip.encode(prompt)
+        finally:
+            clip.cuda()
```
joeyballentine, line 45: what's the point of this being here twice?

EDIT: I think this was meant to be a .cpu() call based on the others below

backend/src/nodes/nodes/stable_diffusion/image2image.py
```diff
@@ -0,0 +1,165 @@
...
+        if width1 != width or height1 != height:
+            input_image = resize(
+                input_image, (width1, height1), InterpolationMethod.AUTO
+            )
+
+        positive = positive or ""
+        negative = negative or ""
+
+        try:
+            vae.cuda()
```
joeyballentine, line 129: ditto the above about the .cuda calls

backend/src/nodes/nodes/stable_diffusion/image2image.py
```diff
@@ -0,0 +1,165 @@
...
+            latent = vae.encode(RGBImage.from_array(input_image, device="cuda"))
+        finally:
+            vae.cpu()
+
+        try:
+            clip.cuda()
```
joeyballentine, line 135: I'm not gonna mark all these lol

backend/src/nodes/nodes/stable_diffusion/vae_encode.py
```diff
@@ -0,0 +1,60 @@
...
+        width1, height1 = nearest_valid_size(
+            width, height, step=64
+        )  # This cooperates with the "image_type" of the ImageOutput
+
+        if width1 != width or height1 != height:
+            image = resize(image, (width1, height1), InterpolationMethod.AUTO)
+
+        try:
+            vae.cuda()
+            img = RGBImage.from_array(image, device="cuda")
```
joeyballentine, line 56: same thing with the device param here. Might be more complex than just changing the .cuda calls, but can probably just pass in the execution options device

backend/src/nodes/properties/inputs/file_inputs.py
```diff
@@ -105,6 +107,28 @@ def PthFileInput(primary_input: bool = False) -> FileInput:
...
+
+def StableDiffusionPtFileInput(primary_input: bool = False) -> FileInput:
+    """Input for submitting a local stable diffusion CLIP or VAE model file"""
+    return FileInput(
+        input_type_name="StableDiffusionPtFile",
+        label="Model",
+        file_kind="pt",
+        filetypes=[".ckpt", ".pt", ".bin", ".pth", ".safetensors"],
+        primary_input=primary_input,
+    )
```
joeyballentine, line 129: Is there any reason these can't just be all combined into the existing Load Model node? At least eventually? IMO it would be ideal to just have one node that can load all these file types and figure out based on model metadata what kind of thing it is. For example right now we don't have different Load Model nodes for each arch we support, we have one that loads any .pth or .pt and figures out which arch it is based on the contents of the model. (This puts the burden of figuring out what the model is on chaiNNer rather than on the user). Plus, any type restrictions off of them can be done still via Navi types.

src/common/dependencies.ts
```diff
@@ -125,6 +125,34 @@ export const getOptionalDependencies = (isNvidiaAvailable: boolean): Dependency[
             description:
                 'ONNX uses .onnx models to upscale images. It also helps to convert between PyTorch and NCNN. It is fastest when CUDA is supported. If TensorRT is installed on the system, it can also be configured to use that.',
         },
+        {
+            name: 'Stable Diffusion',
+            packages: [
+                {
+                    packageName: 'comfylib',
+                    version: '0.0.5',
+                    sizeEstimate: 3 * GB,
```
joeyballentine, line 134: What exactly makes this 3GB? We aren't downloading torch again through this are we?

src/renderer/components/outputs/StableDiffusionModelOutput.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { BackendContext } from '../../contexts/BackendContext';
+import { GlobalContext, GlobalVolatileContext } from '../../contexts/GlobalNodeState';
+import { TypeTags } from '../TypeTag';
+import { OutputProps } from './props';
+
+interface StableDiffusionModelData {
+    arch: string;
+}
+
+export const StableDiffusionModelOutput = memo(
```
joeyballentine, line 16: This isn't anything you need to worry about, I just wanna say this would have been a whole lot easier if this was just a single TaggedOutput. I should probably open an issue for doing that

joeyballentine (commented):
A few initial comments. In general this looks pretty good and clean, and I like that comfylib did most of the heavy lifting so we didn't have to have much code here.

backend/src/nodes/impl/external_stable_diffusion.py
```diff
@@ -120,8 +120,8 @@ def post(path, json_data: Dict) -> Dict:
     return response.json()
 
 
-def nearest_valid_size(width, height):
-    return (width // 8) * 8, (height // 8) * 8
+def nearest_valid_size(width, height, step=8):
```
RunDevelopment, line 123: Since non-external-SD nodes are now using this mehod, it might be good to make this a util.util method.

backend/src/nodes/nodes/stable_diffusion/clip_encode.py
```diff
@@ -0,0 +1,46 @@
...
+from ...properties.inputs.stable_diffusion_inputs import CLIPModelInput
+from ...properties.outputs.stable_diffusion_outputs import ConditioningOutput
+from . import category as StableDiffusionCategory
+
+
+@NodeFactory.register("chainner:stable_diffusion:clip_encode")
+class CLIPEncodeNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = ""
```
RunDevelopment, line 20: So, what does it do? :)

src/renderer/components/outputs/StableDiffusionModelOutput.tsx
RunDevelopment: I half-finished a type broadcast PR a few days ago. I should probably finish it. Then we wouldn't need those output kinds at all, since the python outputs could just declare a type broadcast, which would automatically be handled.

joeyballentine: @adodge Since #1681 is now merged, I believe all of these new frontend output components can be replaced by a tag type broadcast

joeyballentine:
Note: this will now need to be updated to the new package node importing & function-based node format. I can do this myself if you'd prefer

adodge:
> Note: this will now need to be updated to the new package node importing & function-based node format. I can do this myself if you'd prefer

That would be very nice, if it's not too much trouble.  Sorry this is taking so long for me to get back to.  Hoping for this week.

adodge:
I'm working on this today, so I'll take a crack at rebasing it.

joeyballentine:
Sorry, been working on other things and haven't gotten a chance to make that change. If you still want me to do it i can do it today

adodge:
I think it shouldn't take me too long.  Will be a good chance to familiarize myself with the new stuff

adodge:
Rebase done.  Haven't addressed any of the other feedback, but it works again now.  I like the new system!

backend/src/nodes/properties/inputs/file_inputs.py
adodge: I think it's different from having different arches, since each of the model types (SD, CLIP, VAE) is used in a different set of nodes.  There's no "apply model" node that works with any of them, like there is for the pytorch models.  We could still one one type that covers everything and do the rest with type properties or whatever.  I have no opinion either way.

We could have one "Load Model" node that loads any model and returns the appropriate type.  That will come in handy when there's more types of models to load:  LORAs, controlnet things, etc.

The only complication, and the reason there's two different inputs here, is that "Checkpoints" actually have three different models in, so I think need a separate node.  Checkpoints seem to have one set of possible extensions, and others (VAEs, CLIPs, LORAs, etc) have another.

src/common/dependencies.ts
adodge: There is a bunch of other junk that's pulled in, directly or indirectly.  I should do a pass on the comfy code and figure out if there's any dependencies  we can drop.  I believe it should not re-download pytorch, if it's already installed.

With a fresh venv, after installing comfylib the `lib` directory is 4.6GB, and contains these libraries:  https://gist.github.com/adodge/b6722ab4d2229562ef5e2663a85e1daf

backend/src/nodes/nodes/stable_diffusion/__init__.py
adodge: Yeah, I just used the color and icon of whatever node I copied to get started with.    I changed the color to match the external SD one, and changed the icons to something suitably vague.  Hard to find good icons to represent the things these nodes do, at least without designing our own icons.  I tried to use the `SiPytorchLightling` icon from here https://react-icons.github.io/react-icons/search?q=lightning, but it doesn't seem to be part of the icon library we're using.

backend/src/nodes/nodes/stable_diffusion/clip_encode.py
adodge: An excellent question! lol

I took a stab at some hopefully useful descriptions

RunDevelopment: Looking good! Thanks!

backend/src/packages/chaiNNer_stable_diffusion/stable_diffusion/high_level/txt2img.py
```diff
@@ -0,0 +1,142 @@
...
+            "Height",
+            minimum=64,
+            default=512,
+            maximum=2048,
+            slider_step=64,
+            controls_step=64,
+        ),
+        TextAreaInput("Prompt").make_optional(),
+        TextAreaInput("Negative Prompt").make_optional(),
+        group("seed")(NumberInput("Seed", minimum=0, default=42, maximum=4294967296)),
```
RunDevelopment, line 58: We now have a seed input that gives you a seed object. See this example: https://github.com/chaiNNer-org/chaiNNer/blob/04ba6fe7454de042b0351c109c65aed4ef6fc32f/backend/src/packages/chaiNNer_standard/image_filter/noise/add_noise.py#L49

You can then use `seed.to_u32()` to get a number within limits. https://github.com/chaiNNer-org/chaiNNer/blob/04ba6fe7454de042b0351c109c65aed4ef6fc32f/backend/src/nodes/utils/seed.py#L28

backend/src/nodes/nodes/stable_diffusion/__init__.py
joeyballentine: We probably need to update the package

src/common/dependencies.ts
joeyballentine: What's that `nvidia` package?

backend/src/nodes/nodes/stable_diffusion/clip_encode.py
adodge: Done!

backend/src/nodes/nodes/stable_diffusion/__init__.py
adodge: Thank you!

backend/src/nodes/nodes/stable_diffusion/vae_encode.py
adodge: So, I went and replaced the `.cuda()`s with `.to(exec_options.full_device)`.  For some reason, that seems to be set to "cpu" for me.  I don't have the CPU Mode setting set in the options.  I'm not really sure how it's getting set.

adodge: The normal PyTorch nodes seem to be running on GPU.  Odd...

joeyballentine: My guess is that since this isn't pytorch directly (I'm not even sure what this is) it probably doesn't support that syntax. I'd say just check if the device is CUDA and if so do this, otherwise do CPU

adodge: The `.to` calls are getting passed transparently into pytorch.  When I stick a breakpoint in the sampler node, `exec_options.full_device` is set to `cpu`.   When I put the breakpoint in the PyTorch load model node, it's set to `cuda`.  I don't really know why that would be, and I haven't had a chance to dig into it.  It looks like it's supposed to be a global.

image
image

src/common/dependencies.ts
adodge: hrm.  I'm not sure.  That doesn't seem to be installed anymore.  it's possible I didn't do that experiment correctly.

Here's the list of dependencies for comfylib, and all their dependencies, courtesy of `pipdeptree`:  https://gist.github.com/adodge/de2e1686f0ec19727c44bd905b3c119e

When I get a chance, I'll go through and do a dropout test to see which of these are actually being used by the code integrated in chainner.

adodge: When I"m done, I'll add a specific version of all the packages it installs to `dependencies.ts`.

MElhagaly:
@joeyballentine , I came across your tool and discovered this PR when looking for similar functionality. Is the PR still a Work in Progress or is it stale ?

joeyballentine:
@MElhagaly at this point, this PR is stale as adodge never finished it and I don't have the time to redo the work that went into it to fix it. 

If you want node-based SD, just use comfyui

MElhagaly:
@joeyballentine I was considering picking up this PR and finish it. At least, if the functionality is deemed interesting and relevant to the community. 

Thoughts?

joeyballentine:
To be honest, I don't think having SD functionality in chaiNNer is even worth it. I don't have the time to keep up with the constantly changing landscape and constant new features popping up in that space. Even if this PR were to be finished, it would be massively out of date and wouldn't support most of the things people have come to expect in an SD application in the time since this PR was first made. It would take a significant amount of effort to implement all that new stuff on top of this (not to mention you'd need to update ComfyLib first).

Maybe some day once chaiNNer supports extensions/third party nodes, someone can make and maintain their own SD extension independently. That would probably be the most ideal scenario. 

So, if you were to finish this out, that would be great, but it would be sorta a maintenance hell for us unless you're willing to stick with it and keep it up to date for us. And that would include maintaining some version of ComfyLib (if you were to finish out that project as well). If you're willing to do that and can commit to at least keeping it up to date every once in a while, then I say go for it (and I would very much appreciate it).

I don't remember exactly where this left off, but I think one of the biggest blockers were all of the unnecessary dependencies ComfyLib was pulling in. In order for this to be considered, it would need to significantly reduce that.

MElhagaly:
Thank you for your detailed reply. I will take a look at this PR, and try and get it working. I see what you mean with ComfyLib having a huge amount of dependencies. My workflow will be to get everything up and running using @adodge's work. Then see what can eliminated. 

I already noticed that Dependency management has been moved to the backend since the start of this PR. So I will try and evaluate what other changes have been pushes as well.
If you have a chance, you could save my sometime, if you can summarize other major changes that have taken place since. 

In the meantime, I will get started and keep you updated.

joeyballentine:
I think dependency definitions being moved to the backend is the only major thing that changed since then, but I could be forgetting things. After a quick glance at the code changes for this, that seems like the only thing thats super out of date (in terms of chaiNNer code that is)

MElhagaly:
@joeyballentine quick update: I have managed to get this PR to work on the latest version of ChaiNNer. It works as expected. I haven't implemented the dependencies part yet, mainly to figure out how many dependency installs are required. As you stated, it is a lot of (unnecessary) dependencies. 

Next steps for me are:
- Implement dependency installation 
- Fork and Extend ComfyLib to be more of a wrapper/interface API around ComfyUI. Ideally we would want ChaiNNer to simply be extended with ComfyUI Chainner without the tight coupling. What do you think?

I am willing to dedicate some time to get the base up and running as well as maintain it in the long run. 

You already mentioned a future functionality where chaiNNer would allow third party packages / nodes. 
We can implement this for now and once that functionality is implemented we can extract the package from chaiNNer and host it in a separate repo.

Any thoughts?

joeyballentine:
@MElhagaly that sounds good to me. Thanks for the update and I appreciate you being willing to work on this 🙏

joeyballentine:
At this point, this PR is too outdated and it is still unfinished and relies on a package we do not own. 

This will need to be revisited in the future if we ever want SD support.

Thanks for the work that went into it even though we will not be using it.

PR #1603, opened 2023-02-27
percent operator for math node

adodge:
image


RunDevelopment:
All of these are unary operators. Wouldn't it make more sense to conditionally hide the b operate instead of trying to involve it somehow?

adodge:
"Percent" comes from a [suggestion in discord](https://discord.com/channels/930865462852591648/930865780663398451/1079564343819112488).  I can see the utility of combining the division and a multiplication, since taking a percentage of something seems like the most likely thing to do once you have a percent.  (Evidently, at least one person would find it useful.)

"Round up" and "Round down" might need different names, but I think rounding to the nearest `n` is a useful generalization of rounding to the nearest `1`.  Some nodes only work with sizes that are multiples of 8 or 64, for example.

adodge:
this needs a rebase...  I'll take a look later today

RunDevelopment:
My issue with `b` in rounding up and down is that it default to 0, which kinda breaks the node. Users have to actively "fix" the node to get rounding on its own.

adodge:
Ahh, yeah that makes sense.

adodge:
How about something like this?  A separate node for rounding:

image

joeyballentine:
I think it makes sense

backend/src/nodes/nodes/utility/round.py
```diff
@@ -0,0 +1,129 @@
...
+        scale: RoundScale,
+        m: Union[int, float],
+        p: Union[int, float],
+    ) -> Union[int, float]:
+        if operation == RoundOperation.FLOOR:
+            op = math.floor
+        elif operation == RoundOperation.CEILING:
+            op = math.ceil
+        elif operation == RoundOperation.ROUND:
+            op = round
```
RunDevelopment, line 118: Excuse me, but this python. Did you actually think standard `round` works in a sensible way?
```py
>>> round(2.5)
2
>>> round(3.5)
4
```

Rookie mistake. [Couldn't be me](https://github.com/chaiNNer-org/chaiNNer/pull/1161) :)

Anyway, please you our `round_half_up` function from util.

backend/src/nodes/nodes/utility/round.py
```diff
@@ -0,0 +1,129 @@
...
+                    maximum=None,
+                    precision=100,
+                    controls_step=1,
+                )
+            ),
+            conditional_group(enum=2, condition=RoundScale.POWER.value)(
+                NumberInput(
+                    "Power",
+                    default=2,
+                    minimum=(1 + 1e-100),
```
RunDevelopment, line 64: Floating point imprecision.

```py
>>> 1 + 1e-100
1.0
```

backend/src/nodes/nodes/utility/round.py
```diff
@@ -0,0 +1,129 @@
...
+from ...node_factory import NodeFactory
+from ...properties.inputs import EnumInput, NumberInput
+from ...properties.outputs import NumberOutput
+from . import category as UtilityCategory
+
+
+class RoundOperation(Enum):
+    FLOOR = "Down"
+    CEILING = "Up"
+    ROUND = "Either way"
```
RunDevelopment, line 18: "Either way" is a really weird way to refer to normal/half up rounding. Maybe we could find a better way to refer to it? Maybe "To Nearest"?

RunDevelopment (changes requested):

backend/src/nodes/nodes/utility/round.py
adodge: ouch!  thanks

backend/src/nodes/nodes/utility/round.py
adodge: Oops.  Yeah, I should have realized that.  I guess `np.nextafter(1.0, np.inf)` is the technically right way to do this.  I wonder if a better minimum for any practical use would be `2`.

backend/src/nodes/nodes/utility/round.py
adodge: I agree it's very weird.  How about "round up"/"round down"/"round"?

adodge: Or just floor/ceiling/round.  The reason  I didn't do that to begin with is I'm not sure whether normal people know those terms, or if that's math/programmer jargon.

RunDevelopment (approved):

PR #1610, opened 2023-03-01
Change Font to Roboto-Flex

joeyballentine:
It's actually only a subtle difference, but IMO it makes it look a bit nicer. Plus now it wont use system fonts which make it look different on different platforms.

image

and Roboto-Mono for the monospace stuff 
image


src/renderer/global.scss
```diff
@@ -5,6 +5,19 @@
 :not(.chakra-slider__track):not(.chakra-menu__menu-button):not(.react-flow__handle)
 :not(.chakra-slider__track):not(.chakra-menu__menu-button):not(.react-flow__handle) */
 
+@font-face {
+    font-family: 'Roboto-Mono';
```
joeyballentine, line 9:
```suggestion
    font-family: Roboto-Mono;
```

src/renderer/global.scss
```diff
@@ -5,6 +5,19 @@
...
+    font-display: swap;
+    src: url('../public/fonts/Roboto Mono/RobotoMono-VariableFont_wght.ttf') format('truetype');
+}
+
+@font-face {
+    font-family: 'Roboto-Flex';
```
joeyballentine, line 15:
```suggestion
    font-family: Roboto-Flex;
```

joeyballentine (commented):
Trying to fix this from my phone lol

RunDevelopment (approved):

PR #1611, opened 2023-03-01
Added Derive Seed node

RunDevelopment:
This PR implements [the "derive seed" method I described here](https://github.com/chaiNNer-org/chaiNNer/pull/1499#issuecomment-1406462464).

The new Derive Seed node has a (base) seed input and 10 source inputs. The source inputs can be used to connect sources of randomness that will be used to deterministically derive a seed.

image

---

That being said, I think that we should make seeds their own datatype. While seeds may be simple integers under the hood, I think that we should not make any guarantees about that (and we currently don't). Seeds being numbers for the type system also creates problems because obvious errors aren't caught. Example:
image
Of course, the type system also determines which nodes are suggested in the content node selector:
image

These 2 factors combined make it actually quite hard to use seeds correctly.

Note that if users want a random number, they should use the Random Number node to turn a seed into a number.

Thoughts?


joeyballentine:
Yeah i think that makes sense

RunDevelopment:
Okay, then I'll also implement that as part of this PR.

RunDevelopment:
Okay, seeds are now their own datatype. This doesn't change a lot for users, but it did require some interesting code changes.

image

Code changes:
- Fixed the type system not respecting custom adapters. When the user enters a literal value (either string or number), then we have to convert this literal into a type. This might sounds simple, but inputs like the DirectoryInput make this more complex because the directory input wraps strings. To accommodate this, I added input adapters. Instead of the type system blindly saying "it's a number", inputs can have a custom adapter function that take the literal value and returns the real type of that literal. When implementing this, I didn't expect inputs with default adapters to have custom adpaters. This has been fixed.
- Added `SeedInput` (which is just a very special number input) and `SeedOutput`.
- Added a `Seed` class to wrap seeds. This class also contains a few methods to convert seeds to different ranges (e.g. uint32).

joeyballentine (approved):

PR #1618, opened 2023-03-01
Denser node titles using letter spacing

RunDevelopment:
Currently discussed on discord.


joeyballentine:
I don't think we need this

PR #1621, opened 2023-03-03
Add Convert Number to base-10 node

JustNoon:
Small node to go from any text to a base-10 integer.

Criticism and/ or improvements welcome.

### DIFF
--- a/backend/src/nodes/nodes/utility/parse_number.py
+++ b/backend/src/nodes/nodes/utility/parse_number.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import NumberInput, TextInput
+from ...properties.outputs import NumberOutput
+from . import category as UtilityCategory
+
+
+@NodeFactory.register("chainner:utility:parse_number")
+class ParseNumberNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Parses text to base-10."
+        self.inputs = [
+            TextInput("Text", min_length=0),
+            NumberInput("Base", default=10, minimum=2, maximum=36),
+        ]
+        self.outputs = [
+            NumberOutput(
+                "Value",
+                output_type="int & number::parseInt(Input0, Input1)",
+            ).with_never_reason("The given text cannot be parsed into a number."),
+        ]
+
+        self.category = UtilityCategory
+        self.name = "Parse Number"
+        self.icon = "MdCalculate"
+        self.sub = "Value"
+
+    def run(self, text: str, base: int) -> int:
+        return int(text, base)
--- a/backend/src/nodes/nodes/utility/parse_number.py
+++ b/backend/src/nodes/nodes/utility/parse_number.py
@@ -19,7 +19,7 @@ class ParseNumberNode(NodeBase):
         self.outputs = [
             NumberOutput(
                 "Value",
-                output_type="int & number::parseInt(Input0, Input1)",
+                output_type="int & number::parseInt(toString(Input0), Input1)",
             ).with_never_reason("The given text cannot be parsed into a number."),
         ]
 


backend/src/nodes/nodes/utility/convert_number.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@NodeFactory.register("chainner:utility:convert_number")
+class TextLengthNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Converts a text to base-10."
+        self.inputs = [
+            TextInput("Text", min_length=0),
+            NumberInput("Base", default=10, minimum=0, maximum=64),
```
RunDevelopment, line 17:
```py
>>> int("3", 37)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: int() base must be >= 2 and <= 36, or 0
```

As in most languages, integer parsing only goes up to base 36. This is because these parsers are case-insensitive. E.g. in base 16, both `F` and `f` are parsed as 15. 

Since base 1 doesn't exist and "base 0" is highly specific to python, I would also disallow those.

```suggestion
            NumberInput("Base", default=10, minimum=2, maximum=36),
```

backend/src/nodes/nodes/utility/convert_number.py
```diff
@@ -0,0 +1,29 @@
...
+
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import NumberInput, TextInput
+from ...properties.outputs import NumberOutput
+from . import category as UtilityCategory
+
+
+@NodeFactory.register("chainner:utility:convert_number")
+class TextLengthNode(NodeBase):
```
RunDevelopment, line 11:
```suggestion
class ConvertNumberNode(NodeBase):
```

backend/src/nodes/nodes/utility/convert_number.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self):
+        super().__init__()
+        self.description = "Converts a text to base-10."
+        self.inputs = [
+            TextInput("Text", min_length=0),
+            NumberInput("Base", default=10, minimum=0, maximum=64),
+        ]
+        self.outputs = [
+            NumberOutput("Value"),
```
RunDevelopment, line 20: While we currently don't have a way to parse numbers in the type system, we could at least declare that we're only returning ints.
```suggestion
            NumberOutput("Value", output_type="int"),
```

RunDevelopment (changes requested):

JustNoon:
Thank you for the review. I appreciate the explanations too. It's been a while for python for me

RunDevelopment:
I wonder whether we should call the node "Parse Number" instead. "Convert" is quite generic. Thoughts?

JustNoon:
I also prefer a more technical name like parse but when i was workshopping it in discord there was a general reticence towards programmer jargon even if they're more correct.

I like Parse Number a lot more than convert. Makes it easier to quick search for too.

joeyballentine:
Maybe "extract number"?

JustNoon:
"Extract" feels more like a substring manipulation. It'd feel even weirder imo if/ when there's actual string extraction.

Parse is still my vote on verb. It's unambiguous and doesn't overlap with other nodes.

backend/src/nodes/nodes/utility/convert_number.py
```diff
@@ -0,0 +1,29 @@
...
+class ConvertNumberNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Converts a text to base-10."
+        self.inputs = [
+            TextInput("Text", min_length=0),
+            NumberInput("Base", default=10, minimum=2, maximum=36),
+        ]
+        self.outputs = [
+            NumberOutput("Value", output_type="int"),
```
RunDevelopment, line 20: I just added support for parsing numbers in the type system, so we will be able to use `number::parseInt` after #1629 is merged, and you rebased your PR branch.

```suggestion
            NumberOutput(
                "Value",
                output_type="int & number::parseInt(Input0, Input1)",
            ).with_never_reason("The given text cannot be parsed into a number."),
```

JustNoon: Rebased and updated. Along with renaming to Parse Number.

backend/src/nodes/nodes/utility/parse_number.py
```diff
@@ -0,0 +1,32 @@
...
+        super().__init__()
+        self.description = "Parses text to base-10."
+        self.inputs = [
+            TextInput("Text", min_length=0),
+            NumberInput("Base", default=10, minimum=2, maximum=36),
+        ]
+        self.outputs = [
+            NumberOutput(
+                "Value",
+                output_type="int & number::parseInt(Input0, Input1)",
```
RunDevelopment, line 22: Woops, I forgot something.

```suggestion
                output_type="int & number::parseInt(toString(Input0), Input1)",
```

RunDevelopment (approved):

PR #1623, opened 2023-03-03
(hotifx) Seed types should be cacheable

adodge:
currently all SD nodes are broken, since they use the `SeedInput` and also the `@cached` decorator


joeyballentine (approved):

adodge:
in retrospect I should have just used `__hash__` instead of making a new function

joeyballentine:
@adodge feel free to make another PR but if this works I'm fine with it. Will release later today

adodge:
we can change it later.  this is good for a hotfix, since it's probably not going to break anything else

PR #1624, opened 2023-03-04
Node visibility

### LINKED ISSUE #1373: Hide nodes
Brought up multiple times, but we don't seem to have made an actual issue for it. With the ever-growing node list, it would be great if users could hide ones they don't use from the node selector pane and node selector context menu. The clearest and simplest way to do this would probably be to have a tab in settings for node visibility checklist (as well as categories and subcategories for group hiding), sort of like how image editors let you you choose which dockable panes dialogs to display.

theflyingzamboni:
Introduces a toggle for an alternate selector pane mode that turns it into a node visibility settings menu. Nodes can be toggled off/on individually, by subcategory, or by category. All nodes will be visible when this mode is active, but switching back will hide any nodes/categories/subcategories that are not checked. This applies to the context selector pane as well. Newly added nodes should be visible by default.

Current issues:
- I've set it so that the node single-click tooltip does not pop up in node visibility mode, as clicking the checkbox would cause this to trigger. However, if you click a node and then switch quickly back to normal mode, it will still pop up then.
- There's a noticeable latency between clicking a checkbox and the checkbox changing state. This has occurred from the start even before cat and subcat hiding were implemented, and I'm not sure what the slowdown is. It works basically the same way as favoriting nodes, but seems less responsive.
- General aesthetics. I didn't really put much into deciding how to position the checkboxes aside from the nodes themselves, where they replace the favorites icon.

Closes #1373


RunDevelopment:
Without having reviewed the code, a few thoughts and observations.

1. Lag. Entering and leaving edit mode takes almost a second on my machine due to the extreme lag. 730ms to be exact. Making any change in edit mode (so hiding or unhiding any node, sub category, or category) also causes this lag.
    image
1. The rounded corners of the button and the space to the left of it look weird.
   image
1. The proximity of the toggle button to the search bar makes it seem like the toggle has something to do with it. Very confusing.
1. When I collapse the Favorites category and enter edit mode, it then collapses the Image category. In general, when any category is collapsed, and you enter edit mode, the category after it will be collapsed.
1. The arrow up is too close to the checkbox IMO
   image
1. I don't like that checkboxes aren't aligned. Probably hard to change though.
   image
1. Hiding a node hides them *everywhere*. You literally can't use them anymore (if they aren't included in your chain already). You can't even find them in the context selector pane. This means that hiding a node is a much more consequential change than just cleaning up the node selector panel. I honestly think it goes too far. I think my main issue is that you can't find these nodes even if you search them (text search).

theflyingzamboni:
Thanks for reminding me of #4. That was something I had meant to fix, but forgot about.

joeyballentine:
Thanks for the work on this, but at this point it's too outdated. Also, there's probably a better way it could be done if revisited.

PR #1627, opened 2023-03-05
Adds a regex node

ExoPhantasm:
A basic stripped back implementation that does what it needs but I can appreciate that there might be a better format.  

My usage involves BLIP captions and then needing to perform regex modifications on common captioning errors (sci - fi fi fi fi fi etc) or to adjust the output to my needs, for example perhaps you want to change all cakes into bananas. 

This is primarily for stable diffusion users.

regex

### DIFF
--- a/backend/src/nodes/nodes/utility/regex_pattern.py
+++ b/backend/src/nodes/nodes/utility/regex_pattern.py
@@ -25,7 +25,7 @@ class RegexNode(NodeBase):
         self.category = UtilityCategory
         self.name = "Regex Replace"
         self.icon = "MdTextFields"
-        self.sub = "Value"
+        self.sub = "Regex"
 
     def run(self, text: str, reg: str, rep: str) -> str:
         return re.sub(reg, rep, text, count=0, flags=0)
--- a/backend/src/nodes/nodes/utility/regex_pattern.py
+++ b/backend/src/nodes/nodes/utility/regex_pattern.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import re
+
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import TextInput
+from ...properties.outputs import TextOutput
+from . import category as UtilityCategory
+
+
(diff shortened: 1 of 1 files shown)


joeyballentine:
Thanks! 

I do wonder if having separate "regex match" and "text replace" nodes would be a bit better and more modular. For example i can see it being useful to extract text via regex without doing a replacement, or replacing text without regex. And then they could be used together to get this functionality.

RunDevelopment (commented):
I don't know how I like making Python re the de facto regex language of chainner.

I think my main problem is Python re is not standardized. Put simply, the semantics of how a regex actually behaves are left up to the implementation. While common constructs are documented, many edge cases are not. 
My favorite example of this are unorthodox backreferences. AFAICT, the documentation will not tell you how these 2 regexes behave: `\1(a)`, `(?:(a)|\1{2}){2}`.

RunDevelopment:
Resolves #352.

joeyballentine:
@RunDevelopment what should we do about this PR?

RunDevelopment:
Not sure. It would be a useful node, but there are still a few open questions.

The main open question is which regex engine we want to use. I would like to add support for doing regex operations in the type system, so I would prefer something like re2 or rust's regex. Exponential backtracking in the type system would be pretty bad.

We should also think about the replacement syntax. Python uses `\X` to reference groups and JS uses `$X`, so it's not like there's a standard way of doing this. We already use `{X}` syntax in Text Pattern, so I would err towards that.

joeyballentine:
> The main open question is which regex engine we want to use

I say we should just pick one and roll with it. I'd rather this PR get finished and merged than stay open forever.

I'm also good with the {} syntax for replacement

RunDevelopment:
The let's just go with re2 for now. Changing that in the future might be a breaking change, but whatever.

RunDevelopment:
Wait, we have chainner_ext now. Let's just use Rust regex. This would allow us to run regexes in the type system since we already use Rust regex in the frontend.

RunDevelopment:
Closed in favor of #1893.

Issue #1622, opened 2023-03-03
Implement a high pass filter


WalternateBishop:
It'll be great to have a node that would output a high pass version of the image with the ability to tweak the strength of the filter. 
Can be a separate node or a toggle in the existing high boost filter node. 

It's a relatively simple change that will add a massive amount of flexibility to what can be done with the program! 

![image](https://user-images.githubusercontent.com/123200296/222783228-9b65187e-e4b7-4aaf-b01e-f00439cc5656.png)

PR #1630, opened 2023-03-06
High Pass filter

harisreedhar:
create photoshop like highpass filter mentioned in #1622

chaiNNer-Untitled-2023-3-6_20-48

### DIFF
--- a/backend/src/nodes/nodes/image_filter/high_pass.py
+++ b/backend/src/nodes/nodes/image_filter/high_pass.py
@@ -1,0 +1,42 @@
+from __future__ import annotations
+
+import cv2
+import numpy as np
+
+from ...node_base import NodeBase
+from ...node_factory import NodeFactory
+from ...properties.inputs import ImageInput, NumberInput
+from ...properties.outputs import ImageOutput
+from . import category as ImageFilterCategory
+
+
+@NodeFactory.register("chainner:image:high_pass")
+class HighPassFilterNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply High Pass filter on image"
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Radius", minimum=0, default=1, precision=2, controls_step=1),
+        ]
+        self.outputs = [ImageOutput(image_type="Input0")]
+        self.category = ImageFilterCategory
+        self.name = "High Pass"
+        self.icon = "MdOutlineAutoFixHigh"
+        self.sub = "Miscellaneous"
+
+    def run(
+        self,
+        img: np.ndarray,
+        radius: float,
+    ) -> np.ndarray:
+        """High Pass filter"""
+
+        if radius == 0:
+            return img * 0 + 0.5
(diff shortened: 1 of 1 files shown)


backend/src/nodes/nodes/image_filter/high_pass.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@NodeFactory.register("chainner:image:high_pass")
+class HighPassFilterNode(NodeBase):
+    def __init__(self):
+        super().__init__()
+        self.description = "Apply High Pass filter on image"
+        self.inputs = [
+            ImageInput(),
+            NumberInput("Radius", minimum=0, default=1, precision=2, controls_step=1),
```
RunDevelopment, line 20: I think it would be nice if we could use a default value that will produce okay-ish results for most people.

```suggestion
            NumberInput("Radius", minimum=0, default=3, precision=2, controls_step=1),
```

backend/src/nodes/nodes/image_filter/high_pass.py
```diff
@@ -0,0 +1,44 @@
...
+            NumberInput("Contrast", minimum=0, default=1, precision=2, controls_step=1),
```
RunDevelopment, line 21: I think it would be better if we could make contrast a slider.
```suggestion
            SliderInput(
                "Contrast",
                minimum=0,
                maximum=100,
                default=1,
                precision=2,
                controls_step=0.1,
                scale="log-offset",
            ),
```

backend/src/nodes/nodes/image_filter/high_pass.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def run(
+        self,
+        img: np.ndarray,
+        radius: float,
+        contrast: float,
+    ) -> np.ndarray:
+        """High Pass filter"""
+
+        if radius == 0 or contrast == 0:
```
RunDevelopment, line 37: We already know that it's a high pass filter, so there's no need to repeat it here again :)

```suggestion
        if radius == 0 or contrast == 0:
```

RunDevelopment (commented):
Thank you for making this node!

I have a few suggestions to hopefully make it even better.

harisreedhar:
I also removed alpha channel from processing

RunDevelopment:
About alpha channel processing: The implementation is no longer correct for images with more than 4 channels. You can make sure that you only get images with at most 4 channels using the `channels` argument on [ImageInput](https://github.com/chaiNNer-org/chaiNNer/blob/458d1c4658c8daa2413ab03cb5dd7b4aa5b5e711/backend/src/nodes/properties/inputs/numpy_inputs.py#L25). In this case, supporting grayscale, RGB, and RGBA images will probably be enough, so use `channels=[1, 3, 4]`.

harisreedhar:
@RunDevelopment For some reason ``Lint / backend-formatting`` is failing. I had formatted with black though :(

RunDevelopment (approved):

PR #1640, opened 2023-03-09
Replace clear cache calls with safer version

joeyballentine:
Ideally will help the (at least) one person where this crashes for them.

Might also be worth trying to figure out if these are even worth doing. I'm not sure they _Actually_ do anything. I've read that calling clear cache is kinda useless since for the most part pytorch handles the garbage collection just fine, and doing this adds some slowness. However it probably is worth doing in all auto-split scenarios.

### DIFF
--- a/backend/src/nodes/nodes/pytorch/upscale_face.py
+++ b/backend/src/nodes/nodes/pytorch/upscale_face.py
@@ -12,7 +12,12 @@ from sanic.log import logger
 from torchvision.transforms.functional import normalize as tv_normalize
 
 from ...impl.pytorch.types import PyTorchFaceModel
-from ...impl.pytorch.utils import np2tensor, tensor2np, to_pytorch_execution_options
+from ...impl.pytorch.utils import (
+    np2tensor,
+    safe_cuda_cache_empty,
+    tensor2np,
+    to_pytorch_execution_options,
+)
 from ...node_base import NodeBase
 from ...node_factory import NodeFactory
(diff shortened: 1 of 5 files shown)


joeyballentine:
Update on this: it seems for the one person this was errorong for, the calls no longer crashed chaiNNer but froze on the upscaling step. I'm not really sure what to do from here

RunDevelopment:
Does `empty_cache()` even do anything when CUDA isn't available?

joeyballentine:
no idea, but i wanted to be extra sure that these weren't causing issues. I still don't know if they're even necessary anywhere other than splitting, and even then I'm not sure if they matter

backend/src/nodes/nodes/pytorch/upscale_face.py
```diff
@@ -183,6 +188,6 @@ def run(
             logger.error(f"Face Upscale failed: {e}")
             face_helper = None
             del face_helper
-            torch.cuda.empty_cache()
+            safe_cuda_cache_empty()
```
RunDevelopment, line 191: How can pytorch not be available here?

joeyballentine: It's not about pytorch not being available. Someone is having an issue currently where they get a CUDA misalignment issue from the cache clearing calls, so these are all here to make sure that happening doesn't stop execution.

However, it doesn't seem to have helped since their CUDA gets locked up.... Even completely removing the clearing didn't fix that, so i think their system is just f'd up somehow

RunDevelopment (approved):

PR #1642, opened 2023-03-09
Added group for type conditional inputs

RunDevelopment:
As discussed on Discord. The type conditional group is similar to the enum conditional group, but it works with types instead.

I had to change TypeState as well. It now also contains a list of all edges that were used to create it. This was necessary because the type state is updated with a slight delay for performance reasons. During this type, the type state is outdated and might disagree on which inputs are connected. This caused problems with the type conditional group, so now the type state is the only source of truth for the new group.

### Usage

In the following example, the boolean input will only be show if the input with id 0 is compatible with type `Image { channels: 4 }`.

```py
type_conditional_group(0, "Image { channels: 4 }")(
    BoolInput("Separate Alpha", False),
),
```

### DIFF
--- a/src/renderer/components/groups/ConditionalTypeGroup.tsx
+++ b/src/renderer/components/groups/ConditionalTypeGroup.tsx
@@ -1,0 +1,80 @@
+import { NeverType, evaluate, isDisjointWith } from '@chainner/navi';
+import { memo, useMemo } from 'react';
+import { useContext, useContextSelector } from 'use-context-selector';
+import { InputItem, getUniqueKey } from '../../../common/group-inputs';
+import { getChainnerScope } from '../../../common/types/chainner-scope';
+import { fromJson } from '../../../common/types/json';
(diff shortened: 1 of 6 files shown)


src/renderer/components/groups/ConditionalTypeGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+        );
+
+        let isEnabled = true;
+        if (disjoint) {
+            // the condition is not met
+            isEnabled = false;
+        } else {
+            // If the input has not been assigned a value, then it will default to its declaration type.
+            // This means that the the condition is trivially met, but this isn't what we want.
+            // So we will only show the conditional inputs iff the input has been assigned a value.
```
joeyballentine, line 48:
```suggestion
            // So we will only show the conditional inputs if the input has been assigned a value.
```

joeyballentine: im gonna assume thats what this was

joeyballentine (approved):

PR #1643, opened 2023-03-09
Reload starting nodes on file changes

### LINKED ISSUE #732: Reload starting nodes
We currently run and cache starting nodes. However, this caching assumes that any files referenced by the node input data do not change.

Add a way for users to refresh the starting nodes, so that the cached output is up-to-date. This could either be a simple context menu entry "Refresh" or some kind of auto reloading (which would be nice).

RunDevelopment:
Resolves #732.

The heavy lifting is done by [chokidar](https://github.com/paulmillr/chokidar), which is a pretty popular library for file watching. I just added the `useWatchFiles` hook as a light wrapper around chokidar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@react-nano/use-event-source": "^0.12.0",
         "adm-zip": "^0.5.10",
         "bezier-js": "^6.1.0",
+        "chokidar": "^3.5.3",
         "decompress": "^4.2.1",
         "electron-log": "^4.4.8",
         "electron-squirrel-startup": "^1.0.0",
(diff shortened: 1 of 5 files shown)


joeyballentine (commented):
what does this do when the file is renamed?

RunDevelopment:
image

The image inside the node still stays the same. Since running the node didn't succeed, the backend cache will also still have the old value. 

As soon as a new file takes that place old the old file, the node reloads again.

joeyballentine:
I feel like it should instead switch the path to be the renamed file. Thoughts?

RunDevelopment:
I feel like it should instead switch the path to be the renamed file.

No. Firstly, we can't detect this. chokidar only gives us events for 'add', 'change', and 'unlink'. Secondly, that is waaay too much magic. No other program I ever used behaves that way.

joeyballentine (approved):

joeyballentine:
Makes sense

PR #1644, opened 2023-03-09
Disallow ONNX -> NCNN on M1 machines

RunDevelopment:
After our talk on Discord, I noticed that #1561 was slightly wrong. Since ONNX -> NCNN conversion require onnxoptimizer for correctness, the onnxoptimizer package must be present.


RunDevelopment:
Not necessary.

PR #1648, opened 2023-03-12
SD Check status code

adodge:
webui returns a 404 JSON message when --api isn't passed.  I think it used to return invalid JSON, but now it doesn't , so it doesn't fail correctly, so we need to check the status code.

Also if webui gives us an error code, put the error message in our logs.

### DIFF
--- a/backend/src/nodes/impl/external_stable_diffusion.py
+++ b/backend/src/nodes/impl/external_stable_diffusion.py
@@ -33,7 +33,7 @@ def _stable_diffusion_url(path):
     return f"{STABLE_DIFFUSION_PROTOCOL}://{STABLE_DIFFUSION_HOST}:{STABLE_DIFFUSION_PORT}{path}"
 
 
-ERROR_MSG = f"""
+INFO_MSG = f"""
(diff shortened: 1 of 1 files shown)


RunDevelopment (approved):

joeyballentine:
Does this check still work for older versions of the webui?

adodge:
> Does this check still work for older versions of the webui?

That is an excellent question.  I will check and get back to you

joeyballentine:
Any update on this?

joeyballentine:
I'm going to merge this so the nightlies will work with the latest webui

adodge:
Sorry about the delay.  I tested against today's version and the one from the beginning of the month, and they both seem to have the correct behavior with the current patch.

Auto1111: a9fed7c364061ae6efb37f797b6b522cb3cf7aa2 (latest, March 14)

* ---api:  shows SD nodes
* no --api:  does not show SD nodes

Auto1111:  b14d8b61bdc4cc2e6c0ea484a5ba8fd370231227  (March 1)

* ---api:  shows SD nodes
* no --api:  does not show SD nodes

PR #1657, opened 2023-03-15
Make Gloss and Roughness optional

RunDevelopment:
Roughness and Gloss are not strictly necessary for the Metal → Specular and Specular → Metal conversions, so this PR makes them optional. Since we don't have optional outputs, I return a 1x1 standard gray image instead. 

I'm not sure about the 1x1 image btw. We don't have optional outputs (not do I want to add them just for this), so I had to return something.


joeyballentine:
I feel like we should return the resolution of the input rather than 1x1. I'm also wondering if just all black would be better too, since that's what we do with other nodes like split rgba when the image doesn't have an A channel

joeyballentine (approved):
This is fine then. We should probably look into doing dynamic outputs at some point, but it's not really that important rn

PR #1672, opened 2023-03-23
Update PyTorch for Python 3.10

joeyballentine:
~We should wait to merge this until it's been thoroughly tested, but this seems to work well enough from my testing.~

~This resolves the issue of installing pytorch to system python 3.10 via the dependency manager, and gets us access to all the latest pytorch features and optimizations.~

This PR now just updates to PyTorch 1.12.1 when on python 3.10. This should fix the long running problem of people on system python being unable to install pytorch from the dependency manager.


Kim2091:
Did some benchmarks with these settings:
FP16 enabled
SwinIR Large Tile Size: `Auto (estimate)`
Real-ESRGAN Tile Size: `384`
Input image: `768x576`

**SwinIR Large:**

__chaiNNer 0.18.3:__
1: 16.98
2: 14.42
3: 14.48

__Pytorch 2.0:__
1: 15.02
2: 12.37
3: 12.29

**4x Real-ESRGAN:**

__chaiNNer 0.18.3:__
1: 2.794
2: 1.506
3: 1.512

__Pytorch 2.0:__
1: 3.3
2: 1.834
3: 1.844

joeyballentine:
How do we feel about going to 1.13.1 instead?

RunDevelopment:
Not too good. https://discord.com/channels/930865462852591648/1078534152464384152/1088813748107624610

src/common/dependencies.ts
```diff
@@ -43,24 +44,52 @@ const getOnnxRuntime = (canCuda: boolean): PyPiPackage => {
     };
 };
 
-export const getOptionalDependencies = (isNvidiaAvailable: boolean): Dependency[] => {
+const getPytorch = (canCuda: boolean, pythonVersion: Version): PyPiPackage[] => {
+    if (semver.gte(pythonVersion, '3.8.0') && semver.lt(pythonVersion, '3.10.0')) {
```
RunDevelopment, line 48: Our minimum supported python version is 3.8.0, so won't `semver.gte(pythonVersion, '3.8.0')` always be true?

joeyballentine: Yes, that's why i made it there an error otherwise.

RunDevelopment: Fair.

RunDevelopment (approved):

PR #1674, opened 2023-03-23
Replace M1 check with regex match

joeyballentine:
Fixes the arm check for M2 devices, probably

### DIFF
--- a/src/common/env.ts
+++ b/src/common/env.ts
@@ -3,7 +3,9 @@ import path from 'path';
 
 export const isWindows = process.platform === 'win32';
 export const isMac = process.platform === 'darwin';
-export const isM1 = isMac && (os.cpus()[0]?.model.includes('Apple M1') ?? false);
+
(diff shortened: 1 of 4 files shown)


src/common/env.ts
```diff
@@ -3,7 +3,9 @@ import path from 'path';
 
 export const isWindows = process.platform === 'win32';
 export const isMac = process.platform === 'darwin';
-export const isM1 = isMac && (os.cpus()[0]?.model.includes('Apple M1') ?? false);
+
+const cpuModel = os.cpus()[0]?.model || null;
+export const isArmMac = isMac && cpuModel && /Apple M\d/i.test(cpuModel);
```
RunDevelopment, line 8: This isn't guaranteed to be a boolean. It's `null` if `cpuModel` is `null`,
```suggestion
export const isArmMac: boolean = isMac && !!cpuModel && /Apple M\d/i.test(cpuModel);
```

PR #1676, opened 2023-03-24
update random number node

Zarxrax:
Add functionality so that if a seed of -1 is provided, it will generate a new random number each time the chain is run.
This was originally a feature when the random number node was created, but it was removed at the time due to issues with caching in iterators. The derive seed node resolved the issues around iterators, so it makes sense to add this back in.

### DIFF
--- a/backend/src/nodes/nodes/utility/random_number.py
+++ b/backend/src/nodes/nodes/utility/random_number.py
@@ -40,4 +40,7 @@ class RandomNumberNode(NodeBase):
         self.sub = "Random"
 
     def run(self, min_val: int, max_val: int, seed: Seed) -> int:
-        return Random(seed.value).randint(min_val, max_val)
+        if seed.value == -1:
+            return Random().randint(min_val, max_val)
+        else:
+            return Random(seed.value).randint(min_val, max_val)


RunDevelopment:
> The derive seed node resolved the issues around iterators, so it makes sense to add this back in.

Just because Derive Seed provided a solution doesn't mean that it's okay to add the problem back in. Seed -1 still doesn't work within iterators.

In general, non-deterministic nodes don't work (correctly) in chainner right now. See #1677 for more details.

Zarxrax:
> > The derive seed node resolved the issues around iterators, so it makes sense to add this back in.
> 
> Just because Derive Seed provided a solution doesn't mean that it's okay to add the problem back in. Seed -1 still doesn't work within iterators.
> 
> In general, non-deterministic nodes don't work (correctly) in chainner right now. See #1677 for more details.

While I see what you are saying about not working inside of iterators, it still works outside of them, and it can work WITH iterators via the derive seed node.
With the chain in the attached screenshot, one can get completely different random numbers in each run.
image

Unless this were to completely break some things, I think it could be a useful feature until the determinism issue gets figured out.

RunDevelopment:
> it can work WITH iterators via the derive seed node.

One issue is that it doesn't work without Derive Seed. Any node that works outside iterators should also work inside iterators. Otherwise, it's just inconsistent. 

But there is also a broader issue: guarantees. We don't guarantee that we cache or don't cache nodes. Caching is purely a performance optimization. If and how long outputs are cached is up to the executor. You cannot rely on any specific caching strategy.

Or put another way: Right now, random seeds barely and coincidently work.

> until the determinism issue gets figured out

It's the other way around. Right now, non-determinism is not supported, so we can't have any non-deterministic nodes until we figure it out.

Zarxrax:
Alright, I will just close this for now.

PR #1679, opened 2023-03-26
(WIP) New import system proof of concept v2

joeyballentine:
This is based on #1430.

Summary of what I've done on top of that:
- Renamed "subcategories" to "node groups" to be more clear about them and also to prepare for a future in which we use sub-menus
- Move metadata things into the decorator. Now the constructor is only used for input & output definitions and things like saying if a node has side effects. This is nice because it now gives a clear separation between what is metadata and what is functionality.
- Path-like routing. I'm not 100% sold on this system and will probably go back to how it was before, but basically, now categories and node groups have their variables defined in the `__init__.py` files and are generically named so they can be moved around freely to re-categorize them without needing code changes. This means in order to reorder node groups or categories, you need to name the folders in the alphabetical order you want to use them in. Since that's kinda weird, I'll probably go back to the other system where categories and node groups are defined one level up and then imported. in each node. Though, one potential idea to get around that could be just using the `__init__.py` file in each node group folder to import the node group and then export it again as `node_group`. That would have the best of both systems.

Anyway, this is still heavily WIP as I've only ported over the image category, but I want to gett an opinion before working on it more.


RunDevelopment:
> Renamed "subcategories" to "node groups"

👍 

> Move metadata things into the decorator. Now the constructor is only used for input & output definitions and things like saying if a node has side effects. This is nice because it now gives a clear separation between what is metadata and what is functionality.

Input/output declarations, side effects, etc. are also metadata. They describe the behavior of the node. Literally everything except the `run` function is metadata. (This is why we do `NodeClassName().run(*args)` when calling nodes from other nodes. Nothing except the `run` function affects functionality.)

---

One question I would like to raise is whether nodes should be classes. 
By their very nature, classes are stateful. But nodes are expected to be stateless. As I see it, a node is effectively just a function with some metadata. (Metadata isn't state because it's not specific to a class instance, it's essentially a static constant.)

Conceptually, nodes are just function + metadata, so why would we need a class?

Using functions for nodes would also simplify the API. Example:
```py
@register(
    schema_id="chainner:image:load",
    group=node_group,
    name="Load Image",
    description="Load image from specified file.",
    icon="BsFillImageFill",
    inputs=[
        ImageFileInput(primary_input=True),
    ],
    outputs=[
        LargeImageOutput(),
        DirectoryOutput("Image Directory", of_input=0),
        FileNameOutput("Image Name", of_input=0),
    ],
)
def load_image(path: str) -> Tuple[np.ndarray, str, str]:
    ...
```

Of course, internally in the registry, we still need (data)classes to hold the function + metadata pair.

joeyballentine:
I actually completely agree with that. Especially since I've been getting a warning not to use decorators with classes 😆 

This would basically make them the equivalent of function components in react 👍

Side thought: i still would absolutely love if there was a way to type validate the "props" of the run function based on the inputs metadata. I doubt that's possible, but i wish it was.

RunDevelopment:
> Side thought: i still would absolutely love if there was a way to type validate the "props" of the run function based on the inputs metadata. I doubt that's possible, but i wish it was.

Python's type system doesn't have powerful-enough generics and literal types. It would be possible in TypeScript, not python.

We could do runtime checks, though. [The `__annotations__` prop](https://docs.python.org/3/howto/annotations.html?highlight=__annotations__) gives us all typed args + the return type. So we could somewhat verify input and outputs types. The only issue is that it doesn't contain untyped args. At the very least, we could verify the return type.

```py
>>> def foo(a: int, b: str) -> float:
...     return 0
>>> foo.__annotations__
{'a': <class 'int'>, 'b': <class 'str'>, 'return': <class 'float'>}

>>> def bar(a, b: str):
...     pass
>>> bar.__annotations__
{'b': <class 'str'>}
```

joeyballentine:
Theoretically could it be done through a vscode extension or something? So not through python directly

backend/src/api.py
```diff
@@ -0,0 +1,236 @@
...
+from nodes.properties.outputs.base_output import BaseOutput
+
+
+@dataclass
+class Node:
+    schema_id: str = ""
+    description: str = ""
+    name: str = ""
+    icon: str = ""
+    type: NodeType | None = "regularNode"
```
RunDevelopment, line 23: Why does this allow `None`?

backend/src/api.py
```diff
@@ -0,0 +1,236 @@
...
+from nodes.properties.inputs.base_input import BaseInput
+from nodes.properties.outputs.base_output import BaseOutput
+
+
+@dataclass
+class Node:
+    schema_id: str = ""
+    description: str = ""
+    name: str = ""
+    icon: str = ""
```
RunDevelopment, line 22: Why do all fields have default values? Most of those don't even make sense. E.g. `""` isn't a valid schema id.

backend/src/api.py
joeyballentine: So we don't have to explicitly pass it in every time and it can default to "regularNode"

joeyballentine:
I have no idea why the CI is failing. As in, I don't know why the dictionary size is changing during iteration. I don't get that same issue locally

RunDevelopment:
Regarding node, function, and file names, could we have some checks to enforce consistency? Something like function name + ".py" == file name as a hard requirement. We need one node per file anywhere to support partial dependencies, so I think this requirement is fine. 

Additionally, something like function name == node name.to_lower().to_snake_case() also be nice. This would allow us to quickly fine node files, since we just have to search for the node name we see in the UI. This means that renaming a node requires a migration, but I think that's fine.

RunDevelopment:
Then let's not enforce node names for now. Fn name == file name == schema id is already a big win. We can rope in the node name later if we want to.

joeyballentine:
Thoughts on enforcing that these always end with `_node`?

RunDevelopment:
Don't like it. Within a file, it's clear which function is the node by the `@x.register`, and in the file system, nodes are grouped within categories which only contain nodes. So, why do we need the `_node`?

joeyballentine:
> So, why do we need the _node?

It makes it clearer when importing another node, for example in iterators when they import the load model node. That's kinda it though

RunDevelopment:
But does that matter? With iterators, does it matter whether `load_image` is a node or not? Nodes are just functions + some metadata. There is nothing that differentiates nodes from regular functions from the perspective of the caller.

backend/src/run.py
```diff
@@ -164,42 +117,35 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
...
+            "icon": node.icon,
+            "subcategory": sub.name,
+            "nodeType": node.type,
+            "hasSideEffects": node.side_effects,
+            "deprecated": node.deprecated,
         }
-        if node_object.type == "iterator":
-            node_dict["defaultNodes"] = node_object.get_default_nodes()  # type: ignore
+        if node.type == "iterator":
+            node_dict["defaultNodes"] = node.default_nodes  # type: ignore
```
RunDevelopment, line 141: I think we can simplify this if we change the TS type to `readonly defaultNodes?: readonly DefaultNode[] | null;`

```suggestion
        node_dict["defaultNodes"] = node.default_nodes
```

backend/src/run.py
```diff
@@ -68,71 +67,25 @@ def get(app_instance: Sanic) -> "AppContext":
 app.config.RESPONSE_TIMEOUT = sys.maxsize
 CORS(app)
 
+# Join current file path with the path to the packages folder
+packages_dir = os.path.join(os.path.dirname(__file__), "packages")
+sys.path.append(packages_dir)
```
RunDevelopment, line 72: Why? Could we just import e.g. `packages.chaiNNer_standard` and get rid of this? I don't want to mess with python import paths...

backend/src/chain/chain.py
```diff
@@ -37,15 +36,15 @@ def __init__(self, node_id: NodeId, schema_id: str):
...
-            node = NodeFactory.get_node(self.schema_id)
-            assert isinstance(node, IteratorNodeBase), "Invalid iterator node"
+            node = registry.get_node(self.schema_id)
+            assert Node.type == "iterator", "Invalid iterator node"
             self.__node = node
         return self.__node
 
 
-Node = Union[FunctionNode, IteratorNode]
+FunctionalNodes = Union[FunctionNode, IteratorNode]
```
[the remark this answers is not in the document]
joeyballentine: Sure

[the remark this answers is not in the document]
joeyballentine: Actually i don't think we need this iterator node at all, do we?

backend/src/run.py
joeyballentine: We're gonna have to do this for plugins anyway right? Since we can't import outside our package. Might as well make sure it works here. (This is what comfyui does btw)

backend/src/chain/chain.py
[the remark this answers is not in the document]
RunDevelopment: No, `IteratorNode` is necessary for type safety. Merging iterator nodes into `FunctionNode` makes it impossible to prove certain properties with the type system.

backend/src/run.py
RunDevelopment: But they aren't outside our package and this path is not useful for anything besides our internal packages.

joeyballentine: Ok

joeyballentine: I mainly did this for when i was just scanning the dir to import, but since we don't do that anymore it's useless i guess

backend/src/chain/chain.py
```diff
@@ -37,15 +36,15 @@ def __init__(self, node_id: NodeId, schema_id: str):
...
-            node = NodeFactory.get_node(self.schema_id)
-            assert isinstance(node, IteratorNodeBase), "Invalid iterator node"
+            node = registry.get_node(self.schema_id)
+            assert node.type == "iterator", "Invalid iterator node"
             self.__node = node
         return self.__node
 
 
-Node = Union[FunctionNode, IteratorNode]
+Nodes = Union[Node, IteratorNode]
```
RunDevelopment, line 47: "Nodes" implies that this type is some sort of collection, which it is absolutely not. Just keep the old names as `Node` and `FunctionNode`.

joeyballentine: My bad, i thought FunctionNode was something i did and that your comment from earlier meant to completely change it

RunDevelopment:
Nodes within node groups are not sorted by name. This is a pretty minor frontend thing, so I think we can also fix that after merging this.

image

RunDevelopment (approved):

RunDevelopment:
There are still a couple of minor things that we should consider changing (node function names, name enforcement, implementation details of registry, node order), but none are pressing enough to warrant holding this PR back. We can make follow-up PRs for those.

PR #1681, opened 2023-03-27
Type broadcasts

RunDevelopment:
Each output can now broadcast its current Navi type. The frontend is then responsible to deal with that information.

This system has a few advantages:
1. The backend no longer decides whether outputs broadcast. This makes the backend code simpler and the frontend has more information to make such decisions anyway.
2. Since type broadcasts are standardized, types no longer have to be derived from broadcast data per output in the frontend. This makes the implementations of all output elements simpler.


RunDevelopment:
Because they have these tags derived from their broadcast data

We could make an output component specifically for tags, though. It could simply assume that the broadcast data has a `tags: string[]` prop and use that.

joeyballentine:
Yeah that's basically what i was expecting to get out of this

joeyballentine (approved):
LGTM

PR #1692, opened 2023-04-02
Add support for AUTOMATIC1111's extras - image upscaling

mateusz:
Found this bit of AUTOMATIC1111's webui quite useful, and wanted to use it in the interface. "Single Image" is the only bit that's really necessary - your tool can deal with batching via iterators.

Note I tried implementing dynamic size and asserts, but the logic is very convoluted and I couldn't figure it out - and I struggled with the syntax of `ExpressionJson`. The logic is dependent on multiple inputs - if upscaler_1 is None, output is the same shape, if absolute rescaling is on, then shape is different, if clip is on the output is assymetric, plus some rounding to deal with. I tried multiple versions but couldn't figure it out :(

## View of Webui's interface for extras

extras

## View in ChaiNNer

chainner-extras

### DIFF
--- a/backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
+++ b/backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
@@ -1,0 +1,202 @@
+from __future__ import annotations
+
+from enum import Enum
+
+import numpy as np
+
+from nodes.groups import Condition, if_enum_group, if_group
+from nodes.impl.external_stable_diffusion import (
+    STABLE_DIFFUSION_EXTRA_SINGLE_IMAGE_PATH,
+    decode_base64_image,
+    encode_base64_image,
+    get_upscalers,
+    post,
+    verify_api_connection,
+)
+from nodes.node_cache import cached
+from nodes.properties.inputs import (
+    BoolInput,
+    EnumInput,
+    ImageInput,
+    NumberInput,
+    SliderInput,
+)
+from nodes.properties.outputs import ImageOutput
+from nodes.utils.utils import get_h_w_c
+
+from .. import auto1111_group
+
+verify_api_connection()
+upscaler_enum, upscaler_labels = get_upscalers()
+
+
+class UpscalerMode(Enum):
+    SCALE_BY = "ScaleBy"
+    SCALE_TO = "ScaleTo"
+
+
+UPSCALER_MODE_LABELS = {
+    UpscalerMode.SCALE_BY: "Scale by",
+    UpscalerMode.SCALE_TO: "Scale to",
+}
+
+
+@auto1111_group.register(
+    schema_id="chainner:external_stable_diffusion:upscaling",
+    name="Upscale",
+    description="Upscale an image using Automatic1111",
+    icon="MdChangeCircle",
+    inputs=[
+        ImageInput(channels=3),
+        EnumInput(
+            UpscalerMode,
+            default_value=UpscalerMode.SCALE_BY,
+            option_labels=UPSCALER_MODE_LABELS,
+        ).with_id(1),
+        if_enum_group(1, UpscalerMode.SCALE_BY)(
+            SliderInput(
+                "Resize multiplier",
+                minimum=1.0,
+                default=4.0,
+                maximum=8.0,
+                slider_step=0.1,
+                controls_step=0.1,
+                precision=1,
+            ).with_id(2),
+        ),
+        if_enum_group(1, UpscalerMode.SCALE_TO)(
+            NumberInput("Width", controls_step=1, default=512).with_id(3),
+            NumberInput("Height", controls_step=1, default=512).with_id(4),
+            BoolInput("Crop to fit", default=True).with_id(5),
+        ),
+        EnumInput(
+            upscaler_enum,
+            label="Upscaler 1",
+            option_labels=upscaler_labels,
+        ),
+        BoolInput("Use second upscaler", default=False).with_id(7),
+        if_group(Condition.bool(7, True))(
+            EnumInput(
+                upscaler_enum,
+                label="Upscaler 2",
+                option_labels=upscaler_labels,
+            ),
+            SliderInput(
+                "Upscaler 2 visibility",
+                minimum=0.0,
+                default=0.0,
+                maximum=1.0,
+                slider_step=0.001,
+                controls_step=0.001,
+                precision=3,
+            ),
+        ),
+    ],
+    outputs=[
+        ImageOutput(
+            image_type="""
+                def nearest_valid(n: number) = int & floor(n);
+
+                let in_w = Input0.width;
+                let in_h = Input0.height;
+                let ratio_w = width/in_w;
+                let ratio_h = height/in_h;
+                let larger_ratio = if ratio_w>ratio_h { ratio_w } else { ratio_h };
+
+                let mode = Input1;
+                let factor = Input2;
+
+                let crop = Input5;
+                let width = Input3;
+                let height = Input4;
+
+                match mode {
+                    UpscalerMode::ScaleTo => if crop {
+                        Image { width: width, height: height }
+                    } else {
+                        Image {
+                            width: nearest_valid(in_w*larger_ratio),
+                            height: nearest_valid(in_h*larger_ratio)
+                        }
+                    },
+                    UpscalerMode::ScaleBy => Image {
+                        width: nearest_valid(in_w*factor),
+                        height: nearest_valid(in_h*factor)
+                    }
+                }
+            """,
+            channels=3,
+        )
+    ],
+)
+@cached
+def upscale_node(
+    image: np.ndarray,
+    mode: UpscalerMode,
+    upscaling_resize: float,
+    width: int,
+    height: int,
+    crop: bool,
+    upscaler_1: Enum,
+    use_second_upscaler: bool,
+    upscaler_2: Enum,
+    upscaler_2_visibility: float,
+) -> np.ndarray:
+    if mode == UpscalerMode.SCALE_BY:
+        resize_mode = 0
+    else:
+        resize_mode = 1
+
+    if use_second_upscaler:
+        u2 = upscaler_2.value
+    else:
+        u2 = "None"
+
+    request_data = {
+        "resize_mode": resize_mode,
+        "show_extras_results": False,
+        "gfpgan_visibility": 0.0,
+        "codeformer_visibility": 0.0,
+        "codeformer_weight": 0.0,
+        "upscaling_resize": upscaling_resize,
+        "upscaling_resize_w": width,
+        "upscaling_resize_h": height,
+        "upscaling_crop": crop,
+        "upscaler_1": upscaler_1.value,
+        "upscaler_2": u2,
+        "extras_upscaler_2_visibility": upscaler_2_visibility,
+        "upscale_first": False,
+        "image": encode_base64_image(image),
+    }
+    response = post(
+        path=STABLE_DIFFUSION_EXTRA_SINGLE_IMAGE_PATH, json_data=request_data
+    )
+    result = decode_base64_image(response["image"])
+
+    ih, iw, _ = get_h_w_c(image)
+    rh, rw, _ = get_h_w_c(result)
+    ratio_w = width / iw
+    ratio_h = height / ih
+    if ratio_w > ratio_h:
+        larger_ratio = ratio_w
+    else:
+        larger_ratio = ratio_h
+
+    if mode == UpscalerMode.SCALE_TO:
+        if crop:
+            assert (rw, rh) == (
+                width,
+                height,
+            ), f"Expected the returned image to be {width}x{height}px but found {rw}x{rh}px instead "
+        else:
+            assert (rw, rh) == (
+                int(iw * larger_ratio),
+                int(ih * larger_ratio),
+            ), f"Expected the returned image to be {width}x{height}px but found {rw}x{rh}px instead "
+    else:
+        assert (rw, rh) == (
+            int(iw * upscaling_resize),
+            int(ih * upscaling_resize),
+        ), f"Expected the returned image to be {width}x{height}px but found {rw}x{rh}px instead "
+
+    return result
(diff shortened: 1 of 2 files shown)


joeyballentine:
Sorry for the late reply, but thanks for submitting this. @adodge is going to look it over when he has time.

As for the `ExpressionJson` stuff, those are Navi types. It's a custom language for our type system. [You can read more about it here](https://github.com/chaiNNer-org/Navi/blob/main/navi.md). Though, IMO the best way to learn is usually just by looking at a bunch of other nodes that perform a similar function and seeing how they implement it.

RunDevelopment:
Sorry for the delay here too!

I think we should slightly restructure our inputs before we tackle the output type. In the webui, there are 2 modes: "Scale by" and "Scale to". So let's do that too. Let's define another enum for those 2 states. 

```py
class ScaleMode(Enum):
    SCALE_BY = 0
    SCALE_TO = 1
```

We can then use an `if_enum_group` to hide unnecessary inputs depending on the currently selected mode. So e.g. width and height will only be shown when "Scale to" is selected. A good example for how this works is [Create Gradient](https://github.com/chaiNNer-org/chaiNNer/blob/6a782e74f48d078ca7efde68e6c2e46c8ba19e65/backend/src/nodes/nodes/image/create_gradient.py#L41). The first argument of `if_enum_group` is the ID of the enum input we want to check. All inputs get assigned an ID implicitly, but we typically use `with_id(X)` on the enum input to make this more explicit.

```py
         self.inputs = [
            ImageInput(),
            EnumInput(ScaleMode, default_value=ScaleMode.SCALE_BY).with_id(1),
            if_enum_group(1, ScaleMode.SCALE_BY)(
                SliderInput(
                    "Relative resize multiplier",
                    minimum=0,
                    default=4,
                    maximum=8,
                    slider_step=0.1,
                    controls_step=0.1,
                    precision=1,
                ),
            ),
            if_enum_group(1, ScaleMode.SCALE_TO)(
                NumberInput("Width", controls_step=1, default=512),
                NumberInput("Height", controls_step=1, default=512),
                BoolInput("Crop to fit", default=True),
            ),
            ...
        ]
```

adodge:
Very cool!  Thank you for this!

I have a few notes:

1. With the default settings (both upscalers set to None) it doesn't actually do any upscaling.  So there's a third "scale mode" for "no scaling".
2. The name "extras" doesn't immediately mean anything to me. It would if I was used to using it in webui, but I've never really explored that tab (partly because of how vague the name is.)  Maybe we could use a different name for the node, and we can mention that it uses the extras tab in the description, or in a parenthetical.
4.  When I picked LDSR, it downloaded a 2GB file before running the model.  (I know this is just what webui does, but it was surprising to me.  It could be undesired by some users.)
3. I get a server error if I select ScuNET, which I don't have a model for in webui.  Maybe we could catch this and show a better message.  Alternatively, (or additionally) there's [this endpoint](http://127.0.0.1:7860/sdapi/v1/upscalers), which we could use to get a list of the models they actually have installed.  This is a similar problem to picking an SD model in the text-to-image node.  We punted there because we could just not specify and it would use whatever they have loaded already, but I don't think we can do that here.  If we find a good solution to this, we could also probably use it in the other node.  Or we could just restrict this to models that are available by default or can be automatically downloaded.

It looks like what it does is some number (maybe zero) of the following operations (I believe in this order):

A. Upscales with one upscaler
B. Upscales with another upscaler and blends it with the output
C. Applies GFPGAN and blends it with the output.
D. Applies CodeFormer and blends it with the output

Maybe we could have separate nodes that each do one of these steps:  Upscale, GFPGAN, and CodeFormer.  That might make it easier to understand what they're doing.  It's also more modular and node-ish.

Another point that I don't really have a strong opinion on but is probably worth noting:  I believe all of these things are already supported natively by ChaiNNer.  I can think of a few scenarios where someone would want to use this instead of the built-in chainner nodes:

1. They're running webui on a different machine
2. They're already accustomed to the webui interface and they just want to use it from chainner (in which case we wouldn't want to break it up into separate nodes like I suggest above, as the whole point is to give the same UI they're used to)
3. They've already got the models they want installed in webui, and they don't want to bother with having to find those files again in the "Load Model" node.  Or they're taking advantage of the automatic download in webui.
4. There's some model where webui is faster or better than chainner, or a model that webui supports that chainner doesn't

Once we have internal SD, assuming we keep the webui integration, we'll have sort of retroactively established a precedent for webui nodes that duplicate built-in behavior, so it's not automatically a deal-breaker for me.  The scenarios above are compelling reasons to have this node, IMO.

mateusz:
Thank you for an in-depth review and tons of references to look up.

I appreciate this PR might not make sense. As @adodge mentioned: "all of these things are already supported natively by ChaiNNer." 

To elaborate on the use case: I wanted to use LDSR for upscaling in the middle of a pipeline, and LDSR specifically because it gives me the best results.

The context is that webui/LDSR works quite well, I tried SDv2 [superresolution](https://github.com/Stability-AI/stablediffusion#image-upscaling-with-stable-diffusion) and it gave mediocre results (for my use case), so I have a feeling someone tuned that LDSR model pretty hard. Webui implementation looks pretty hacked though, and I have a feeling it'd be hard to actually run in chaiNNer :)

Let me know if I should continue a bit further. I'm not advocating either way, happy for you to close this PR - I've done my thing by having it hacked in :)

adodge:
Actually, I don't see LDSR on the list of things ChaiNNer supports, so I take it back that it's all stuff we already have.

For me the only sticking point is figuring out how we want to translate it into nodes (one or several) and how to handle picking from the models the user has installed.  I'm certainly not arguing for closing the PR.  I think this is a cool feature.

mateusz:
> the only sticking point is figuring out how we want to translate it into nodes

Ok I'll take this PR one step further, I think I'll just reduce it to Upscaling via webui, because GFPGAN and CodeFormer are already supported via pytorch.

mateusz:
I've followed your suggestions and done the following (see the animated gif too):

- tidied up naming, e.g. renamed the block to "Upscale" and file to `upscaling.py`
- added dynamic fetching of upscalers from `sdapi/v1/upscalers` endpoint at import time
- removed GFPGAN and CodeFormer on the premise this is not upscaling, and it's supported elsewhere
- removed "None" upscaler option on the assumption it was only needed for GFPGAN and CodeFormer - it unnecessarily confuses the interface
- hide Upscaler 2 behind a checkbox, to simplify the UI
- hide "scale to" section (width, height, crop) unless selected from dropdown
- implemented navi spec for all possibilites
- added asserts for all possibilities

chainner-upscale

adodge:
Nice!

I get a Navi error on load:

```
12:32:14.658 › Error: Unable to add type definitions of chainner:external_stable_diffusion:upscaling > Upscaler 1 (id: 6):
Error: SyntaxError: At 1:77: no viable alternative at input 'DynamicEnum::4'
Type definitions: let DynamicEnum = DynamicEnum::Lanczos | DynamicEnum::Nearest | DynamicEnum::4X-ultrasharp | DynamicEnum::4XFoolhardyRemacri | DynamicEnum::Esrgan4X | DynamicEnum::Ldsr | DynamicEnum::R-esrgan4X+ | DynamicEnum::R-esrgan4X+Anime6B | DynamicEnum::Swinir4X
```

Probably need more aggressive sanitization of the upscaler name.  I'm not sure what the set of valid names is.  Maybe we could just hash the names to hex, prepend an "x" so it doesn't start with a number, and use that as the ID in Navi.  That seems safe.

I was going to test what happens when we load a chain that references an upscaler that doesn't exist.  I don't know what it _should_ do, but we should make sure it does something reasonable.

adodge:
Also we should call "DynamicEnum" something else, in case we want to use this strategy in other places.  (like to specify models for the text-to-image node)

RunDevelopment:
Awesome work @mateusz!

> I get a Navi error on load:

Oh god. I didn't know that it was possible to dynamically create enums like this. `EnumInput` assume that all enum variants have SNAKE_CASE names. So names that don't follow this will cause it to produce invalid Navi code.

RunDevelopment:
> I was going to test what happens when we load a chain that references an upscaler that doesn't exist. I don't know what it should do, but we should make sure it does something reasonable.

In that case, [`Dropdown` will reset the input](https://github.com/chaiNNer-org/chaiNNer/blob/4f0d0afe3f4c053d7432b799142a184bd0a0aee6/src/renderer/components/inputs/elements/Dropdown.tsx#L16) to its default value (typically the first variant).

mateusz:
@adodge I haven't run into error while testing and I'm just wondering how I missed it - is there anything specific I need to do before `npm run make`? Do I need to wipe `rm -fr out`? 

> call "DynamicEnum" something more specific

Oh true, missed that, fixed.

> all enum variants have SNAKE_CASE names

Sanitised and converted to SNAKE_CASE.

joeyballentine:
This will need to be updated/rebased with the new node importing system btw

adodge:
> I haven't run into error while testing and I'm just wondering how I missed it - is there anything specific I need to do before `npm run make`? Do I need to wipe `rm -fr out`?

I think I probably have a different set of upscalers installed in webui.  I think the one it was complaining about was called `4x_foolhardy_Remacri`, which turned into an invalid token name in Navi.  I suspect it doesn't like names that start with numbers.  We should also make sure it doesn't choke on names like `R-ESRGAN 4x+ Anime6B`, with special characters in the middle.

The name of the model is based on whatever the filename is on the user's machine, so we should try to be resilient to all sorts of possible strings.  `אַפּסקייל.pt`, etc

mateusz:
Another push:

- rebased onto new main (new node importing system)
- [switched](https://github.com/chaiNNer-org/chaiNNer/pull/1692/files#diff-141916a155c04573f70aba6e4c9ce0d7ea1282f7cf2a0f195c96e72eaa3caff8R165) to use `UPSCALER_` + md5 labels to support 4x_foolhardy_Remacri and hebrew models :stuck_out_tongue:
- added second ret val to get_upscalers, to provide explicit labels (previously, id would be used instead of the human-friendly name)

backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
```diff
@@ -0,0 +1,201 @@
...
+        larger_ratio = ratio_w
+    else:
+        larger_ratio = ratio_h
+
+    if mode == UpscalerMode.SCALE_TO:
+        if crop:
+            assert (rw, rh) == (
+                width,
+                height,
+            ), f"Expected the returned image to be {width}x{height}px but found {w}x{h}px instead "
```
RunDevelopment, line 189: I agree with the linter. Those should be `rw` and `rh`. Same for the other 2 error messages.

```suggestion
            ), f"Expected the returned image to be {width}x{height}px but found {rw}x{rh}px instead "
```

backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
```diff
@@ -0,0 +1,201 @@
+from __future__ import annotations
+
+from enum import Enum
+
+import numpy as np
+from nodes.groups import Condition, if_enum_group, if_group
```
RunDevelopment, line 6: isort won't like that. 

@mateusz Are you using VSCode? We set up isort to automatically run on save in VSCode, so I would like to know whether it didn't work for you or whether you're using another IDE.

```suggestion
import numpy as np

from nodes.groups import Condition, if_enum_group, if_group
```

backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
mateusz: For sure, copy-paste error.

backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
mateusz: I'm using VSCode, but I must admit I haven't put serious thought into it and been installing random tools to get it to pass :) It might be related to Microsoft's "Black Formatter v2022.4.0" idiosyncracy?

In any case, I've then installed `isort v2022.8.0` from VSCode plugins too, and that did not reformat the file.
Then I installed `pip install isort` (isort-5.12.0) and ran the exact command your pipeline is running and the following fix was applied:

```diff
--- a/backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
+++ b/backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/upscaling.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from enum import Enum
 
 import numpy as np
+
 from nodes.groups import Condition, if_enum_group, if_group
 from nodes.impl.external_stable_diffusion import (
     STABLE_DIFFUSION_EXTRA_SINGLE_IMAGE_PATH,
 ```

RunDevelopment: That is weird.

> Black Formatter v2022.4.0
> isort v2022.8.0

Those are quite old versions. Maybe that's the issue?

RunDevelopment (approved):
I can't test this, since I don't have AUTOMATIC1111's webui installed. So someone that has it and has used it before, please test the new node. 
@joeyballentine @adodge

The code itself looks good.

mateusz:
Hey folks - just wanted to mention, don't feel bad if you want to reject and close the PR, if it doesn't suit the philosophy of chaiNNer. I don't mind, really. It was fun to hack something together :)

joeyballentine:
I'm getting errors when trying to use any of the models. It's always this same error, saying it can't import the model 
image

joeyballentine:
I'm guessing this probably doesn't have to do with your code here, and is probably due to me using a somewhat older version of the webui, but if its at all fixable we should probably try to do something about it

mateusz:
Similar problem here, although only with SwinIR model https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/6395 . Possibly something to do with dependencies for webui, maybe update pip deps? Do you get the same error when you try to use the UI directly?

Unless you mean we should handle webui errors better in chaiNNer?

joeyballentine:
I just wasn't sure if there was maybe some config thing wrong. But yeah it just looks like a webui issue. So I'm good with calling this good

PR #1698, opened 2023-04-04
Include NCNN model resize scale when interp is last op

### LINKED ISSUE #1697: realesr-animevideov3 model scales do not function correctly
**Description**
When loading realesr-animevideov3-x2 (2x) into chaiNNer through NCNN, the model is still detected and processed as a 4x model. The same occurs with 3x. Refer to this discussion: https://discord.com/channels/547949405949657098/547949405949657100/1092840573808943226

theflyingzamboni:
The scale on realesranime_v3 was not being computed correctly because the resize scaling factor was only included when interp was not the last op. Added scale factor calculation into IndexError except statement to incorporate end-of-model resizes.

Closes #1697

### DIFF
--- a/backend/src/nodes/impl/ncnn/model.py
+++ b/backend/src/nodes/impl/ncnn/model.py
@@ -731,7 +731,7 @@ class NcnnModelWrapper:
                     ):
                         scale *= checked_cast(float, layer.params[1].value)
                 except IndexError:
-                    pass
+                    scale *= checked_cast(float, layer.params[1].value)
             elif layer.op_type == "PixelShuffle":
                 scale *= checked_cast(int, layer.params[0].value)
                 pixel_shuffle *= checked_cast(int, layer.params[0].value)


joeyballentine:
why is it necessary to catch the index error in order to detect this? I'm fine with it if it works and this is just the way, but that just seems odd to me

theflyingzamboni:
> why is it necessary to catch the index error in order to detect this? I'm fine with it if it works and this is just the way, but that just seems odd to me

There's some case that I don't recall anymore where Interp followed by a BinaryOp means something other than a change in model scale, as far as I can remember. Because of that, I have to do a forward check in the loop, which throws an index error on the last item. Before I was just ignoring it, but since realesranime does a resize at the end, that needs to be accounted for.

The alternative would be to if-check that it wasn't the last item, but try/except should be more efficient here because the except is only ever going to be triggered once.

RunDevelopment:
https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

> should be more efficient

Because exception have a stack trace + some other info + potentially use CPU interrupts, they are typically 10~100x more expensive than normal control flow. So no, a simple `if` would be orders of magnitudes faster. Not that performance really matters here.

theflyingzamboni:
> Because exception have a stack trace + some other info + potentially use CPU interrupts, they are typically 10~100x more expensive than normal control flow. So no, a simple `if` would be orders of magnitudes faster. Not that performance really matters here.

If the except block were being triggered often enough, that would be true. But it will trigger exactly once at the end, whereas an if statement will cost every iteration. A lot of these models have several hundred layers to iterate through.

Here's an example loop on a list of similar size timed:

```py
stmt = """
for i, a in enumerate(x):
 if i+1<len(x):
  a+x[i+1]"""
timeit.timeit(setup="x=[a for a in range(600)]", stmt=stmt)
78.60759079999843

stmt2 = """
for i, a in enumerate(x):
 try:
  a+x[i+1]
 except IndexError:
  pass"""
timeit.timeit(setup="x=[a for a in range(600)]", stmt=stmt2)
44.634887100000924
```

RunDevelopment:
The fact that the `try/except` is faster is insane to me. But I guess it's not really that surprising in python. Everything in python is slow, so the fixed cost of exceptions is comparatively low in python. I was comparing them to Java, C#, C++ exceptions, but all those languages are quite a bit better at optimizing than python. My bad.

> EAFP

First time I heard of that. While I see its usefulness in the motivating examples with file access (and resources in general), using it as too broadly is not good IMO. Our code is a good example:

```py
                try:
                    if (
                        model.layers[i + 1].op_type != "BinaryOp"
                        and model.layers[i + 1].params[0].value != 0
                    ):
                        scale *= checked_cast(float, layer.params[1].value)
                except IndexError:
                    scale *= checked_cast(float, layer.params[1].value)
```

Which part of the code throws the IndexError? Answer: you can't know. We assume that it's `model.layers[i + 1]`, but it could also be `layer.params[1]` or even `.op_type` (getters exist), the `!=` (operator overloading), or `checked_cast` (bug in its implementation). 

Exceptions provide no way of distinguishing these cases. So when do (or don't) you want to catch this exception and what for? In this code, I have to guess what your intentions and assumptions are. On the other hand, `i + 1 < len(model.layers)` makes it every explicit what you expect to be the case. 

The `IndexError` can come from a number of places, which makes your intentions unclear. This is in contrast to the motivating examples for EAFP:

```py
try:
    f = open(my_file)
except IOError as e:
    print("File can't be accessed")
else:
    with f:
        print(f.read())
```

joeyballentine (approved):

PR #1699, opened 2023-04-05
Basic static runtime type-checking between python type hints and our inputs & outputs

joeyballentine:
Branched this off of #1679 because I don't want to add anything else there but needed the existing type length checking stuff. In fact, I completely removed it from that PR and moved it here. 

I mainly just wanted to see how possible this was, and it was very possible. Not only does it work pretty well, but it also caught a bug.

~~This is still a WIP as I add more types. Feel free to add to this.~~ Added all types lol

i think i'll make this run in 3 modes based on a cli arg, so that we dont run this in the deployed app, we run it in warning mode in run dev, and we run it in error mode in CI

This does not check outputs, I'll wait for feedback before continuing. I got a little carried away already.


backend/src/api.py
```diff
@@ -119,6 +121,81 @@ def inner_wrapper(wrapped_func: T) -> T:
...
+                            try:
+                                # Allows us to use pipe unions still
+                                if "|" in py_type:
+                                    py_type = f'Union[{py_type.replace(" |", ",")}]'
+                                global_vals = wrapped_func.__globals__
+                                # Gotta add these to the scope
+                                global_vals.update({"Union": Union})
+                                global_vals.update({"np": np})
+                                # pylint: disable=eval-used
+                                evaluated_py_type = eval(py_type, global_vals)
```
RunDevelopment, line 178: Could we use locals instead of changing global scope? https://docs.python.org/3/library/functions.html#eval

backend/src/api.py
```diff
@@ -92,6 +95,128 @@ class NodeData:
...
+        return
+
+    # We don't want to run this check on iterator helpers as they can have different input/output metadata than what they actually run
+    if node_type == "iteratorHelper":
+        return
+
+    ### Return type validation
+
+    # Pop the return type key off the dict
+    return_type = ann.pop("return", None)
```
RunDevelopment, line 169: I think this might be changing the dict associated with function. Please make a copy of the annotations so we don't change the function metadata.

backend/src/api.py
```diff
@@ -92,6 +95,128 @@ class NodeData:
...
+    py_var: str,
+    py_type: Union[str, Any],
+    associated_type: Type,
+):
+    evaluated_py_type = py_type
+    if isinstance(py_type, str):
+        try:
+            # Allows us to use pipe unions still
+            if "|" in py_type:
+                py_type = f'Union[{py_type.replace(" |", ",")}]'
```
RunDevelopment, line 110: I think there's an easier way to do this. https://stackoverflow.com/questions/50074956/how-exactly-do-eval-and-exec-interact-with-future

We could move this into its own file to make the connection between `from __future__ import annotations` and the type eval more explicit.

joeyballentine: I don't think this helps us. The current behavior seems to be that `eval` does not take `__future__` imports into account at all. All this StackOverflow says is that it _does_ take them into account (which is not the behavior I see) and how to disable it, which I don't need.

RunDevelopment: Alright, looking at the docs, we could so this: 
```py

code_obj = compile(source_str, '<string>', 'eval', flags=annotations.compiler_flag)
eval(code_obj , globals, locals)
```

joeyballentine: that didn't do anything.

RunDevelopment (dismissed):
Please move the type validation code into its own file.

backend/src/packages/chaiNNer_external/external_stable_diffusion/automatic1111/img2img.py
```diff
@@ -35,6 +35,7 @@
 verify_api_connection()
 
 
+@cached
```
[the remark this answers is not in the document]
joeyballentine: We have a problem then because putting this below the register decorator breaks the type checking

.github/workflows/test-backend.yml
```diff
@@ -51,3 +51,5 @@ jobs:
                   python-version: ${{ matrix.python-version }}
             - run: npx ts-node ./scripts/install-required-deps.ts
             - run: python ./backend/src/run.py --no-run
+              env:
+                  TYPE_CHECK_LEVEL: 'error'
```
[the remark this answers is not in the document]
joeyballentine: The intended behavior was failing CI. I wasn't getting that one locally though so I'm not sure why you were getting that. It should be ignoring the variable arg ones

backend/src/api.py
```diff
@@ -228,9 +232,14 @@ def load_nodes(self, current_file: str):
...
-                        logger.warning(f"Failed to load {module}: {e}")
+                        logger.warning(f"Failed to load {module} ({file_path}): {e}")
                     except ValueError as e:
-                        logger.warning(f"Failed to load {module}: {e}")
+                        logger.warning(f"Failed to load {module} ({file_path}): {e}")
+                    except TypeMismatchError as e:
+                        type_errors.append(e)
+
+        if len(type_errors) > 0:
+            raise RuntimeError(type_errors)
```
RunDevelopment, line 242: Right now, we just get this in the logs:
```
RuntimeError: [TypeMismatchError("Error in chainner:utility:text_append: Input type of Text A '<class 'str'>' is not assignable to varargs type 'typing.Optional[str]'"), TypeMismatchError("Error in chainner:image:z_stack: Input type of Image A '<class 'numpy.ndarray'>' is not assignable to varargs type 'typing.Optional[numpy.ndarray]'")]
```

This isn't readable at all, so I think that we should always log type errors.

```suggestion
                    except TypeMismatchError as e:
                        logger.error(e)
                        type_errors.append(e)

        if len(type_errors) > 0:
            raise RuntimeError(f"Type errors occurred in {len(type_errors)} node(s)")
```

backend/src/api.py
```diff
@@ -114,10 +100,27 @@ def register(
...
             p_inputs, group_layout = _process_inputs(inputs)
-            p_output = _process_outputs(outputs)
+            p_outputs = _process_outputs(outputs)
+
+            TYPE_CHECK_LEVEL = os.environ.get("TYPE_CHECK_LEVEL", "none")
+
+            if TYPE_CHECK_LEVEL != "none":
+                try:
+                    typeValidateSchema(wrapped_func, node_type, p_inputs, p_outputs)
+                except Exception as e:
```
RunDevelopment, line 114: So we are catching more than just type errors?

backend/src/api.py
```diff
@@ -114,10 +100,27 @@ def register(
...
+                    full_error_message = f"Error in {schema_id}: {e}"
+                    if TYPE_CHECK_LEVEL == "warn":
+                        logger.warning(full_error_message)
+                    else:
+                        raise TypeMismatchError(full_error_message) from e
```
RunDevelopment, line 119: I think the `from e` is detrimental here. We don't care about the stack trace of any type errors, so this just makes the new error more complex without adding useful information.

backend/src/type_checking.py
```diff
@@ -0,0 +1,208 @@
...
+    for t2 in types[1:]:
+        t = Union[t, cast(Any, t2)]
+    return t
+
+
+def union_to_set(t: _Ty) -> Set[_Ty]:
+    if str(t).startswith("typing.Union["):
+        return set(get_args(t))
+    else:
+        return {t}
```
RunDevelopment, line 72: The fix for what we talked about on Discord.

```suggestion
def union_to_set(t: _Ty) -> Set[_Ty]:
    s = str(t)
    if s.startswith("typing.Union["):
        return set(get_args(t))
    elif s.startswith("typing.Optional["):
        return {*union_to_set(get_args(t)[0]), _Ty(type(None))}
    else:
        return {t}
```

backend/src/api.py
```diff
@@ -114,10 +100,27 @@ def register(
...
         deprecated: bool = False,
         default_nodes: List[DefaultNode] | None = None,
+        decorators: List[Callable] | None = None,
     ):
         def inner_wrapper(wrapped_func: T) -> T:
             p_inputs, group_layout = _process_inputs(inputs)
-            p_output = _process_outputs(outputs)
+            p_outputs = _process_outputs(outputs)
+
+            TYPE_CHECK_LEVEL = os.environ.get("TYPE_CHECK_LEVEL", "none")
```
RunDevelopment, line 109: Please do some more validation for `TYPE_CHECK_LEVEL`. Right now, literally any string that isn't `"none"` or `"warn"` will behave like `"error"`, e.g. `"WARN"`. Maybe we could add a function `def get_type_checking_level() -> Literal["none", "warn", "error"]`?

RunDevelopment:
`get_type_check_level` doesn't work.

```py
>>> def get_type_check_level(s: str) -> TypeCheckLevel:
...     type_check_level = s
...     if type_check_level not in TypeCheckLevel.__dict__.values():
...         return TypeCheckLevel.NONE
...     type_check_level = cast(TypeCheckLevel, type_check_level)
...
>>> get_type_check_level("foo")
<TypeCheckLevel.NONE: 'none'>
>>> get_type_check_level("warn")
<TypeCheckLevel.NONE: 'none'>
```

joeyballentine:
😬

joeyballentine:
TIL how python enums work

RunDevelopment (approved):
Let's go!!

PR #1710, opened 2023-04-06
Slightly more efficient `normalize`

RunDevelopment:
We can use `np.clip`'s `out` parameter after we copied the original image with `astype`. This makes the normalization faster for some dtypes. This won't affect any of our commonly-used dtypes like float32 and uint8.

### DIFF
--- a/backend/src/nodes/impl/image_utils.py
+++ b/backend/src/nodes/impl/image_utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import itertools
 import os
 import random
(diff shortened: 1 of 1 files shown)


backend/src/nodes/impl/image_utils.py
```diff
@@ -78,17 +80,27 @@ def convert_to_BGRA(img: np.ndarray, in_c: int) -> np.ndarray:
...
             if info.min == 0:
                 # we don't need to clip
                 return img
-        except:
-            img = img.astype(np.float32)
+
+        # we own `img`, so it's okay to write to it
+        return np.clip(img, 0, 1, out=img)
+
     return np.clip(img, 0, 1)
```
joeyballentine, line 104: So why is overwriting img for the astype not an issue but doing out=img here would be?

RunDevelopment: `img = img.astype(np.float32)` creates a copy, so the clip with `out=img` writes to the copy.

But if the image is already float32, then we don't copy, so `out=img` would change the original image. This would break the current API contract of `normalize` since some callers may rely on `normalize` not changing the passed-in image.

joeyballentine:
Actually, are we sure we don't wanna just make it clip the original instead of making a copy in that final return? In what case would we _not_ want it to be clipped? If it's already clipped, it doesn't matter. If it isn't clipped, it needs to be. Not overwriting the original, while maybe more correct, doesn't seems to me like an actual issue

RunDevelopment:
> doesn't seems to me like an actual issue

Well, you just changed the behavior of `normalize` in a breaking way. So please review and test all calls of `normalize` in the entire code base to verify that they still behave correctly :)

We literally just fixed a similar breaking change that was "probably alright" (- me) with #1655 and #1706. I learned my lesson.

joeyballentine (approved):

PR #1711, opened 2023-04-06
Added type checking to inputs' `enforce`

RunDevelopment:
I added `object` as the type hint for `value` in `enforce`. This forces all subclasses of `BaseInput` to check the type of `value` before using it. This makes all `enforce` implementations type-checkable, which is good for correctness. The forced type check also means that inputs enforce correct values better now.

### DIFF
--- a/backend/src/nodes/properties/inputs/base_input.py
+++ b/backend/src/nodes/properties/inputs/base_input.py
@@ -67,12 +67,12 @@ class BaseInput:
         self.id: InputId = InputId(-1)
 
     # This is the method that should be created by each input
-    def enforce(self, value):
+    def enforce(self, value: object):
         """Enforce the input type"""
         return value
 
     # This is the method that should be called by the processing code
-    def enforce_(self, value):
+    def enforce_(self, value: object | None):
         if self.optional and value is None:
             return None
         assert value is not None, (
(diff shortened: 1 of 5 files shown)


joeyballentine:
I don't really know how i feel about potentially causing errors if something comes in as the wrong type. The point of the enforce methods was to cast invalid values to avoid errors. For example if a number input gets a string that's a number it should cast it. It should definitely log that there's a bug there, but it shouldn't prevent the user from using it if we can get around that. With this PR, a valid scenario that isn't type correct would prevent a user from running their chain

RunDevelopment:
> With this PR, a valid scenario that isn't type correct would prevent a user from running their chain

That's not true. This PR is purely about the *implementations* of `enforce` methods. The goal isn't that the chain type checks or to disallow casts, but to verify that the `enforce` methods are implemented correctly.

The best examples here are the ONNX outputs. They need to access the `.bytes` prop of `value`. But what happens when `value` isn't an `ONNXModel`? Then python will throw some generic field not valid error AT BEST. At worst, the value just happens to be of a class that happens to have a `.bytes` prop and the checks succeeds.

Again, this PR is about ensuring that the `enforce` methods are implemented correctly, not about restricting *what* they can implement. You can still implement that str -> number cast with this PR.

joeyballentine:
Ahhhhhh, ok I understand now. This is great then :) sorry about the confusion

RunDevelopment:
No worries. This PR is literally about type checking type checking code :)

joeyballentine (approved):

PR #1712, opened 2023-04-07
Fix ESRGAN-2c2 scale detection

joeyballentine:
I was doing a cube root, which worked perfectly fine for 4x, but broke at 2x. Turns out I actually needed to do a /4 then a sqrt.

### DIFF
--- a/backend/src/nodes/impl/pytorch/architecture/RRDB.py
+++ b/backend/src/nodes/impl/pytorch/architecture/RRDB.py
@@ -82,7 +82,7 @@ class RRDBNet(nn.Module):
         c2x2 = False
         if self.state["model.0.weight"].shape[-2] == 2:
             c2x2 = True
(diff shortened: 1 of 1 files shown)


joeyballentine:
Would it be better to use math.sqrt for this?

RunDevelopment:
> Would it be better to use math.sqrt for this?

I don't think it really matters. sqrt will be faster, but `** 0.5` is clearer IMO.

---

Nothing issue: Taking the square root doesn't exactly guarantee integers and `int(x)` will truncate. Is rounding down really the intended rounding method? If the result *should* be an integer before truncation, we might get an off-by-one error because the float will be 1.99999999 or something.

joeyballentine:
Oh i thought int rounded to the nearest whole. Should I use math.round then?

RunDevelopment:
Yeah, `round` should be good.

RunDevelopment (approved):

PR #1713, opened 2023-04-07
Give iterators side effects

joeyballentine:
This fixes an issue i accidentally introduced in #1679


RunDevelopment:
But most iterators *don't* have side effects. E.g. the model iterator nodes only have side effects if one of their child nodes has side effects.

Anyway, we're gonna replace iterators some day anyway, so I think that this is good enough for now.

RunDevelopment (approved):

joeyballentine:
@RunDevelopment that doesnt matter from the perspective of the node optimizer, which doesn't run the iterators at all (including the children) unless they have side effects.

PR #1714, opened 2023-04-07
Add weight support to codeformer when using face upscale

### LINKED ISSUE #1475: Codeformer settings not available?
Hello,

Is there a way to tweak Codeformer settings within Chainner? 

I'm getting much better results with this implementation: https://huggingface.co/spaces/sczhou/CodeFormer probably because I can tweak the fidelity settings there but I can't seem to do it within Chainner.

Is there some way to select those within the interface?

Thank you!

joeyballentine:
image

Fixes #1475

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/pytorch/restoration/upscale_face.py
+++ b/backend/src/packages/chaiNNer_pytorch/pytorch/restoration/upscale_face.py
@@ -121,14 +121,7 @@ def upscale(
         NumberInput(
             label="Output Scale", default=8, minimum=1, maximum=8, unit="x"
         ).with_id(3),
-        if_group(
-            Condition.type(0, 'PyTorchModel { arch: "CodeFormer" }')
-            & (
-                Condition.type(
-                    0, 'PyTorchModel { arch: "GFPGAN" | "RestoreFormer" }'
-                ).__invert__()
-            )
-        )(
+        if_group(Condition.type(0, 'PyTorchModel { arch: "CodeFormer" }'))(
             SliderInput(
                 label="Weight",
                 default=0.7,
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_pytorch/pytorch/restoration/upscale_face.py
```diff
@@ -120,6 +121,24 @@ def upscale(
...
             label="Output Scale", default=8, minimum=1, maximum=8, unit="x"
         ).with_id(3),
+        if_group(
+            Condition.type(0, 'PyTorchModel { arch: "CodeFormer" }')
+            & (
+                Condition.type(
+                    0, 'PyTorchModel { arch: "GFPGAN" | "RestoreFormer" }'
+                ).__invert__()
+            )
+        )(
```
RunDevelopment, line 131: We talked about this on discord. But I thought about this a bit more, and negating type conditions is actually incorrect in general. We can't really enforce this, but I'll make an issue about this for now.
```suggestion
        if_group(Condition.type(0, 'PyTorchModel { arch: "CodeFormer" }'))(
```

joeyballentine: I don't want this showing up unless codeformer is connected. It's not even useful to have this being conditional without that. If it's gonna show up without anything connected to it i might as well just make it show up all the time, because it goes from "oh there's a different option when i connect codeformer" to "hey where did that option go when i connected gfpgan wtf"

joeyballentine: Actually fine i give up on this. Let's do it your way and see what happens

RunDevelopment (approved):

backend/src/packages/chaiNNer_pytorch/pytorch/restoration/upscale_face.py
RunDevelopment: > If it's gonna show up without anything connected to it

It won't show up unless *something* is connected, don't worry. We had this discussion before when we first added type conditions :)

joeyballentine: That's not how it was functioning last night, it was showing up all the time

RunDevelopment: You're right... wtf. I'll look into it. This is not how they are supposed to behave.

PR #1725, opened 2023-04-08
Consider output error for connectability

### LINKED ISSUE #1673: Input assignability should consider output type `never`.
**Motivation**
As pointed out [here](https://github.com/chaiNNer-org/chaiNNer/issues/1662#issuecomment-1480396443), if you can connect 2 things, it should just work. However, this is currently not the case because of output types. Output types can become `never` which signifies invalid inputs. These types of errors are currently not accounted for when checking whether for compatibility.

**Description**
Check for output type `never` when checking assignability. If assigning some input type causes any node output to become `never`, then the assignment should be invalid.

RunDevelopment:
Resolves #1673.

If PR makes it so that inputs are not connectable of the assigned type would cause an output error. So if the target node is valid (no output errors), but the new connections would create output errors, then we don't allow that connection. Examples:

image
image

While this enforces the general idea "if it's connectable, it's valid," it also makes it a little harder to edit chains.

E.g. in the Guided Upscale example, I initially had source and guide the other way around by accident. Chainner would not let me assign the other input with the intended value until I removed the old (invalid) connection.

Frankly, I don't know whether I like this. A new user might not realize that they have to remove old invalid connections before they can connect their intended value. This might "trap" new users and lead to frustration. But maybe I'm overthinking here.

**Please test this feature and see whether this feature is desirable.**

### DIFF
--- a/src/common/types/function.ts
+++ b/src/common/types/function.ts
@@ -606,4 +606,16 @@ export class FunctionInstance {
         // we say that types A is assignable to type B if they are not disjoint
         return !isDisjointWith(iType, this.definition.convertInput(inputId, type));
     }
+
+    /**
+     * Returns a new function instance with the given input assigned to the given type.
+     */
+    withInput(inputId: InputId, type: NonNeverType): FunctionInstance {
+        return FunctionInstance.fromPartialInputs(this.definition, (id) => {
+            if (id === inputId) {
+                return type;
+            }
+            return this.inputs.get(id);
+        });
+    }
 }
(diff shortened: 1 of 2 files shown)


joeyballentine:
In general i think the idea is good, but yeah i can see how it might "trap" users. Maybe we could have a popup instead that says it's invalid, but give them the option to cancel or remove the other connections making it invalid?

joeyballentine:
Any more thoughts on what we should do with this?

RunDevelopment:
The user experience with the feature is not good. It's too restrictive.

Instead of disallowing connections that will cause errors, I think it might be better to allow it, but show a warning. Not sure about *how* though. 
We could give the handle something like an orange warning color and show a tooltip on how. But our handles are already pretty colorful, so it might not be clear enough.

joeyballentine:
We could highlight the handle and edge with red maybe? We specifically haven't used red yet so it could be used for error stuff

PR #1728, opened 2023-04-09
Add handle to text area

### LINKED ISSUE #1727: Text areas should be text connectable
Title.

joeyballentine:
Fixes #1727

### DIFF
--- a/backend/src/nodes/properties/inputs/generic_inputs.py
+++ b/backend/src/nodes/properties/inputs/generic_inputs.py
@@ -240,6 +240,7 @@ class TextAreaInput(BaseInput):
         super().__init__("string", label, has_handle=False, kind="text")
         self.resizable = True
         self.default = default
+        self.has_handle = True
 
     def enforce(self, value) -> str:
         if isinstance(value, float) and int(value) == value:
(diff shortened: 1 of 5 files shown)


RunDevelopment:
Uhhhmmmm, no. You just added handles to *all* text areas, including the one for notes. Please add a `has_handle: bool` option to `TextAreaInput` so nodes can opt into/out of (depends on the default) handles. 

Conservatively speaking, I think that the default should be the current behavior and nodes can opt into handles.

Also:
image

backend/src/nodes/properties/inputs/generic_inputs.py
```diff
@@ -236,10 +236,13 @@ def toDict(self):
...
 
-    def __init__(self, label: str = "Text", default: Union[str, None] = None):
-        super().__init__("string", label, has_handle=False, kind="text")
+    def __init__(
+        self, label: str = "Text", default: Union[str, None] = None, has_handle=False
+    ):
+        super().__init__("string", label, has_handle=has_handle, kind="text")
         self.resizable = True
         self.default = default
+        self.has_handle = has_handle
```
RunDevelopment, line 245: We don't need this, do we?

```suggestion
```

joeyballentine: We do not. Good catch

RunDevelopment (approved):

PR #1730, opened 2023-04-11
Add ETA to iterators

joeyballentine:
This ended up being a bit more complicated than I thought it would be. I ended up removing the existing progress (which we put in nodeData for some reason) and moved it to the execution context with a fancy hook. Something about my implementation doesn't seem quite right, but this at least works :)

image

TBH probably wasn't really worth spending the time on since we plan on changing iterators, but since I have no idea when that would be I guess it makes sense to add this considering it's been requested several times.


backend/src/process.py
```diff
@@ -118,20 +119,29 @@ async def run_iteration(self, index: int, total: int):
...
                     "running": list(self.chain.nodes.keys()),
                 },
             }
         )
 
         try:
+            start = time.time()
             await executor.run_iteration(self.chain)
+            end = time.time()
+            self.avg_time = (self.avg_time * index + (end - start)) / (index + 1)
```
RunDevelopment, line 135: Since indexes may be repeated, so this is incorrect. I think it would be better to use a list of durations instead. This would also make it possible to use something like exponential smoothing to give more weight to recent durations.

backend/src/events.py
```diff
@@ -40,6 +40,9 @@ class NodeFinishData(TypedDict):
 
 class IteratorProgressUpdateData(TypedDict):
     percent: float
+    index: int
+    total: int
```
RunDevelopment, line 44: You don't use those 2 in the frontend, so why transmit them?

Btw. it's good that you don't use them in the frontend, because you send slightly wrong indexes in a few places.

src/renderer/components/node/IteratorNode.tsx
```diff
@@ -31,17 +32,14 @@ export const IteratorNode = memo(({ data, selected }: IteratorNodeProps) => (
 
 const IteratorNodeInner = memo(({ data, selected }: IteratorNodeProps) => {
     const { schemata, categories } = useContext(BackendContext);
+    const { useIteratorProgress } = useContext(ExecutionContext);
+
+    const { id, inputData, isLocked, schemaId, iteratorSize, minWidth, minHeight } = data;
+
+    const { getIteratorProgress } = useIteratorProgress(id);
+
+    const { percent, eta } = getIteratorProgress();
```
RunDevelopment, line 41: Why expose `useIteratorProgress` when only `getIteratorProgress` is ever used outside `ExecutionContext`?

joeyballentine: When i was writing this i thought i needed these and i just didn't remove them

backend/src/events.py
joeyballentine: which places do I send the wrong index?

backend/src/process.py
```diff
@@ -118,20 +124,29 @@ async def run_iteration(self, index: int, total: int):
...
             await executor.run_iteration(self.chain)
+            end = time.time()
+            self.times.append(end - start)
         finally:
             await self.executor.queue.put(
                 {
                     "event": "iterator-progress-update",
                     "data": {
                         "percent": (index + 1) / total,
+                        "index": index,
```
RunDevelopment, line 147:
```suggestion
                        "index": index + 1,
```

backend/src/process.py
```diff
@@ -203,6 +218,9 @@ async def run_while(
                     "event": "iterator-progress-update",
                     "data": {
                         "percent": 1,
+                        "index": index,
+                        "total": length_estimate,
```
RunDevelopment, line 222: This is the case where our length estimate was too large. So `index` contains the actual length.
```suggestion
                        "index": index,
                        "total": index,
```

RunDevelopment (commented):
Regarding progress events:
- You forgot to send the 0% event in `run_while`.
- We should probably extract out the code for sending the 0% and 100% events into 2 functions respectively. I want to avoid code duplication for them.
- I think we need to change how to handle `length_estimate` in `run_while`. If the estimate is too small, then the UI will get stuck on e.g. `200/200` and then suddenly jump to `250/250` with the final 100% event. So instead of `run_iteration(min(index, length_estimate - 1), length_estimate)`, we should do `run_iteration(index, max(length_estimate, index + 1))`.
   This will cause it to count up, so the user will see `200/200` `201/201` `202/202` ... `250/250`, which is a much better progress indicator.

If you don't like the `200/200` `201/201` behavior, then we just have to change `run_iteration` to send events before doing the iteration. Then the users would see `199/200` `200/201` and so on. So it would get stuck on 99% instead of 100%, which is better, I think. This would also have the advantage that we wouldn't need to send the 0% event in `run` and `run_while`, since `run_iteration` would always do that for us.

backend/src/process.py
```diff
@@ -109,34 +110,52 @@ def __create_iterator_executor(self) -> Executor:
...
+        executor = self.__create_iterator_executor()
+
+        await self.progress.suspend()
+
+        start = time.time()
+        await executor.run_iteration(self.chain)
+        end = time.time()
+        self.times.append(end - start)
+
+        await self.__update_progress(index + 1, total)
```
RunDevelopment, line 158: `executor.run_iteration` might throw an error, but we still have to update progress. So I would suggest doing progress updates before calling the executor. This has a few other advantages as well, as I outlined in my previous comment.

```suggestion
        await self.progress.suspend()
        await self.__update_progress(index, total)

        start = time.time()
        await executor.run_iteration(self.chain)
        end = time.time()
        self.times.append(end - start)
```

backend/src/process.py
```diff
@@ -146,6 +165,8 @@ async def run(
         items = list(collection)
         length = len(items)
 
+        await self.__update_progress(0, length)
+
```
RunDevelopment, line 169: If we do a progress update with `index, length`, then we don't need to do 0% explicitly.

```suggestion
```

backend/src/process.py
```diff
@@ -176,6 +199,9 @@ async def run_while(
     ):
         errors: List[str] = []
         index = -1
+
+        await self.__update_progress(0, length_estimate)
+
```
RunDevelopment, line 204: Same here.

```suggestion

```

backend/src/process.py
```diff
@@ -197,17 +221,7 @@ async def run_while(
...
-                {
-                    "event": "iterator-progress-update",
-                    "data": {
-                        "percent": 1,
-                        "iteratorId": self.iterator_id,
-                        "running": None,
-                    },
-                }
-            )
+        await self.__finish_progress(max(length_estimate, index + 1))
```
RunDevelopment, line 224: This has to be just `index`.

```suggestion
        await self.__finish_progress(index)
```

backend/src/process.py
```diff
@@ -109,34 +110,52 @@ def __create_iterator_executor(self) -> Executor:
...
-                }
-            )
+    async def __finish_progress(self, length: int):
+        await self.executor.queue.put(
+            {
+                "event": "iterator-progress-update",
+                "data": {
+                    "percent": 1,
+                    "index": length + 1,
+                    "total": length + 1,
```
RunDevelopment, line 140: Why?

```suggestion
                    "index": length,
                    "total": length,
```

backend/src/process.py
joeyballentine: I tried that, and it was getting stuck at length - 1. Putting it after fixed it. Can't really explain that

backend/src/process.py
joeyballentine: Was trying to get it to not stop at length - 1

RunDevelopment (approved):

PR #1734, opened 2023-04-12
Fixed iterator progress

RunDevelopment:
> [...] it was getting stuck at length - 1. Putting it after fixed it. Can't really explain that

@joeyballentine Here's the explanation: iterator progress was ignored unless we also sent a list of running nodes. This was a bug you introduced in the first commit of #1730. So the last event you sent to set it to 100% was ignored. 

Changes:
- Fixed that progress updates were ignored unless they included a list of nodes.
- Fixed the length+1 in `__finish_progress`.
- Send progress updates before the running the iteration. This has better behavior than doing it after it.
- Fixed the length passed to `__finish_progress` in `run_while`.
- Record iteration duration even if the iteration errors.

I haven't tested it with video frame iterator yet (no videos on the disk of my laptop), so please verify that it behaves correctly there as well.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -136,8 +136,8 @@ class IteratorContext:
                 "event": "iterator-progress-update",
                 "data": {
                     "percent": 1,
-                    "index": length + 1,
-                    "total": length + 1,
+                    "index": length,
+                    "total": length,
                     "eta": 0,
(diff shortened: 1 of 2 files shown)


joeyballentine:
works 👍

joeyballentine (approved):

PR #1738, opened 2023-04-15
add generate_hash node

Zarxrax:
This node takes an image input and outputs a string containing a hash of the image data in hexadecimal or url-safe base64 encoding. There is a slider to adjust the digest size, which controls the length of the output string.

pyright was complaining about the type of the numpy image array on line 38, and I was not able to figure out how to resolve that.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/generate_hash.py
+++ b/backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/generate_hash.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+import base64
+import hashlib
+from typing import Tuple
+
+import numpy as np
+
+from nodes.properties.inputs import ImageInput, SliderInput
+from nodes.properties.outputs import TextOutput
+
+from .. import miscellaneous_group
+
+
+@miscellaneous_group.register(
+    schema_id="chainner:image:generate_hash",
+    name="Generate Hash",
+    description="Generate a hash from an image using the BLAKE2b hashing algorithm.",
+    icon="MdCalculate",
+    inputs=[
+        ImageInput(),
+        SliderInput(
+            "Digest Size (in bytes)",
+            minimum=1,
+            maximum=64,
+            default=8,
+            precision=1,
+            controls_step=1,
+        ),
+    ],
+    outputs=[
+        TextOutput("Hex", output_type="string"),
+        TextOutput("Base64", output_type="string"),
+    ],
+)
+def run(img: np.ndarray, size: int) -> Tuple[str, str]:
+    """Generate a hash from the input image. The digest size determines the length of the hash that is output."""
+    h = hashlib.blake2b(img, digest_size=size)  # type: ignore
+    return h.hexdigest(), base64.urlsafe_b64encode(h.digest()).decode("utf-8")


RunDevelopment:
About the image to hash: Chainner uses float32 images, so I wonder whether we should convert them to uint8 before hashing. 

float32 is inherently susceptible to rounding errors. If we ever so slightly changed the implementation of a node, we may get images that *look* identical, but have different hashes. [Floating point addition is neither commutative nor associative](https://www.quora.com/Is-floating-point-addition-commutative-and-associative), so this type of change occurs quite frequently in chainner, because we *assume* that it is (for various reasons). But since users will get totally different hashes, this PR effectively turned a lot of minor changes/optimizations to nodes into breaking changes since users may be relying upon getting the same hashes every time.

Even converting to uint8 won't perfectly solve this issue, but it will help a lot. If we do convert to uint8, I would suggest using the `to_uint8` method from image utils.

> pyright was complaining about the type of the numpy image array

That's because blake2b wants a [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object), which seems to be something only C types can implement, and so this requirement can't be expressed using Python types. So the type ignore is fine.

Zarxrax:
> About the image to hash: Chainner uses float32 images, so I wonder whether we should convert them to uint8 before hashing.
> 
> float32 is inherently susceptible to rounding errors. If we ever so slightly changed the implementation of a node, we may get images that _look_ identical, but have different hashes. [Floating point addition is neither commutative nor associative](https://www.quora.com/Is-floating-point-addition-commutative-and-associative), so this type of change occurs quite frequently in chainner, because we _assume_ that it is (for various reasons). But since users will get totally different hashes, this PR effectively turned a lot of minor changes/optimizations to nodes into breaking changes since users may be relying upon getting the same hashes every time.
> 
> Even converting to uint8 won't perfectly solve this issue, but it will help a lot. If we do convert to uint8, I would suggest using the `to_uint8` method from image utils.
> 

Well that's interesting to know. I did a completely unscientific test, and it did seem to agree that converting to uint8 might be better.

I took a folder of over 2000 images, and ran the hash on them, and saved the result with the hash as the filename. There were 4 versions - one where I did no processing, one where I inverted the image twice, once where I flipped the image horizontally twice, and once where I separated and re-merged the alpha channel. These were just some operations that I intuitively felt should not have any effect on the output.

From that, I compared the filenames across each of the 4 folders. For the images that had been inverted twice, none of the hashes matched up with the ones that had not been processed. But for the other folders, every file matched.
I then did a hash of the files themselves (outside of chainner), saved the new hash as the filename, and compared again. This time all 4 folders were perfect matches.

So what was interesting there, is that when I inverted the file twice, the hash ended up changing, even though when I hashed the actual resulting file, it didn't actually change. Could this be explained by the rounding error?

I also repeated the whole test except changed the code to convert to uint8 as you recommended. This time, all 4 folders of files matched perfectly.

RunDevelopment:
> So what was interesting there, is that when I inverted the file twice, the hash ended up changing, even though when I hashed the actual resulting file, it didn't actually change. Could this be explained by the rounding error?

Yup, that's rounding. Inverting is `1 - image`, so we have to do some calculations on the actual float32 values. It shouldn't have surprised me, but yes `1-(1-x) != x` for floating point numbers. Flipping only moves the pixels around and doesn't do any floating point calculations, so it's good that the hash didn't change there.

You didn't see any hash changes with the actual image file because Save Image and Copy To Clipboard also use `to_uint8` for rounding.

backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/generate_hash.py
```diff
@@ -0,0 +1,41 @@
...
+            minimum=1,
+            maximum=64,
+            default=8,
+            precision=1,
+            controls_step=1,
+        ),
+    ],
+    outputs=[
+        TextOutput("Hex", output_type="string"),
+        TextOutput("Base64", output_type="string"),
```
RunDevelopment, line 34: `string` is the default, so no need to specify it.

```suggestion
        TextOutput("Hex"),
        TextOutput("Base64"),
```

RunDevelopment (approved):

RunDevelopment (approved):

Issue #1686, opened 2023-03-30
Median Stack Node


joeyballentine:
A suggestion from a user on Discord.

An image median node. Basically, stack _n_ images (probably just 4 for now), sort each pixel, along the stack axis, and then take the middle one (or average of the middle two if an even number). This would probably be pretty inefficient to do pixel by pixel but I bet numpy has a function for this.

RunDevelopment:
What's the median of a list of pixels? Let's say we have 3 pixels, how do we define what their median is? Do we calculate the median per RGB channel, do we do it in a perceptual color space, or do we define some selector function that turns each pixel into a single numerical value (e.g. lightness)?

joeyballentine:
This is how it was described to me: https://discord.com/channels/930865462852591648/930865463318179952/1087449448906502224

JustNoon:
In imagemagick and photoshop which is probably the gold standard, median is just intensity in RGB colorspace.

[This ](https://petapixel.com/2013/05/29/a-look-at-reducing-noise-in-photographs-using-median-blending/)is an excellent write-up on the subject.

JustNoon:
Implemented in #1745

PR #1745, opened 2023-04-17
Add blend stacking node (Z-Stack)

JustNoon:
Implements #1686 as well as mean, minimum, and maximum.

I chose 2-15 images with 2 being a minimum.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_utility/compositing/z_stack.py
+++ b/backend/src/packages/chaiNNer_standard/image_utility/compositing/z_stack.py
@@ -1,0 +1,79 @@
+from __future__ import annotations
+
+from enum import Enum
+from typing import List, Union
+
+import numpy as np
+
+from nodes.group import group
+from nodes.properties.inputs import EnumInput, ImageInput
+from nodes.properties.outputs import ImageOutput
+from nodes.utils.utils import ALPHABET, get_h_w_c
+
+from .. import compositing_group
+
+
+class Expression(Enum):
+    MEDIAN = "median"
+    MEAN = "mean"
+    MIN = "minimum"
+    MAX = "maximum"
+
+
+def align_images(images: List[np.ndarray]) -> np.ndarray:
+    center = np.mean(images, axis=0)
+    center_h, center_w, center_c = get_h_w_c(center)
+    aligned_images = []
+    for image in images:
+        h, w, c = get_h_w_c(image)
+        diff_y, diff_x = (center_h - h) // 2, (center_w - w) // 2
+        aligned_image = np.zeros_like(center)
+        aligned_image[diff_y : (diff_y + h), diff_x : (diff_x + w)] = image
+        aligned_images.append(aligned_image)
+    return np.array(aligned_images)
+
+
+@compositing_group.register(
+    schema_id="chainner:image:z_stack",
+    name="Z-Stack Images",
+    description="""Aligns multiple images and evaluates them in relation to each other.""",
+    icon="BsLayersHalf",
+    inputs=[
+        EnumInput(Expression),
+        ImageInput("Image A"),
+        ImageInput("Image B"),
+        group("optional-list")(
+            *[
+                ImageInput(f"Image {letter}").make_optional()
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_standard/image_utility/compositing/z_stack.py
```diff
@@ -0,0 +1,72 @@
...
+from nodes.group import group
+from nodes.properties.inputs import EnumInput, ImageInput
+from nodes.properties.outputs import ImageOutput
+from nodes.utils.utils import ALPHABET, get_h_w_c
+
+from .. import compositing_group
+
+class Expression(Enum):
+    MEDIAN = "median"
+    MEAN = "mean"
```
RunDevelopment, line 17: Which [mean](https://en.wikipedia.org/wiki/Mean#Types_of_means) is it?

JustNoon: Arithmetic. Should be the same results as if you had blended n layers with 1/n opacity each (with most transparent at the top).

JustNoon:
I think I've finally pleased the lint and formatting demons so I've rebased to current commit.

Any review corrections I'll add as separate commits going forward.

RunDevelopment:
This does not work if the images aren't all the same shape.

image
image


Why don't we just require that all images need to have the same size and number of channels? Would this limit the use case for this node?

JustNoon:
> This does not work if the images aren't all the same shape.
> 
> image image
> 
> Why don't we just require that all images need to have the same size and number of channels? Would this limit the use case for this node?

It certainly wouldn't be a problem for my use-cases but I didn't know how people would use it. It could always be expanded to handle dimensions and channels.

backend/src/packages/chaiNNer_standard/image_utility/compositing/z_stack.py
```diff
@@ -0,0 +1,68 @@
...
+    outputs=[ImageOutput()],
+)
+def z_stack(
+    expression: Expression,
+    *args: Union[np.ndarray, None],
+) -> np.ndarray:
+    """Align and evaluate images"""
+
+    inputs: List[Union[np.ndarray, None]] = [*args]
+    images = [x for x in inputs if x is not None]
```
RunDevelopment, line 48: Just cause it's bothering me. 
```suggestion
    *inputs: np.ndarray | None,
) -> np.ndarray:
    """Align and evaluate images"""

    images = [x for x in inputs if x is not None]
```

backend/src/packages/chaiNNer_standard/image_utility/compositing/z_stack.py
```diff
@@ -0,0 +1,68 @@
...
+        ImageInput("Image A"),
+        ImageInput("Image B"),
+        group("optional-list")(
+            *[
+                ImageInput(f"Image {letter}").make_optional()
+                for letter in ALPHABET[2:14]
+            ],
+        ),
+    ],
+    outputs=[ImageOutput()],
```
RunDevelopment, line 39: Okay, this is going to be just a little painful. So the type system doesn't support arrays yet, so we have to reference each input image individually...
```suggestion
    outputs=[
        ImageOutput(
            image_type="""
                def conv(i: Image | null) = match i { Image => i, _ => any };

                Input1 & Input2
                    & conv(Input3)
                    & conv(Input4)
                    & conv(Input5)
                    & conv(Input6)
                    & conv(Input7)
                    & conv(Input8)
                    & conv(Input9)
                    & conv(Input10)
                    & conv(Input11)
                    & conv(Input12)
                    & conv(Input13)
                    & conv(Input14)
            """
        ).with_never_reason(
            "All input images much have the same size and number of channels."
        ),
    ],
```

RunDevelopment (commented):
Last suggestions, then gtg.

JustNoon:
Thanks for the help. Seriously.

The parameters got that way in my fight with pywright and I didn't want to touch it anymore after we had a ceasefire.

joeyballentine (approved):

PR #1748, opened 2023-04-18
Added Image Statistic node

RunDevelopment:
A node to get the min, max, mean, media, n-th percentile of an image. This was requested by JCtheMC on Dicord.

Right now, only grayscale images are supported, because they guarantee that the result will be a single number. The resulting number is output as a floating point number in 0-255 range.

Open questions:
- Should the node output a color instead? This would make it pretty easy to extend the node to multiple channels, but would limit the number of operations we can do. E.g. a median color makes sense, but variance doesn't.
- Should the output format be 0-255?
- Should the node be called "Image Statistic"? I got this from numpy docs, which lists `np.min`, `np.max`, `np.media`, etc. under a "Statistics" category. It makes sense for numpy, but I don't know whether it's the right name for us.

image


theflyingzamboni:
I would use "Statistics" rather than "Statistic". I get why you used the singular (it outputs one at a time), but it feels a little off in English? It would feel more natural for it to refer to the full scope of what it contains. Broadly, the term for these values is "descriptive statistics", so I think Image Statistics makes sense as a node name. 

I think this node would make more sense if all the statistics were separate outputs instead of dropdown options. This would bring it in line with the Image Metrics and Get Dimensions nodes, and be much more convenient, since you wouldn't need a node for each stat you wanted. I realize the percentile adjustment complicates that.

...Is there a way we could group an input with a certain output?

RunDevelopment:
> I would use "Statistics" rather than "Statistic"

👍 

> I think this node would make more sense if all the statistics were separate outputs instead of dropdown options.

True. I originally did this for performance reasons to avoid useless computations, but all stats are super fast, so I think it's okay.

joeyballentine:
> I originally did this for performance reasons to avoid useless computations

Makes me wonder if we could have a way of conditionally computing stuff like that, so that it only computes an output if there's something attached to it (for certain nodes)

theflyingzamboni:
I was wondering the same thing as soon as I read that.

joeyballentine (approved):

joeyballentine:
No reason to keep this open, gonna merge it

PR #1750, opened 2023-04-19
Added lazy outputs

RunDevelopment:
This adds support for lazily-computed node outputs via the new `Lazy[T]` class.

Caveats:
- Broadcasts are not supported for lazy outputs.
- Since lazy output values are computed by `enforce_input`, the time it takes to compute them counts towards the execution time of the target node. This might make measuring the performance of nodes a bit more confusing, because we have to keep in mind which outputs are eagerly and which are lazily computed.

@joeyballentine Have fun with type validation :)

### DIFF
--- a/backend/src/lazy.py
+++ b/backend/src/lazy.py
@@ -1,0 +1,28 @@
+from __future__ import annotations
+
+from typing import Callable, Generic, TypeVar, cast
+
+from sanic.log import logger
+
+T = TypeVar("T")
+
+
+class Lazy(Generic[T]):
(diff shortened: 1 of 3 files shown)


RunDevelopment:
It's all coming back...

joeyballentine:
@RunDevelopment Instead of doing this, how about we make the metrics node only output one number, and you just select from a dropdown which metric you want to calculate?

That would make that node more akin to others of that nature. Imagine if Save image saved every format rather than having you pick one, or if resize output every algorithm instead of having you pick one.

Then, we solve the problem of calculating all 3 metrics, but we also don't need lazy outputs.

PR #1766, opened 2023-04-28
Add missing features from <Text/> to <TextArea/>

joeyballentine:
When testing #1762, I noticed that we don't currently disable or show type previews on TextAreas like we do for Text inputs normally. So I added that stuff in.


RunDevelopment:
At this point, is there much of a difference between our text and text area inputs? Maybe we should unify them into a single input (in both frontend and backend) and have a `multiline: bool` option to choose between `<input type=text/>` and `<textarea>` in the frontend.

But if we do, then it can be a follow up PR.

RunDevelopment (approved):

joeyballentine:
Yeah we probably should

PR #1777, opened 2023-05-04
Pixel Blur

harisreedhar:
chaiNNer-pixel_blur-2023-5-4_21-15

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_filter/blur/pixel_blur.py
+++ b/backend/src/packages/chaiNNer_standard/image_filter/blur/pixel_blur.py
@@ -14,8 +14,8 @@ from .. import blur_group
     icon="MdBlurOn",
     inputs=[
         ImageInput(),
-        NumberInput("Size X", controls_step=1),
-        NumberInput("Size Y", controls_step=1),
+        NumberInput("Size X", default=10, precision=0, minimum=1, controls_step=1),
+        NumberInput("Size Y", default=10, precision=0, minimum=1, controls_step=1),
     ],
     outputs=[ImageOutput(image_type="Input0")],
 )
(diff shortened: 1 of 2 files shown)


RunDevelopment:
Maybe I'm misreading the code, but won't the dimensions of the output image be a multiple of the given block size? E.g. with an input image of width 100 and a block size of 16, won't the width of the output image be 96?

harisreedhar:
> Maybe I'm misreading the code, but won't the dimensions of the output image be a multiple of the given block size? E.g. with an input image of width 100 and a block size of 16, won't the width of the output image be 96?

You are right. I have added padding.

RunDevelopment:
I have no idea, why the CI fails. Pyright suddenly says that it can't find `numpy.bool8` anymore in totally unrelated files. I thought that it might be that a new numpy version 1.24.3 broke something, but we already have other successful CI runs with 1.24.3, so it's not that.

---

About the node itself: I think the implementation looks fine now, but I'm not sure about the name and category. Is this a blur? I'm not sure. For reference, Paint.net has its Pixelate effect under the Distortion category. Why did you call this node Pixel *Blur*? (I'm not saying it's bad, I just want to know.)

harisreedhar:
Blur literally means making an image unclear and less distinguishable. This node does the same thing, so I put it in the blur category. Different programmes use different names and categories. In Gimp, it is under the blur category called "pixelize," and in Photoshop, it is called "mosaic."

joeyballentine:
> in Photoshop, it is called "mosaic."

It's also not under the blur category, it's in "Pixelate". Tbh I'm not sure what to do here. Pixelation doesn't scream "blur" to me but I can see how it fits in with the idea of obscuring information.

harisreedhar:
Do you have any suggestion?

RunDevelopment:
Maybe we could call the node Pixelate and move it into Image Filters > Misc? We don't have a "Distortion" category yet, so misc seems fitting. I think that the node name should have "pixel" in it somewhere (for searchability), so I'm against calling it "Mosaic."

RunDevelopment (approved):
Thank you @harisreedhar!

PR #1814, opened 2023-05-22
add type hint for function `timed_supplier_async`

shispt:
This PR adds parameter type and return type for function `timed_supplier_async`, just like function `timed_supplier`


RunDevelopment (approved):
Thank you @shispt!

PR #1819, opened 2023-05-23
More group functions

RunDevelopment:
This adds function for the seed and optional list groups. I wanted to document the groups in python, so people can find and use them more easily.


joeyballentine:
Should group types be an enum maybe?

RunDevelopment:
I was thinking about that too, but then options should be typed as well, which is difficult to do in python (since the object type depends on the key type).

I also think that making them functions is better for documentation purposes. Making them functions will also allow us to build abstractions around them. Like how `if_group` and `if_enum_group` are for `"conditional"`.

joeyballentine (approved):

Issue #1677, opened 2023-03-25
Non-deterministic nodes


RunDevelopment:
**Motivation**
Some features we would like to add as nodes are inherently non-deterministic. This includes relatively simple things like "what's the current time?", "how many times has the user run this chain?", and "give me a random number." These example, while very simple, are very useful, so it would be nice if we could add them.

The problem with non-deterministic nodes is that they break chaiNNer. Currently, pretty much everything assumes that nodes are deterministic, and even more systems will make that assumption in the future. Here are some things that non-deterministic nodes break right now or will break in the future:
- The current chain executor assumes that all nodes are deterministic and uses this assumption to cache nodes.
- Trivial optimizations (such as iterator outlining) aren't correct anymore. 
- Type broadcasts aren't correct, because a static type system requires determinism to propagate types.
- Live execution (planned feature) relies on caching intermediate results from previous runs, and non-deterministic nodes can't be cached.

**Description**
Find a way to support non-deterministic nodes without breaking chaiNNer. This includes answering the following questions:
- What are the **semantics** of the outputs of non-deterministic nodes? 
    This is mostly about what down-stream nodes see. E.g. `R` be a non-deterministic node that generates a random number, and let `R` be connected to 2 nodes `A` and `B` like so:
    ```
    R ─> A
    └──> B
    ```
    Do `A` and `B` see the same random number? If so, *why* would this change if `B` is inside an iterator?
- How do these semantics interact with caching and optimization?
- Are there non-deterministic nodes that are acceptable to cache, e.g. for live execution? If so, what property do they have that makes it acceptable?

jackcarey:
1. Semantics: nodes could implement non-deterministic behaviour, but the outputs should still be *predictable*. IMO, each output should pass the same value to all connected nodes. If more than one random number is needed for example, then the extendable list pattern could be taken from 'Text Append' to add more outputs with different values. For iterators: if the non-deterministic node isn't part of the iteration scope itself, I'm of the opinion that it *shouldn't* recalculate each time, because that's the flow would essentially have to either "work backwards" or "pull the non-deterministic node forward" to create a new output.

2. Caching & Optimizations: in a similar position to the toggle for enabling and disabling a node at the moment, how would you feel about a button to 'freeze' node outputs? Or making that toggle handle freezing as well? This could be part of the base node definition as a 'freezable' attribute? I think, with some refinement, this could be a workable solution. It'd have to be intuitive as to how it works though, which will take a bit more thought.

[...]

PR #1820, opened 2023-05-24
Added Emoji icon support, current time node

jackcarey:
* Introduced allowing single characters if the `icon` attribute on a node has a [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) length of 1 - useful for emoji or other non-latin characters as an alternative to react-icons
* Added URLInput as a generic input. Uses `urllib.parse` and an allowlist to check if a string is a URL and optionally, if it matches a regex pattern. (not used yet, but intended for future use)
* Added a 'Current Datetime' node under `utility\value\` which includes some formatting options. Useful with 'Text Append' to create unique file names every time a flow is run.


RunDevelopment:
Oh no... You picked the worst possible feature to get started with chainner 😄 

Chainner currently assumes that all nodes are deterministic, and returning the current data time is certainly not deterministic. We don't really support non-deterministic nodes right now. See #1677 for more details.

jackcarey:
That's a shame but there are good questions on that issue so I'm happy for this to be closed. Are the Emoji icons something you'd accept on its own? If so, I'll re-submit that separately

RunDevelopment:
> Are the Emoji icons something you'd accept on its own?

I wonder. Emojis have the disadvantage that their appearance is platform dependent, so we lose control over how they are displayed. Platforms may also choose to give them colors on their own and ignore the theme color we set. As I see it, the main advantages of Emojis are that they are easy to pick, and that we don't pay for them (the part of React Icons that we use is 4MB IIRC).

What do you think @joeyballentine?

joeyballentine:
I'm not totally against the emoji icon support in general, but i wouldn't want to use it for any of our built in nodes unless we can guarantee the appearance on all platforms

joeyballentine:
Are we able to set a custom emoji font?

jackcarey:
The display issue could be solved with something like [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji) and it's [black and white fallback](https://fonts.google.com/noto/specimen/Noto+Emoji) to make them consistent. Color can be controlled using [palletes](https://css-tricks.com/colrv1-and-css-font-palette-web-typography/) if you want to get fancy, or plain font color if the monochromatic version is used.

RunDevelopment:
Yup. Let's do it :)

jackcarey:
emoji icons to be modified and re-submitted

RunDevelopment:
Thank you!

PR #1829, opened 2023-05-25
Possibly fix portable building on linux and macos

joeyballentine:
- Places the `portable` file in (what I assume) is the correct path for macos and linux

To verify this, I needed to make it build and upload artifacts to CI, so this PR also:
- Makes it so we can run the build workflow by committing a commit with "build" in the message
- Makes it so the build workflow uploads artifacts to the action

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -15,9 +15,8 @@ on:
 
 jobs:
   test-build-release:
-    # Set this to `true` to enable the workflow.
-    # This workflow is disabled by default because it takes multiple minutes in CI which slows down development.
-    if: false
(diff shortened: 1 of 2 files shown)


RunDevelopment (approved):
I assume that you have tested the portable stuff and that it works correctly.

> Makes it so we can run the build workflow by committing a commit with "build" in the message

That's really clever, I like that.

joeyballentine:
> I assume that you have tested the portable stuff and that it works correctly.

Nope, because i don't use macos or Linux

joeyballentine:
But it's definitely currently broken so there's no reason not to try this

PR #1830, opened 2023-05-25
Fix argument parsing to allow passthrough of extra args

### LINKED ISSUE #1735: Uable to install v.0.18.5 on Windows
Hey, did anybody encounter this issue as well? When trying to install it the first time, Windows had a pop saying something about "not trustworthy". I clicked "trust anyways" and from there on nothing happens after starting the exe file. 

Are there any steps I have to do beforehand etc.? Or is there a way to fix it?

### LINKED ISSUE #1746: [Feature Request]: Something in Windows Installer to let you know it's installed.
**Motivation**
Please explain the problem you're having/why you propose this feature. Ex. I'm always frustrated when [...]

When I downloaded and ran the Windows installer "chaiNNer-0.18.6-x64-windows-setup.exe", today, nothing happened--_I thought_. I assumed it was crashing, but saw that something called "updater.exe" was trying to access the internet, so I looked further. It wasn't until I checked the Issues list that I realized it had installed. It runs so fast, the thought didn't occur to me and there's no notification or anything, so I assumed it had crashed. 

**Description**
Group the installer and the stuff chaiNNer does on first run (installing dependencies) and make a window with something like:
"Installing chaNNer..." 
then when done successfully, something like: 
"Installation complete! Click "chaiNNer" in Start Menu to run." etc... 

Alternatively, you could make a note in the install instructions eluding to how the install works. This is a kinda small issue, but I don't want you guys to lose even a single user because he/she thought the program hadn't worked. Thank you for making this amazing tool!

### LINKED ISSUE #1818: `--no-sandbox` Electron flag causes ChaiNNer to error
**Information:**

- Chainner version: v0.18.7
- OS: Debian Bullseye

**Description**

The `--no-sandbox` Electron flag is necessary on some platforms (e.g. ppc64le) where the Chromium sandbox is unavailable. Unfortunately, it seems that its presence confuses ChaiNNer.

**Steps to reproduce**

`./chainner --no-sandbox`

**Expected results**

ChaiNNer should launch the GUI as usual (Electron will disable the sandbox).

**Observed results**

`Unknown argument: sandbox`

**Logs**

Nothing is logged when ChaiNNer is launched in this way.

joeyballentine:
Theoretically fixes #1818 
For sure fixes #1746 and fixes #1735

I can't believe it was this easy all along....

Need to regression test CLI mode with this, but theoretically it should still work fine.

### DIFF
--- a/src/main/arguments.ts
+++ b/src/main/arguments.ts
@@ -68,7 +68,7 @@ export const parseArgs = (args: readonly string[]): ParsedArguments => {
                 hidden: true,
             },
         })
-        .strict()
+        .parserConfiguration({ 'unknown-options-as-args': true })
         .parseSync();
 
     const options: ArgumentOptions = {


JeremyRand:
Wow, I definitely wasn't expecting my ppc64le bug report to be the key to solving the Windows bug your release notes have warned about for months. Not complaining though. Let me know once there's a binary for me to test and I'll give it a spin. :)

RunDevelopment (approved):

joeyballentine:
@RunDevelopment did you test CLI mode with this? I have not because i don't know its intended behavior that well

RunDevelopment:
Argument parsing for GUI and CLI mode are practically identical. If it works for GUI mode, it'll work for CLI mode as well.

PR #1831, opened 2023-05-25
Added support for single character (emoji) icons

jackcarey:
Aimed at non-latin chars (incl. emoji) using [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter). [Noto Emoji](https://fonts.google.com/noto/specimen/Noto+Emoji) is used to keep appearance consistent across platforms.

Screenshot 2023-05-25 084513


RunDevelopment (commented):
Thanks for the PR!

CI is failing, because we apparently don't use lib ES2022. I made a PR for that #1833, so this will be possible soon.

src/renderer/components/CustomIcons.tsx
```diff
@@ -139,7 +139,29 @@ export const IconFactory = memo(
                     />
                 );
             default:
-            // nothing
+                break;
+        }
+
+        // using segmenter to account for non-latin and emoji characters
+        const isSingleCharacter = [...new Intl.Segmenter().segment(icon)].length == 1;
```
RunDevelopment, line 146: ESLint complains :)

```suggestion
        const isSingleCharacter = [...new Intl.Segmenter().segment(icon)].length === 1;
```

src/renderer/components/CustomIcons.tsx
```diff
@@ -139,7 +139,29 @@ export const IconFactory = memo(
...
+                    aria-hidden="true"
+                    className="chakra-heading"
+                    role="presentation"
+                    style={{
+                        color: accentColor,
+                        'font-weight': 'bold',
+                        'text-align': 'center',
+                        'vertical-align': 'middle',
+                        'text-rendering': 'geometricPrecision',
+                        'font-family': 'Noto Emoji, Open Sans, sans-serif',
```
joeyballentine, line 159:
```suggestion
                        fontWeight: 'bold',
                        textAlign: 'center',
                        verticalAlign: 'middle',
                        textRendering: 'geometricPrecision',
                        fontFamily: 'Noto Emoji, Open Sans, sans-serif',
```
React's style objects use camel case instead of kebab kase

jackcarey:
Open Sans and Noto Emoji alone aren't enough to support all non-latin characters. Bundling a lot of font files for this will bloat the application size. I will look to add local Noto variants for: latin, CJK, arabic, devanagari, bengali, cyrillic with the view that more can be added as and when needed.

RunDevelopment:
> Open Sans and Noto Emoji alone aren't enough to support all non-latin characters.

It's good enough. Let's make some nodes that actually use single character icons before trying to make the feature perfect ;)

RunDevelopment (approved):

joeyballentine:
Thanks again for this! Sorry for things being a bit slow lately, I'm hoping to get more development rolling soon

PR #1832, opened 2023-05-25
Upgrade to TypeScript v5

RunDevelopment:


RunDevelopment:
I also updated Jest to make sure that TS-jest works fine. Jest changed its snapshot format to be more JSON-like, so I had to update snapshots as well.

joeyballentine (approved):

PR #1834, opened 2023-05-25
Backend-defined Dependencies

joeyballentine:
Mostly finished. I just need to review everything because I'm sure I left tons of commented-out code everywhere and probably just need to clean stuff up as well. But functionally its working.

This PR defines dependencies on the backend side. This means dependencies are linked directly to a package now.

The python process also installs all the core dependencies itself through pip commands in a subprocess. This allows someone to run the backend remotely and still have all the core deps be installed. The core deps just get auto-installed the way they did before.

`run.py` now contains some imports that are required to be in that order. Basically, it imports the stuff that checks and installs the dependencies before doing anything else. It also force installs sanic, sanic-cors, and semver before doing any others.

This also implements the nvml python package to check for an nvidia GPU on the backend side rather than on the frontend. this allows us to do the same GPU checks as before, only without needing to search for nvidia smi.

What this doesn't have is the ability to install optional dependencies through the backend. I'll do that in a separate PR.


backend/src/packages/chaiNNer_standard/__init__.py
```diff
@@ -1,8 +1,29 @@
...
+        # Dependency("appdirs", "appdirs", "1.4.4", 13.5 * KB),
+        # Dependency("FFMPEG", "ffmpeg-python", "0.2.0", 25 * KB, import_name="ffmpeg"),
+        # Dependency("Requests", "requests", "2.28.2", 452 * KB),
+        # Dependency("re2", "google-re2", "1.0", 275 * KB, import_name="re2"),
+        # Dependency("scipy", "scipy", "1.9.3", 42 * MB),
+    ],
+)
+
+if is_arm_mac:
+    package.add_dependency(Dependency("Pasteboard", "pasteboard", "0.3.3", 19 * KB))
```
joeyballentine, line 22: These need to be moved to the other file

backend/src/packages/chaiNNer_standard/__init__.py
```diff
@@ -1,8 +1,29 @@
...
-from api import add_package
+from api import KB, MB, Dependency, add_package, is_arm_mac, is_windows
 
-package = add_package(__file__, name="chaiNNer_standard", dependencies=[])
+package = add_package(
+    __file__,
+    name="chaiNNer_standard",
+    description="The standard set of nodes for chaiNNer.",
+    dependencies=[
+        # Dependency("OpenCV", "opencv-python", "4.7.0.68", 30 * MB, import_name="cv2"),
```
joeyballentine, line 10: Comments should be removed. Though this code might be able to actually go there instead of using the hardcoded non-Dependency-class version

backend/src/setup_deps.py
```diff
@@ -0,0 +1,116 @@
...
+        #             f"{dependency['package_name']}=={dependency['version']}",
+        #         ]
+        #     )
+        #     installed_packages[dependency["package_name"]] = dependency["version"]
+        install_dependency(dependency["package_name"], dependency["version"])
+
+
+install_required_dependencies()
+
+other_required_dependencies = [
```
joeyballentine, line 70: These might be able to use the Dependency dataclass. But i wonder if that's even worth it tbh since these don't need to be able to have their data sent to the frontend

src/main/backend/setup.ts
```diff
@@ -178,65 +171,56 @@ const ensurePythonDeps = async (
...
     try {
-        const pipList = await runPipList(pythonInfo);
-        const installedPackages = new Set(Object.keys(pipList));
-
-        const requiredPackages = requiredDependencies.flatMap((dep) => dep.packages);
-        const optionalPackages = getOptionalDependencies(hasNvidia, pythonInfo.version).flatMap(
-            (dep) => dep.packages
-        );
-
+        // const pipList = await runPipList(pythonInfo);
```
joeyballentine, line 174: I can remove all this stuff

joeyballentine (commented):
TLDR: TODO is to clean up come comments and unused code. I also need to test what happens on a fresh start.

backend/src/setup_deps.py
```diff
@@ -0,0 +1,116 @@
...
+initial_required_dependencies = [
+    {
+        "package_name": "sanic",
+        "version": "23.3.0",
+    },
+    {
+        "package_name": "Sanic-Cors",
+        "version": "2.2.0",
+    },
+    {"package_name": "pynvml", "version": "11.5.0"},
```
joeyballentine, line 42: Also so i don't forget, this can be moved to the other deps stuff. Idk why it's here

joeyballentine:
So this has an issue: Startup up the backend for the first time takes some time due to the deps being installed. I'm not really sure what to do about that, since we time out the backend if it doesn't respond within a certain time. gonna have to think about this a bit

src/common/dependencies.ts
```diff
@@ -18,196 +13,6 @@ export interface PyPiPackage {
 }
 export interface Dependency {
     name: string;
-    packages: PyPiPackage[];
+    dependencies: PyPiPackage[];
```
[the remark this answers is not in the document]
RunDevelopment: Fair.

backend/src/dependencies/install_required_deps.py
```diff
@@ -0,0 +1,46 @@
...
+# We can't rely on using the package's __version__ attribute because not all packages actually have it
+try:
+    pip_list = subprocess.check_output(
+        [python_path, "-m", "pip", "list", "--format=json"]
+    )
+    # set_installed_packages({p["name"]: p["version"] for p in json_parse(pip_list)})
+    for p in json_parse(pip_list):
+        installed_packages[p["name"]] = p["version"]
+except Exception as e:
+    installed_packages = {}
```
RunDevelopment, line 17: How is this even allowed by the language?

backend/src/gpu.py
```diff
@@ -0,0 +1,59 @@
...
+    nv.nvmlShutdown()
+except nv.NVMLError as e:
+    logger.info("No Nvidia GPU found, or invalid driver installed.")
+except Exception as e:
+    logger.info(f"Unknown error occurred when trying to initialize Nvidia GPU: {e}")
+
+
+class NvidiaHelper:
+    def __init__(self):
+        self.nvidia_is_available = nvidia_is_available
```
RunDevelopment, line 18: This field is never used.

backend/src/gpu.py
```diff
@@ -0,0 +1,59 @@
...
+                }
+            )
+
+    def __del__(self):
+        if nvidia_is_available:
+            nv.nvmlShutdown()
+
+    def list_gpus(self):
+        if not nvidia_is_available:
+            return None
```
RunDevelopment, line 44: None of those checks make sense. If an Nvidia GPU wasn't available, we wouldn't have been able to construct this class instance.

backend/src/dependencies/install_required_deps.py
```diff
@@ -0,0 +1,46 @@
...
+        "version": "1.23.2",
+    },
+    {
+        "package_name": "opencv-python",
+        "version": "4.7.0.68",
+    },
+    {
+        "package_name": "Pillow",
+        "version": "9.2.0",
+    },
```
RunDevelopment, line 40: I think we should move those 2 into `chainner_standard` or other deps.

backend/src/dependencies/install_other_deps.py
```diff
@@ -0,0 +1,58 @@
...
+        "version": "1.4.4",
+    },
+    {
+        "package_name": "ffmpeg-python",
+        "version": "0.2.0",
+    },
+    {
+        "package_name": "requests",
+        "version": "2.28.2",
+    },
```
RunDevelopment, line 37: This is only used by external SD.

RunDevelopment (commented):
We should also get rid of all pip calls in the frontend and move them into the backend. Since the backend is now responsible for deps, it should be the single source of truth for them. Basically, I want `common/pip.ts` gone.

joeyballentine:
@RunDevelopment see where i said "What this doesn't have is the ability to install optional dependencies through the backend. I'll do that in a separate PR." 

I want to do that separately because I'm order to get progress i wanna attempt using the pseudo terminal stuff and i don't wanna worry about that as a part of this PR. I also have other backend stuff i wanna change and i went to get this PR in first before i do that and i don't want more code changes holding it up

backend/src/dependencies/store.py
```diff
@@ -0,0 +1,30 @@
...
+def install_dependency(package_name, version):
+    if package_name not in installed_packages:
+        subprocess.check_call(
+            [
+                python_path,
+                "-m",
+                "pip",
+                "install",
+                "--upgrade",
+                f"{package_name}=={version}",
```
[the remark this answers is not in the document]
joeyballentine: I wasn't aware of that. For some reason i was thinking there was a problem with the way we currently do it which is why i did it this way, but I'll change that

backend/src/dependencies/install_required_deps.py
joeyballentine: Just python things

backend/src/dependencies/install_required_deps.py
joeyballentine: Can't because the API needs them. I explained this in discord. To make that not happen I'd need to move around a ton of files and i don't want to do that in this PR

backend/src/dependencies/install_other_deps.py
joeyballentine: If i move that over, then that means external SD will be an optional installed package (which is fine) but i'd rather just make this required and have a separate method for making external integrations optional, rather than relying on this one dependency to do that. especially since we could have any number of external connections in the future

src/common/Backend.ts
```diff
@@ -173,6 +174,10 @@ export class Backend {
     pythonInfo(): Promise<PythonInfo> {
         return this.fetchJson('/python-info', 'GET');
     }
+
+    dependencies(): Promise<{ name: string; dependencies: PyPiPackage[]; description: string }[]> {
```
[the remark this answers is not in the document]
joeyballentine: yes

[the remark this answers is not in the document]
joeyballentine: actually, no. it's everything including the 0-dep ones. it gets filtered in DependencyContext:

```
const res = await backend.dependencies();
return res.filter((d) => d.dependencies.length > 0);
```

backend/src/api.py
```diff
@@ -268,8 +301,12 @@ def _refresh_nodes(self):
                         self.nodes[node.schema_id] = node, sub
 
 
+# pylint: disable=global-at-module-level
+global registry
```
RunDevelopment, line 305: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/global-at-module-level.html
> Used when you use the "global" statement at the module level since it has no effect.

I don't think we need this then.

backend/src/api.py
```diff
@@ -174,11 +178,34 @@ def toDict(self):
         }
 
 
+@dataclass
+class Dependency:
+    display_name: str
+    pypi_name: str
+    version: str
+    size_estimate: int | float
+    auto_update: bool = False
```
RunDevelopment, line 187: `auto_update` is only used for NCNN. Why?

backend/src/gpu.py
```diff
@@ -0,0 +1,51 @@
+import pynvml as nv
+from sanic.log import logger
+
+# pylint: disable=global-at-module-level
+global nvidia_is_available
```
RunDevelopment, line 5: Same.

backend/src/gpu.py
```diff
@@ -0,0 +1,51 @@
...
+    def get_current_vram_usage(self, gpu_index=0):
+        info = nv.nvmlDeviceGetMemoryInfo(self.__gpus[gpu_index]["handle"])
+
+        return info.total, info.used, info.free
+
+
+__all__ = [
+    "nvidia_is_available",
+    "NvidiaHelper",
+]
```
RunDevelopment, line 51: Do we need this?

backend/src/server.py
```diff
@@ -0,0 +1,557 @@
...
+            install_version_checked_dependencies(dep_info)
+        except Exception as ex:
+            logger.error(f"Error installing dependencies: {ex}")
+
+    # Manually import built-in packages to get ordering correct
+    # Using importlib here so we don't have to ignore that it isn't used
+    importlib.import_module("packages.chaiNNer_standard")
+
+    # for dep in next(iter(api.registry.packages.values())).dependencies:
+    #     install_dep(dep)
```
RunDevelopment, line 420: Please deal with the commented-out code.

src/common/Backend.ts
RunDevelopment, line 178: > { name: string; dependencies: PyPiPackage[]; description: string }

Let's add a type for this. Maybe `ChainnerPackage`?

RunDevelopment: Oh, you called it `Package` in dependencies.ts. That works too. Please use the type.

src/common/dependencies.ts
```diff
@@ -1,13 +1,8 @@
...
 import { Version } from './common-types';
-import { isArmMac, isMac, isWindows } from './env';
-
-const KB = 1024 ** 1;
-const MB = 1024 ** 2;
-const GB = 1024 ** 3;
 
 export interface PyPiPackage {
-    packageName: string;
+    displayName?: string;
```
RunDevelopment, line 4: This is no longer optional.
```suggestion
    displayName: string;
```

src/common/dependencies.ts
```diff
@@ -16,198 +11,8 @@ export interface PyPiPackage {
     sizeEstimate: number;
     autoUpdate?: boolean;
```
RunDevelopment, line 12: This isn't optional.

```suggestion
    autoUpdate: boolean;
```

src/common/dependencies.ts
```diff
@@ -1,13 +1,8 @@
...
-const KB = 1024 ** 1;
-const MB = 1024 ** 2;
-const GB = 1024 ** 3;
 
 export interface PyPiPackage {
-    packageName: string;
+    displayName?: string;
+    pypiName: string;
     version: Version;
     findLink?: string;
```
RunDevelopment, line 7: This type is slightly wrong, probably from before. The link can be string or null, not string or undefined.

```suggestion
    findLink?: string | null;
```

src/renderer/splash.tsx
```diff
@@ -14,11 +20,41 @@ const Splash = memo(() => {
...
     const [overallProgress, setOverallProgress] = useState(0);
 
+    const [port, setPort] = useState<number>(8000);
+    useAsyncEffect(
+        () => ({ supplier: () => ipcRenderer.invoke('get-port'), successEffect: setPort }),
+        []
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [eventSource, eventSourceStatus] = useBackendEventSource(port);
```
RunDevelopment, line 30: Why is this stilll here?

backend/src/api.py
joeyballentine: It's like this currently in chaiNNer. since we control the NCNN package and have introduced features side-by-side with chaiNNer updates, we figured having it auto update (if installed) was the right call to avoid issues. We obviously don't want this happening with something like pytorch and it's not really necessary for anything else. But since we did things like change our ncnn implementation to use new functions we added to the bindings, we need ncnn to stay up to date

backend/src/gpu.py
joeyballentine: not really, but i kinda think this is something we should start doing to avoid leaking imported things as exports. python's default implementation is "export everything" which in the past has caused us to use things in one file that we import via importing something else (which while that might be ok, I personally hate). Using this is like how nodejs does explicit exports. Without this, it would be exporting sanic's logger and the pynvml package

src/renderer/splash.tsx
joeyballentine: forgot to remove it

backend/src/gpu.py
RunDevelopment: alright. It's good as is then.

backend/src/api.py
RunDevelopment: Is the version pinning we do not good enough for that?

backend/src/dependencies/store.py
```diff
@@ -0,0 +1,47 @@
+import subprocess
+import sys
+from typing import List, TypedDict, Union
+
+python_path = sys.executable
+
+installed_packages = {}
```
RunDevelopment, line 7: Please add a type.

backend/src/api.py
joeyballentine: It's an optional dependency, so a user has to manually initiate an update for it to install it at all.

src/renderer/main.tsx
```diff
@@ -143,12 +146,21 @@ export const Main = memo(({ port }: MainProps) => {
                 return prev;
             });
         }
+    }, [response, data, loading, error, backendReady, sendAlert, t]);
 
+    useIpcRendererListener('backend-started', () => {
+        // We know the backend is good at this point, so we can set the retry count to a high number
+        setRetryCount(9999);
+    });
```
RunDevelopment, line 154: What exactly is this?

src/renderer/splash.tsx
```diff
@@ -18,7 +19,7 @@ const Splash = memo(() => {
     useEffect(() => {
         ipcRenderer.on('splash-setup-progress', (event, progress) => {
             setOverallProgress(progress.totalProgress);
-            setStatusProgress(progress.statusProgress > 0 ? progress.statusProgress : null);
+            setStatusProgress(progress.statusProgress > 0 ? progress.statusProgress / 2 : null);
```
RunDevelopment, line 22: Why?

backend/src/api.py
joeyballentine: but with this auto update flag, if the dependency is installed, it will then from that point on auto update if there is an update to perform, without the user needing to manually trigger it. this is to prevent ncnn from becoming outdated

src/renderer/splash.tsx
joeyballentine: oops, i did that before learning about the progress slices

src/renderer/main.tsx
joeyballentine: hmmm i had originally put this here to avoid it erroring when /nodes wasnt available, but I don't think its needed anymore, since /nodes should be available at the same time the backend starts

src/renderer/splash.tsx
```diff
@@ -5,6 +5,7 @@ import './i18n';
 import { useTranslation } from 'react-i18next';
 import { ipcRenderer } from '../common/safeIpc';
 import { ChaiNNerLogo } from './components/chaiNNerLogo';
+
```
RunDevelopment, line 8:
```suggestion
```

src/renderer/main.tsx
```diff
@@ -29,6 +29,7 @@ import { ExecutionProvider } from './contexts/ExecutionContext';
 import { GlobalProvider } from './contexts/GlobalNodeState';
 import { SettingsProvider } from './contexts/SettingsContext';
 import { useAsyncEffect } from './hooks/useAsyncEffect';
+
```
RunDevelopment, line 32:
```suggestion
```

RunDevelopment (approved):

PR #1839, opened 2023-05-31
Fix backend script path detection with external Electron

JeremyRand:
If an external Electron binary in `~/electron/` is being used to run a packaged chaiNNer zip (via the `ELECTRON_FORCE_IS_PACKAGED=1` environment variable), chaiNNer was calculating the Python script as `~/electron/resources/src/run.py`, which results in "No such file or directory".

This PR uses `app.getAppPath()` instead of `process.resourcesPath`; `app.getAppPath()` behaves correctly regardless of whether an external Electron binary is being used.

### DIFF
--- a/src/main/backend/process.ts
+++ b/src/main/backend/process.ts
@@ -1,14 +1,29 @@
 import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 import { app } from 'electron';
+import { existsSync } from 'fs';
 import path from 'path';
 import { getBackend } from '../../common/Backend';
 import { PythonInfo } from '../../common/common-types';
 import { sanitizedEnv } from '../../common/env';
 import { log } from '../../common/log';
+import { lazy } from '../../common/util';
 
-const backendPath = app.isPackaged
-    ? path.join(process.resourcesPath, 'src', 'run.py')
-    : './backend/src/run.py';
+const getBackendPath = lazy((): string => {
+    const candidates: string[] = [
+        path.join(process.resourcesPath, 'src', 'run.py'),
+        path.join(app.getAppPath(), '..', 'src', 'run.py'),
+        './backend/src/run.py',
+    ];
+
+    for (const candidate of candidates) {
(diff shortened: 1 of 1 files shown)


JeremyRand:
Combined with https://github.com/chaiNNer-org/chaiNNer/pull/1830 , I *think* this should be sufficient to get chaiNNer running on ppc64le. Note that I only tested on Linux (I don't have easy access to any Windows or macOS machines to test this on). I'd be surprised if this PR breaks anything on Windows or macOS, but you may want to test that prior to merging.

RunDevelopment:
Q: Instead of trying to perfectly determine the path of `run.py` and hoping that we don't break anything, couldn't we just define a list of possible paths and then use the first one that exists? This would guarantee that we don't break existing anything that is currently working. 

I was thinking about replacing the `backendPath` variable with a cached function like this:
```ts
const getBackendPath = lazy((): string => {
    const candidates: string[] = app.isPackaged
        ? [
              path.join(process.resourcesPath, 'src', 'run.py'),
              path.join(app.getAppPath(), '..', 'src', 'run.py'),
          ]
        : ['./backend/src/run.py'];

    for (const candidate of candidates) {
        if (fs.existsSync(candidate)) {
            return candidate;
        }
    }

    log.error('Unable to find backend path from the following candidates:', candidates);
    throw new Error('Unable to find backend path');
});
```

joeyballentine:
That sounds good to me

JeremyRand:
@RunDevelopment That sounds reasonable. Any reason to keep the special-case logic for `./backend/src/run.py`, or should we simply add it to the list of paths to try?

RunDevelopment:
You're right. That could be just one array. Then let's make it one list.

RunDevelopment (approved):

joeyballentine:
@RunDevelopment shouldn't the dev path be first so it overwrites the resource paths? Otherwise run dev will just use the one in appdata

joeyballentine (approved):

PR #1840, opened 2023-06-03
Add a wcmatch.glob field to image iterator

zeptofine:
This makes it easier to filter out images based on their filename, without having to separate the files manually. Nothing changes if no regex is given
I also added an exception during the iterator process to disambiguize the "division by zero" error when the list is empty.
Problems:
- I've heard Regex Injection is a thing and I don't know if it's a concern, or how to handle it
- I am conflicted whether to use re.match or re.fullmatch


zeptofine:
I've found a project that might help with the safety stuff. Not sure how it could be implemented though https://www.npmjs.com/package/safe-regex
I also heard about google-RE2 while making this but it doesn't seem to allow the same features so Idk if it's a good alternative

jackcarey:
Did you consider using the `glob` package for extracting specific images from a folder? It might not be as capable as regular expressions, but, I think it could do a good job and you can avoid RE safety issues.

https://docs.python.org/3/library/glob.html

zeptofine:
I thought about that. It also has the upside of being easier to understand lol. I will ask the person on Discord who proposed it
https://docs.python.org/3/library/fnmatch.html#module-fnmatch
Sounds like the more user friendly option

zeptofine:
That's awesome
image
extension checking might have to be separated from the glob expression lol

zeptofine:
I migrated to [wcmatch.glob](https://facelessuser.github.io/wcmatch/glob/). I posted a few examples on some expressions on Discord but here they are for those not in there:
- the same as how it is now: `{*,**/*}` this will be the default
- recursive & inclusive search for images with extensions png, and jpg: `{*,**/*}.{png,jpg}`
- non recursive: `*.{png,jpg}`
- recursive & inclusive images not including exr or png: `{*,**/*}.!(exr|png)`
- files with a string inside it: `{*,**/*}string1*`
- same thing but multiple strings `{*,**/*}{string1,string2}*`
POSIX character classes are also available.

zeptofine:
ahh... idk if I should edit the [src/common/dependencies.ts requiredDependencies for this](https://github.com/zeptofine/chaiNNer/blob/98d67d62ae9757502b182db1f3c9a26347ec3e28/src/common/dependencies.ts#L160)

jackcarey:
Yep, you will have to edit the `requiredDependencies.ts` file to include `wcmatch`. Also, it'd help to rename the input so it says `Glob Expression (WCMatch)` or something similar, so users who know glob know it isn't standard.

```
{
        name: 'WCMatch',
        packages: [{ packageName: 'wcmatch', version: '8.4.1', sizeEstimate: 40 * KB }],
    },
    ```

RunDevelopment:
Thank you for the PR @zeptofine, and thank you for the help @jackcarey!

@joeyballentine is currently refactoring how we handle required dependencies, so we might want to wait with adding glob to `requiredDependencies.ts` until #1834 is merged. Alternatively, we can also just let Joey deal with the small conflict will create.

---

I know that we moved away from regexes, but I wanted to answer the regex-related questions just in case it comes up again:

> https://www.npmjs.com/package/safe-regex

Please don't use this. The analysis done by safe-regex is just wrong. It cannot detect even the most simple cases of exponential backtracking and produces numerous false positives.

I've talked with the maintainer of safe-regex and a few security researches before, and we all agree that safe-regex has zero value when it comes to helping security.

> I also heard about google-RE2 while making this but it doesn't seem to allow the same features so Idk if it's a good alternative

re2 is great. We actually use its python bindings already. It can't do backreferences or lookaheads, but it's guaranteed to be safe. 

In our case, we used python `re` before, and run into a regex with O(n^5) runtime, so we needed re2 to get back to that sweet sweet O(n).

joeyballentine:
@zeptofine My backend dependency PR has been merged, if you'd like to proceed with this.

zeptofine:
Nice, I'll add the dependency using the new system
My main problem is that I don't know if this is important enough to validate adding a new required dependency to the program

RunDevelopment:
As I see it, this PR is basically finished. But we still need advanced options to hide the glob expression by default.

zeptofine:
Could that be accomplished here by simply adding a "Use glob expression" checkbox connected to an input group? When you pointed me towards them in Discord I looked into them and I think it may fit better here.
I made a branch [here](https://github.com/zeptofine/chaiNNer/tree/iterator-regex-groups) to test that
image
I still understand why that may not be enough though

joeyballentine:
That actually looks perfect IMO

zeptofine:
> That actually looks perfect IMO

alright, should I merge the checkbox branch here?

RunDevelopment (approved):

joeyballentine:
Thanks again for this!

PR #1841, opened 2023-06-04
Added CAS to High Boost Filter

RunDevelopment:
This adds support contrast adaptive sharpening to High Boost Filter. CAS is an addition to the classical high boost filter. Its main effect is that it prevents oversharpening artifacts in regions of high contrast. This causes the sharpening to be applied more evenly.

image

Example:

| Original | High boost | High boost + CAS |
| --- | --- | --- |
| image  | image | image |
| image | image | image |

CAS is normally implemented by adjusting the per-pixel sharpen kernel. This is not an option for us, because we how to implement this in python and not shader code. So I implemented CAS as a linear blend mask between the original and sharpened image. This seems to yield a good-enough effect, as can be seen in the examples above.

### DIFF
--- a/backend/src/nodes/impl/cas.py
+++ b/backend/src/nodes/impl/cas.py
@@ -15,7 +15,7 @@ def _luminance(img: np.ndarray) -> np.ndarray:
     return np.dot(img[..., :3], [0.2126, 0.7152, 0.0722])
 
 
-def create_cas_mask(img: np.ndarray, kernel) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


RunDevelopment:
I also added a bias slider now. Higher values will bias CAS towards sharpening more. In the previous examples, I used a bias of 1. The original CAS work uses a bias of 2, so I made this the default here too.

image

joeyballentine (approved):

PR #1849, opened 2023-06-09
Add `--devtools` argument

JeremyRand:
Currently, chaiNNer assumes that Electron's DevTools should be automatically opened on launch if and only if an external Electron binary is in use. While this assumption holds quite often (an external Electron binary is used with `npm run start`), it is not always correct. For example, users might want to use a distro-packaged Electron binary on platforms (e.g. Linux/ppc64le) where chaiNNer does not distribute binaries.

This commit replaces the `app.isPackaged` condition with an explicit `--devtools` command-line flag, and tweaks `package.json` to provide this flag in the situations where an external Electron binary would normally have been used (e.g. `npm run start`), thus fixing the issue.

As a side benefit, users of official chaiNNer binaries who, for whatever reason, want to automatically launch Electron's DevTools, can now do so by passing the new command-line flag.


JeremyRand:
Good catch by the linter, will fix the warning ASAP.

JeremyRand:
There we go, CI is passing now.

JeremyRand:
> As a side benefit, users of official chaiNNer binaries who, for whatever reason, want to automatically launch Electron's DevTools, can now do so by passing the new command-line flag.

On further testing, it looks like the flag only works when an external Electron binary is used; the DevTools pane doesn't appear when using chaiNNer's bundled Electron binary regardless of the flag. I think this PR is still clearly an improvement (e.g. people can download a chaiNNer binary zip, and use it with an external Electron binary to get the pane without needing to clone Git), and I'm not sure how best to deal with that edge case beyond that. Maybe just add a note to the help text for that flag so that affected users know why it doesn't work?

RunDevelopment:
Thanks for the PR @JeremyRand!

Code looks good, but I still have to test it.

> the DevTools pane doesn't appear when using chaiNNer's bundled Electron binary regardless of the flag

Interesting. Is DevTools completely removed/non-functional or does it just fail to open? In any case, we should add this fact to the description of the devtools flag.

JeremyRand:
> Interesting. Is DevTools completely removed/non-functional or does it just fail to open? In any case, we should add this fact to the description of the devtools flag.

Hmm, now I can't reproduce that issue anymore; DevTools works fine for me now with chaiNNer's bundled Electron binary. Probably I did something wrong earlier, might have been an artifact of running it in a VM without enough RAM. Carry on, I guess.

PR #1852, opened 2023-06-10
Update forge config to remove pyc files

joeyballentine:
This won't affect CI builds at all, but for local builds this avoids having the pycache files included, which especially bloat the zip builds and make extracting take much longer.


RunDevelopment:
I was initially going to suggest using include/exclude glob patterns to exclude these files, but Electron forge config docs are such a clusterfuck that I don't want to waste both of our time.

RunDevelopment (approved):

PR #1855, opened 2023-06-11
Use async/await for .pyc removal forge hook

joeyballentine:
When i submitted my last PR, I "cleaned it up" by removing some callbacks. I assumed it was fine because I didn't get any linting warnings. What I didn't realize though, is that this isn't a typescript file, so that actually broke the function and didn't tell me.


RunDevelopment:
I should have noticed that when reviewing it: This doesn't work. All of these fs function are "async" with callbacks. So whether forge actually deletes these files before packing solely depends on how fast your machine can remove them. Too slow and they'll make it in.

I think the best way to solve this is to make `deletePycFiles` async and to only use the promise versions of the fs functions.

joeyballentine:
i merged this without approval thinking that it was necessary for the nightly to build, but apparently it wasn't

PR #1856, opened 2023-06-11
Use tuples instead of semver package

joeyballentine:


backend/src/dependencies/store.py
```diff
@@ -39,6 +40,33 @@ def install_dependencies(dependency_info_array: List[DependencyInfo]):
         installed_packages[package_name] = version
 
 
+def coerce_semver(version: str) -> Tuple[int, int, int]:
+    regex = r"(\d+)\.(\d+)\.(\d+)"
```
RunDevelopment, line 44: The .minor and .patch numbers need to be optional. E.g. pywin32 has version "303".

```suggestion
    regex = r"(\d+)(?:\.(\d+)(?:\.(\d+))?)?"
```

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #1858, opened 2023-06-12
New min dependencies CI test

RunDevelopment:
This PR changes the min deps test to include starting up the server to actually install dependencies. This means that it now simulates the bootstrap process of a fresh install.

All of this relies on the new `--close-after-start` flag. If set, it will cause the server to start up and immediately close after being fully started.


RunDevelopment:
Alright, we can see by 9793f30fc2ca2afd639047164784de14551ef85a, the CI will fail if we mess up the bootstrap process in any way. So we can now change it more confidently.

joeyballentine (approved):

PR #1859, opened 2023-06-12
Use util `get_h_w_c`

RunDevelopment:
This removes the indirect OpenCV import from `process.py`.


RunDevelopment:
Found 2 more.

joeyballentine (approved):

PR #1860, opened 2023-06-12
Removed old unused Category class

RunDevelopment:
Seems like we forgot to remove it at some point. 

We now have exactly 0 backend test.


RunDevelopment:
Uhhhmm, pytest doesn't like no tests.

https://docs.pytest.org/en/7.1.x/reference/exit-codes.html#:~:text=Exit%20code%205,No%20tests%20were%20collected&text=If%20you%20would%20like%20to,using%20the%20pytest%2Dcustom_exit_code%20plugin.

joeyballentine (approved):

PR #1864, opened 2023-06-12
Add weight loading for NCNN Scale op and fix InnerProduct bug

theflyingzamboni:


RunDevelopment (approved):
I'm assuming this is correct and works perfectly.

PR #1865, opened 2023-06-14
Use `chainner_ext` for Fill Alpha

RunDevelopment:
I already shared the improved performance on Discord, but I changed the settings a little since. The Rust implementation of `fill_alpha_extend_color` is around 100x faster than the Python implementation (not counting the overhead of copying a numpy image to Rust), so I increased the number of iterations to fill the whole image. The performance cost for this is relatively small even for large images. Even with the new settings, Extend Texture is around 5x faster and Extend Color is around 15x faster.

Obviously, this PR also adds [`chainner_ext`](https://pypi.org/project/chainner-ext/) as a dependency. I only published v0.0.0 (a test version) right now. Before we include `chainner_ext` in the next release, we have to make sure that it works correctly for our users. I published pre-built wheels for all major architectures and python versions (3.8-3.11), but I haven't tested anything except cp3.8 Win x64 and cp3.9 Win x64 yet.


RunDevelopment:
See #1866 regarding the failing CI.

joeyballentine (approved):

PR #1867, opened 2023-06-15
Support NCNN CPU inference

JeremyRand:
This is a useful option on systems with strong CPU and weak GPU, where PyTorch is nontrivial to install (e.g. ppc64le systems).

Currently CPU inference is keyed on which NCNN package is installed; a future PR could allow CPU inference even when the Vulkan NCNN package is installed.

### DIFF
--- a/backend/src/nodes/impl/ncnn/auto_split.py
+++ b/backend/src/nodes/impl/ncnn/auto_split.py
@@ -3,7 +3,15 @@ from __future__ import annotations
 import gc
 
 import numpy as np
-from ncnn_vulkan import ncnn
+
+try:
+    from ncnn_vulkan import ncnn
+
+    use_gpu = True
+except ImportError:
+    from ncnn import ncnn
+
+    use_gpu = False
 from sanic.log import logger
 
 from ...utils.utils import get_h_w_c
@@ -22,9 +30,10 @@ def ncnn_auto_split(
 ) -> np.ndarray:
     def upscale(img: np.ndarray, _):
         ex = net.create_extractor()
-        ex.set_blob_vkallocator(blob_vkallocator)
-        ex.set_workspace_vkallocator(blob_vkallocator)
-        ex.set_staging_vkallocator(staging_vkallocator)
+        if use_gpu:
+            ex.set_blob_vkallocator(blob_vkallocator)
+            ex.set_workspace_vkallocator(blob_vkallocator)
+            ex.set_staging_vkallocator(staging_vkallocator)
         # ex.set_light_mode(True)
         try:
             lr_c = get_h_w_c(img)[2]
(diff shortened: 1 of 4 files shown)


JeremyRand:
(Will fix the lint errors ASAP)

JeremyRand:
Linters are passing now; sorry that took so many force-pushes.

joeyballentine:
Odd you test this? There's no way this should work. Ncnn_vulkan is my fork of NCNN that has multiple changes we rely on. Model loading shouldn't even work properly using the regular package

JeremyRand:
I upscaled a few images and it worked fine here AFAICT. Also NCNN was a lot faster than PyTorch on POWER9 once I fixed some POWER9-specific optimization flags in the NCNN build scripts; upscaling my test image on POWER9 took 1:05 with NCNN compared to 1:32 with PyTorch (that's including Electron startup/shutdown overhead, so the actual inference speed difference is a bit larger). (Planning to submit those optimization patches to upstream NCNN soon.) I have no idea how NCNN compares to PyTorch on x86 in terms of CPU inference speed.

joeyballentine:
Huh, well in that case it sounds like it's fine. I wanna test some stuff myself first though

JeremyRand:
Worth noting that NCNN's CPU inference is supposedly very well-optimized for ARM64, so if someone has an ARM64 machine that they could test this on (and compare to PyTorch speed), that might be interesting.

JeremyRand:
Yes, I was intending to submit such a feature as a follow-up PR. There are a few reasons why some users might prefer to use upstream NCNN:

* Easier to build from source since there are fewer dependencies.
* Updated more frequently.
* More eyes on the code.

In particular, I couldn't get ncnn_vulkan to build on ppc64le, and while that is probably fixable, I don't think it's desirable to force users to install ncnn_vulkan fork if they don't intend to use Vulkan. The chaiNNer changes needed to support upstream NCNN are tiny, as shown in this PR. But yes, I am supportive of supporting CPU inference with ncnn_vulkan as well.

theflyingzamboni:
The lack of load_param_mem and load_model_mem and associated hackiness with the temporary files isn't ideal. I don't know why they left those out of the pybinds in the official repo.

JeremyRand:
Agreed that the tempfile workaround is a bit annoying, and that it would be desirable for upstream NCNN to add the missing functions. That said, the workaround is only a few lines of code, and isn't particularly bad for readability, so I think it's an acceptable cost of supporting upstream.

joeyballentine:
We could submit a PR with those pybinds added and wait for an official update from that

JeremyRand:
Is there a particular commit from ncnn_vulkan containing those pybinds that I could cherry-pick onto upstream master and submit to them?

joeyballentine:
@theflyingzamboni did that part so it's up to him on whether he wants to do that or if he'd rather you do it -- i can check later what commit it was

theflyingzamboni:
I'll take a look soon.

joeyballentine (approved):

joeyballentine:
Even though this isn't an ideal implementation, it'll take way too long to get the stuff we'd need added to the official ncnn, so i think we're fine just using this for now. It won't affect regular users.

PR #1869, opened 2023-06-16
Move most dependencies from install_core_deps to chaiNNer_standard Dependency definition

joeyballentine:
these feel like they make more sense here.

### DIFF
--- a/backend/src/dependencies/install_core_deps.py
+++ b/backend/src/dependencies/install_core_deps.py
@@ -1,46 +1,8 @@
 from typing import List
 
-from system import is_arm_mac, is_windows
-
 from .store import DependencyInfo, install_dependencies
 
-# I'm leaving this here in case I can use the Dependency class in the future, so I don't lose the extra info
-
-# dependencies=[
-#         Dependency("OpenCV", "opencv-python", "4.7.0.68", 30 * MB, import_name="cv2"),
-#         Dependency("NumPy", "numpy", "1.23.2", 15 * MB),
-#         Dependency("Pillow (PIL)", "Pillow", "9.2.0", 3 * MB, import_name="PIL"),
-#         Dependency("appdirs", "appdirs", "1.4.4", 13.5 * KB),
-#         Dependency("FFMPEG", "ffmpeg-python", "0.2.0", 25 * KB, import_name="ffmpeg"),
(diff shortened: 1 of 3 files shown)


joeyballentine:
this apparently doesn't work? it works fine in npm start but with npm run dev I get this... 
image

backend/src/server.py
```diff
@@ -493,16 +505,15 @@ async def setup(sanic_app: Sanic):
     await setup_queue.put_and_wait(
         {
             "event": "backend-status",
```
RunDevelopment, line 507: Could we use `update_progress` for these "backend-status" events too?

backend/src/server.py
```diff
@@ -419,7 +419,8 @@ def install_deps(dependencies: List[api.Dependency]):
     for package in api.registry.packages.values():
         logger.info(f"Checking dependencies for {package.name}...")
         if package.name == "chaiNNer_standard":
-            continue
+            await update_progress_cb("Installing core dependencies...", 0.65)
+            install_deps(package.dependencies)
```
RunDevelopment, line 423: Why not push them into `to_install`?

backend/src/packages/chaiNNer_standard/__init__.py
```diff
@@ -1,14 +1,30 @@
...
+from api import KB, MB, Dependency, add_package
+from system import is_arm_mac, is_windows
 
 package = add_package(
     __file__,
     name="chaiNNer_standard",
     description="The standard set of nodes for chaiNNer.",
-    dependencies=[],
+    dependencies=[
+        Dependency("OpenCV", "opencv-python", "4.7.0.68", 30 * MB, import_name="cv2"),
```
RunDevelopment, line 11: Could you please use kw args here like we do in other packages?

backend/src/dependencies/install_core_deps.py
```diff
@@ -49,23 +11,10 @@
...
-    {
-        "package_name": "opencv-python",
-        "version": "4.7.0.68",
-    },
-    {
-        "package_name": "Pillow",
-        "version": "9.2.0",
-    },
     {
         "package_name": "chainner_ext",
```
RunDevelopment, line 15: I think you can move this over too.

joeyballentine:
Assuming this is good now, going to merge so I can open my next PR

Issue #1875, opened 2023-06-19
Support remote backends


RunDevelopment:
**Motivation**
We have been getting a lot of requests to run chainner on a server, so this issue to intended to track progress towards that. 

The main goal of this is to support offloading the heavy computational cost of some operations (e.g. upscaling) to cloud servers such that low-end and mid-range devices can use the full range of chainner's features. Users would be able to use the frontend to connect to some remote backend and use chainner as normal while the remote server does the computations.

**Description**
Remote backends should support the following:

1. Running the backend without the GUI.
2. Supporting Docker.
3. Allowing the backend to connect to remote servers that host a backend.

As I see it, the main technical challenges will be the following:

1. File paths. Right now, the backend and frontend assume that they have access to the same file system. This obviously isn't the case with remote backends, so we have to send over files to share them between the backend and frontend. \
    This mainly affects the Load Image, Load Model, Save Image, Save Model, and View Image (External) nodes.
2. Authentication. With chainner being a visual programming language, we are essentially running untrusted code on those remote backends (not even accounting for third-party plugins). We need to be able to authenticate users such that only allowed users can connect to a backend.
3. Multiple tenants. There might be multiple users connected to a single backend at once. This is currently not supported at all. \
    It should be noted that this problem is equivalent to supporting multiple chains (#607).

Note: We may be able to get away with not supporting multiple tenants at first.

Splendide-Imaginarius:
It seems like one stepping stone on the way to this goal would be to change how the `--no-backend` flag works. Currently, it takes no arguments (it's just a boolean flag), and if specified, it assumes the backend is at `http://localhost:8000/`. It might be helpful to allow specifying the backend address as an argument, e.g. `chainner --remote-backend http://localhost:8000/` would be the equivalent of the current behavior of `chainner --no-backend`. (This would also be independently helpful for use cases where the user wants to run the backend on a different port or different loopback address, e.g. if the current default port is already in use.)

Would such a change be welcome?

joeyballentine:
Yes that would be a welcome addition for sure

PR #1877, opened 2023-06-19
Get list of available Nvidia GPUs from the backend

### LINKED ISSUE #1744: GPU selector sets wrong GPU
**Information:**

- Chainner version: 0.18.6
- OS: Windows 11

**Description**
The GPU selection under Settings doesn't match the GPU order in nvidia-smi. I have a 4090 and a 3090, if I select the 4090 and run pytorch it's actually using the 3090 and vice-versa.

joeyballentine:
Has potential to fix #1744 but mostly just something to get us closer to #1875

### DIFF
--- a/backend/src/gpu.py
+++ b/backend/src/gpu.py
@@ -27,7 +27,6 @@ class NvidiaHelper:
                     "name": nv.nvmlDeviceGetName(handle),
                     "uuid": nv.nvmlDeviceGetUUID(handle),
                     "index": i,
-                    "handle": handle,
                 }
             )
 
--- a/backend/src/gpu.py
+++ b/backend/src/gpu.py
@@ -13,7 +13,7 @@ except Exception as e:
     logger.info(f"Unknown error occurred when trying to initialize Nvidia GPU: {e}")
 
 
-class NvidiaHelper:
+class _NvidiaHelper:
     def __init__(self):
         nv.nvmlInit()
 
@@ -27,6 +27,7 @@ class NvidiaHelper:
                     "name": nv.nvmlDeviceGetName(handle),
                     "uuid": nv.nvmlDeviceGetUUID(handle),
                     "index": i,
+                    "handle": handle,
                 }
             )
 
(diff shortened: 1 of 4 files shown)


backend/src/server.py
```diff
@@ -365,6 +365,26 @@ async def list_ncnn_gpus(_request: Request):
...
+    """Lists the available GPUs for NCNN"""
+    await nodes_available()
+    try:
+        # pylint: disable=import-outside-toplevel
+        from gpu import NvidiaHelper, nvidia_is_available
+
+        if not nvidia_is_available:
+            return json([])
+
+        nv = NvidiaHelper()
```
RunDevelopment, line 379: Should we really create a new helper each time? Wouldn't it be better to just reuse a single instance? I'm just worried that the nvml `init` and `shutdown` functions might be expensive.

RunDevelopment:
Doesn't this mean that we can completely get rid of Nvidia SMI now?

joeyballentine:
We currently still use the nvidia-smi stuff for vram reporting, which I can move over to using pynvml in a separate pr

backend/src/server.py
joeyballentine: Yeah you're right

RunDevelopment:
Right, I didn't think of that.

backend/src/gpu.py
```diff
@@ -14,6 +14,11 @@
 
 
 class NvidiaHelper:
+    def __new__(cls):
+        if not hasattr(cls, "instance"):
+            cls.instance = super(NvidiaHelper, cls).__new__(cls)
+            return cls.instance
```
RunDevelopment, line 20: Man, this is why I hate python. Who in the language design team though that allowing users to change the very nature of class instantiation is a good idea?

Let's please not do this. I had to go on a 10min detour just to figure out wtf this is supposed to be. I don't blame you, because the guides I found really like this black magic, but I don't. 

Let's just have a little helper function `get_nvidia_helper() -> NvidiaHelper | None` that returns `None` when there are no Nvidia GPUs and a cached instance otherwise. This gets rid of the voodoo and also has the advantage that we can't forget to check `nvidia_is_available`.

joeyballentine: FYI this is just a singleton pattern for a class

RunDevelopment: [This](https://www.baeldung.com/java-singleton#singleton) is how you do a singleton is any other language. `__new__` is just voodoo.

backend/src/server.py
```diff
@@ -365,6 +365,29 @@ async def list_ncnn_gpus(_request: Request):
...
+        # pylint: disable=import-outside-toplevel
+        from gpu import get_nvidia_helper, nvidia_is_available
+
+        if not nvidia_is_available:
+            return json([])
+
+        nv = get_nvidia_helper()
+
+        if nv is None:
+            return json([])
```
RunDevelopment, line 382:
```suggestion
        from gpu import get_nvidia_helper

        nv = get_nvidia_helper()

        if nv is None:
            return json([])
```

RunDevelopment (approved):
Just one thing. Otherwise gtg

PR #1878, opened 2023-06-19
Bundle py3-none-any server and core deps

joeyballentine:
Not 100% sure we actually want to do this. It didn't really save any startup time for me


joeyballentine:
merging this for now. we can always undo it later if we decide we don't want it, but I can't bundle our custom pip without it

PR #1880, opened 2023-06-20
Node Documentation

joeyballentine:
Initial implementation of node documentation as discussed on discord. 

Current issues:
- ~~The pages themselves don't scroll on overflow~~
- ~~When using the context menu to open the docs for a specific node, the sidebar doesn't scroll into place like it should~~

Remaining questions:
- ~~Is there anything else we need to consider this complete for now?~~
- ~~Do we need to have that example of the node at the bottom?~~

image

image


joeyballentine:
Updated layout:
image

joeyballentine:
Update: 
image

Still stuff to do but wanted to share where it's at

joeyballentine:
It properly scrolls now. Also, I removed the footer -- it made it look odd not going down all the way, and it was kinda useless without any other button there.
image

src/renderer/components/NodeDocumentationModal.tsx
```diff
@@ -0,0 +1,492 @@
...
+                                            </Box>
+                                            <Box position="relative">
+                                                <Heading
+                                                    mb={1}
+                                                    size="sm"
+                                                    userSelect="text"
+                                                >
+                                                    Inputs
+                                                </Heading>
+                                                {selectedSchema.outputs.length > 0 ? (
```
RunDevelopment, line 325: So that's why some nodes were shown as having no inputs.
```suggestion
                                                {selectedSchema.inputs.length > 0 ? (
```

src/renderer/components/NodeDocumentationModal.tsx
```diff
@@ -0,0 +1,492 @@
...
+                                                                    <Text
+                                                                        fontWeight="bold"
+                                                                        userSelect="text"
+                                                                    >
+                                                                        {output.label}
+                                                                    </Text>
+                                                                    {output.description && (
+                                                                        <ReactMarkdown
+                                                                            components={ChakraUIRenderer(
+                                                                                customMarkdownTheme
```
RunDevelopment, line 402: We are 20 indentation levels deep here. Maybe we could extract the input/output lists into their own components?

src/renderer/components/NodeDocumentationModal.tsx
joeyballentine: LOL

joeyballentine:
Added the ability to link to other pages.

image

joeyballentine:
I also made iterators show their helper nodes in the docs as well

image

RunDevelopment:
1. Color contrast for links is too low.
    image
1. You implemented links in a quite simple way, but I don't like that changing the title of a node now means that we have to go through all docs and update it there as well. Ideally, we would just need the schema ID to link to another node.

joeyballentine:
> but I don't like that changing the title of a node now means that we have to go through all docs and update it there as well. Ideally, we would just need the schema ID to link to another node.

I'll just remove the fallback text and implement the same automatic name system for the tooltip descriptions as well

joeyballentine:
Merging this since on discord you said it was good besides the two things, which I have now fixed

PR #1884, opened 2023-06-22
Add "Get Bounding Box" node

zeptofine:
This essentially does the same thing as the 'crop content' node but instead returns the dimensions needed to do such an operation.
box
The threshold represents the minimum value a pixel must be to be considered part of the mask.
This example uses the data to simply crop the original image, like 'crop content': simple-preview
Having these as separate node operations give access to more complex node setups, like this version: 
full-preview
This crops the image, superscales the specific section then moves it back into place.

Edit: changed the complex example to use the mask as transparency.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
@@ -31,8 +31,8 @@ from .. import utility_group
     outputs=[
         NumberOutput("X", output_type="min(uint, Input0.width - 1)"),
         NumberOutput("Y", output_type="min(uint, Input0.height - 1)"),
-        NumberOutput("Width", output_type="min(uint, Input0.width)"),
-        NumberOutput("Height", output_type="min(uint, Input0.height)"),
+        NumberOutput("Width", output_type="min(uint, Input0.width) & 1.."),
+        NumberOutput("Height", output_type="min(uint, Input0.height) & 1.."),
     ],
 )
 def get_bounding_box_node(
@@ -45,14 +45,9 @@ def get_bounding_box_node(
     h, w, _ = get_h_w_c(img)
 
     r = np.any(img > thresh, 1)
-    if r.any():
-        c = np.any(img > thresh, 0)
-        x, y = c.argmax(), r.argmax()
-        return (
-            int(x),
-            int(y),
-            int(w - x - c[::-1].argmax()),
-            int(h - y - r[::-1].argmax()),
-        )
-    else:
-        return 0, 0, w, h
+    c = np.any(img > thresh, 0)
+    if not (r.any() and c.any()):
+        raise RuntimeError("Resulting bounding box is empty.")
+
+    x, y = c.argmax(), r.argmax()
+    return int(x), int(y), int(w - x - c[::-1].argmax()), int(h - y - r[::-1].argmax())
--- a/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
@@ -14,7 +14,7 @@ from .. import utility_group
 @utility_group.register(
     schema_id="chainner:image:get_bbox",
     name="Get Bounding Box",
-    description="Get the X, Y, Height, and Width of a bounding box in the image in px.",
+    description="Gets a bounding box (X, Y, Height, and Width) of the white area of a mask.",
     icon="BsRulers",
     inputs=[
         ImageInput(channels=1),
--- a/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
@@ -46,7 +46,7 @@ def get_bounding_box_node(
 
     r = np.any(img > thresh, 1)
     c = np.any(img > thresh, 0)
-    if not (r.any() and c.any()):
+    if not r.any():
         raise RuntimeError("Resulting bounding box is empty.")
 
     x, y = c.argmax(), r.argmax()
--- a/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
@@ -15,7 +15,7 @@ from .. import utility_group
     schema_id="chainner:image:get_bbox",
     name="Get Bounding Box",
     description="Gets a bounding box (X, Y, Height, and Width) of the white area of a mask.",
-    icon="BsRulers",
+    icon="BsBoundingBox",
     inputs=[
         ImageInput(channels=1),
         SliderInput(
--- a/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from typing import Tuple
+
+import cv2
+import numpy as np
+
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
```diff
@@ -0,0 +1,57 @@
...
+    schema_id="chainner:image:get_bbox",
+    name="Get Bounding Box",
+    description="Gets a bounding box (X, Y, Height, and Width) of the white area of a mask.",
+    icon="BsRulers",
+    inputs=[
+        ImageInput(channels=1),
+        SliderInput(
+            "Threshold",
+            precision=1,
+            minimum=1,
```
RunDevelopment, line 24: Please specify the maximum. I don't have the default values for all inputs in my head all the time.

```suggestion
            minimum=1,
            maximum=100,
```

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
```diff
@@ -0,0 +1,57 @@
...
+from nodes.utils.utils import get_h_w_c
+
+from .. import utility_group
+
+
+@utility_group.register(
+    schema_id="chainner:image:get_bbox",
+    name="Get Bounding Box",
+    description="Gets a bounding box (X, Y, Height, and Width) of the white area of a mask.",
+    icon="BsRulers",
```
RunDevelopment, line 18: Ruler? Maybe there's the better icon for it... Not sure what though.

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
```diff
@@ -0,0 +1,57 @@
...
+        ),
+    ],
+    outputs=[
+        NumberOutput("X"),
+        NumberOutput("Y"),
+        NumberOutput("Width"),
+        NumberOutput("Height"),
+    ],
+)
+def get_dimensions_node(
```
RunDevelopment, line 37: It's still the old name.
```suggestion
def get_bounding_box_node(
```

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
```diff
@@ -0,0 +1,57 @@
...
+    icon="BsRulers",
+    inputs=[
+        ImageInput(channels=1),
+        SliderInput(
+            "Threshold",
+            precision=1,
+            minimum=1,
+            controls_step=1,
+            slider_step=1,
+            default=1,
```
RunDevelopment, line 27: Is that a reasonable default?

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
RunDevelopment, line 24: Why is the minimum 1? 

Threshold=0 would mean that the bounding box is the entire image, right? If so, then it should be allowed. Having neutral values can be very useful. E.g. the Math node doesn't forbid addition with 0 just because it doesn't do anything.

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
```diff
@@ -0,0 +1,57 @@
...
+        ),
+    ],
+    outputs=[
+        NumberOutput("X"),
+        NumberOutput("Y"),
+        NumberOutput("Width"),
+        NumberOutput("Height"),
```
RunDevelopment, line 34: I think we should add output types to these. We might not know their exact values, but we know (1) their range and (2) that they will be ints. The type system can then use this information to detect invalid connections between nodes.

About the range, while the range for X and Y is clear (between 0 and image.width-1, and 0 and image.height-1 respectively), I don't know about Width and Height. Can they be 0? Or a more direct question: what happens when no pixels pass the threshold?

```suggestion
        NumberOutput("X", output_type="min(uint, Input0.width - 1)"),
        NumberOutput("Y", output_type="min(uint, Input0.height - 1)"),
        NumberOutput("Width", output_type="min(uint, Input0.width)"),
        NumberOutput("Height", output_type="min(uint, Input0.height)"),
```

RunDevelopment (commented):
Good idea for a node!

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: Ah yeah, sorry I forgot xd

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: Ah, I was confused about how the output_types worked. Thanks.
When no pixels pass the threshold, it returns 0 for the x and y and the dimensions of the image. I thought it was better than returning a box with no size, so maybe the output_type should go from 0 to the actual width and height?

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: That was a remnant of when I was tinkering with cv2's actual system for thresholds. I'll change that to 0.

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: This was also a remnant of the cv2.threshold idea. I'm not entirely sure, but with the way I tested it using the example, there was no visible outline that occurred in higher thresholds when I used 1 so I kept it.

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: ah wait no that makes sense I'm dumb

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
zeptofine: Where do the icons come from? I may be able to find a better one.

zeptofine: Nevermind, I found the React Icons. I'll try and find one

zeptofine: image
How does this look?

RunDevelopment: Sorry for not telling you sooner. We use React Icons.

Yes, that icon looks good!

backend/src/packages/chaiNNer_standard/image_dimension/utility/get_bounding_box.py
RunDevelopment: > When no pixels pass the threshold, it returns 0 for the x and y and the dimensions of the image

This doesn't really make sense. Conceptually, we should return an empty bounding box. Unfortunately, the empty bounding box doesn't have a unique representation (any BB with width=0 or height=0 is empty), so it's not clear what the values should be.

We could also go the route Crop Content went. It just throws an error if the BB is empty. I think this would make sense here too. Users probably want to use the BB to crop an image, but width=0 and height=0 aren't valid inputs for our Crop nodes.

If we go with throwing an error, then the output type of Width and Height has to be changed to reflect the fact that they won't ever be 0. Use `min(uint, Input0.width) & 1..` for Width and the same with `.height` for Height.

RunDevelopment (approved):
I finally tested it, and it works well. Thank you @zeptofine!

PR #1885, opened 2023-06-23
Pip download progress reporting

joeyballentine:
This PR adds the ability for us to report on native pip downloads, via a build of my pip fork. This can be replaced in the future when the change gets merged into pip and released.

The bundled pip is slightly modified from my fork, in that I renamed every instance of "pip" to "chainner_pip" so it wouldn't overwrite the normal pip.

I tested it, and it works pretty well.


backend/src/dependencies/store.py
```diff
@@ -65,12 +66,12 @@ def get_deps_to_install(dependencies: List[DependencyInfo]):
 
 def install_dependencies_sync(
     dependencies: List[DependencyInfo],
-):
+) -> int:
     dependencies_to_install = get_deps_to_install(dependencies)
     if len(dependencies_to_install) == 0:
-        return
+        return -1
```
RunDevelopment, line 72: Why would "there are no deps that need to be installed" be an error?

src/common/Backend.ts
```diff
@@ -267,6 +267,7 @@ export interface BackendEventMap {
     'backend-status': {
         message: string;
         percent: number;
+        statusPercent?: number;
```
RunDevelopment, line 270: I just noticed this. Neither of those numbers are percentages (0-100). They are both 0-1. Maybe we should just rename them to `totalProgress` and `statusProgress` to be consistent with my `Progress` interface (TS).

Also, the type of `statusPercent` is wrong. It can be null, so it has to be `statusPercent?: number | null`.

RunDevelopment (commented):
What exactly is the status code even for? It's not like it's checked anywhere. We currently just log it, but that seems pretty useless. If an error occurs, then the status code likely won't tell you much.

backend/src/dependencies/store.py
joeyballentine: Sorry i should have taken this out when i was done debugging it

backend/src/dependencies/store.py
```diff
@@ -158,20 +167,40 @@ def get_progress_amount():
...
-    await update_progress_cb(f"Finished installing dependencies...", 1)
+    await update_progress_cb(f"Finished installing dependencies...", 1, None)
 
     for dep_info in dependencies_to_install:
         installing_name = dep_info["package_name"]
         version = dep_info["version"]
         installed_packages[installing_name] = version
 
+    return
+
```
RunDevelopment, line 203:
```suggestion

```

backend/src/server.py
```diff
@@ -427,7 +427,8 @@ async def install_deps(dependencies: List[api.Dependency]):
                 }
                 for dep in dependencies
             ]
-            await install_dependencies(dep_info, update_progress_cb)
+            status = await install_dependencies(dep_info, update_progress_cb, logger)
+            logger.info(f"Install status: {status}")
```
RunDevelopment, line 431: This returns None now.

backend/src/dependencies/store.py
```diff
@@ -86,10 +88,13 @@ def install_dependencies_sync(
         version = dep_info["version"]
         installed_packages[package_name] = version
 
+    return status
```
RunDevelopment, line 91: This still isn't used.

src/common/Backend.ts
```diff
@@ -266,7 +266,8 @@ export interface BackendEventMap {
     };
     'backend-status': {
         message: string;
-        percent: number;
+        progress: number;
+        statusProgress?: number;
```
RunDevelopment, line 270: This type is still wrong.
```suggestion
        statusProgress?: number | null;
```

src/main/gui/main-window.ts
```diff
@@ -360,10 +360,15 @@ export const createMainWindow = async (args: OpenArguments) => {
         const backendStatusProgressSlice = SubProgress.slice(progressController, 0.5, 0.95);
         sse.addEventListener('backend-status', (e: MessageEvent<string>) => {
             if (e.data) {
-                const data = JSON.parse(e.data) as { message: string; percent: number };
+                const data = JSON.parse(e.data) as {
+                    message: string;
+                    progress: number;
+                    statusProgress?: number;
+                };
```
RunDevelopment, line 367: I like my code [super dry](https://www.youtube.com/watch?v=fZC0Utqi-Ro).
```suggestion
                const data = JSON.parse(e.data) as BackendEventMap['backend-status'];
```

backend/src/dependencies/store.py
```diff
@@ -70,7 +72,7 @@ def install_dependencies_sync(
     if len(dependencies_to_install) == 0:
         return
 
-    subprocess.check_call(
+    status = subprocess.check_call(
```
joeyballentine, line 75:
```suggestion
    subprocess.check_call(
```

backend/src/dependencies/store.py
joeyballentine, line 91:
```suggestion
```

backend/src/dependencies/store.py
```diff
@@ -86,10 +88,13 @@ def install_dependencies_sync(
         version = dep_info["version"]
         installed_packages[package_name] = version
 
+    return status
+
```
joeyballentine, line 92:
```suggestion
```

backend/src/server.py
```diff
@@ -427,7 +427,8 @@ async def install_deps(dependencies: List[api.Dependency]):
                 }
                 for dep in dependencies
             ]
-            await install_dependencies(dep_info, update_progress_cb)
+            status = await install_dependencies(dep_info, update_progress_cb, logger)
```
joeyballentine, line 430:
```suggestion
            await install_dependencies(dep_info, update_progress_cb, logger)
```

backend/src/server.py
joeyballentine, line 431:
```suggestion
```

joeyballentine:
Assuming this is good. Merging so I can work on the optional dep installation with progress reporting

PR #1890, opened 2023-06-23
Add fast NL means denoiser node

adegerard:
This PR adds a Non-local Means Denoising algorithm.

Note: the "Color strength" input is only used when the input image is RGB/RGBA: is it possible to hide this field when the connected image is a grayscale image?

chaiNNer-fast-nlmeans-2023-6-23_20-07


joeyballentine:
> is it possible to hide this field when the connected image is a grayscale image?

Yes actually. You would use an "if_group" with a type condition that matches RGB/RGBA, similar to this: https://github.com/chaiNNer-org/chaiNNer/blob/880441d6367d73f582081fe42c199d65184ebfb4/backend/src/packages/chaiNNer_pytorch/pytorch/restoration/upscale_face.py#L125

adegerard:
> Yes actually. You would use an "if_group" with a type condition that matches RGB/RGBA, similar to this:
> 
OK. I'll modify my code.

adegerard:
Done.
chaiNNer-fast-nlmeans-2023-6-23_21-06

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
```diff
@@ -0,0 +1,92 @@
...
+            maximum=50.,
+            default=3.0,
+            precision=1,
+            controls_step=0.1,
+            slider_step=0.1,
+        ),
+        if_group(
+            Condition.type(0, "Image { channels: 3 }")
+            | Condition.type(0, "Image { channels: 4 }")
+        )(
```
RunDevelopment, line 34: We can move the OR into the type expression.

```suggestion
        if_group(Condition.type(0, "Image { channels: 3 | 4 }"))(
```

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
```diff
@@ -0,0 +1,92 @@
...
+    schema_id="chainner:image:fast_nlmeans",
+    name="Fast NL means",
+    description="Use the fast Non-Local Means algorithm to denoise an image.",
+    icon="CgEditNoise",
+    inputs=[
+        ImageInput("Image", channels=[1, 3, 4]),
+        SliderInput(
+            "Luminance strength",
+            minimum=0.,
+            maximum=50.,
```
RunDevelopment, line 25: We don't care whether those values are floats or ints. They will be sent to the frontend via JSON, so this information is lost during the JSON conversion.

However, I think that extra dot is annoying, so could we please get rid of it?

```suggestion
            minimum=0,
            maximum=50,
```

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
```diff
@@ -0,0 +1,92 @@
...
+    h: float,
+    h_color: float,
+    template_window_size: int,
+    search_window_size: int,
+) -> np.ndarray:
+
+    if not (template_window_size % 2):
+        raise ValueError("\'Template window size\' must be an odd value.")
+    if not (search_window_size % 2):
+        raise ValueError("\'Search window size\' must be an odd value.")
```
RunDevelopment, line 71: Instead of asserting odd values here, I think we should instead think of a way to allow all inputs. The main advantage of all allowing all values in the input range is that the user experience is better. The type system cannot enforce odd values, so users cannot know that e.g. 20 is not valid until they run their chain and see it crash.

One idea would to use a radius instead of a size. Then we can then do 2r+1 to get an odd size.

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
```diff
@@ -0,0 +1,87 @@
...
+                default=3.,
+                precision=1,
+                controls_step=0.1,
+                slider_step=0.1,
+            )
+        ),
+        NumberInput("Patch radius",
+            minimum=1,
+            default=3,
+            maximum=30,
```
RunDevelopment, line 45: Why is the minimum 1? We use 2r+1, so shouldn't the radius' minimum be 0, so we can get a size of 1?

Why is maximum 30? It was 49 before, which would translate to a 24. Is the maximum just an arbitrary number? I'm not against having a defined maximum, not allowing inputs that don't make sense is a good thing, I'm just curious why this number.

Same for search radius.

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
```diff
@@ -0,0 +1,87 @@
...
+        rgb = image_array[:, :, :3]
+        if c == 4:
+            alpha = image_array[:, :, 3]
+
+        denoised = cv2.fastNlMeansDenoisingColored(src=rgb, h=h, hColor=h_color,
+            templateWindowSize=patch_window_size,
+            searchWindowSize=search_window_size)
+
+        if c == 4:
+            denoised = np.dstack((denoised, alpha))
```
RunDevelopment, line 84: PyRight can't statically prove that this way of handling alpha is correct. The issue is that it doesn't know that the value of `alpha` depends on `c == 4`, so when we do `np.dstack((denoised, alpha))`, PyRight thinks that `alpha` might not have been assigned a value just (since it's only conditionally assigned).

The fix is to replace the second `c == 4` with a condition that depends on the value of `alpha`. This makes it possible for PyRight to see that the code is correct.
```suggestion
        alpha = None
        if c == 4:
            alpha = image_array[:, :, 3]

        denoised = cv2.fastNlMeansDenoisingColored(src=rgb, h=h, hColor=h_color,
            templateWindowSize=patch_window_size,
            searchWindowSize=search_window_size)

        if alpha is not None:
            denoised = np.dstack((denoised, alpha))
```

RunDevelopment:
Formatting on tests currently fail. To fix the formatting, run the command `npm run lint-fix` - after installing our dev dependencies of course (`npm ci` and `pip install -r requirements.txt`). The tests fail because you added a new test case (the .chn file) but didn't update the test snapshots. Run `npm test -- -u` to update them.

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
adegerard: 1. error from my side. However, not sure it makes sense as I have never saw this values in any paper.
https://www.apeer.com/app/modules/Non-local-means-denoising/da13051a-db17-4b97-a260-8faa9ac341b7
let me know.

2. Maximum is an arbitrary nb.Never seen such high values in papers. (window size  = 2*radius+1)
https://jivp-eurasipjournals.springeropen.com/articles/10.1186/s13640-018-0346-y
I do not see why patch size would be > search size, but did not implement a condition as I am not an expert.
 I am open to any suggestions.

adegerard:
> Formatting on tests currently fail. To fix the formatting, run the command `npm run lint-fix` - after installing our dev dependencies of course (`npm ci` and `pip install -r requirements.txt`). The tests fail because you added a new test case (the .chn file) but didn't update the test snapshots. Run `npm test -- -u` to update them.

I didn't know. I am a beginner. Would it be possible to document the different steps before creating a pull request ? Thanks.

RunDevelopment:
> Would it be possible to document the different steps before creating a pull request ?

Good point. I'll add it to our contributing docs.

backend/src/packages/chaiNNer_standard/image_filter/noise/fast_nlmeans.py
RunDevelopment: > However, not sure it makes sense as I have never saw this values in any paper.

Then let's keep it at 1.

> Never seen such high values in papers.

Good enough for me. I was just curious.

RunDevelopment (approved):
Thank you @adegerard!

PR #1902, opened 2023-06-30
Improved node docs

RunDevelopment:
Changes:
- Removed examples API.
- Added "See Also" section. It just contains links to other nodes.
- Changed python API for descriptions. All methods that take secription now support lists. Lists items are joined together using `\n\n`. This makes it easy to write paragraphs. E.g. `withDocs("para1", "para2")`
- Renamed `widthDocumentation` to `withDocs`.
- Changed links to other nodes. Instead of using empty links, I switched to inline code. E.g. `` `chainner:ncnn:load_model` ``. This is easier to use, and it gives us something to fall back on when the schema id is invalid (e.g. because the node isn't installed).
- Added more documentation to a few nodes.

image


src/renderer/components/NodeDocumentation/NodeDocs.tsx
```diff
@@ -33,65 +34,75 @@ import { NodeFooter } from '../node/NodeFooter/NodeFooter';
...
-                        onClick={() => openNodeDocumentation(nodeSchema.schemaId)}
-                    >
-                        {nodeSchema.name}
-                    </Text>
-                );
-            }
-        }
-
+    // eslint-disable-next-line react/prop-types
+    a: memo(({ children, href }) => {
```
joeyballentine, line 73: I don't think we need to put the link thing here, it was just for the node links which have moved to code. The default should work fine.

joeyballentine (dismissed):

joeyballentine:
I didn't mean to approve that. I think you missed the markdown customization in the node description tooltips -- those have to properly convert the schema ID code block to a node name as well

joeyballentine (approved):

PR #1903, opened 2023-07-01
Add algorithms to the "color transfer" node

adegerard:
- Currently implemented algorithm is named Std+Mean
- Added Linear Histogram Matching algorithm
- Added  Principal Component Color Matching algorithm

Notes:
1) Removed the possibility to use a grayscale image for the input image: this was a bug which causes an exception. I assume that the input image is colored (alpha layer accepted)
2) The current default algorithm is the "std+mean" even if it may not be the best when the input img is the same as the ref image (except colors). Reason: users won't understand why the generated image is different compared to previous versions
3) In "@correction_group.register" function, I hard-coded the input id of the algorithm to 5 to not break current implementation. If not hard-coded, This would mean to define the migration (rules? settings?). Do I have to do this just for this 'small' improvment? Or is there a way to track this for a later "migration"?

chaiNNer-color-transfer-2023-7-1_18-25


joeyballentine:
> In "@correction_group.register" function, I hard-coded the input id of the algorithm to 5 to not break current implementation.

If I understand correctly, you did the right thing. The .with_id() function is meant for reordering inputs without breaking the node and needing a migration.

backend/src/nodes/impl/color_transfer/linear_histogram.py
```diff
@@ -0,0 +1,53 @@
...
+    using the Linear Histogram Matching..
+
+    This implementation is based on to the Hertzmann, Aaron. "Algorithms
+    for Rendering in Artistic Styles." Ph.D., New York University, 2001.
+    """
+
+    shape = img.shape
+
+    # Convert HxWxC image to a (H*W)xC matrix.
+    content = to_uint8(img).reshape(-1, shape[-1])
```
RunDevelopment, line 36: Question: do we really need to convert our images to uint8 here just to convert them back later? I just don't see why we wouldn't get the same result without it?

backend/src/nodes/impl/color_transfer/principal_color.py
```diff
@@ -0,0 +1,56 @@
...
+        SPIE, 1998.
+    - Kotera, Hiroaki. "A Scene-Referred Color Transfer for Pleasant Imaging
+        on Display." In IEEE International Conference on Image Processing 2005,
+        2:II-5, 2005.
+    """
+
+    shape = img.shape
+
+    # Convert HxWxC image to a (H*W)xC matrix.
+    content = to_uint8(img).reshape(-1, shape[-1])
```
RunDevelopment, line 37: Same here.

backend/src/packages/chaiNNer_standard/image_filter/correction/color_transfer.py
```diff
@@ -19,42 +41,55 @@
             best results.""",
     icon="MdInput",
     inputs=[
-        ImageInput("Image", channels=[1, 3, 4]),
+        ImageInput("Image", channels=[3, 4]),
```
RunDevelopment, line 44: That is a breaking change. Why did you remove support for grayscale images?

adegerard: As I mentionned it above, there was a bug:
image

joeyballentine: You could have it convert to RGB automatically instead

backend/src/nodes/impl/color_transfer/linear_histogram.py
adegerard: My bad. I added this code before knowing the img.dtype was float: I then forgot to optimize this.
It works without the conversion.
Waiting for your feedback on the following conversation before pushing

backend/src/nodes/impl/color_transfer/principal_color.py
adegerard: same here! ;-)

backend/src/packages/chaiNNer_standard/image_filter/correction/color_transfer.py
adegerard: I could.
But results are so bad that I think that it will be a bad user experience. And this does not work for the 2 added algos.
This is to transfer colors, not to colorize grayscale images.
chaiNNer-color-transfer-2023-7-1_22-32

adegerard: "But results are so bad"  <- No, this is completely logical that it does not work.
Color transfer means that the color information is still here before the transfer. The 'color' information is lost during the BGR->grayscale conversion.

joeyballentine: True. In that case, we can either do a migration or just not do that and hope nobody was doing gray with this before. I doubt anyone was, and if it doesn't work with it gray then it's not really a loss at all

RunDevelopment: I would say that we just declare supporting grayscale images a bug. Colorization isn't the primary purpose of this node anyway.

RunDevelopment:
When I use an image that is black and white as a reference image with either of the 2 new methods, I get the following warnings:

```
C:\Users\micha\Git\chaiNNer\backend\src\nodes\impl\color_transfer\linear_histogram.py:16: RuntimeWarning: invalid value encountered in sqrt
  return eig_vec.dot(np.diag(np.sqrt(eig_val))).dot(eig_vec.T)
C:\Users\micha\Git\chaiNNer\backend\src\nodes\impl\image_utils.py:129: RuntimeWarning: invalid value encountered in cast
  return (img * 255).round().astype(np.uint8)
```

They both mean that some float is nan or +-inf. I'm not sure whether this is even something that we can fix. This might just be an inherent problem with the algorithms at hand.

Regardless, we should not output nan or +-inf. I don't know how, but it seems like those values survive the `.clip(0, 1)`, because the second warning comes from inside View Image (it converts the image to uint8 to display it). Could you please investigate this?

My chain:
image

adegerard:
> 
Linear Histogram: I reproduced the problem (RuntimeWarning) with a ref img which has only black and white pixels. It does not happen if there is at least one other pixel with a different value and then out img seems to be correct.
I currently don't know how to fix it.
I will try (1) to remove the sqrt "runtime warning" by checking the output of the covariance and (2) to understand why the output image is not clipped.

RunDevelopment:
> to understand why the output image is not clipped.

It seems like this is a just a property of `np.clip`. I tested it:

```py
>>> a = np.array([0, 1, 3, float("nan"), float("inf"), -float("inf")], dtype=np.float32)
>>> a
array([  0.,   1.,   3.,  nan,  inf, -inf], dtype=float32)
>>> np.clip(a, 0, 1)
array([ 0.,  1.,  1., nan,  1.,  0.], dtype=float32)
```

Since `np.clip` preserves `nan`, let's use `np.nan_to_num` to get rid of it. E.g. the following map `nan` to 0.

```py
>>> np.nan_to_num(np.clip(a, 0, 1))
array([0., 1., 1., 0., 1., 0.], dtype=float32)
```

I think mapping NaN to 0 is an okay behavior for us. NaN doesn't make sense for us in any way, so replacing it with any value is better than having it.

adegerard:
> It seems like this is a just a property of `np.clip`. I tested it:

The value of eig_val shall be >=0, otherwise the sqrt returns nan
```python
def matrix_sqrt(X):
    eig_val, eig_vec = np.linalg.eig(X)
    return eig_vec.dot(np.diag(np.sqrt(eig_val.clip(min=0)))).dot(eig_vec.T)
```
tested on lhm, works. on pccm, it works but the result is weird...

adegerard:
> Since `np.clip` preserves `nan`, let's use `np.nan_to_num` to get rid of it. E.g. the following map `nan` to 0.

I made some improvments so that anormal uses cases won't generate exceptions or NaN values, that's why I did not use the nan_to_num function.
I continue testing without using it to be sure to not mask any further errors.

RunDevelopment:
It seems to work perfectly now. Thank you @adegerard!

Principal color sometimes produces weird results, but that just seems to be the method working as intended.

RunDevelopment (approved):

PR #1905, opened 2023-07-02
Run `npm audit fix` + uninstall unnecessary deps

RunDevelopment:
Both deps I removed were unused.


RunDevelopment:
Great. Some of our deps declared their deps incorrectly or made a breaking change while ignoring semver, and now we have a type error. I'll fix it.

joeyballentine (approved):

PR #1909, opened 2023-07-03
Change `--no-backend` boolean arg to `--remote-backend <url>`

Splendide-Imaginarius:
Non-default ports work without issues in my testing (both CLI and GUI), so this is (unless I missed something?) strictly an improvement over the status quo. Non-default hosts partially work, but there are a few places (mainly in the GUI) where I wasn't able to add the needed plumbing, so `127.0.0.1` is hardcoded in those places for now. I think the extra plumbing could be added in a follow-up PR. Let me know if this looks OK, or if the structure needs changes.

Refs https://github.com/chaiNNer-org/chaiNNer/issues/1875.


src/main/backend/process.ts
```diff
@@ -28,6 +28,7 @@ const getBackendPath = lazy((): string => {
 interface BaseBackendProcess {
     readonly owned: boolean;
     readonly port: number;
+    readonly url: string;
```
RunDevelopment, line 31: I think we should remove `port`. `port` assumed localhost, so what's the point? Anything that uses `port` is incorrect anyway.

RunDevelopment (commented):
Questions and suggestions:

1. Should the option really be called `--remote-backend`? localhost isn't very remote. Maybe we could call it `--backend-address` or even just `--backend`.
2. Should the backend address be a URL or just a domain + port?
3. We used `127.0.0.1` specifically because `localhost` doesn't always work for SSE for some reason. So please use `127.0.0.1` in the run scripts in `package.json`.

> I think the extra plumbing could be added in a follow-up PR.

Let's please do it in this one. Having both the old port and new URL system is really unclean. We need to change [the ICP `get-port`](https://github.com/chaiNNer-org/chaiNNer/blob/b31655980b0a66b8337c817f15a732d73b9c919e/src/common/safeIpc.ts#L27) to `get-backend-url` and then we can use the new system in the frontend as well. After that is done, we just need to propagate the change from port to url through the components and contexts. TS should help you with that, since it will give you errors for all places you need to change.

Splendide-Imaginarius:
>     1. Should the option really be called `--remote-backend`? localhost isn't very remote. Maybe we could call it `--backend-address` or even just `--backend`.

The intent was to convey that it's not a locally managed backend spawned by chaiNNer's frontend. I think `--backend-address` could be misconstrued to mean "this is the address that the backend should be spawned on". You're right that having a "remote" backend on localhost is a little bit odd; suggestions welcome if you have a way to avoid both ambiguities.

>     2. Should the backend address be a URL or just a domain + port?

The intent there was to future-proof in case someone wants to run a backend with TLS, or on a subpath of a reverse proxy.

>     3. We used `127.0.0.1` specifically because `localhost` doesn't always work for SSE for some reason. So please use `127.0.0.1` in the run scripts in `package.json`.

OK, I'll make that change. I was a little confused why both `127.0.0.1` and `localhost` showed up in the code; standardizing on `127.0.0.1` sounds reasonable, and will maybe avoid issues on systems with IPv6.

> Let's please do it in this one. Having both the old port and new URL system is really unclean. We need to change [the ICP `get-port`](https://github.com/chaiNNer-org/chaiNNer/blob/b31655980b0a66b8337c817f15a732d73b9c919e/src/common/safeIpc.ts#L27) to `get-backend-url` and then we can use the new system in the frontend as well. After that is done, we just need to propagate the change from port to url through the components and contexts. TS should help you with that, since it will give you errors for all places you need to change.

OK, sounds good. I'll try that, hopefully will be fairly easy. It might take me a few days; my free time is mildly scarce this week.

RunDevelopment:
I took over @Splendide-Imaginarius :) Thank you for the great work!

Closed in favor of #1957.

Splendide-Imaginarius:
Awesome, thank you for finishing it @RunDevelopment! (Wish I had more free time lately.)

PR #1912, opened 2023-07-04
Don't check whether the directory exists in Save Image

### LINKED ISSUE #1904: Check that base directory exists before running a chain
- Chainner version: 0.18.9
- OS: Windows 11

Description:
Currently Chainner does not check if the base directory for saving exists before running the chain. It would useful to check before the processing starts as if you have moved or deleted the directory before running an iterator, each instance will be processed without presenting an error. At the end of the process multiple error lines will be displayed in the error output as the directory does not exist.

Logs:
[main.zip](https://github.com/chaiNNer-org/chaiNNer/files/11929966/main.zip)

RunDevelopment:
Resolves #1904.

Save Image will now just create the Base Directory if it doesn't exist already.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/io/save_image.py
+++ b/backend/src/packages/chaiNNer_standard/image/io/save_image.py
@@ -66,7 +66,7 @@ class JpegSubsampling(Enum):
     icon="MdSave",
     inputs=[
         ImageInput(),
-        DirectoryInput(has_handle=True),
+        DirectoryInput(must_exist=False, has_handle=True),
         TextInput("Subdirectory Path")
         .make_optional()
         .with_docs(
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_standard/image/io/save_image.py
```diff
@@ -66,7 +66,7 @@ class JpegSubsampling(Enum):
     icon="MdSave",
     inputs=[
         ImageInput(),
-        DirectoryInput(has_handle=True),
+        DirectoryInput(must_exist=False, has_handle=True),
```
joeyballentine, line 69: Why not change that here as well?

RunDevelopment: What are you talking about?

joeyballentine: My bad, I misread this. Don't do code reviews at 4am 😂

joeyballentine (approved):

PR #1913, opened 2023-07-04
Added simple Pick Color node

RunDevelopment:
This PR adds a very simple color picker nodes. It takes an image and absolute pixel coordinates, and outputs the color at that position.

---

In the future, we may want to extend this node to be a more general sampler node. Additional features may include:
- Normalized coords.
- Abyss policy (wrap, reflect, etc.).
- Interpolation method (NN, bi-linear, bi-cubic, etc.).


joeyballentine (approved):

joeyballentine:
Perfect

PR #1923, opened 2023-07-07
Add handle colors to input/output docs

joeyballentine:
For some reason, I couldn't get linear gradients to work, assumedly because this is inside a bulleted list? I'm not really sure, but this is the solution I ended up with.

image

Multiple:

image

With optional tag:

image


RunDevelopment:
Should we show this for inputs without handles? It would be more consistent, but most inputs without handles are dropdowns which are gray and I don't think that color tells you a whole lot.

joeyballentine:
> Should we show this for inputs without handles?

Nope. Good call

RunDevelopment (approved):

PR #1925, opened 2023-07-08
Use fake handles on node examples

joeyballentine:


src/renderer/helpers/nodeState.ts
```diff
@@ -71,6 +71,7 @@ export interface NodeState {
     readonly connectedOutputs: ReadonlySet<OutputId>;
     readonly type: TypeInfo;
     readonly testCondition: (condition: Condition) => boolean;
+    readonly useFakeHandles?: boolean;
```
RunDevelopment, line 74: I don't like that the prop is optional here and in all prop interfaces. We might have missed passing it somewhere and TS wouldn't be able to help us.

Could we make this a required boolean everywhere?

RunDevelopment (approved):

RunDevelopment:
Looking good!

PR #1926, opened 2023-07-08
File input docs improvements

joeyballentine:
- Displays supported filetypes using type tags instead of in the description. Also applies automatically now instead of needing to be done via the backend
- Indicates if an input is the primary input for its type, and lets the user know what that means.

I don't necessarily like having all those booleans in InputOutputItem, but we can maybe have some kind of doc component selector thing in the future if we need to add more of those for different input types.

image

### DIFF
--- a/src/renderer/components/NodeDocumentation/NodeDocs.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeDocs.tsx
@@ -49,7 +49,7 @@ const InputOutputItem = memo(({ type, item }: InputOutputItemProps) => {
 
     return (
         <ListItem my={2}>
-            <HStack>
+            <HStack mb={1}>
                 <Text
                     fontWeight="bold"
                     userSelect="text"
@@ -85,7 +85,7 @@ const InputOutputItem = memo(({ type, item }: InputOutputItemProps) => {
 
             <VStack
                 alignItems="start"
-                mb={2}
+                mb={1}
                 w="full"
             >
                 {isFileInput && supportedFileTypes.length > 0 && (
--- a/src/renderer/components/NodeDocumentation/NodeDocs.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeDocs.tsx
@@ -35,7 +35,8 @@ interface InputOutputItemProps {
 }
 
 const InputOutputItem = memo(({ type, item }: InputOutputItemProps) => {
-    if ('optional' in item && item.optional) {
+    const isOptional = 'optional' in item && item.optional;
+    if (isOptional) {
         // eslint-disable-next-line no-param-reassign
         type = withoutNull(type);
     }
@@ -55,7 +56,7 @@ const InputOutputItem = memo(({ type, item }: InputOutputItemProps) => {
                 >
                     {item.label}
                 </Text>
-                {'optional' in item && item.optional && (
+                {isOptional && (
                     <TypeTag
                         isOptional
                         fontSize="small"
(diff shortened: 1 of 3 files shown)


src/renderer/components/NodeDocumentation/docsMarkdown.tsx
```diff
@@ -12,7 +12,7 @@ const getDocsMarkdownComponents = (interactive: boolean): Components => {
         p: ({ children }) => {
             return (
                 <Text
-                    my={2}
+                    // my={2}
```
RunDevelopment, line 15: Why?

src/renderer/components/NodeDocumentation/NodeDocs.tsx
```diff
@@ -30,61 +30,98 @@ import { NodeExample } from './NodeExample';
...
+    }
 
-        return (
-            <ListItem my={2}>
-                <HStack>
+    const handleColors = getTypeAccentColors(type);
+
+    const isFileInput = 'kind' in item && item.kind === 'file';
+    const supportedFileTypes = isFileInput ? item.filetypes : [];
+    const isPrimaryInput = isFileInput && 'primaryInput' in item && item.primaryInput;
```
RunDevelopment, line 47: Why do you even need to do these `in` checks?

src/renderer/components/NodeDocumentation/NodeDocs.tsx
```diff
@@ -30,61 +30,98 @@ import { NodeExample } from './NodeExample';
...
+    return (
+        <ListItem my={2}>
+            <HStack>
+                <Text
+                    fontWeight="bold"
+                    userSelect="text"
+                >
+                    {item.label}
+                </Text>
+                {'optional' in item && item.optional && (
```
RunDevelopment, line 58: Not a fan of this code duplication (see without null code). Could make an `optional` variable for it?

src/renderer/components/NodeDocumentation/docsMarkdown.tsx
joeyballentine: I put everything in a VStack that has this margin instead.

src/renderer/components/NodeDocumentation/NodeDocs.tsx
joeyballentine: TS complains. See this: https://stackoverflow.com/questions/58974640/typescript-property-does-not-exist-on-union-type

Basically, since its a union, .optional, .kind, etc don't exist on outputs and therefore TS yells at you if you don't type narrow first

RunDevelopment: Hence my question. Both `Input` and `Output` have a `kind` property, so this check shouldn't be necessary. And the `'primaryInput' in item` also shouldn't be necessary because TS already narrowed it down to FileInput because of the `isFileInput &&`.

src/renderer/components/NodeDocumentation/docsMarkdown.tsx
RunDevelopment: That's the issue. You can't remove the margin here, because that is necessary for markdown formatting. You just removed all spacing between paragraphs. 

It's cool that your VStack has margin, but now nothing else has :)

joeyballentine: Oooooh oops

src/renderer/components/NodeDocumentation/NodeDocs.tsx
joeyballentine: Hmm lemme try

joeyballentine: yeah it's fine without them. I guess this is one of those cases where whatever way I was originally doing it required something like that, and when I split it up and moved things I kept some unnecessary things

joeyballentine:
going to assume this is good now. if it isn't, we can always fix it in another pr

PR #1929, opened 2023-07-09
Add more doc information for other input types

joeyballentine:
image
image
image
image

### DIFF
--- a/src/renderer/components/NodeDocumentation/NodeDocs.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeDocs.tsx
@@ -137,9 +137,16 @@ const InputOutputItem = memo(({ type, item }: InputOutputItemProps) => {
                         fontSize="md"
                         userSelect="text"
                     >
-                        {`A ${item.multiline ? 'multi-line ' : ''}string between ${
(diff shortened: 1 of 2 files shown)


RunDevelopment:
My thoughts in no particular order:

- I like listing dropdown variants.
- We use minimum text length 0 vs 1 inconsistently, so we should probably treat them the same. Also, I don't think that showing min len = 0/1 has much value, because it pretty much means that there is no required minimum length.
- I think we shouldn't show numeric precision. All our number inputs with precision >=1 can deal with any number of decimal digits (I think). As I see it, limiting the number of decimal digits is a pure UI thing, and not part of the function of a node.
- Numeric minimum and maximum are already part of the type, so I don't see why we would repeat it.

About numbers again: if you think that number types are too unclear to new users (and I agree), then we could add tooltips to types that explain them. E.g. `int(10..)` could have the tooltip text "An integer that is at least 10."

RunDevelopment:
image
image


Not the most natural of language.

joeyballentine:
image

joeyballentine:
Assuming this is good now and merging

PR #1930, opened 2023-07-09
Docs search

joeyballentine:
image

Minisearch is great

### DIFF
--- a/src/renderer/components/NodeDocumentation/NodeDocumentationModal.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeDocumentationModal.tsx
@@ -3,7 +3,6 @@ import {
     HStack,
     Icon,
     IconButton,
-    Input,
     Modal,
     ModalBody,
     ModalCloseButton,
@@ -25,7 +24,6 @@ interface NodeDocumentationModalProps {
 }
 
 export const NodeDocumentationModal = memo(({ isOpen, onClose }: NodeDocumentationModalProps) => {
-    const [searchQuery, setSearchQuery] = useState('');
     return (
         <Modal
             isCentered
@@ -56,13 +54,6 @@ export const NodeDocumentationModal = memo(({ isOpen, onClose }: NodeDocumentati
                             />
                         </Box>
                         <Box whiteSpace="nowrap">Node Documentation</Box>
-                        <Input
-                            placeholder="Search..."
-                            value={searchQuery}
-                            w="full"
-                            onChange={(e) => setSearchQuery(e.target.value)}
-                        />
-                        <Box w={8} />
                     </HStack>
                 </ModalHeader>
                 <ModalCloseButton />
@@ -75,7 +66,7 @@ export const NodeDocumentationModal = memo(({ isOpen, onClose }: NodeDocumentati
                         h="full"
                         w="full"
                     >
-                        <NodesList searchQuery={searchQuery} />
+                        <NodesList />
                         <NodeDocs />
                     </HStack>
                 </ModalBody>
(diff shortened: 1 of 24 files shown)


src/renderer/components/NodeDocumentation/NodesList.tsx
```diff
@@ -1,18 +1,43 @@
...
+    miniSearch.addAll(schema);
+
+    const searchResult = miniSearch.search(searchQuery, {
+        boost: { name: 2 },
+        fuzzy: 0.2,
+        prefix: true,
+        combineWith: 'AND',
+    });
+    const matchingSchemaIds = useMemo(
+        () => [...new Set(searchResult.map((s) => String(s.id)))],
```
RunDevelopment, line 31: How are duplicates possible? Does mini search not guarantee that ids are unique in the search result list?

src/renderer/components/NodeDocumentation/NodesList.tsx
```diff
@@ -1,18 +1,43 @@
...
         useContext(NodeDocumentationContext);
 
     const { schemata, categories } = useContext(BackendContext);
     const schema = schemata.schemata;
-    const byCategories = useMemo(() => getNodesByCategory(schema), [schema]);
+
+    const miniSearch = new MiniSearch({
+        idField: 'schemaId',
+        fields: ['category', 'description', 'name', 'subcategory'],
+    });
```
RunDevelopment, line 20: `fields` is just a string array and I can already see us changing the name of some key (e.g. "subcategory" is still the old name of nodes groups) and breaking search without realizing it. I think we should extract it out into an array with a better type.

```suggestion
    const idField: keyof NodeSchema = 'schemaId';
    const fields: (keyof NodeSchema)[] = ['category', 'description', 'name', 'subcategory'];
    const miniSearch = new MiniSearch({ idField, fields });
```

RunDevelopment:
I think the layout is a bit confusing. Without using it, it's not clear that the search box only filters the entries in the nodes list. There's just no visual connection between the two.

Maybe we could put it above the nodes list? This would make the connection very clear. The node selector already uses this layout, so users should be able to immediately know what it does. 
The only inconsistency between the two would be that the node selector uses a different search algorithm and will return different results. This might not be an issue though.

src/renderer/components/NodeDocumentation/NodesList.tsx
joeyballentine: Honestly the set thing was a copilot suggestion and I didn't even question it...

joeyballentine:
image

src/renderer/components/NodeDocumentation/NodeDocumentationModal.tsx
```diff
@@ -11,7 +11,7 @@ import {
     ModalOverlay,
     Tooltip,
 } from '@chakra-ui/react';
-import { memo } from 'react';
+import { memo, useState } from 'react';
```
joeyballentine, line 14:
```suggestion
import { memo } from 'react';
```

RunDevelopment:
I think we should include input/output descriptions+names in the search. It's confusing that they aren't included.

image

RunDevelopment:
image

I'm mean, I know.

I'm honestly not sure what the best way of fixing this is, but I can think of 2 approaches:

1. We go in deep on describing the page. Basically, we create an intermediate representation (IR) for the docs of a single node. This IR would already have most extra infos (e.g. supported file types, etc) rendered as text. Search would then be done on the IR. Rendering the IR should be fairly simple.
    
   However, this does mean that it will be harder to add features to the docs, since we have to funnel everything through the IR first. So something like "add a button here" would be very difficult.


2. We abuse React. We'll use React to (1) create the rendered pages of all nodes, (2) extract their text, and then (3) search on the extracted text. Step (1) and (3) are easy. Step (2) is the issue. We need a way to only extract the text that we are interested in. Just taking all text might be an okay approximation, or we might need something smarter. (Step 1 and 2 are cached btw.)

    The main advantage of this approach is that we don't have to worry about text search ever again, because it will search in everything. It will just work.

---

Anyway, do we want to do this now? If not, then feel free to merge. It's not prefect, but it's good enough as is.

joeyballentine:
Yeah I don't really wanna worry about that right now. I don't think that's too big of a deal at the moment. We should definitely do something about it at some point but I think it's fine as-is for now

PR #1935, opened 2023-07-11
Update checking improvement

joeyballentine:
- Switch to using the official GitHub REST API npm package 
- New UI
    - new update indicator next to the version tag
    - new modal that appears when you click the update indicator
        - Markdown formatted changelog of versions from current to latest
        - Ability to open the github release page or chaiNNer.app
    - new alert dialog to alert the user of an update
- Ability to ignore individual updates
- Changed the keyname of the update check setting to reset it for users. Mild inconvenience, but I'd rather them know that there's a new system and that they can ignore individual updates now.

image

image


src/renderer/components/Header/AppInfo.tsx
```diff
@@ -14,21 +47,185 @@ export const AppInfo = memo(() => {
...
+                                Close
+                            </Button>
+                            <Button
+                                isExternal
+                                as={Link}
+                                colorScheme="green"
+                                href={updateVersion?.html_url}
+                            >
+                                Download from GitHub
+                            </Button>
```
RunDevelopment, line 160: I don't think we need the GH download link. Our download page is more user-friendly anyway.

src/renderer/contexts/SettingsContext.tsx
```diff
@@ -60,7 +60,7 @@ export const SettingsProvider = memo(({ children }: React.PropsWithChildren<unkn
     const useSystemPythonLocation = useMemoArray(
         useLocalStorage<string | null>('system-python-location', null)
     );
-    const useCheckUpdOnStrtUp = useMemoArray(useLocalStorage('check-upd-on-strtup', true));
+    const useCheckUpdOnStrtUp = useMemoArray(useLocalStorage('check-upd-on-strtup-2', true));
```
RunDevelopment, line 63: Do we even need a setting for update checks anymore? The update "notification" is very non-intrusive now, why shouldn't we just always check for updates?

src/renderer/components/Header/AppInfo.tsx
joeyballentine: I agree that our page is more user friendly but I'd still like to link this. I see no reason not to

src/renderer/contexts/SettingsContext.tsx
joeyballentine: I agree, but I guarantee you somebody will complain if it's removed completely

src/renderer/components/Header/AppInfo.tsx
```diff
@@ -14,21 +47,185 @@ export const AppInfo = memo(() => {
...
+                                >
+                                    Close
+                                </Button>
+                                <Button
+                                    onClick={() => {
+                                        if (updateVersion?.tag_name) {
+                                            localStorage.setItem(
+                                                `ignore-update:${updateVersion.tag_name}`,
+                                                'true'
+                                            );
```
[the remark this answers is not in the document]
joeyballentine: I went with option 2 because it's cleaner

src/renderer/components/Header/AppInfo.tsx
RunDevelopment: Why do want to link it? Or phrased differently: what additional value does it provide?

I just see it as an option that could confuse new users. We want people to update, so I think that we should make that process simple. This is why I like the website: it auto selects the correct OS. I just don't think that new users will have an easy time picking the correct download from our GH releases.

As for advanced users: they should be able to navigate to the GH releases themselves. If you think that's too many clicks, then we could add a "View this release on GitHub" link to the download page.

src/renderer/contexts/SettingsContext.tsx
RunDevelopment: Alright. I can see that.

joeyballentine:
Assuming this is good now

PR #1937, opened 2023-07-13
Improved Upscaling Node Docs + Input Hints

joeyballentine:
- Upscaling node doc improvements
- The hint feature discussed on discord. It applies to both labeled and generic inputs, though I really only ever see it being useful for labeled ones, particularly dropdowns,

image

image

### DIFF
--- a/backend/src/nodes/base_input.py
+++ b/backend/src/nodes/base_input.py
@@ -89,6 +89,7 @@ class BaseInput:
 
         # Optional documentation
         self.description: str | None = None
+        self.hint: bool = False
 
     # This is the method that should be created by each input
     def enforce(self, value: object):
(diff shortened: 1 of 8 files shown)


RunDevelopment:
Here's what I found:

1. The (i) icon is too small. It's readable on default zoom. image I'm not sure how to fix this.
2. Please add an opening delay to the tooltip. Having a random tooltip pop open every time to move your mouse over a node is annoying.
3. The text is rendered differently in the tooltip and in the docs. This is because the tooltip just ignores Markdown. We already have the code to render markdown in tooltip (for node descriptions), so please use that.
image

joeyballentine:
image

image

I changed it to one that can (IMO) be more easily identified at a smaller scale.

joeyballentine:
image
and now this uses markdown

RunDevelopment:
> I changed it to one that can (IMO) be more easily identified at a smaller scale.

It's even harder to read for me now 😄 

How about we make the original icon bigger? It was smaller than an upper case letter, so maybe we could make the size of the "T"/"S"?

joeyballentine:
image

RunDevelopment (approved):
That works. Thanks!

PR #1938, opened 2023-07-13
Update splash.ts -disable alwaysOnTop

jabbawack0:
Disable splash screen's alwaysOnTop property. 
Splash screen blocks user from doing anything else while starting up.  
Lasts up to 40 seconds on HDD.

### DIFF
--- a/src/main/gui/splash.ts
+++ b/src/main/gui/splash.ts
@@ -21,7 +21,7 @@ export const addSplashScreen = (monitor: ProgressMonitor) => {
         minimizable: true,
         maximizable: false,
         closable: false,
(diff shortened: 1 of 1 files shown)


RunDevelopment:
Honestly, I agree. The splash screen being on top always bugged me. The installations we do during the first start-up might take a while, so we do block the user from doing anything else during that time?

@joeyballentine What do you think?

> Lasts up to 40 seconds on HDD.

Oh god.

joeyballentine:
Yeah we talked about it on discord already in EE @RunDevelopment

joeyballentine (approved):

PR #1939, opened 2023-07-14
Fix iterator glob on windows

joeyballentine:
Check #help-troubleshooting on discord for context


zeptofine:
oh what a difference two lines can make

joeyballentine:
I don't think this really needs a review, so I'm merging it to make sure it gets into the nightly

PR #1943, opened 2023-07-16
Add trigger for site build to release workflow

joeyballentine:
This for sure works. The issue is I had to use a personal token secret rather than the default GH token since that one is repo-scoped. Should be fine though.


joeyballentine:
image

RunDevelopment (approved):
Seems to work, so sure.

PR #1945, opened 2023-07-16
Switch Node

joeyballentine:
Did what I could to make this suggestion happen.

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/utility/value/switch.py
+++ b/backend/src/packages/chaiNNer_standard/utility/value/switch.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+from enum import Enum
+from typing import Union
+
+from nodes.groups import optional_list_group
+from nodes.properties.inputs import AnyInput, EnumInput
+from nodes.properties.outputs import BaseOutput
+from nodes.utils.utils import ALPHABET
+
+from .. import value_group
+
+
+class ValueIndex(Enum):
+    A = 0
+    B = 1
+    C = 2
+    D = 3
+    E = 4
+    F = 5
+    G = 6
+    H = 7
+    I = 8
+    J = 9
+
+
+@value_group.register(
+    schema_id="chainner:utility:switch",
+    name="Switch",
+    description="Allows you to pass in multiple inputs and then change which one passes through to the output.",
+    icon="BsShuffle",
+    inputs=[
+        EnumInput(ValueIndex),
+        AnyInput(label="Value A"),
+        AnyInput(label="Value B"),
+        optional_list_group(
+            *[
+                AnyInput(label=f"Value {letter}").make_optional()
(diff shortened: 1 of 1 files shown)


joeyballentine:
Open question: Is there something we can do with Navi to restrict all the types to be the same?

backend/src/packages/chaiNNer_standard/utility/value/switch.py
```diff
@@ -0,0 +1,67 @@
...
+            ValueIndex::C => Input3,
+            ValueIndex::D => Input4,
+            ValueIndex::E => Input5,
+            ValueIndex::F => Input6,
+            ValueIndex::G => Input7,
+            ValueIndex::H => Input8,
+            ValueIndex::I => Input9,
+            ValueIndex::J => Input10,
+            _ => never
+        }""",
```
RunDevelopment, line 58: We can do a bit better. Since the node throws an error on null, we can also replicate this behavior here.

This is how we can do it in the type system. Please also add a never reason telling users that they selected an unassigned value.

```suggestion
            output_type="""
            let value = match Input0 {
                ValueIndex::A => Input1,
                ValueIndex::B => Input2,
                ValueIndex::C => Input3,
                ValueIndex::D => Input4,
                ValueIndex::E => Input5,
                ValueIndex::F => Input6,
                ValueIndex::G => Input7,
                ValueIndex::H => Input8,
                ValueIndex::I => Input9,
                ValueIndex::J => Input10,
                _ => never
            };
            
            match value {
                null => never,
                _ => value
            }
            """,
```

backend/src/packages/chaiNNer_standard/utility/value/switch.py
```diff
@@ -0,0 +1,67 @@
...
+            label="Value",
+        )
+    ],
+    see_also=["chainner:utility:pass_through"],
+)
+def switch_node(selection: ValueIndex, *args: Union[object, None]) -> object:
```
RunDevelopment, line 64: Isn't None ⊆ object?

RunDevelopment (changes requested):

RunDevelopment:
> Open question: Is there something we can do with Navi to restrict all the types to be the same?

One step back: should we do that? We have a few inputs that accept values of multiple kinds of types (e.g. TextInputs accept numbers and strings), so this restriction would limit utility. We also don't really need the restriction. Since everything is strongly typed, users can't use this node to create invalid chains.

To answer the question: kinda? A type is just a set of values for Navi. So `12` (the set containing only 12) is a perfectly good type, but enforcing that all inputs are 12 isn't very useful. So obviously we have to widen the type somewhat, but how far do we go? Do we allow all numbers? Just integers?

joeyballentine:
> One step back: should we do that?

ehhh probably not

RunDevelopment (approved):

PR #1947, opened 2023-07-17
Highlight node documentation search query terms

joeyballentine:
I could see this being a bit annoying, so I almost wonder if it should be a toggleable thing. I'm open to adding that as part of this if you agree. I do think this helps visibly show how a node matches what you've searched for.

Also, part of this PR makes it auto-select the first node in the list when you search.

image
image

Also while doing this I realized we don't show node groups (subcategories) anywhere on here, so I'll address that in a separate PR.

### DIFF
--- a/src/renderer/components/NodeDocumentation/NodesList.tsx
+++ b/src/renderer/components/NodeDocumentation/NodesList.tsx
@@ -50,6 +50,21 @@ export const NodesList = memo(() => {
         [byCategories, categories, scoreMap, searchQuery]
     );
 
+    const highestNode = useMemo(
+        () =>
+            byCategories
+                .get(categoriesByMaxNodeScore[0].name)
+                ?.sort(
+                    (a, b) => (scoreMap.get(b.schemaId) ?? 0) - (scoreMap.get(a.schemaId) ?? 0)
+                )[0] ?? undefined,
+        [byCategories, categoriesByMaxNodeScore, scoreMap]
+    );
+
+    useEffect(() => {
+        openNodeDocumentation(highestNode?.schemaId);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [highestNode]);
+
     const selectedElement = useRef<HTMLDivElement>(null);
 
     const [selectScrollTrigger, setSelectScrollTrigger] = useState(false);
--- a/src/renderer/components/NodeDocumentation/NodesList.tsx
+++ b/src/renderer/components/NodeDocumentation/NodesList.tsx
@@ -1,8 +1,6 @@
 import { Box, Center, HStack, Text, VStack } from '@chakra-ui/react';
-import MiniSearch from 'minisearch';
 import { memo, useEffect, useMemo, useRef, useState } from 'react';
 import { useContext } from 'use-context-selector';
-import { NodeSchema } from '../../../common/common-types';
 import { BackendContext } from '../../contexts/BackendContext';
 import { NodeDocumentationContext } from '../../contexts/NodeDocumentationContext';
 import { getNodesByCategory } from '../../helpers/nodeSearchFuncs';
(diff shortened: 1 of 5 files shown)


src/renderer/components/NodeDocumentation/docsMarkdown.tsx
```diff
@@ -1,25 +1,40 @@
...
 const getDocsMarkdownComponents = (interactive: boolean): Components => {
     return {
-        p: ({ children }) => {
+        p: memo(({ children }: PropsWithChildren<unknown>) => {
+            const { useNodeDocumentationSearch } = useContext(NodeDocumentationContext);
+            const { searchTerms } = useNodeDocumentationSearch;
+
+            const stringChildren = Array.isArray(children)
+                ? (children as string[]).join('')
+                : String(children);
```
RunDevelopment, line 19: You just destroyed any and all formatting within paragraphs. No more bold, italics, links, inline-code, anything.

image

src/renderer/components/NodeDocumentation/NodesList.tsx
```diff
@@ -79,6 +50,21 @@ export const NodesList = memo(() => {
...
+                .get(categoriesByMaxNodeScore[0].name)
+                ?.sort(
+                    (a, b) => (scoreMap.get(b.schemaId) ?? 0) - (scoreMap.get(a.schemaId) ?? 0)
+                )[0] ?? undefined,
+        [byCategories, categoriesByMaxNodeScore, scoreMap]
+    );
+
+    useEffect(() => {
+        openNodeDocumentation(highestNode?.schemaId);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
RunDevelopment, line 65: Why ignore the warning?

src/renderer/components/NodeDocumentation/docsMarkdown.tsx
```diff
@@ -1,25 +1,40 @@
...
                 >
-                    {children}
+                    <Highlight
+                        query={searchTerms}
+                        styles={{
+                            backgroundColor: 'yellow.300',
+                        }}
+                    >
+                        {stringChildren}
+                    </Highlight>
```
RunDevelopment, line 34: You should probably make a component that wraps around `Highlight` and handles its styling + handling inner elements. I don't want to see `styles` being repeated everywhere, especially since we might want to change it.

RunDevelopment (changes requested):

src/renderer/components/NodeDocumentation/NodesList.tsx
joeyballentine: Because I only want it to run when that one changes

joeyballentine: It very much doesn't work properly with all the deps there, or at least it didn't back when I was doing this slightly differently. Maybe that changed though, so I'll try it again

src/renderer/components/NodeDocumentation/docsMarkdown.tsx
joeyballentine: Ugh.... That's gonna be a pain to solve

src/renderer/components/NodeDocumentation/NodesList.tsx
RunDevelopment: > It very much doesn't work properly with all the deps there

If you want to ensure that an effect only runs when a specific value changed, use a ref and store the last value in there. You can then do a `if (ref.current !== newValue)` inside the effect. Boilerplatey, but it's the only thing that is correct.

src/renderer/contexts/NodeDocumentationContext.tsx
```diff
@@ -25,11 +33,53 @@ export const NodeDocumentationProvider = memo(({ children }: React.PropsWithChil
...
+        searchResult,
+        searchTerms,
+    };
+
     const contextValue = useMemoObject({
         selectedSchemaId,
         isOpen,
         openNodeDocumentation,
         onClose,
+        nodeDocsSearchState,
```
RunDevelopment, line 82: You effectively disabled memoing because `nodeDocsSearchState` is a new object every time.

RunDevelopment:
I quickly fixed memoing myself, since it was still non-functional.

RunDevelopment (dismissed):

RunDevelopment:
Okay, I think we have to rethink the approach.

Funneling search terms from the *global* node docs context into `getDocsMarkdownComponents` is fundamentally incorrect because we use those components elsewhere.

Making the search query part of the global context also causes it to be sticky/remembered. This is an issue because suddenly `openNodeDocumentation` opens with the last search query. It also just fundamentally doesn't make sense that the search query is global.

I found something called [mark.js](https://www.npmjs.com/package/mark.js). I'll trying cooking something up with this.

joeyballentine:
Closing in favor of #1956

PR #1948, opened 2023-07-18
Move system stats to backend

### LINKED ISSUE #1105: Linux/MacOS RAM usage is inaccurate
Windows's RAM usage meter is accurate. However, Linux's isn't. This is because we look at the "free" memory rather than the "available" memory. 

[Discord discussion about this](https://discord.com/channels/930865462852591648/930875396277280788/1025840124644307084)

I'm still not sure why macos's is off though. If I knew how much it was off by, I could determine if we just need to recalculate it a bit.

joeyballentine:
Probably fixes #1105.

- Moves the system information collection to a backend route which the fronted polls using react-query. Requires psutil as a new dependency, which compared to what I was using for system stats with nodejs, is _so_ much better lol. 
- Removes all the nvidia smi grossness from the frontend. Good riddance.


backend/src/server.py
```diff
@@ -401,6 +401,35 @@ async def python_info(_request: Request):
...
 
 
+@dataclass
+class SystemStat:
+    label: str
+    percent: float
+
+
+@app.route("/system-stats", methods=["GET"])
+async def system_stats(_request: Request):
```
RunDevelopment, line 411: Why did you call the end point "system stats"? Wouldn't "system usage" be better? 

When I first read the title of this PR, I thought that you moved the system info collection for error logs into the backend.

RunDevelopment (changes requested):
See #1951 for more.

joeyballentine:
lol i clicked "cancel review" and it posted the wip review I was making earlier from my phone. thanks github

RunDevelopment (approved):

PR #1950, opened 2023-07-18
Better types in docs

RunDevelopment:
Changes:
- Don't show type for dropdowns. Showing all options is sufficient. We don't allow connections to dropdowns anyway.
- Added "Type:" before the type.
- Made "Options:" italic.
- Improved type tooltips. They are now derived from the type and use the same underlying explaining logic as type errors. So any improvement we make towards explaining doc types will as improve error messages. In fact, I had to improve the explaining and type error messages now use natural language more often.

image

image
image
image
image
image
image

### DIFF
--- a/src/common/types/explain.ts
+++ b/src/common/types/explain.ts
@@ -1,0 +1,169 @@
+import {
+    IntIntervalType,
+    NumberPrimitive,
+    NumericLiteralType,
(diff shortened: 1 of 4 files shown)


joeyballentine (approved):
Awesome

PR #1952, opened 2023-07-18
Added input conditions to docs

RunDevelopment:
This PR adds basic support for showing the condition of conditional inputs in node docs.

Future PRs should aim to improve the visual presentation. Right now, there is a lot of duplication because conditions are typically shared between inputs.

### Examples

image
image
image
image


joeyballentine:
The fact that both "condition" and the input name are italic is a little odd. IMO those need more contrast

RunDevelopment:
I agree. I initially made them bold, so they look just like input labels, but that was a bit too much IMO.

image

joeyballentine:
Maybe bold with a smaller weight?

joeyballentine:
Might be worth trying making the input names a larger font size and/or more bold or something. idk

RunDevelopment:
I put more space between inputs to make things clearer.

image

joeyballentine:
I think that looks good personally 👍

joeyballentine (approved):

PR #1954, opened 2023-07-19
Changed empty-string behavior of text inputs

RunDevelopment:
Changes:
- The default value of `min_length` is now 0.
- Text inputs will now strictly enforce length requirements.
- Text inputs will now interpret an empty input as "no input". Except if `allow_empty_string` is true, then they will treat an empty input as the empty string.
    - Only a few text inputs allow the empty string. It's not just the one in the Text node. I tested all text nodes and allowed empty strings where the empty string is very useful (e.g. Text Replace with replacement="" to remove text).

Open question: current save files.
Current save files can have the empty string in text inputs that previously already had min_length=0 and that now do not allow the empty string. This creates a weird situation where the empty string is a valid input as long as you don't touch it (then it will be removed).
How should we deal with this? Should we just live with this weirdness? Should we reset non-allowed empty string input to "no value" and break some chains? Should we make a migration?


joeyballentine:
> This creates a weird situation where the empty string is a valid input as long as you don't touch it (then it will be removed).

This sounds fine to me. It won't break existing chains but it'll go to the new behavior as soon as the chain is changed. That sounds unobtrusive enough

RunDevelopment:
I added a migration

joeyballentine (approved):

PR #1957, opened 2023-07-20
Change `--no-backend` boolean arg to `--remote-backend=<url>`

RunDevelopment:
This PR is based on #1909. I simply finished it and tested everything.

I choose to go with `--remote-backend=http://127.0.0.1:8000`, because it's simpler for now. The option is still internal, so we can change the API later.


joeyballentine:
> The option is still internal, so we can change the API later.

I plan on mentioning this in the next release, so are you sure about that?

joeyballentine (approved):

PR #1958, opened 2023-07-20
Fixed wide dropdowns

RunDevelopment:
This fixes wide dropdowns. Dropdowns will now always have their size determined by their container. Long options no longer cause them to expand.

image


joeyballentine (approved):
Yessss

PR #1959, opened 2023-07-20
Added Stretch Contrast node

### LINKED ISSUE #1789: Auto contrast and/or autolevels.
Just an idea for something I often use in image editing software which is some form of automatic contrast/levels that basically maximizes contrast.

RunDevelopment:
Resolves #1789.

I basically added [Stretch Contrast from Gimp](https://docs.gimp.org/2.10/en/gimp-filter-stretch-contrast.html) with a few extras.

image

The node has 3 modes:
- Auto: Behaves just like Gimp. The node will use the min/max values to stretch the image.
- Percentile. Similar to Auto, but we can deviate from the true min/max a little to account for noise.
- Manual. The user can provide a stretch min/max. This gives complete control over the stretching. In the example, I used it to 8x amplify a weak signal.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/stretch_contrast.py
+++ b/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/stretch_contrast.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from enum import Enum
+from typing import List, Tuple
+
+import numpy as np
+
+from nodes.groups import if_enum_group
+from nodes.properties.inputs import BoolInput, EnumInput, ImageInput, SliderInput
+from nodes.properties.outputs import ImageOutput
+from nodes.utils.utils import get_h_w_c
+
+from .. import adjustments_group
+
+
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/stretch_contrast.py
```diff
@@ -0,0 +1,112 @@
...
+
+class StretchMode(Enum):
+    AUTOMATIC = 0
+    PERCENTILE = 1
+    MANUAL = 2
+
+
+@adjustments_group.register(
+    schema_id="chainner:image:stretch_contrast",
+    description="Automatically stretches the histogram values in the given image.",
```
joeyballentine, line 33: You should mention here that it is similar to auto levels, since I'm sure that's something someone would search for

RunDevelopment: Done.

joeyballentine (approved):

PR #1960, opened 2023-07-21
Remove pip-list call from the frontend

joeyballentine:
The most we can do in terms of getting pip out of the frontend right now.

It's a little jank since the restart makes the failed to fetch error appear. That just needs to be fixed in general but we already discussed that on discord.


src/renderer/contexts/DependencyContext.tsx
```diff
@@ -302,17 +289,27 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
...
-                const res = await backend.dependencies();
-                return res.filter((d) => d.dependencies.length > 0);
-            },
-            successEffect: setAvailableDeps,
-        }),
-        [backend]
-    );
+    const {
+        data: depList,
+        isLoading,
```
joeyballentine, line 294:
```suggestion
```

src/renderer/contexts/DependencyContext.tsx
```diff
@@ -302,17 +289,27 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
...
+        isFetching,
```
joeyballentine, line 295:
```suggestion
```

backend/src/server.py
```diff
@@ -432,7 +432,17 @@ async def get_dependencies(_request: Request):
...
     for package in api.registry.packages.values():
-        pkg_dependencies = [x.toDict() for x in package.dependencies]
+        pkg_dependencies = []
+        for pkg_dep in package.dependencies:
+            installed_version = installed_packages.get(pkg_dep.pypi_name, None)
+            pkg_dep_item = {
+                **pkg_dep.toDict(),
+            }
+            if installed_version is None:
+                pkg_dep_item["installed"] = False
```
RunDevelopment, line 442: Why False?

joeyballentine: Because it's not installed. Idk, I guess it could be None instead

RunDevelopment (approved):

PR #1972, opened 2023-07-23
Stack many images

RunDevelopment:
This PR extends Stack Images to allow for up to 14 images, just like Z-stack Images.

I also used direct `return` statements in the node, which allows pyright to prove that we checked all `orientation` variants, which makes the `else: assert False` unnecessary.


joeyballentine (approved):
Those are some gross types

RunDevelopment:
> Those are some gross types

No array types :(

PR #1978, opened 2023-07-23
Changed Surface blur diameter to radius

RunDevelopment:
I noticed that the Diameter input of the Surface Blur node (#1292) behaved weirdly. In my testing, I set both sigmas to 1000 (max), which makes the node behave like a simple Gaussian blur. This makes it easy to see the effects of the diameter. I saw that D=1,2,3 produced the same image. Same with D=4,5 and D=6,7 etc. In general, only odd numbers >=3 were actually used by OpenCV's `bilateralFilter`.

To fix this, I switched to radius instead. The diameter is then calculated as 2r+1. This makes the node behave as expected and is consistent with all other blur nodes. I also set 100 as the max radius, because Surface blur is extremely slow for large radii (anything R>30 takes multiple seconds at 100% CPU on all cores).

Since both linear and log scale didn't feel right for the radius slider, I added a new sqrt scale. We can use this scale for other stuff in the future, but it's only used here right now.

image

Migration: Since this change is breaking, I added a migration. Migrating input values is easy, but connections are hard. 

In fact, 3 nodes necessary to approximately convert from diameter to radius (D=1 is incorrectly mapped to R=0):
image

And it takes 6 nodes for a correct conversion:
image

Do I have to code up a migration for connections, or can we just list this under "breaking changes" in the next release?

### DIFF
--- a/backend/src/nodes/properties/inputs/numeric_inputs.py
+++ b/backend/src/nodes/properties/inputs/numeric_inputs.py
@@ -119,7 +119,7 @@ class SliderInput(NumberInput):
         ends: Tuple[Union[str, None], Union[str, None]] = (None, None),
         hide_trailing_zeros: bool = False,
         gradient: Union[List[str], None] = None,
-        scale: Literal["linear", "log", "log-offset"] = "linear",
+        scale: Literal["linear", "log", "log-offset", "sqrt"] = "linear",
     ):
         super().__init__(
             label,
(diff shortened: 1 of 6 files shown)


joeyballentine:
> Do I have to code up a migration for connections, or can we just list this under "breaking changes" in the next release?

I'm fine with it just breaking a breaking change tbh

RunDevelopment:
> I'm fine with it just breaking a breaking change tbh

Then let's do that.

joeyballentine (approved):

PR #1983, opened 2023-07-24
Added linked inputs group

RunDevelopment:
As mentioned on Discord, I made a group to link the values of inputs. While not perfect, it's functional. The main issue is that values are linked via `inputData`, so sliders aren't synced in real time, but only when a value is entered (slider is released). 

image

---

A for the design: it's not done yet. I'll do it tomorrow.

### DIFF
--- a/backend/src/nodes/groups.py
+++ b/backend/src/nodes/groups.py
@@ -202,7 +202,7 @@ def optional_list_group(*inputs: BaseInput | NestedGroup):
     return group("optional-list")(*inputs)
 
 
-def linked_inputs_group(*inputs: BaseInput | NestedGroup):
+def linked_inputs_group(*inputs: BaseInput):
     """
     This group wraps around inputs of the same type. It ensures that all inputs have the same
     value.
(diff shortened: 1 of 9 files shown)


joeyballentine:
How hard would it be to add a way to define the relationship between the inputs? For example if one needs to be double the other, of if two linked file paths should have a different extension?

RunDevelopment:
> How hard would it be to add a way to define the relationship between the inputs?

Not very. The only hurdle is that you need to be able to describe this relationship in a JSON-API-compatible way. But we don't need this right now, so why add it?

joeyballentine:
> But we don't need this right now, so why add it?

Specifically I'd like to use that for the NCNN load model node, but I guess it's not too big of a deal right now

RunDevelopment:
Design: 
image

joeyballentine:
> Huh? NCNN Load Model already syncs up .bin and .param

Yeah but I've always wanted to replace that with a generic implementation. Right now it's just hardcoded on the frontend

joeyballentine (approved):

PR #1996, opened 2023-07-27
Unified crop node

RunDevelopment:
This unifies the Crop Border/Edges/Offsets nodes into one Crop node.

image

Other changes:
- Removed Crop Border/Edges/Offsets and added a migration for the new Crop node.
- Renamed "Crop (Content)" to "Crop Content".
- Changed order of inputs compared to Crop Edges/Offsets. Inputs are now consistently in order X then Y.


joeyballentine:
I think we can do even better. Crop (border) is just crop (edges) with all the sizes linked together. We have an input group that can link number inputs now dont we?

RunDevelopment:
> I think we can do even better. Crop (border) is just crop (edges) with all the sizes linked together. We have an input group that can link number inputs now dont we?

Linked inputs don't work for connections. So you would need to manually make 4 connections to remove a constant border.

That being said, maybe linked inputs should affect connections? So if you connect an edge to an input that is linked with other inputs, chainner automatically creates edges to all inputs. Same for removing edges.

If we do go with linked inputs, then Create Border/Edges would be trivial to unify.

joeyballentine:
That sounds good to me and makes sense to do

theflyingzamboni:
Request to change Crop Content to Crop To Content. A bit more accurate.

RunDevelopment:
> Request to change Crop Content to Crop To Content. A bit more accurate.

Done.

joeyballentine (approved):

PR #2008, opened 2023-07-29
Refactor BackendContext to handle nodes request/state + fix restart bug

joeyballentine:
Now we no longer get the backend failed to fetch error during an intentional restart.

### DIFF
--- a/src/renderer/contexts/BackendContext.tsx
+++ b/src/renderer/contexts/BackendContext.tsx
@@ -192,7 +192,7 @@ const useNodes = ({ backend, isRestarting }: { backend: Backend; isRestarting: b
     );
 
     let connectionState: 'connecting' | 'connected' | 'failed' = 'connecting';
-    if (nodesQuery.status === 'success') {
+    if (nodesQuery.status === 'success' && nodesInfo !== undefined) {
         connectionState = 'connected';
     } else if (nodesQuery.status === 'error') {
         connectionState = 'failed';
--- a/src/renderer/contexts/BackendContext.tsx
+++ b/src/renderer/contexts/BackendContext.tsx
@@ -198,13 +198,13 @@ const useNodes = ({ backend, isRestarting }: { backend: Backend; isRestarting: b
         connectionState = 'failed';
     }
 
-    return useMemoObject({
+    return {
         nodesInfo,
         schemaInputs,
         scope,
         refreshNodes,
         connectionState,
-    });
+    };
 };
 
 export const BackendProvider = memo(
--- a/src/renderer/contexts/BackendContext.tsx
+++ b/src/renderer/contexts/BackendContext.tsx
@@ -2,7 +2,7 @@ import { Scope } from '@chainner/navi';
 import isDeepEqual from 'fast-deep-equal';
 import React, { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { UseQueryResult, useQuery, useQueryClient } from 'react-query';
+import { useQuery, useQueryClient } from 'react-query';
 import { createContext, useContext } from 'use-context-selector';
 import { Backend, BackendNodesResponse, getBackend } from '../../common/Backend';
 import { Category, PythonInfo, SchemaId } from '../../common/common-types';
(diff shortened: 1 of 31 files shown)


src/renderer/contexts/BackendContext.tsx
```diff
@@ -29,6 +36,8 @@ interface BackendContextState {
     scope: Scope;
     restartingRef: Readonly<React.MutableRefObject<boolean>>;
     restart: () => Promise<void>;
+    nodesQuery: UseQueryResult<BackendNodesResponse, unknown>;
+    nodesInfo: NodesInfo | undefined;
```
RunDevelopment, line 40: No. `main.tsx` basically uses these 2 to deduce the current loading state. This is a suuuuper leaky abstraction, or rather no abstraction at all. So let's fix that. Maybe something like `connectionState: "connecting" | "connected" | "failed"` could work.

src/renderer/contexts/BackendContext.tsx
```diff
@@ -37,27 +46,130 @@ export const BackendContext = createContext<Readonly<BackendContextState>>(
...
-    }: React.PropsWithChildren<BackendProviderProps>) => {
+    };
+};
+
+export const BackendProvider = memo(
+    ({ url, pythonInfo, children }: React.PropsWithChildren<BackendProviderProps>) => {
+        const { t } = useTranslation();
         const backend = getBackend(url);
 
+        const { sendAlert, forgetAlert } = useContext(AlertBoxContext);
```
RunDevelopment, line 77: I'm not a fan of all this code just existing here. Could we extract all the code you inserted out into a function (hook)? I think it only needs to return `nodesInfo` and the current connection state. Feel free to keep it in this file, I just don't want to make `BackendProvider` this complex.

src/renderer/contexts/BackendContext.tsx
```diff
@@ -120,19 +237,39 @@ export const BackendProvider = memo(
...
+            schemata: nodesInfo?.schemata ?? new SchemaMap([]),
             schemaInputs,
             pythonInfo,
-            categories,
-            categoriesMissingNodes,
-            functionDefinitions,
+            categories: nodesInfo?.categories ?? [],
+            categoriesMissingNodes: nodesInfo?.categoriesMissingNodes ?? [],
+            functionDefinitions:
+                nodesInfo?.functionDefinitions ?? new Map<SchemaId, FunctionDefinition>(),
```
RunDevelopment, line 267: You disabled memoizing during loading with those empty arrays and maps. Use `EMPTY_ARRAY` and `EMPTY_MAP` instead. For `SchemaMap` we should probably add something like `SchemaMap.EMPTY`.

src/renderer/contexts/BackendContext.tsx
```diff
@@ -37,25 +46,169 @@ export const BackendContext = createContext<Readonly<BackendContextState>>(
...
+    );
+
+    let connectionState: 'connecting' | 'connected' | 'failed' = 'connecting';
+    if (nodesQuery.status === 'success') {
+        connectionState = 'connected';
+    } else if (nodesQuery.status === 'error') {
+        connectionState = 'failed';
+    }
+
+    return useMemoObject({
```
RunDevelopment, line 201: You immediately destructure this object. No need to memo it.

src/renderer/main.tsx
```diff
@@ -207,18 +64,9 @@ export const Main = memo(({ url }: MainProps) => {
...
-            successEffect: setPythonInfo,
-        }),
-        []
-    );
+    if (connectionState === 'failed') return null;
 
-    if (nodesQuery.isError) return null;
-
-    if (!nodesInfo || !pythonInfo || !ready) {
+    if (connectionState === 'connecting' || schemata.schemata.length === 0 || !ready) {
```
RunDevelopment, line 69: Why `schemata.schemata.length === 0`?

joeyballentine: Is there a better way to check if no nodes are available? If that's not what you mean, I mentioned on discord why I did this.

RunDevelopment: > it's setting the connectionState to 'connected' before everything is actually defined, so like it connects and goes through everything, but the schemamap is still empty

If you mean this, then I have another question: why did you not fix `connectionState` being `connected` before the data is available? This is a bug in `BackendContext` and shouldn't be fixed here with a work around.

joeyballentine: ah ok that makes sense

RunDevelopment (approved):

PR #2010, opened 2023-07-29
Update ONNX + ONNXRuntime + fix mac issues

### LINKED ISSUE #2000: Image upscale. So slooooooooow.
I’m running OS 13.5 Ventura. MacBook Pro 16-inch, 2021, Apple M1 Max, 64GB RAM but Chainner has been going for over 3 hours without producing a single one of the 2 files in the specified folder. The Dock icon suggests its about a 3rd of the way through but after so long I'm beginning to doubt that it's actually doing anything at all. I had such great hopes for this level of automated image manipulation. Am I being too ambitious? Any suggestions?
Screenshot 2023-07-27 at 11 45 08 PM

joeyballentine:
Should close #2000

I updated onnxruntime. Since it supports apple silicon now, we no longer need onnxruntime-silicon.

I also updated onnxoptimizer, which also conveniently supports macos now. This was the main issue preventing #2000 from being able to convert to NCNN.

Also updated onnx. Doesn't get us anything, but an update won't hurt.

### DIFF
--- a/backend/src/nodes/impl/onnx/session.py
+++ b/backend/src/nodes/impl/onnx/session.py
@@ -4,8 +4,6 @@ from weakref import WeakKeyDictionary
 
 import onnxruntime as ort
 
-from system import is_arm_mac
-
 from ...utils.exec_options import ExecutionOptions
 from .model import OnnxModel
 
--- a/backend/src/nodes/impl/onnx/session.py
+++ b/backend/src/nodes/impl/onnx/session.py
@@ -13,12 +13,7 @@ from .model import OnnxModel
 def create_inference_session(
     model: OnnxModel, exec_options: ExecutionOptions
 ) -> ort.InferenceSession:
-    if is_arm_mac:
-        providers = [
-            "CoreMLExecutionProvider",
-            "CPUExecutionProvider",
-        ]
-    elif exec_options.onnx_execution_provider == "TensorrtExecutionProvider":
+    if exec_options.onnx_execution_provider == "TensorrtExecutionProvider":
         providers = [
             (
                 "TensorrtExecutionProvider",
--- a/backend/src/nodes/impl/onnx/session.py
+++ b/backend/src/nodes/impl/onnx/session.py
@@ -4,6 +4,8 @@ from weakref import WeakKeyDictionary
 
 import onnxruntime as ort
 
+from system import is_arm_mac
+
 from ...utils.exec_options import ExecutionOptions
 from .model import OnnxModel
 
@@ -11,7 +13,12 @@ from .model import OnnxModel
 def create_inference_session(
     model: OnnxModel, exec_options: ExecutionOptions
 ) -> ort.InferenceSession:
-    if exec_options.onnx_execution_provider == "TensorrtExecutionProvider":
+    if is_arm_mac:
+        providers = [
+            "CoreMLExecutionProvider",
+            "CPUExecutionProvider",
+        ]
+    elif exec_options.onnx_execution_provider == "TensorrtExecutionProvider":
         providers = [
             (
                 "TensorrtExecutionProvider",
--- a/backend/src/nodes/impl/onnx/utils.py
+++ b/backend/src/nodes/impl/onnx/utils.py
@@ -2,13 +2,9 @@ from __future__ import annotations
 
 from typing import Literal, Tuple
 
+import onnxoptimizer
 import onnxruntime as ort
 from onnx.onnx_pb import ModelProto
-
-try:
-    import onnxoptimizer
-except ImportError:
-    onnxoptimizer = None
 
 OnnxInputShape = Literal["BCHW", "BHWC"]
 
@@ -54,8 +50,9 @@ def safely_optimize_onnx_model(model_proto: ModelProto) -> ModelProto:
     """
     Optimizes the model using onnxoptimizer. If onnxoptimizer is not installed, the model is returned as is.
     """
-
-    if onnxoptimizer is not None:
+    try:
         passes = onnxoptimizer.get_fuse_and_elimination_passes()
         model_proto = onnxoptimizer.optimize(model_proto, passes)
+    except:
+        pass
     return model_proto
--- a/backend/src/nodes/impl/onnx/utils.py
+++ b/backend/src/nodes/impl/onnx/utils.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 from typing import Literal, Tuple
 
-import onnx
 import onnxruntime as ort
+from onnx.onnx_pb import ModelProto
 
 try:
     import onnxoptimizer
@@ -50,7 +50,7 @@ def get_output_shape(
     return parse_onnx_shape(session.get_outputs()[0].shape)
 
 
-def safely_optimize_onnx_model(model_proto: onnx.ModelProto) -> onnx.ModelProto:
+def safely_optimize_onnx_model(model_proto: ModelProto) -> ModelProto:
     """
     Optimizes the model using onnxoptimizer. If onnxoptimizer is not installed, the model is returned as is.
     """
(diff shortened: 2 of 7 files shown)


joeyballentine:
Something that might be an issue is the fact that onnxruntime-silicon is still installed. We might get an import conflict due to it using the same import name as regular onnxruntime. If this does become an issue, we may need to have some kind of conflicting dependency declaration where we can automatically remove conflicts before installing new ones. But I don't wanna worry about that right now.

backend/src/packages/chaiNNer_onnx/__init__.py
```diff
@@ -2,49 +2,26 @@
...
-            import_name="onnxruntime",
-        )
-    elif nvidia_is_available:
+    if nvidia_is_available:
         return Dependency(
             display_name="ONNX Runtime (GPU)",
             pypi_name="onnxruntime-gpu",
-            version="1.13.1",
+            version="1.15.1",
             size_estimate=110 * MB,
```
RunDevelopment, line 13: It grew in size.

```suggestion
            size_estimate=120 * MB,
```

backend/src/packages/chaiNNer_onnx/__init__.py
```diff
@@ -2,49 +2,26 @@
...
             import_name="onnxruntime",
         )
     else:
         return Dependency(
             display_name="ONNX Runtime",
             pypi_name="onnxruntime",
-            version="1.13.1",
+            version="1.15.1",
             size_estimate=5 * MB,
```
RunDevelopment, line 21: Same
```suggestion
            size_estimate=6 * MB,
```

joeyballentine:
Updated to theoretically use the CoreMLExecutionProvider. No clue if that works, but according to #2013 it should

backend/src/nodes/impl/onnx/session.py
```diff
@@ -4,14 +4,21 @@
...
 
 def create_inference_session(
     model: OnnxModel, exec_options: ExecutionOptions
 ) -> ort.InferenceSession:
-    if exec_options.onnx_execution_provider == "TensorrtExecutionProvider":
+    if is_arm_mac:
+        providers = [
+            "CoreMLExecutionProvider",
+            "CPUExecutionProvider",
+        ]
```
RunDevelopment, line 20: So we're just going to ignore execution options?

backend/src/packages/chaiNNer_onnx/__init__.py
```diff
@@ -53,10 +39,15 @@ def get_onnx_optimizer():
         Dependency(
             display_name="ONNX",
             pypi_name="onnx",
-            version="1.13.0",
+            version="1.14.0",
             size_estimate=12 * MB,
         ),
-        *get_onnx_optimizer(),
+        Dependency(
+            display_name="ONNX Optimizer",
```
RunDevelopment, line 46: Does it support ARM mac now?

backend/src/nodes/impl/onnx/session.py
joeyballentine: Oh hmm.. I guess I need to add it as an option there

backend/src/packages/chaiNNer_onnx/__init__.py
joeyballentine: Yes, I think. Unless I misread the wheel list

joeyballentine: Confirmed it has universal wheels now

backend/src/nodes/impl/onnx/utils.py
```diff
@@ -2,8 +2,8 @@
 
 from typing import Literal, Tuple
 
-import onnx
 import onnxruntime as ort
+from onnx.onnx_pb import ModelProto
 
 try:
     import onnxoptimizer
```
RunDevelopment, line 9: Do we want to keep this import as optional?

Also, we should definitely keep `safely_optimize_onnx_model`. It nicely abstracts away the dependency.

stonerl:
I tested the PR but at them moment it still falls back to the CPU. I think this is because no GPU can be selected.

CleanShot 2023-07-30 at 14 47 46@2x


For reference, in the NCNN settings tap my GPU is selected, and it uses it for processing.

CleanShot 2023-07-30 at 14 48 15@2x

src/renderer/components/SettingsModal.tsx
```diff
@@ -417,6 +417,7 @@ const PythonSettings = memo(() => {
                       },
                   ]
                 : []),
+            ...(isArmMac ? [{ label: 'CoreML', value: 'CoreMLExecutionProvider' }] : []),
```
stonerl, line 420: Wouldn't it be better to call it **GPU** instead of **CoreML**? It would make the UI more consistent. Since in all other parts of the UI hardware acceleration is also referred to as **GPU*

joeyballentine: How about GPU (CoreML)

joeyballentine: I want people to know that it's actual a MacOS GPU inference by using the actual name of the thing

stonerl: > How about GPU (CoreML)

Sounds good to me.

stonerl:
Regarding the missing GPU, it doesn't seem to be the only problem that is causing the CPU fallback.

I changed these lines from:

https://github.com/chaiNNer-org/chaiNNer/blob/4a6c99e819b47a40522e414e8bf26359a18f3b47/src/renderer/components/SettingsModal.tsx#L586-L588

to

```
isDisabled={ncnnGpuList.length === 0}
options={ncnnGpuList.map((gpu, i) => ({
```

I can now select the GPU, but it still falls back to the CPU.

joeyballentine:
> I tested the PR but at them moment it still falls back to the CPU. I think this is because no GPU can be selected.

@stonerl

Where it's actually initializing the coremlexecutionprovider, it just ignores the GPU selector entirely as there is no configuration for device id. If it's not working at all, idk what else I can do since onnxruntime-silicon doesn't give any documentation for how to use this feature.

joeyballentine:
if you make a little python script that just runs this, what does it output?

```py
import onnxruntime as ort
print(ort.get_available_providers())
```

joeyballentine:
And if that doesn't work, maybe we could use this? https://pypi.org/project/onnxruntime-coreml/#description

stonerl:
> if you make a little python script that just runs this, what does it output?
> 
> ```python
> import onnxruntime as ort
> print(ort.get_available_providers())
> ```

```['CoreMLExecutionProvider', 'CPUExecutionProvider']```

**Comment by joeyballentine** [comment]
hmmmm well IMO it definitely should be working then. can you make sure that you don't have the regular onnxruntime installed alongside onnxruntime-silicon? (I don't think that's the issue but I just wanna be sure. i don't think you'd get the CoreMLExecutionProvider shown in that output if it was that issue).

I'm not really sure what else to have you try... I can't debug this myself since I don't have an M1 device. Maybe if you modify the code to not put the CPUExecutionProvider at the end of the list it will at least give you some kind of helpful error?

**Comment by stonerl** [comment]
Hm, did some digging. Could it be that to make use of CoreML some Arguments need to passed?

https://onnxruntime.ai/docs/execution-providers/CoreML-ExecutionProvider#usage

**Comment by joeyballentine** [comment]
Could be. Do you have the ability to try all of those options out and see if anything works?

**Comment by stonerl** [comment]
So I spent some time tinkering around. I think the reason CoreML isn't enabled by default on upstream is simply because they currently haven't implemented all the functions.

A specific model for upscaling I used triggers the following error:

```
[W:onnxruntime:, helper.cc:66 IsInputSupported] Dynamic shape is not supported for now, for input
```

Which means it falls back to the `CPUExecutionProvider`.

As you suggested earlier, the best one could do at them moment is to switch from `onnxruntime-silicon` to `onnxruntime` and wait for upstream to officially enable `CoreMLExecutionProvider`.

It seems that `CoreML` in `onnxruntime-silicon` is only suitable for very specific use cases right now.

RunDevelopment (approved):

PR #2012, opened 2023-07-29
Use a later version of pytorch that supports MPS for Arm MacOS

joeyballentine:


backend/src/packages/chaiNNer_pytorch/__init__.py
```diff
@@ -4,11 +4,28 @@
...
 def get_pytorch():
+    # 1.13.1 can take advantage of MPS
+    if is_arm_mac:
+        return [
+            Dependency(
+                display_name="PyTorch",
+                pypi_name="torch",
+                version="1.13.1",
+                size_estimate=140 * MB,
+            ),
```
RunDevelopment, line 21: No special link this time?

joeyballentine: no, because its on pypi. only the cuda version needs the findlink

RunDevelopment (approved):

Issue #1752, opened 2023-04-19
Error report: Object has been destroyed


bedalton:
```
TypeError: Object has been destroyed

    at click (/Applications/chaiNNer.app/Contents/Resources/app/.webpack/main/index.js:2:498857)
    at MenuItem.click (node:electron/js2c/browser_init:2:30166)
    at a._executeCommand (node:electron/js2c/browser_init:2:35562)
```

ChaiNNer: chaiNNer 0.18.6 (and I believe all versions I have used since starting using the app a month or two ago)
OS: macOS 10.18

Not sure if this is a duplicate of #1665, #1691 , and #1700 and/or if this was also fixed by #1723, but on macOS Ventura, (and on the version I was on before, I think Monterey)
If you:
1. Close the main window
2. Don't quit the app (as it keeps running and does not quit automatically)
3. Click on File->New for a new window

The app shows the error with the options: Report, Exit and Ignore; No option opens a new window

joeyballentine:
Ah! Thank you for providing steps to recreate it, that totally makes sense for what's happening. 

So on MacOS, you can close the main window but it still runs? I didn't know it would do that. Are you able to open the window again after you close it?

PR #2018, opened 2023-07-31
Update electron to v25

joeyballentine:
Not planning on merging until after the next release.

I'm hoping this fixes the #1752 issue. I'm assuming that problem is electron related and that a fix has since been included. No guarantees though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,7 +92,7 @@
         "copy-webpack-plugin": "^11.0.0",
         "cross-env": "^7.0.3",
         "css-loader": "^6.7.1",
-        "electron": "^23.0.0",
+        "electron": "^25.3.2",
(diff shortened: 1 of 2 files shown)


RunDevelopment:
Why did you only update the infrastructure but not the actual `electron` package?

joeyballentine:
> Why did you only update the infrastructure but not the actual `electron` package?

Uhhhhh I guess I overwrote my package.json.... the first commit of this PR was just the electron update, then I realized i didn't save my changes for the infrastructure stuff, but then obviously I didn't realize that it didn't have the electron change in the opened package.json when I saved it. Oops

RunDevelopment (dismissed):
Looks good now. Feel free to merge after the next release.

PR #2023, opened 2023-07-31
use traffic-light colors for CPU/RAM

stonerl:


joeyballentine:
Ooo interesting. As much as I like the color blue, this change feels like it makes a lot of sense. 👍

src/renderer/components/SystemStats.tsx
```diff
@@ -28,27 +28,38 @@ export const SystemStats = memo(() => {
...
-                            </CircularProgress>
-                        </Box>
-                    </Tooltip>
-                ))}
+                data.map((usage) => {
+                    let color;
+                    if (usage.percent < 50) {
+                        color = 'green.400';
+                    } else if (usage.percent < 90) {
+                        color = 'orange.400';
```
theflyingzamboni, line 36: Request to change this to a yellow rather than an orange. I checked color-blindness filters, and while green.400 seems to have plenty of blue to stand out, the orange and red end up overly similar for certain types of colorblindness.

src/renderer/components/SystemStats.tsx
```diff
@@ -28,27 +28,38 @@ export const SystemStats = memo(() => {
...
-                                value={usage.percent}
-                            >
-                                <CircularProgressLabel>{usage.label}</CircularProgressLabel>
-                            </CircularProgress>
-                        </Box>
-                    </Tooltip>
-                ))}
+                data.map((usage) => {
+                    let color;
+                    if (usage.percent < 50) {
```
RunDevelopment, line 33: Isn't 50% a bit low for what is pretty much a warning color? E.g. my RAM typically hovers around 50%, but I still have 16GB of RAM left, so there's no issue at all. 

Maybe we could raise it to 75%? That feels more like something that would warrant getting the user's attention.

stonerl: No problem, I will raise it to 75%.

joeyballentine (approved):
Thanks again for this!

PR #2027, opened 2023-08-01
building x64 and arm64 ZIP/DMG for macOS

stonerl:
Building a universal app is currently only possible on Apple Silicon Macs. But creating a separate x64 and arm64 on Intel Macs should work.

make-mac-arm64 creates a dmg with arm64 binaries namend: chaiNNer-version-arm64.dmg
make-mac-x64 creates a dmg with x64 binaries namend: chaiNNer-version-x64.dmg


package.json
```diff
@@ -12,10 +12,15 @@
...
+    "make-mac-x64": "cross-env NODE_ENV=production electron-forge make --arch=x64",
     "make-linux-zip": "cross-env NODE_ENV=production electron-forge make --targets @electron-forge/maker-zip --platform linux",
     "make-win-zip": "cross-env NODE_ENV=production electron-forge make --targets @electron-forge/maker-zip --platform win32",
-    "make-mac-zip": "cross-env NODE_ENV=production electron-forge make --targets @electron-forge/maker-zip --platform darwin",
-    "publish": "cross-env NODE_ENV=production electron-forge publish",
+    "make-mac-arm64-zip": "cross-env NODE_ENV=production electron-forge make --targets @electron-forge/maker-zip --arch arm64",
+    "make-mac-x64-zip": "cross-env NODE_ENV=production electron-forge make --targets @electron-forge/maker-zip --arch x64",
+    "make-mac-all": "npm run make-mac-arm64 && npm run make-mac-x64",
+    "publish-mac": "npm run make-mac-all && npm run publish",
+    "publish": "cross-env NODE_ENV=production electron-forge publish --dry-run",
```
joeyballentine, line 23:
```suggestion
    "publish": "cross-env NODE_ENV=production electron-forge publish",
```

forge.config.js
```diff
@@ -65,7 +65,6 @@ const config = {
             name: '@electron-forge/maker-dmg',
             config: {
                 format: 'ULFO',
-                name: 'chaiNNer',
```
joeyballentine, line 68: is there a reason you removed this?

stonerl: Yes, when the name is set to chaiNNEr the resulting DMG is always called `chaiNNer.dmg`, regardless of the architecture it is built for.

So building arm64 creates a DMG called `chaiNNEr.dmg`. When the x64 version is built, it also creates a DMG called `chaiNNEr.dmg`. Which basically overwrites the arm64 build when running `npm run make-mac-all`.

By not providing a name, the DMGs are named automatically: `chaiNNer-<version>-<architecture>.dmg`. 
Example for the current version:
```
chaiNNer-0.19.0-arm64.dmg
chaiNNer-0.19.0-x64.dmg
```

package.json
stonerl: Ups, that was an oversight.

forge.config.js
joeyballentine: Oh cool

joeyballentine (approved):
I'm trusting that this will work. Thanks again

RunDevelopment:
> I'm trusting that this will work.

Never again. 

Soooooo, this PR broke our release workflow in 2 ways:

1. The bash script broke windows containers. Fixed in #2056.
2. The macos build is broken. See https://github.com/chaiNNer-org/chaiNNer/actions/runs/5765328561/job/15631071808#step:5:122.

stonerl:
Dang. Sorry for that. I'll have a look and do a test-workflow on my fork.

RunDevelopment:
Don't worry about it. It's not like we tested this beforehand... Anyway, I reverted this PR in #2059 for now. 

To prevent this from happening again going forward, I think that we should avoid having multiple `publish` commands. The platform-dependent stuff is hard enough to test already, so we should keep it to a minimum. Do you think it would possible to integrate the functionality you split across multiple commands into forge? Ideally `electron-forge make --targets @electron-forge/maker-zip --platform darwin` would just build for all mac archs. Maybe that's just wishful thinking on my part though.

stonerl:
I'll take a look. Maybe it can be integrated into forge. If I find a solution, you'll get a PR.

RunDevelopment:
Thank you!

PR #2030, opened 2023-08-01
increase failure count to 1

stonerl:
This should make sure that the Critical Error windows not appear during dependencies installation


RunDevelopment:
> during dependencies installation

When exactly is that? Did you install deps in the dependency manager?

stonerl:
@RunDevelopment I made a short screencast. This error appears on all installation and removal processes, for all dependencies.

CleanShot 2023-08-02 at 15 00 04

stonerl:
And this is how it looks with the change I made:
CleanShot 2023-08-02 at 15 08 03

RunDevelopment:
Interesting. I cannot reproduce this. Does it log any errors? I don't see any reason why fetching `/nodes` should fail here.

Also, sorry for delaying this PR. Since increasing the allowed number of failures would hide the underlying issue, I want to figure it out and fix it first. We will merge your PR later anyway, because it's just a good idea to allow one failure.

stonerl:
> Also, sorry for delaying this PR.

No worries.

I'll open an issue and report the log there.

joeyballentine (approved):

PR #2031, opened 2023-08-01
use native clipboard on macos

### LINKED ISSUE #958: Pasteboard has no arm64 wheels
We should consider removing copy to clipboard support for M1 MacBooks. It tries to build from source and crashes chaiNNer

### LINKED ISSUE #1590: (bug) MacOS copy/paste/duplicate happens twice with keyboard shortcuts
Mentioned by a user in the discord. The events fire twice on MacOS

stonerl:
On macOS we need the native Edit menu to make all keyboard shortcuts accessable.

The shortcuts can copy/cut/paste text and nodes. But using entries from the menu itself only interacts with text.

Therefore on macOS a new menu is added: "Nodes"

Menu structure:

```
*Nodes
|
|- Undo
|- Redo
|--------
|- Cut
|- Copy
|- Paste
|--------
|- Duplicate
|- Duplicate with Connections
```
This menus sole purpose is to give the user a menu to interact with Nodes.

closes #1590

### DIFF
--- a/src/main/gui/menu.ts
+++ b/src/main/gui/menu.ts
@@ -139,6 +139,22 @@ export const setMainMenu = ({ mainWindow, menuData, enabled = false }: MainMenuA
             ? [
                   { role: 'editMenu' },
                   {
+                      label: 'View',
+                      submenu: [
+                          { role: 'reload', enabled },
+                          { role: 'forceReload', enabled },
+                          { type: 'separator' },
+                          { role: 'resetZoom', enabled },
+                          { role: 'zoomIn', enabled },
+                          { role: 'zoomOut', enabled },
+                          { type: 'separator' },
+                          { role: 'togglefullscreen' },
+                          ...(!app.isPackaged
+                              ? [{ type: 'separator' }, { role: 'toggleDevTools' }]
+                              : []),
+                      ],
+                  },
+                  {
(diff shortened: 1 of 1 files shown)


stonerl:
Just to give you a comparison.

## New implementation
CleanShot 2023-08-01 at 21 47 29

## Current implementation
CleanShot 2023-08-01 at 21 52 46

joeyballentine:
FYI this does not close #958. That issue is with the copy to clipboard node, not frontend copy/paste. That issue will be resolved when we switch from pasteboard to a custom rust binding

stonerl:
Ah, I see. Removed the issue from the message.

So I better re-add pasteboard for the moment?

joeyballentine:
yes, you should not remove pasteboard in this.

stonerl:
Okay, reverted the changes. Pasteboard is back in.

stonerl:
Here is the new menu on macOS in action.

CleanShot 2023-08-02 at 11 33 30

RunDevelopment:
> the native Edit menu

I never used macOS, so I had no idea such a thing even exists. Thank you for fixing this!

RunDevelopment (approved):

RunDevelopment:
I can't test this, but the code looks good!

PR #2033, opened 2023-08-02
add .nvmrc and .envrc

stonerl:
When nvm and direnv are installed, cd into the project folder loads the correct node.js version.

This is just a QOL improvement. If a developer has neither of the tools installed, then these files will be ignored.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,8 @@
+NODEVERSION=18
+nvmrc=~/.nvm/nvm.sh
+if [ -e $nvmrc ]; then
+  source $nvmrc
+  nvm use $NODEVERSION
+fi
+
+PATH_add node_modules/.bin


RunDevelopment:
Is there a way to do this without adding 2 files into the root directory? 

Our GH repo page is like our main landing page for new people, and making them scroll down a bunch until they can find actual useful information isn't good. We already have a lot of files in root, so I'd like to avoid adding more. Of course, this is mostly a problem caused by GH's page layout, but it's not like we can change that...

stonerl:
I changed it in a way that it now uses `.envrc` only.

RunDevelopment:
Good enough. We can try to get rid of other config files if we want to.

RunDevelopment (approved):
Thank you @stonerl!

.envrc
```diff
@@ -0,0 +1,8 @@
+NODEVERSION=16
```
joeyballentine, line 1: Shouldn't we default to 18? @RunDevelopment

RunDevelopment: Probably. We only support >=16.14 anyway.

joeyballentine, line 1:
```suggestion
NODEVERSION=18
```

stonerl: Done. I used version `16`  because that is what you currently use in the workflow.

https://github.com/chaiNNer-org/chaiNNer/blob/877d03249e1a9a1179b5cf0fd8e4557412a223c1/.github/workflows/release.yml#L33

RunDevelopment (approved):

PR #2036, opened 2023-08-02
Improved header layout

RunDevelopment:
This PR improves the header layout to work better with small window sizes. The entire header is now useable on all allowed window sizes.

Changes:
- Hide "chaiNNer" and "Alpha" for small windows. The break point for this is 992px, so our name will be visible even on half-screen-sized windows on a 1920x1080 FullHD screen.
- Use flex instead of grid to center the execution buttons.
- Removed min width/height from main. This has nothing to do with the header, but it fixes the node selector moving slightly off-screen for small window sizes.

image
image
image
image

### DIFF
--- a/src/renderer/components/Header/AppInfo.tsx
+++ b/src/renderer/components/Header/AppInfo.tsx
@@ -6,7 +6,6 @@ import {
     AlertDialogFooter,
     AlertDialogHeader,
     AlertDialogOverlay,
-    Box,
     Button,
     HStack,
     Heading,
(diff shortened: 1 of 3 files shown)


joeyballentine:
> Use flex instead of grid to center the execution buttons.

No. This now means the buttons aren't actually centered, but rather they're just centered between the stuff to their side. I switched to grid to make them actually in the center

RunDevelopment:
If we use grid, then there is no way to use the available space on small screens. Perfectly centered < useable UI.

joeyballentine:
Ok fine

PR #2042, opened 2023-08-02
macOS Enhancements

mafiosnik777:
- Changed macOS .icns to make it fit in better on newer macOS versions

- Fixed "Error: Cannot find module 'appdmg'" when building on macOS (Added appdmg package to devDependencies)
Note: appdmg doesn't build on node.js >=16, because something got deprecated in newer node.js versions, hence gyp fails when trying to install it

- Fixed the lowercase executable name on macOS and Windows, Linux doesn't pass with uppercase letters in executableName


joeyballentine:
I think you might need to move appdmg to optionalDependencies

joeyballentine (approved):
🤝

PR #2043, opened 2023-08-02
Delay `/nodes` during restarts

### LINKED ISSUE #2038: Critical error when un-/installing dependencies
**Information:**

- Chainner version: [0.19.0]
- OS: [macOS 13.5]

**Description**
When installing or uninstalling dependencies, a critical error appears and disappears after a short while. Workaround in #2030

CleanShot 2023-08-02 at 15 00 04

**Logs**
[localhost-1690992276719.log](https://github.com/chaiNNer-org/chaiNNer/files/12242258/localhost-1690992276719.log)

RunDevelopment:
This hopefully fixes #2038. 

The idea is to delay the actual fetch until the backend is no longer restarting. Spin to win, basically.

@stonerl Could you please test whether this fixes the issue? I can't reproduce the bug locally, so I can't test it myself.

### DIFF
--- a/src/renderer/contexts/BackendContext.tsx
+++ b/src/renderer/contexts/BackendContext.tsx
@@ -1,6 +1,14 @@
 import { Scope } from '@chainner/navi';
 import isDeepEqual from 'fast-deep-equal';
-import React, { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+    MutableRefObject,
+    memo,
+    useCallback,
+    useEffect,
+    useMemo,
+    useRef,
+    useState,
+} from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


joeyballentine:
Should this have a max waiting time in case something goes wrong?

RunDevelopment:
I was thinking about that too. I essentially just added the next deadlock. If anything goes wrong with `restartingRef`, we're in deep shit. Buuuut pytorch can take an hour to install, so a timeout won't help much. We'll just have to be careful with `restartingRef`. To hopefully prevent future bugs, I moved the `restartingRef.current = false` into a `finally` block.

joeyballentine:
sounds good

joeyballentine (approved):

stonerl:
@RunDevelopment 

Sadly, that didn't fix it.

PR #2044, opened 2023-08-03
Define output previewer node in backend

joeyballentine:
Now the backend defines what node the preview button should make for its preview. An output class can define a previewer_schema_id, which the frontend will then use.

The nice thing about this, is that the preview button is now no longer limited to just images. I don't think this is useful for any other type of data chaiNNer currently supports at the moment (maybe for colors?), but in the future it could be used to add preview buttons for other data types.

Another idea I had was to be able to define an array of buttons, each with a schema id and an icon, so you could theoretically add shortcut buttons for certain outputs to do things -- and it could even be defined per-output on a node (for example making a text append button or something that could go on text that typically would get appended onto). Not sure if we should do that, but its just an idea I had.


RunDevelopment:
> preview buttons for other data types

Should we have preview nodes for each data type or just one preview node for all in the future? I personally would prefer having a single node, just because it's easier to use.

joeyballentine:
I suppose we could do that, but then you'd have the node changing size based on what it needs to preview (think audio vs images)

PR #2045, opened 2023-08-03
Make linked-inputs validation input-kind agnostic

joeyballentine:


RunDevelopment:
1. `inputs` might contain nested groups. You have to check that this isn't the case.
2. The placeholders of text inputs might be different.
3. You forgot to update the type of linked inputs in `DeclaredGroupInputs`.

Here's how I would fix 1 and 2:
```ts
        const [ref] = inputs;
        if (ref.kind === 'group') return `Expected linked inputs to not contain groups`;
        if (!allInputsOfKind(inputs, ref.kind))
            return `Expected all inputs to be ${ref.kind} inputs`;

        type Keys<T> = T extends unknown ? keyof T : never;
        const ignoreKeys = new Set<Keys<Input>>(['id', 'label', 'description', 'placeholder']);
        for (const i of inputs) {
            for (const [key, value] of Object.entries(i)) {
                // eslint-disable-next-line no-continue
                if (ignoreKeys.has(key as Keys<Input>)) continue;
                if (JSON.stringify(value) !== JSON.stringify(ref[key as never]))
                    return `Expected all inputs to have the same ${key} value`;
            }
        }
```

src/common/group-inputs.ts
```diff
@@ -35,7 +34,7 @@ type DeclaredGroupInputs = InputGuarantees<{
     'ncnn-file-inputs': readonly [FileInput, FileInput];
     'optional-list': readonly [InputItem, ...InputItem[]];
     seed: readonly [NumberInput];
-    'linked-inputs': readonly [SliderInput, SliderInput, ...SliderInput[]];
+    'linked-inputs': readonly [InputItem, InputItem, ...InputItem[]];
```
RunDevelopment, line 37:
```suggestion
    'linked-inputs': readonly [Input, Input, ...Input[]];
```

RunDevelopment (approved):

PR #2047, opened 2023-08-03
Update descriptions for dependencies on a system basis

stonerl:
The description for the package is now shown on a system basis, e.g., a Mac user with Apple Silicon will not see AMD or NVIDIA info.

The Package size is moved out of the button and shown next to the package in a bubble.

Install/Upgrade/Uninstall buttons now have the same width and are aligned with Show console.

### DIFF
--- a/backend/src/packages/chaiNNer_onnx/__init__.py
+++ b/backend/src/packages/chaiNNer_onnx/__init__.py
@@ -2,6 +2,20 @@ from sanic.log import logger
 
 from api import KB, MB, Dependency, add_package
 from gpu import nvidia_is_available
+from system import is_arm_mac
+
+general = "ONNX uses .onnx models to upscale images."
+conversion = "It also helps to convert between PyTorch and NCNN."
+
+if is_arm_mac:
+    package_description = f"{general} {conversion} However, it does not support CoreML."
+    inst_hint = general
+else:
+    package_description = (
+        f"{general} {conversion} It is fastest when CUDA is supported. If TensorRT is"
+        " installed on the system, it can also be configured to use that."
+    )
+    inst_hint = f"{general} It does not support AMD GPUs."
 
 
 def get_onnx_runtime():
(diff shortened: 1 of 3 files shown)


stonerl:
CleanShot 2023-08-03 at 13 56 13@2x

backend/src/packages/chaiNNer_onnx/__init__.py
```diff
@@ -2,6 +2,23 @@
...
+    inst_hint = "ONNX uses .onnx models to upscale images."
+else:
+    package_description = (
+        "ONNX uses .onnx models to upscale images. It also helps to convert between"
+        " PyTorch and NCNN. It is fastest when CUDA is supported. If TensorRT is"
+        " installed on the system, it can also be configured to use that."
+    )
+    inst_hint = (
+        "ONNX uses .onnx models to upscale images. It does not support AMD GPUs."
+    )
```
RunDevelopment, line 21: There is a lot of redundancy here. Could we maybe extract the bits that are same out into variables? Same for PyTorch and NCNN.

RunDevelopment (commented):
I like the new descriptions. Thanks @stonerl!

However, I'm not sure how I feel about the new install size tag. With that bg color and font size, it feels weirdly attention grabbing. What makes it so important to deserve that kind of highlighting?

backend/src/packages/chaiNNer_onnx/__init__.py
stonerl: I will remove the redundancy.

stonerl:
> However, I'm not sure how I feel about the new install size tag. With that bg color and font size, it feels weirdly attention grabbing. 

How about this size:

CleanShot 2023-08-03 at 14 10 07@2x

RunDevelopment:
IMO, it's not really that much better... The issue is that we use this general design for buttons everywhere, but the size tag isn't a button. It communicates the wrong thing. Right now, that tag looks very clickable to me.

There's another issue as well. Moving the size out of the Install/Update button and over to the label of the package, makes it seem like the size refers to the package as a whole. In your screenshot, "PyTorch (4 packages) (1.3 MB)" could easily be interpreted as the 4 pytorch packages having an install size of 1.3 MB. The only thing associating the size tag with the Install/Update button is the background color, and you already know how I feel about that...

For now, I would suggest moving the dep manager design changes into a separate PR and iterate on the design there. There's no reason from those changes to hold back the system-specific descriptions.

stonerl:
I removed the design changes and going to move them into another branch.

RunDevelopment (approved):

PR #2049, opened 2023-08-03
More macOS bugfixes/enhancements

mafiosnik777:
After closing the window on macOS, the app just stays in the dock & you can click it but it won't reopen. It will call `app.quit()` now once all window instances are closed.

gif

SwinIR throws an error with the mps backend, because the aten::roll operator isn't implemented for mps yet. 
I've set the environment variable to fall back to CPU on unsupported operators. This fixes the error & allows running SwinIR on the GPU, instead of fully on the CPU and brings a performance improvement of 3x.

image

### DIFF
--- a/backend/src/server.py
+++ b/backend/src/server.py
@@ -52,10 +52,6 @@ from response import (
 )
 from server_config import ServerConfig
 
-# enable mps fallback on apple silicon
-if is_arm_mac:
-    os.environ['PYTORCH_ENABLE_MPS_FALLBACK'] = '1'
-
 class AppContext:
     def __init__(self):
         self.config: ServerConfig = None  # type: ignore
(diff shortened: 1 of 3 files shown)


RunDevelopment:
Joey already merged, but I'll tell you about anyway. The `app.quit()` was probably caused by this https://github.com/chaiNNer-org/chaiNNer/blob/fbd9480f4bbd8a8e87709cee31c4fe0c0e8851d0/src/main/gui/create.ts#L82

mafiosnik777:
@RunDevelopment Yeah, didn't see. Not that familiar with the codebase yet.
Nonetheless still undesired behavior, because if you leave them in the dock, it should open a new instance on clicking. Which it didn't do, it just sat there as a dangling process.

Someone should remove L81-86 from `create.ts`.

RunDevelopment:
> Someone should remove L81-86 from create.ts.

Will do.

PR #2053, opened 2023-08-03
Features

RunDevelopment:
This PR adds support for features. Features are a new kind of requirement that is assigned per-node. While they are generally intended to check for external connections/installations, they can also be used as a feature gate. E.g. I intend to use a feature to disable the Copy To Clipboard node on systems that we do not support a clipboard for.

Right now, there is only one feature for the SD web UI.

image

Changes:

- Added features to `api.py`. Features are added to packages using the `package.add_feature(...)` method.
- All packages now have an internal id. This should probably be the package name in the future, but I just hard-coded them for now.
- Added the `web_ui` feature to `chaiNNer_external`. This feature checks whether we can connect to the SD web UI, and sets the current API connection is it finds one.
- Moved all code related to external SD into the `chaiNNer_external` package. I also significantly changed the API of this code. The WebUI API is now accessed via the `get_api()` function. This makes the whole connection management a lot cleaner.
- Changed Upscalers enum. The SD Upscale node used to dynamically fetch its upscaler options from the web UI. Since the web UI connection is now optional, this is no longer possible. Instead, I simply hard-coded the upscalers my installation of the web UI supported. I have a bare-bones stock installation, so there shouldn't be anything in there that other installations don't support.
- Since it bugged me a lot, I renamed some API endpoints. `/clearcache/individual` -> `/clear-cache/individual`, `/listgpus/ncnn` -> `/list-gpus/ncnn`, `/listgpus/nvidia` -> `/list-gpus/nvidia`.
- Removed the `/dependencies` endpoint. I replaced it with `/packages` and `/installed-dependencies` endpoints. These 2 endpoint represent the data we have a backend more closely, and are a better abstraction boundary IMO. `/packages` was also necessary to get the features each package holds.
- Added a `/features` endpoint. This endpoint returns the current state (enabled/disabled) for all features. This doesn't include information about the features themselves (e.g. name, description).
- `Backend#fetchJson` will now automatically detect server error (500), and convert them into exceptions. This makes error handling a lot easier.
- Removed SD web ui check from frontend. The backend now does this check via its `web_ui` feature.
- Removed `runPipList` it's not used anymore anyway.
- Unified Markdown renderer. I added a new `Markdown` component that handles rendering Markdown everywhere in the UI this ensures the MD is rendered consistently everywhere. I did this in this PR, because I needed it to render feature descriptions.


RunDevelopment:
Ah, I forgot to write documentation.

joeyballentine:
> The SD Upscale node used to dynamically fetch its upscaler options from the web UI. Since the web UI connection is now optional, this is no longer possible. Instead, I simply hard-coded the upscalers my installation of the web UI supported. I have a bare-bones stock installation, so there shouldn't be anything in there that other installations don't support.

This is not good IMO. You've essentially just locked people into only using those stock models for this. IDK if anyone actually uses this feature, but if they do I'm sure they won't be happy about this loss of functionality.

Not only that, but if we have more external connections in the future, dynamically fetched dropdowns could be used more or even 100% needed. IMO this is just something we need to make work for these things to be properly supported.

RunDevelopment:
> You've essentially just locked people into only using those stock models for this.

Well, yes. The prior solution was dependent upon there being a connection that exists when the backend starts and persists throughout the lifetime of the backend. The entire idea of features is that this connection might become (un)available at any time.

If you want to dynamically change the inputs of a node depending on the features currently available. Since reloading nodes isn't something we currently support, we would either have to add support for that or add support for nodes to change the inputs on the fly. Of course, since feature checking is asynchronous, we would also need a mechanism to notify the frontend about node changes. We will need to change a lot more than I already did.
Of course, that's just *how* we could implement dynamic inputs. Whether dynamic inputs are a good idea in the first place is a separate question.

That's why I limited it to stock upscalers for now.

joeyballentine (approved):
I don't see anything wrong with any of the code. Going to merge it so we can use it for a while and see if we notice anything weird from it.

I still think a good follow-up to this would be to have API-powered dropdowns somehow. We can brainstorm that over discord but I think I have an idea for how to achieve it.

PR #2054, opened 2023-08-03
some UI fine tuning for the dependency manager

stonerl:
The Package size is moved out of the button and shown next to the package in a bubble.

Install/Upgrade/Uninstall buttons now have the same width and are aligned with Show console.

### DIFF
--- a/src/renderer/contexts/DependencyContext.tsx
+++ b/src/renderer/contexts/DependencyContext.tsx
@@ -5,6 +5,7 @@ import {
     AccordionIcon,
     AccordionItem,
     AccordionPanel,
+    Box,
     Button,
     Center,
     Collapse,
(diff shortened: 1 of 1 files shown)


stonerl:
Here is a screenshot with the changes:

CleanShot 2023-08-03 at 18 22 27@2x

stonerl:
Just to give you an explanation as to why I changed the design.

I think it is good UI practice to have vertically aligned buttons with the same size and not change its size. That is why I moved the download size out of the button.

Now it shows the number of total packages that are installed or the number of packages to be installed and the size of these packages. When a dependency has outdated packages, it shows the number of them and the size the update requires. After updating them, it then shows the total number of installed packages again.

But nothing is set in stone here @RunDevelopment already gave me valuable feedback on my first attempts.

stonerl:
@RunDevelopment forgot about your remark regarding that the tags looking clickable. I could change the color, of course. So that it stand out from the button.

stonerl:
For example:

CleanShot 2023-08-03 at 18 55 00@2x

theflyingzamboni:
I actually find this confusing. Putting the size next to the package name implies that it's the size of the installed package rather than the download. When I first looked at this, I thought NCNN didn't have a listed size because it _wasn't_ installed. It was only after looking at the buttons that I realized it was the opposite. I think keeping the size with the install/update buttons is more clear as to what the number represents.

stonerl:
I actually need to wrap my head around this.

From my point of view this needs to be changed but I'm on board with your concerns. @theflyingzamboni

joeyballentine:
Thanks for the PR but I think a better thing would be to completely redo the way dependencies and the dependency manager work. There's a specific way I want to redo it, but I'm not sure I'll ever actually get around to doing it.

PR #2055, opened 2023-08-04
Option to enable GUI hardware acceleration

stonerl:
Add a settings option in the 'Advanced Tab' to allow hardware acceleration.

### DIFF
--- a/src/main/gui/create.ts
+++ b/src/main/gui/create.ts
@@ -3,6 +3,7 @@ import electronLog from 'electron-log';
 import { log } from '../../common/log';
 import { lazy } from '../../common/util';
 import { OpenArguments } from '../arguments';
+import { settingStorage } from '../setting-storage';
 import { createMainWindow } from './main-window';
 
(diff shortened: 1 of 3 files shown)


joeyballentine:
We actually intentionally removed this. We used to have an option for it, but we found it massively decreases upscaling performance, as well as having the potential to cause visual artifacts for some people (for whatever reason). We figured it wasn't worth it for people to accidentally gimp themselves.

Do you have a good reason for wanting this back?

stonerl:
I guessed that this was the reason.

I added it in case someone has to use the CPU for processing. In that case offloading the GUI to the CPU would give a little more headroom.

I could either change the description or I'll close the PR.

I leave this up to you.

joeyballentine:
Before it was a "disable hardware acceleration". This being an "enable" makes more sense as that way the default is the way it is now. I think just adding a warning about decreasing GPU performance with it on should be enough, and then I'm good with adding this.

stonerl:
Changed the description.

PR #2064, opened 2023-08-07
slight adjustments for light mode

stonerl:
Some screenshots for reference:

CleanShot 2023-08-07 at 12 39 40@2x
CleanShot 2023-08-07 at 12 39 23@2x

### DIFF
--- a/src/renderer/colors.scss
+++ b/src/renderer/colors.scss
@@ -14,8 +14,14 @@
     /* color palette */
     --gray-50: #f7fafc;
     --gray-100: #{$gray-100};
+    --gray-125: #{color.mix($gray-100, $gray-200, 25%)};
+    --gray-150: #{color.mix($gray-100, $gray-200, 50%)};
+    --gray-175: #{color.mix($gray-100, $gray-200, 75%)};
     --gray-200: #{$gray-200};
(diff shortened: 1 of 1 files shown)


stonerl:
should have been a draft

joeyballentine:
A definite improvement over what's there now. I kinda feel like it's just too bright in general though. One thing I've always wanted to try was making the background more or a gray and the nodes lighter rather than the other way around. Would you mind trying that out?

stonerl:
I had to increase brightness to give more contrast to individual elements.

I can try to implement your suggestion. But it could be very difficult since it might sacrifice overall contrast again.

As @RunDevelopment said, since the light theme was an afterthought, it is quite difficult to properly implement it within the current confinements.

In the mid to long run, it would be a good idea to rethink theming and adding more css attributes to have a more nuanced way to theme the app.

I'll set this on my todo list if I find some time.

joeyballentine:
Do you think the changes you have a good enough for now? I'd like to get this in if you think it's ready

stonerl:
> Do you think the changes you have a good enough for now?

Yes, this can go in.

joeyballentine (approved):

PR #2070, opened 2023-08-08
Make ncnn memory budget configurable

JeremyRand:
#1867 didn't support automatic estimation of tile size, which was not great for UX. This PR adds that missing support, by allowing the user to choose a custom memory budget. Choosing a memory budget is likely to be better UX than choosing a tile size (since users are likely to have a better knowledge of how much RAM or VRAM they have, than what tile size will work with their hardware and the model they picked). This should also work fine with Vulkan (and is likely to help with [this UX issue](https://github.com/chaiNNer-org/chaiNNer/blob/main/docs/troubleshooting.md#vkqueuesubmit-error-with-ncnn)), though I wasn't able to test that.

### DIFF
--- a/backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
+++ b/backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
@@ -70,18 +70,33 @@ def upscale_impl(
     net = get_ncnn_net(model, exec_options)
     # Try/except block to catch errors
     try:
-        if use_gpu:
-            vkdev = ncnn.get_gpu_device(exec_options.ncnn_gpu_index)
 
-            def estimate_gpu():
+        def estimate():
+            heap_budget = exec_options.ncnn_budget_limit
+            model_size_estimate = model.model.bin_length
+
+            if use_gpu:
                 if is_mac:
                     # the actual estimate frequently crashes on mac, so we just use 256
                     return MaxTileSize(256)
 
-                heap_budget = vkdev.get_heap_budget() * 1024 * 1024 * 0.8
-                return MaxTileSize(
-                    estimate_tile_size(heap_budget, model.model.bin_length, img, 4)
+                heap_budget = min(
+                    heap_budget, vkdev.get_heap_budget() * 1024 * 1024 * 0.8
                 )
+            else:
+                # Empirically determined
+                model_size_estimate = model_size_estimate * 5 / 3
+                if net.opt.use_winograd_convolution:
+                    model_size_estimate = model_size_estimate * 11 / 5
+                elif net.opt.use_sgemm_convolution:
+                    model_size_estimate = model_size_estimate * 40 / 5
+
+            return MaxTileSize(
+                estimate_tile_size(heap_budget, int(model_size_estimate), img, 4)
+            )
+
+        if use_gpu:
+            vkdev = ncnn.get_gpu_device(exec_options.ncnn_gpu_index)
 
             with ncnn_allocators(vkdev) as (
                 blob_vkallocator,
@@ -94,16 +109,9 @@ def upscale_impl(
                     output_name=output_name,
                     blob_vkallocator=blob_vkallocator,
                     staging_vkallocator=staging_vkallocator,
-                    tiler=parse_tile_size_input(tile_size, estimate_gpu),
+                    tiler=parse_tile_size_input(tile_size, estimate),
                 )
         else:
-
-            def estimate_cpu():
-                # TODO: Improve tile size estimation in CPU mode.
-                raise ValueError(
-                    "Tile size estimation not supported with NCNN CPU inference"
-                )
-
             return ncnn_auto_split(
                 img,
                 net,
@@ -111,7 +119,7 @@ def upscale_impl(
                 output_name=output_name,
                 blob_vkallocator=None,
                 staging_vkallocator=None,
-                tiler=parse_tile_size_input(tile_size, estimate_cpu),
+                tiler=parse_tile_size_input(tile_size, estimate),
             )
     except (RuntimeError, ValueError):
         raise
@@ -125,7 +133,7 @@ def upscale_impl(
     schema_id="chainner:ncnn:upscale_image",
     name="Upscale Image",
     description="Upscale an image with NCNN. Unlike PyTorch, NCNN has GPU support on all devices, assuming your drivers support Vulkan. \
-            Select a manual number of tiles if you are having issues with the automatic mode.",
+            Select a manual number of tiles or set a memory budget limit if you are having issues with the automatic mode.",
     icon="NCNN",
     inputs=[
         ImageInput().with_id(1),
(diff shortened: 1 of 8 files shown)


JeremyRand:
(I'll fix the Pyright error ASAP.)

JeremyRand:
CI is now passing.

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
```diff
@@ -125,7 +133,7 @@ def estimate_cpu():
     schema_id="chainner:ncnn:upscale_image",
     name="Upscale Image",
     description="Upscale an image with NCNN. Unlike PyTorch, NCNN has GPU support on all devices, assuming your drivers support Vulkan. \
-            Select a manual number of tiles if you are having issues with the automatic mode.",
+            Select a manual number of tiles or set a memory budget limit if you are having issues with the automatic mode.",
```
RunDevelopment, line 136: > or set a memory budget limit

1. This doesn't tell users how to do that.
2. This solution is factually incorrect. Setting a memory budget won't help at all in GPU mode.

src/renderer/components/SettingsModal.tsx
```diff
@@ -591,6 +594,28 @@ const PythonSettings = memo(() => {
                             value={ncnnGPU}
                             onChange={setNcnnGPU}
                         />
+                        <SettingsItem
+                            description="Maximum memory to use for NCNN inference."
```
RunDevelopment, line 598: Again, this description is incorrect since the memory budget doesn't apply to GPU mode.

src/renderer/contexts/SettingsContext.tsx
```diff
@@ -46,6 +47,7 @@ export const SettingsProvider = memo(({ children }: React.PropsWithChildren<unkn
     const useIsFp16 = useMemoArray(useLocalStorage('is-fp16', false));
     const usePyTorchGPU = useMemoArray(useLocalStorage('pytorch-gpu', 0));
     const useNcnnGPU = useMemoArray(useLocalStorage('ncnn-gpu', 0));
+    const useNcnnBudgetLimit = useMemoArray(useLocalStorage('ncnn-budget-limit', 1024 ** 5));
```
RunDevelopment, line 50: Why did you make the default 1 Petabyte?

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
JeremyRand: I'm not sure I follow; why would it not help in Vulkan mode?

src/renderer/contexts/SettingsContext.tsx
JeremyRand: The intent was to pick a default that is practically equivalent to "no limit". I find it very unlikely that any usage will go over 1 PiB; 1 TiB is low enough that some users might hit it (some of my test workflows hit 600 GiB or so, and I'm not as much of a power user as some people).

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
RunDevelopment: Sorry, my bad. I didn't see the `heap_budget` in `min(heap_budget, vkdev.get_heap_budget() * 1024 * 1024 * 0.8)`...

Point 1 still stands though.

src/renderer/components/SettingsModal.tsx
RunDevelopment: No, I'm just dumb.

src/renderer/contexts/SettingsContext.tsx
JeremyRand: My workstation has around 300 GiB of RAM, and I typically run chaiNNer in a VM on that workstation that has around 200 GiB of RAM assigned to it. I wouldn't be surprised if there exists some power user out there with 1 TB of RAM, and I'm trying to avoid surprising behavior in such (admittedly rare) situations.

That said -- am I correct in assuming that your concern here is that 1 PiB looks "weird" in the GUI as the default, as opposed to having a concern about the default being "no effective additional limit"? If so, would it be OK to instead modify the backend to special-case a limit of 0 so that 0 means "no additional limit", and make 0 the default?

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
JeremyRand: I don't see instructions on switching between GPU's in the backend metadata (though there is a mention of this in the README), which is why I didn't add similar instructions for this. Would you be favorable to adding a note in the backend docs metadata about making sure the integrated GPU isn't selected, along with a note about setting a memory budget? (Seems to me that documenting both of them in-app would be a UX improvement.)

JeremyRand:
Putting this on hold until #2088 is merged.

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
```diff
@@ -70,18 +70,33 @@ def upscale_impl(
...
                     return MaxTileSize(256)
 
-                heap_budget = vkdev.get_heap_budget() * 1024 * 1024 * 0.8
-                return MaxTileSize(
-                    estimate_tile_size(heap_budget, model.model.bin_length, img, 4)
+                heap_budget = min(
+                    heap_budget, vkdev.get_heap_budget() * 1024 * 1024 * 0.8
                 )
+            else:
+                # Empirically determined
```
theflyingzamboni, line 87: Since you have so much RAM, I'm wondering if you'd be willing to run some tests. I've done some testing on NCNN VRAM estimation in the past (though only with ESRGAN), and I can already tell you that this isn't really correct, and neither is what we already have in place. I've only got 8GB VRAM, so I could never extend my tests far enough to gather fully complete data.

What I can say is that:
1) Model size is not strongly correlative with VRAM usage. This is because a model can be made larger simply by performing more convolutions, which does not matter for VRAM usage because they are being done in sequence. The only thing it definitely correlates with is how much VRAM it takes to store the model itself.
2) Individual weight sizes have a correlation with VRAM usage when running a model.
3) Scale needs to be accounted for, which our estimation does not currently do. This estimation is based around a 4x scale, but an 8x model will blow past it.

I abandoned this back in the day because there seemed to be further factors I couldn't account for with the data I had, but maybe we can finally figure it out. Unfortunately, I seem to have deleted the set of different scale/nf/nb ESRGAN models I generated for these tests. If I can remember how I generated them all, I could send them to you.

theflyingzamboni (changes requested):

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
JeremyRand: @theflyingzamboni I would definitely be interested. Can we split this out into its own issue so that it doesn't get lost once this PR is merged?

JeremyRand:
Superseded by https://github.com/chaiNNer-org/chaiNNer/pull/2351

PR #2071, opened 2023-08-10
Add Quantize to Reference node

joeyballentine:
I'm open to calling this node something else because I don't think it's very descriptive called this.

Anyway, this was an idea I had a while back when someone was trying to apply a color palette to an image but keep the colors as close to the original as possible (such as being in the same spots).

I accomplished this by just doing a basic convolution across the original image, taking the unique colors from the kernel, applying the palette to the equivalent piece of the recipient image, and then creating a result from piecing those together. This restricts the possible colors to only the colors that could possibly be in that part of the image, and as a result, manages to index the image pretty well in a way that keeps it faithful.

image
It's obviously not perfect, but it did what this person wanted more than dither (palette) did. 

The biggest downside is it's pretty slow (On^2 time complexity), so maybe this is another good one to move to rust in the future. Alternatively, something could probably be made using the flatten/reshape stuff like the quantizing function does, but I don't wanna have to figure that out lol


joeyballentine:
Oh yeah, i was going to restrict the reference image type to be a multiple/ratio of the input image, but I couldn't figure that out

joeyballentine:
No. That will apply the palette from the entire reference image to the entire target image. That's not what this does. This restricts the palette to just a small area at a time so there's no change it'll use a color where it's not supposed to.

For example if the top third of an image is green colors, the middle third is reds, and the bottom is blues, it might use the green colors from the top section in the bottom blue section if it decides some of the colors are closer (for example if an upscale has color shift or something). 

That's obviously a bit of an unlikely example, but in something like the image I posted above, you can see how those smooth gradients might end up having different colors from random points of the image, which we want to avoid.

joeyballentine:
Maybe `Quantize to Reference` would be a better name?

RunDevelopment:
I played around with it for a bit, and found an unfortunate issue. Since we only consider a neighborhood of exactly 3x3 pixels with a hard cutoff, the local color palette abruptly changes when we move the kernel. This change can be seen in the final image. Here, I upscaled the MC cobblestone texture 16x:

image
image

So I implemented the following idea to fix this: we add x,y normalized coords to each pixel. This will make far away pixels more distant when calculating the distance, which will prevent the abrupt cutoff.

Here's the result:
image
image


The artifacts are gone, but we lost a lot of detail. The solution for is to increase the kernel size and to decrease the weight of the x,y coords. Then we get this:

image
image

Now that's a good result. No artifacts and good contrast.

The spacial scale also gives us a lot of control over the palette. In the following images, the kernel radius is always 2:
image

Note that scale=0 is the same as not having spacial coords at all.

I'll make PR with my changes.

RunDevelopment:
Is this ready?

joeyballentine:
No, I want to rename it to Quantize to Reference. I can't do that right now though, but I'll do it ASAP

joeyballentine:
Wait, i can abuse reviews

backend/src/packages/chaiNNer_standard/image_filter/quantize/quantize_local.py
```diff
@@ -0,0 +1,139 @@
...
+    quantized_img_flat = flat_palette[closest_palette_idx]
+
+    # Reshape the quantized pixels to the original image shape
+    quantized_img = quantized_img_flat.reshape(image.shape)
+
+    return quantized_img
+
+
+@quantize_group.register(
+    schema_id="chainner:image:quantize_local",
```
joeyballentine, line 39:
```suggestion
    schema_id="chainner:image:quantize_to_referece",
```

backend/src/packages/chaiNNer_standard/image_filter/quantize/quantize_local.py
```diff
@@ -0,0 +1,139 @@
...
+    name="Quantize (Local)",
```
joeyballentine, line 40:
```suggestion
    name="Quantize to Reference",
```

backend/src/packages/chaiNNer_standard/image_filter/quantize/quantize_local.py
```diff
@@ -0,0 +1,139 @@
...
+                height: max(Input0.height, Input1.height),
+                channels: Input0.channels,
+            } & if valid { any } else { never }""",
+            assume_normalized=True,
+        ).with_never_reason(
+            "Target image must be larger than reference image in both dimensions, must have dimensions that are a multiple of each other, and must have the same number of channels."
+        )
+    ],
+)
+def quantize_local_node(
```
joeyballentine, line 96:
```suggestion
def quantize_to_reference_node(
```

joeyballentine:
Ah damn the file name, can't change that via review :(

RunDevelopment:
There ya go

RunDevelopment (approved):

PR #2076, opened 2023-08-11
Add percent input node

adegerard:
**Description**
Add a 'percent' node to have an easy way to modify a percent input.
Easier than using the 'number' node.

example

**Choices and limitations**
- The output value is an integer: I guess it is not necessary to have more precision.
- The output value is not the mathematical value. i.e. for 20% (=0.2), the ouput value is equal to 20.


RunDevelopment (approved):
Nice and simple. Very good! Thank you @adegerard!

Issue #1934, opened 2023-07-10
16 Bits per color support.


maxpiva:
**Motivation**
Currently, saving images only support saving 8 bits per color (24+alpha). 
Png & Tiff supports loading and saving (48+alpha) bit images. 
The current 8-bit per color causes issues mostly on gradients in anime, sawed colors.

**Description**
16-bit per color (R, G, B) support, in loading (not sure) and saving images, and the conversion from/to fp formats.

0x4E69676874466F78:
Related: #819

RunDevelopment:
> The current 8-bit per color causes issues mostly on gradients in anime, sawed colors.

Are you sure this is caused by the 8-bit quantization? If so, then we should probably add an option to enable dithering to Save Image. I added general support for dithering a while back, but we haven't exposed that option yet.

maxpiva:
Yes, is the 8-bit quantization. 
If making the pipe 16-bit is not feasible, dither is the second best. The only tool that was able to do the full 16-bit pipe, is waifu2x-Caffe with custom models. Gradients come out perfect. The other is, the node may or may not support 16 bits. If there are OpenCV limitations in some operations.

IE:
10-Bit Anime -> 16 Bit PNG -> waifu2x-Caffe -> 16 Bit PNG -> 10-Bit Anime.

Issue #1791, opened 2023-05-09
Add more image types for saving and loading


joeyballentine:
People keep requesting support for weird formats. We should see what formats we're able to read and write using opencv and pil, and then just add support for everything we can.

RunDevelopment:
What weird formats do people request? I mostly see EXR, and we can't add support for that (#1628).

joeyballentine:
Someone just requested BMP, and i figure we may as well support at least reading any of those random other formats that pillow and opencv support (if we don't already)

RunDevelopment:
I changed the issue to focus on saving images. We already use PIL and OpenCV to their fullest when loading images.

PR #2077, opened 2023-08-12
Support BMP in Save Image

RunDevelopment:
Progress for #1791.

Changes:
- I refactored the image extensions dropdown to use `EnumInput` instead. This makes image formats type safe, which gives me more confidence to change Save Image.
- Renamed "Image Extension" to "Image Format"
- Added `BMP` to image formats.

image

I also refactored Save Image a little in general. This is preparation for 16bit images (#1934), which I plan to add after this PR.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/io/save_image.py
+++ b/backend/src/packages/chaiNNer_standard/image/io/save_image.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 from enum import Enum
+from typing import Dict, List
 
 import cv2
 import numpy as np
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_standard/image/io/save_image.py
```diff
@@ -35,6 +35,38 @@
...
 
+
+class ImageFormat(Enum):
+    PNG = "png"
+    JPG = "jpg"
+    GIF = "gif"
+    BMP = "bmp"
+    TIFF = "tiff"
+    WEBP = "webp"
+    WEBP_LOSSLESS = "webp-lossless"
```
joeyballentine, line 46: Could we get rid of webp-lossless and instead just add a lossless checkbox that hides the quality slider?

RunDevelopment: We can, but not in this PR.

joeyballentine (approved):

PR #2087, opened 2023-08-13
Add position to blend node

adegerard:
**Description**
- Align the overlay layer according to the choosen position
- Add option to crop the result to the base layer size

example

**Choices and limitations**
- The default settings makes the node behavior the same as previously
- When the user chooses a custom position and to crop to fit the base layer: if the overlay layer position is completely outside the base layer, then the node returns the base image. It also does not display any warning, the user has to understand by himself that he made something '_wrong_'

**Other**
Would you please review the code and give advices to improve the performance?


backend/src/packages/chaiNNer_standard/image_utility/compositing/blend_images.py
```diff
@@ -23,18 +77,116 @@
         ImageInput("Base Layer", channels=[1, 3, 4], allow_colors=True),
         ImageInput("Overlay Layer", channels=[1, 3, 4], allow_colors=True),
         BlendModeDropdown(),
+        if_group(Condition.type(0, "Image") & Condition.type(1, "Image"))(
+            EnumInput(
+                BlendOverlayPosition,
+                label="Overlay position",
+                option_labels=BLEND_OVERLAY_POSITION_LABELS,
+                default_value=BlendOverlayPosition.CENTERED,
```
RunDevelopment, line 85: CI is failing because I changed the name of this parameter to `default` today.

```suggestion
                default=BlendOverlayPosition.CENTERED,
```

backend/src/packages/chaiNNer_standard/image_utility/compositing/blend_images.py
```diff
@@ -23,18 +77,116 @@
...
+                    "Y offset",
+                    controls_step=1,
+                    minimum=None,
+                    maximum=None,
+                    default=0,
+                    unit="px",
+                ),
+            ),
+        ),
+        BoolInput("Crop to fit base layer", default=False),
```
RunDevelopment, line 126: Shouldn't this one be inside the `if_group` as well? What does "Crop to fit base layer" even mean when one of the layers is a color?

backend/src/packages/chaiNNer_standard/image_utility/compositing/blend_images.py
```diff
@@ -44,53 +196,88 @@ def blend_images_node(
...
+        v = [
+            round((base_width - overlay_width) * x_percent / 100),
+            round((base_height - overlay_height) * y_percent / 100),
+        ]
+    else:
+        v = [base_width - overlay_width, base_height - overlay_height]
+    x0, y0 = np.array(
+        (
+            BLEND_OVERLAY_X0_Y0_POSITION[overlay_position][0] * v
+            + BLEND_OVERLAY_X0_Y0_POSITION[overlay_position][1] * [x_px, y_px]
```
RunDevelopment, line 236: You cannot use `x_px` and `y_px` here. Unless `overlay_position == BlendOverlayPosition.PIXEL_OFFSET` their value is not defined and may be any value allowed by the input. 

This is because we keep the input values of hidden inputs. E.g. if the user selects "Offset (pixels)" and sets the x offset to 200, but then changes their mind and selects "Offset (%)" instead, then `x_px` will still have the value 200.

backend/src/packages/chaiNNer_standard/image_utility/compositing/blend_images.py
adegerard: Right, mistake when I moved the checkbox at the bottom of the node.
No meaning when an input is a color

backend/src/packages/chaiNNer_standard/image_utility/compositing/blend_images.py
adegerard: Unless the x_px/y_px are None, this is working because `BLEND_OVERLAY_X0_Y0_POSITION[overlay_position][1]` is equal to `[0,0]` except when overlay_position==BlendOverlayPosition.PIXEL_OFFSET.
I wanted to have one single formula to avoid too many if...elif...else. PERCENT_OFFSET is unfortunately an exception.

Because it is unclear for quick understanding, I may change this to 
```python
    if overlay_position == BlendOverlayPosition.PERCENT_OFFSET:
        x0, y0 = [
            round((base_width - overlay_width) * x_percent / 100),
            round((base_height - overlay_height) * y_percent / 100),
        ]
    elif overlay_position == BlendOverlayPosition.PIXEL_OFFSET:
        x0, y0 = [x_px, y_px]        
    else:
        v = [base_width - overlay_width, base_height - overlay_height]
        x0, y0 = np.array(BLEND_OVERLAY_X0_Y0_POSITION[overlay_position][0] * v).astype("int")
```
What is your opinion?

RunDevelopment: > I may change this to

That is a lot more understandable. Let's go with that.

joeyballentine (approved):
This seems to work really well. Thanks a lot for this!

PR #2088, opened 2023-08-13
add a UI setting for the amount of ram reserved for the system

stonerl:
this is just the UI part and contains no logic.


stonerl:
Just an idea.

CleanShot 2023-08-13 at 18 52 00

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -71,19 +87,27 @@ def estimate():
 @processing_group.register(
     schema_id="chainner:pytorch:upscale_image",
     name="Upscale Image",
-    description="Upscales an image using a PyTorch Super-Resolution model. \
-            Select a manual number of tiles if you are having issues with the automatic mode. ",
+    description=(
+        "Upscales an image using a PyTorch Super-Resolution model.             Select a"
+        " manual number of tiles if you are having issues with the automatic mode. "
```
RunDevelopment, line 92: What is this formatting?

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -39,13 +40,28 @@ def upscale(
...
+                        model_bytes,
+                        img,
+                        element_size,
+                    )
+                )
+
+            if is_arm_mac:
+                total_memory = psutil.virtual_memory().total
+                reserved_system_memory = options.reserved_system_memory * (1024**2)
+                budget = int(total_memory - reserved_system_memory)
```
RunDevelopment, line 63: I would suggest lowering the budget to 80% just like we do for CUDA. 

`estimate_tile_size` really does give back an estimate. The estimated tile size might use a lot more RAM than we allowed. This isn't a huge issue with CUDA because we just have to restart upscaling when we run out of VRAM, but it's a big issue when the same thing happens with RAM. 

Look at it like this: when we estimate a tile size smaller than optimal, upscaling just takes a bit longer, but when we estimate a tile size that is too large, the whole system will stutter (or whatever happens when a mac runs out of memory).

RunDevelopment:
> I'd also like to implement this for NCNN and would reuse some code snippets from https://github.com/chaiNNer-org/chaiNNer/pull/2070. Haven't looked into it yet, but is there something similar for ONNX?

We don't support estimated tile sizes for ONNX. We just never implemented it.

As for NCNN: it's complicated. We do estimate for NCNN, but that estimation frequently crashed the backend on macs for some reason, so I turned it off (#2006). It will now always estimate a tile size of 256 on mac. This is a good default, because it only needs 3~4GB of (V)RAM for most models.

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
stonerl: That was my auto-formatter going crazy. fixed.

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
stonerl: The budget is determined by the value set in `reserved_system_memory`.

On my system 32 GB, it could be 20% when it set it to the max possible value (25.6 GB) and 75% when set to the min possible value (8 GB).

But you're right, could be more than 80% on systems with 64 GB and higher.

src/renderer/components/SettingsModal.tsx
```diff
@@ -297,11 +298,54 @@ const EnvironmentSettings = memo(() => {
...
+
     return (
         <VStack
             divider={<StackDivider />}
             w="full"
         >
+            {isArmMac ? (
+                <SettingsItem
+                    description={`The amount of RAM reserved for the system, that will not be used by chaiNNer for upscaling. (${
+                        minValue / 1024
```
stonerl, line 315: @RunDevelopment 

I should calculate the min value here so that it can never be more than 20% of the system's total memory.

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
stonerl: Fixed, the budget can now be between 20% and 80% of the system memory.

stonerl: I also added a fail-safe. So the budget will always be between 20% and 80% of the total system memory.

stonerl:
@joeyballentine 

Here is a comprehensive comparison:

https://www.sir-apfelot.de/en/compare-shared-memory-unified-memory-41023/

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -54,6 +56,27 @@ def estimate():
...
+                total_memory = psutil.virtual_memory().total
+                reserved_system_memory = options.reserved_system_memory * (1024**2)
+                pre_budget = int(total_memory - reserved_system_memory)
+
+                if pre_budget > total_memory * 0.8:
+                    budget = total_memory * 0.8
+                elif pre_budget < total_memory * 0.2:
+                    budget = total_memory * 0.2
+                else:
+                    budget = pre_budget
```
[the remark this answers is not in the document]
stonerl: Done. Much cleaner 😄

RunDevelopment:
I think Intel integrated graphics also uses the unified memory model. https://www.intel.com/content/www/us/en/support/articles/000020962/graphics.html

stonerl:
Interesting. We would need someone with such a system to test this.

stonerl:
For testing, I added a new switch that overwrites the memory settings and uses instead 80% of the available memory. Any thoughts.

CleanShot 2023-08-14 at 17 37 02@2x

stonerl:
After some thorough thinking, I realized that the users would have far too much rope to hang themselves. Now a saner approach.

By the default, users can choose a percentage of freely available memory (20% – 80%). If they really want to live on the edge, they now have an extra option where they can explicitly use the percentage of total available system memory.

CleanShot 2023-08-14 at 21 45 50@2x

stonerl:
Added calculation for NCNN MaxTileSize on Apple Silicon Macs. Left it 256 for Intel Macs, though.

It seems `vkdev.get_heap_budget` was reporting values that were far too high.

On my system with `vkdev.get_heap_budget`:

```
[2023-08-14 22:49:01.734] [info]  Backend: [92171] [INFO] Estimating memory required: 170.55 GB, 17.07 GB available. Estimated tile size: 1024
```

And with the `psutil.virtual_memory().available` and max memory set to 80%.

```
[2023-08-14 22:44:33.013] [info]  Backend: [88582] [INFO] Estimating memory required: 170.55 GB, 10.78 GB available. Estimated tile size: 512
```

JeremyRand:
Can you clarify how this PR compares to #2070? At first glance it looks like it  may have some similar logic (but possibly used in different places?).

stonerl:
@JeremyRand 

Of course. This one is foremost for Apple Silicon Macs, at least for the moment. But it could be extended to other platforms.

It is not only meant for NCNN upscaling but also for PyTorch. I'm also working on another PR that builds upon this one that implements tile size estimation for ONNX.

Since on an Apple Silicon Mac one has not to differentiate between video and system memory, it only needs one setting.

In general, this is a broader approach that does not focus on one specific backend.

JeremyRand:
(Using a percentage of RAM rather than an explicit number as #2070 does is probably better UX.)

JeremyRand:
Oh, I see. That wasn't obvious to me from the screenshot. Yeah, so this UX seems better than #2070 for my use cases. I'd probably be OK with putting #2070 on hold until this is merged, and then I could rebase #2070 so that it just applies your UI's settings to CPU inference too. Thoughts?

stonerl:
> Oh, I see. That wasn't obvious to me from the screenshot.

Maybe I need to reword this. Instead of **Use total system memory** something like **Apply % to total system memory**

JeremyRand:
> > Oh, I see. That wasn't obvious to me from the screenshot.
> 
> Maybe I need to reword this. Instead of **Use total system memory** something like **Apply % to total system memory**

Maybe also helpful to show the amount of free RAM, e.g. "Up to 22.4 GB out of a total of 32 GB RAM (26 GB free) will be used."

Also minor nit; GB is not really the right unit for this, GiB is preferred for this purpose. (GB is a power of 1000, GiB is a power of 1024 per international standards.)

stonerl:
> Maybe also helpful to show the amount of free RAM, e.g. "Up to 22.4 GB out of a total of 32 GB RAM (26 GB free) will be used."

I was thinking about this, but free RAM differs so much, that showing the user a number would let them assume that it might always be the displayed value. That's why I didn't add it.


> Also minor nit; GB is not really the right unit for this, GiB is preferred for this purpose. (GB is a power of 1000, GiB is a power of 1024 per international standards.)

Done

JeremyRand:
> I was thinking about this, but free RAM differs so much, that showing the user a number would let them assume that it might always be the displayed value. That's why I didn't add it.

Yeah that's fair I guess.

JeremyRand:
How would you feel about making the percentage precise to 0.1% rather than 1%? On systems with a lot of RAM (my main system has around 300 GiB), being more precise than 1% would be nice. I realize the power-of-2 tile size estimation itself introduces more imprecision than that, but that's an orthogonal issue that will hopefully be resolved separately.

stonerl:
> How would you feel about making the percentage precise to 0.1% rather than 1%?

Done.

CleanShot 2023-08-15 at 17 45 12@2x

stonerl:
This isn't Apple Silicon only anymore. 

RAM settings now also apply for CPU upscaling with PyTorch on all platforms.

@JeremyRand you only need to add your CPU settings for NCNN. All the other bits are already in place.

JeremyRand:
Very nice, I'll rebase #2070 after this is merged.

stonerl:
Okay, so basically every bit of memory usage is now configurable. I added an option to set the amount of VRAM that should be used. Not used on Macs, though.

Here's a screenshot:

CleanShot 2023-08-15 at 19 47 09@2x

stonerl:
Can't test this, though. Since I don't have a PC.

RunDevelopment:
Okay, so here are my thoughts on the settings:

1. I like the use of percentages.
1. I don't like the "Use total system memory" toggle. Just allow 100%. Why split one number into 2 settings?
1. I think the "X GB out of a total of Y GB RAM" part of should be part of the RAM percentage setting.

stonerl:
Thanks for your feedback. I think the main problem was the wording of the settings. I changed this. (screenshot ant the end)

Let me answer your question from back to front:

> 3. I think the "X GB out of a total of Y GB RAM" part of should be part of the RAM percentage setting.

@JeremyRand also asked for this, but the thing is, free/available RAM is very volatile. So if we would display a value there, this would only be true for that single point in time when looking at the settings. The next time you open the settings, the values would be different. It could be 8 GiB and the next time 14 GiB.

> 2. I don't like the "Use total system memory" toggle. Just allow 100%. Why split one number into 2 settings?

The idea here is that instead of allocating a max percentage of free memory `psutil.virtual_memory().available` we apply this percentage to the total memory available `psutil.virtual_memory().total`. These values usually vary a lot, e.g., on my system 80% of total system memory would be 25.6 GiB but 80% of free/available memory when I use my system as I usually do it would be around 12 GiB.

I hope the rewording makes this clear.

CleanShot 2023-08-16 at 13 10 23@2x

JeremyRand:
Suggestion: reword "Apply memory limit to total system memory" to "Apply memory limit percentage to total system memory". I think that's a bit less likely to confuse people (it certainly would have confused me less).

RunDevelopment:
> The idea here is that instead of allocating a max percentage of free memory `psutil.virtual_memory().available` we apply this percentage to the total memory available `psutil.virtual_memory().total`.

I see. I misunderstood and thought that the limits were already a percentage of the total memory.

However, is that actually a good idea to have that toggle, though? 
I guess I just don't see how it would be useful. As you said, using a percentage of total memory is essentially just a rope to hang yourself with. So why would we hand people that rope?

---

Also, different topic: is a GPU memory limit useful? Why wouldn't we want to max out VRAM?

RunDevelopment:
> > However, is that actually a good idea to have that toggle, though?
> > I guess I just don't see how it would be useful. As you said, using a percentage of total memory is essentially just a rope to hang yourself with. So why would we hand people that rope?
> 
> @RunDevelopment AFAICT it's necessary if you want to use multiple chaiNNer instances at once; otherwise the first one to start executing a chain will gobble up all the RAM for itself, and the other instances will have nothing left.

Uhhhmmm, how would the toggle help in that scenario? If the first instance takes all the RAM, then your memory limit is too high. How does using a percentage of total memory help in a way that using a percentage of free memory doesn't?

stonerl:
I did some fine-tuning regarding the warning dialog. 

CleanShot 2023-08-18 at 16 51 47@2x

@RunDevelopment, would you be okay if we let the toggle in, but present the users a warning?

JeremyRand:
> Uhhhmmm, how would the toggle help in that scenario? If the first instance takes all the RAM, then your memory limit is too high. How does using a percentage of total memory help in a way that using a percentage of free memory doesn't?

@RunDevelopment I want each instance to use the same absolute amount of RAM, which means they should use the same fraction of total RAM. If each instance instead uses the same fraction of free RAM, then that would result in different instances using different absolute amounts of RAM, with the earlier-starting instances grabbing a higher amount.

stonerl:
I squashed the commits and rebased it on main. This is how it looks now.

## Windows and Linux
CleanShot 2023-08-19 at 00 39 21@2x
CleanShot 2023-08-19 at 00 39 37@2x

## macOS
CleanShot 2023-08-19 at 00 37 39@2x
CleanShot 2023-08-19 at 00 37 49@2x

## Log file
```
2023-08-19 00:41:40.080] [info]  Backend: [70296] [INFO] Memory limit set to 80.0% of total system memory. (32.0 GB)
[2023-08-19 00:41:40.080] [info]  Backend: [70296] [INFO] Estimating memory required: 14.67 GB, 25.60 GB available. Estimated tile size: 1024
[2023-08-19 00:41:40.084] [info]  Backend: [70296] [INFO] Auto split image (1024x1024px @ 3) with initial tile size (1024, 1024).
```

JeremyRand:
What's still needed before we can get this merged?

RunDevelopment (commented):
I am so sorry. I forgot to submit my review. My comments just sat there for a week...

joeyballentine:
Yeah, I'll try to get that done as soon as I can

stonerl:
@JeremyRand I'm currently very busy IRL. If you want to you can take this over and incorporate your changes. I'm not sure when I can continue working on this.

JeremyRand:
@stonerl If I take this PR over, what remaining work needs to be done? I admit I haven't kept up with the code review on this PR, so it's hard for me to evaluate whether the remaining work is something I can do efficiently.

JeremyRand:
(Or maybe @joeyballentine would be able to answer that?)

joeyballentine:
@JeremyRand I'm not sure what still needs to be done for this. I can take a look tomorrow and try to figure that out though

joeyballentine:
@JeremyRand sorry, we've been focused on the iterator rewrite and now on extracting out our model detection code into a separate package. After we're done with that we gotta finish the iterator rewrite (it's taken way longer than we thought).

After that, we should be able to work on stuff like this

JeremyRand:
Sure, no worries. The iterator rewrite looks really cool, I'm fine with waiting on this PR if it gets the iterator rewrite done faster. :)

PR #2089, opened 2023-08-13
Add setting for allowing multiple concurrent instances

joeyballentine:
Long overdue

### DIFF
--- a/src/main/gui/create.ts
+++ b/src/main/gui/create.ts
@@ -57,7 +57,7 @@ export const createGuiApp = (args: OpenArguments) => {
 
     const isAllowMultipleInstances = settingStorage.getItem('allow-multiple-instances') === 'true';
 
-    const hasInstanceLock = app.requestSingleInstanceLock() || isAllowMultipleInstances;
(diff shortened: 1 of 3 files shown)


src/main/gui/create.ts
```diff
@@ -55,7 +55,9 @@ export const createGuiApp = (args: OpenArguments) => {
         app.disableHardwareAcceleration();
     }
 
-    const hasInstanceLock = app.requestSingleInstanceLock();
+    const isAllowMultipleInstances = settingStorage.getItem('allow-multiple-instances') === 'true';
+
+    const hasInstanceLock = app.requestSingleInstanceLock() || isAllowMultipleInstances;
```
RunDevelopment, line 60: `app.requestSingleInstanceLock()` likely has numerous side effects, so let's swap those 2 conditions.
```suggestion
    const hasInstanceLock = isAllowMultipleInstances || app.requestSingleInstanceLock();
```

RunDevelopment (changes requested):

PR #2102, opened 2023-08-15
Add SCUNet architecture

adegerard:
scunet

**Choices and limitations**
- Disabled tiling as SCUNet uses both local and non local properties. Drawback: OOM may occur depending on the image size, image content, NVRAM, type and amount of noise, (other?). Here are the artifacts that may occur if using tiling [issue opened at stable-diffusion-ui](https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/12524)

**Other**
- Would you mind checking the way I detect that the model architecture
- As usual, all feedback welcomed


RunDevelopment (approved):
The code looks good. I'm assuming that the model itself is correct, since I can't review this.

adegerard:
> The code looks good. I'm assuming that the model itself is correct, since I can't review this.

I added links to the models in the README file. For model detection, I only use a subset of the state dict keys.
@joeyballentine would you please also have a look to this as it is the first time I'm doing this.

joeyballentine:
Yes I am planning on looking at this, I just haven't had time

backend/src/nodes/impl/pytorch/architecture/SCUNet.py
```diff
@@ -0,0 +1,455 @@
...
+
+class SCUNet(nn.Module):
+    def __init__(
+        self,
+        state_dict,
+        in_nc=3,
+        config=[4, 4, 4, 4, 4, 4, 4],
+        dim=64,
+        drop_path_rate=0.0,
+        input_resolution=256,
```
joeyballentine, line 282: Generally I prefer detecting things like these settings using the state dict to future-proof a bit in case anyone trains any custom models with different parameters, but I won't make you do that.

joeyballentine (approved):

PR #2104, opened 2023-08-16
Backend-defined package settings

joeyballentine:
This is a draft of my settings refactor. Package settings are now defined on the backend per-package.

I wanted to open this as a draft to get some feedback on how I approached 1) the settings definitions themselves, 2) how I approached how you get settings for a package during a run, and 3) how I typed the settings (since it's impossible to do typescript-like type magic in python to do it automatically)

I have moved everything to the new settings system, but I have only modified the pytorch nodes to use the new execution options system. I am yet to do it for ncnn and onnx.


joeyballentine:
I'd rebase those broken commits out, but that causes a big merge conflict I don't want to deal with at the moment

backend/src/packages/chaiNNer_onnx/__init__.py
```diff
@@ -68,8 +76,68 @@ def get_onnx_runtime():
...
+        )
+    )
+
+execution_providers = []
+try:
+    import onnxruntime as ort
+
+    execution_providers = ort.get_available_providers()
+except:
+    pass
```
RunDevelopment, line 104: That's just not a good idea. What's the point of having these settings when they are functionally useless? 

I think we should only define settings when the package dependencies are met. The trick I used for features could be used here as well. We make a file called `package/settings.py` which adds the settings to the package and exports a `get_settings()` function. This function will be the only way to access settings, so nodes that need settings will import that function. So settings are only defined when nodes that use those settings are available. This also means that we can use all package dependencies in `settings.py`, so no optional imports.

backend/src/api.py
```diff
@@ -433,5 +457,53 @@ def add_package(
...
+class ToggleSetting:
+    label: str
+    key: str
+    description: str
+    default: bool = False
+    disabled: bool = False
+    type: str = "toggle"
+
+
+DropdownOption = TypedDict("DropdownOption", {"label": str, "value": str})
```
RunDevelopment, line 476: You're supposed to use `TypedDict` like this:

```suggestion
class DropdownOption(TypedDict):
    label: str
    value: str
```

backend/src/packages/chaiNNer_pytorch/__init__.py
```diff
@@ -115,8 +116,68 @@ def get_pytorch():
...
+    {
+        "cpu_mode": bool,
+        "fp16_mode": bool,
+        "gpu": int,
+    },
 )
 
+
+def get_pytorch_settings() -> PyTorchSettings:
+    return cast(PyTorchSettings, package.get_execution_settings())
```
RunDevelopment, line 178: I really don't like this. It's both wrong (`gpu` isn't defined for ARM macs) and inconvenient to use (`.get("some-key", default)`; I hope you use the same default everywhere). 

Instead, I think that `get_settings` should parse the raw settings dict. Combined with my `settings.py` suggestions, we could have something like this:

```py
# settings.py
import torch
from . import package

package.add_setting(...)

@dataclass
class PyTorchSettings:
    cpu_mode: bool
    fp16: bool
    gpu: int

    def get_device(self) -> torch.device: ...

def get_settings() -> PyTorchSettings:
    raw = package.get_execution_settings()
    return PyTorchSettings(
        cpu_mode=raw.get("cpu_mode", false),
        ...
    )
```

This also adds a nice abstraction layer between the way we internally store settings and the way nodes consume settings.

backend/src/nodes/impl/pytorch/utils.py
```diff
@@ -6,18 +6,17 @@
 import torch
 from torch import Tensor
 
-from ...utils.exec_options import ExecutionOptions
 from ..image_utils import as_3d
 from ..onnx.np_tensor_utils import MAX_VALUES_BY_DTYPE, np_denorm
 
 
-def to_pytorch_execution_options(options: ExecutionOptions):
+def get_pytorch_device(use_cpu: bool = False, gpu_index: int = 0) -> torch.device:
```
RunDevelopment, line 13: This function isn't used anymore.

backend/src/api.py
```diff
@@ -309,6 +311,11 @@ class Package:
     dependencies: List[Dependency] = field(default_factory=list)
     categories: List[Category] = field(default_factory=list)
     features: List[Feature] = field(default_factory=list)
+    settings: List[
+        ToggleSetting | DropdownSetting | NumberSetting | CacheSetting
```
RunDevelopment, line 315: Please make this a type alias.

src/common/common-types.ts
```diff
@@ -308,12 +308,48 @@ export interface Feature {
     readonly description: string;
 }
 
+export interface ISetting {
```
RunDevelopment, line 311: No `IName`. Common convention is to use `NameBase` and make the interface non-exported.
```suggestion
interface SettingBase {
```

src/common/common-types.ts
```diff
@@ -308,12 +308,48 @@ export interface Feature {
...
+    readonly type: string;
```
RunDevelopment, line 312:
```suggestion
    readonly type: Setting["type"];
```

src/renderer/contexts/SettingsContext.tsx
```diff
@@ -42,21 +35,7 @@ interface Settings {
...
-    const useOnnxExecutionProvider = useMemoArray(
-        useLocalStorage('onnx-execution-provider', 'CUDAExecutionProvider')
-    );
-    const useOnnxShouldTensorRtCache = useMemoArray(
-        useLocalStorage('onnx-should-tensorrt-cache', false)
-    );
-    const useOnnxShouldTensorRtFp16 = useMemoArray(
-        useLocalStorage('onnx-should-tensorrt-fp16', false)
-    );
+    const useBackendSettings = useMemoArray(useLocalStorage('backend-settings', {}));
```
RunDevelopment, line 38: Settings migrations?

src/common/common-types.ts
```diff
@@ -409,3 +445,12 @@ export interface RgbaColorJson {
...
 }
+
+export interface CacheSettingValue {
+    enabled: boolean;
+    location: string;
+}
+
+export interface PackageSettings {
+    [packageName: string]: Record<string, string | number | boolean | CacheSettingValue>;
+}
```
RunDevelopment, line 456: Man, I don't like cache setting. It's so specialized... Not sure what to do about it though.

src/renderer/contexts/SettingsContext.tsx
joeyballentine: I was just going to announce that python related settings are being reset, but yeah I guess I could try adding a migration for settings

src/common/common-types.ts
joeyballentine: I also don't like it. The alternative was to have a standard path convention that the frontend and backend would both find independently, but I didn't like the potential desync. Idk what a better solution would be

src/renderer/contexts/SettingsContext.tsx
joeyballentine: Man, how would i even go about migrating this?

RunDevelopment: Yeah, then let's not do this.

RunDevelopment (approved):

PR #2107, opened 2023-08-16
Disable multi-instance setting on macOS and improve viewport to clipboard shortcut hint

stonerl:
- The multi-instance settings has no effect on macOS ATM.
- shortcut change for viewport to clipboard on control button.

### DIFF
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -673,9 +673,11 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                 <Controls>
                     <ControlButton
                         disabled={nodes.length === 0}
-                        title={'Export viewport as PNG\nCtrl+Click to export to clipboard instead'}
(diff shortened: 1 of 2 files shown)


src/renderer/components/ReactFlowBox.tsx
```diff
@@ -673,9 +673,11 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                 <Controls>
                     <ControlButton
                         disabled={nodes.length === 0}
-                        title={'Export viewport as PNG\nCtrl+Click to export to clipboard instead'}
+                        title={`Export viewport as PNG file\n\nHold ${
+                            isMac ? '⌥' : 'Ctrl'
+                        } to export to clipboard`}
```
RunDevelopment, line 678: The new description makes it sound like simply holding <kbd>Ctrl</kbd>/<kbd>⌥</kbd> will export to clipboard. Could we bring back the "+Click"?

stonerl: done.

PR #2114, opened 2023-08-18
add listener to better handle hotkey presses

### LINKED ISSUE #1639: F5 doesn't work at first
**Information:**

- Chainner version: 0.18.3
- OS: Windows 10, x64

**Description**
After loading a saved .chn file pressing F5 does nothing.  If I want to start upscaling the textures already in the proper folder I have to click on the Run (F5) icon at the top of the window.
On subsequent iterations F5 works as intended.

[logs.zip](https://github.com/chaiNNer-org/chaiNNer/files/10912080/logs.zip)

stonerl:
fixes #1639

### DIFF
--- a/src/renderer/contexts/ExecutionContext.tsx
+++ b/src/renderer/contexts/ExecutionContext.tsx
@@ -20,7 +20,6 @@ import {
 } from '../hooks/useBackendEventSource';
 import { useBackendExecutionOptions } from '../hooks/useBackendExecutionOptions';
 import { useBatchedCallback } from '../hooks/useBatchedCallback';
-import { useHotkeys } from '../hooks/useHotkeys';
 import { useMemoObject } from '../hooks/useMemo';
 import { AlertBoxContext, AlertType } from './AlertBoxContext';
 import { BackendContext } from './BackendContext';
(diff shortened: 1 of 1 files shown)


RunDevelopment:
This causes all commands to be executed twice. This can be seen with an empty chain. Make an empty chain and press F5. You'll get 2 identical errors:

image

You can probably just remove the hotkey handlers in `ExecutionContext.tsx` to fix this. I would even recommend moving your `useEffect` there.

Also, please use the same logic as the current hotkey handlers. This includes accounting for the current execution status, and disabling hotkeys when an alert is open. See `HotkeysContext` and `useHotkeys`.

stonerl:
I hope it is correct now.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -389,59 +388,57 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
-                    assertNever(status);
+    // This makes sure keystrokes are executed even if the focus is on an input field
+    useEffect(() => {
+        const handleKeyPress = (event: KeyboardEvent) => {
+            if (event.code === 'F5') {
+                switch (status) {
+                    case ExecutionStatus.READY:
+                    case ExecutionStatus.PAUSED:
+                        event.preventDefault();
+                        (run as () => void)();
```
RunDevelopment, line 399: Please don't use type casts like this. If we ever change the signature of those functions, TS will not be able to catch the now incorrect call.

You likely did this because ESLint doesn't like floating promises. The easiest way around this is to do what ESLint wants and add a `catch`.
```suggestion
                        run().catch(log.error);
```

RunDevelopment (changes requested):

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -389,59 +388,57 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
...
+                        break;
+                    case ExecutionStatus.RUNNING:
+                    case ExecutionStatus.KILLING:
+                        break;
+                    default:
+                        log.warn('Unknown status for Run:', status);
```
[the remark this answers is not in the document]
stonerl: done.

src/renderer/contexts/ExecutionContext.tsx
stonerl: done

RunDevelopment (approved):
Thank you @stonerl!

PR #2125, opened 2023-08-19
Add "text as image" node

adegerard:
**Motivation**
This node transforms a text into an image.
It creates an image and draw the text with the specified alignement
It calculates the font size to fit the canva: `width` and `height` input values.

text_as_image_example

**Choices and limitations**
- Font style (regular, blod, italic): via a dropdown.
- Text alignement via a dropdown
- Position in the canva via a dropdown

As discussed on Discord, would be great to change the dropdown by grids of 'checkable buttons' (checkboxes)

For better user experience, PR #2087  should be merged in main


joeyballentine:
This seems fine to me, but I'll let @RunDevelopment confirm

joeyballentine:
I think this node probably fits better in the "Create Images" group in the "Images" category, since it involves image creation. Thoughts?

adegerard:
I don't know because, to me, it would be more a converter, however, in chaiNNer, I would agree with with you to put it in "create images" group. It makes more sens.

What about node title? "Create from text" ? it might be confused with IA...

RunDevelopment:
Is there no better way than to vendor fonts? Can't we use fonts installed on the current system?

joeyballentine:
If there's no built in way, we could probably just manually scan the system fonts folder and build a list of all the available valid fonts

adegerard:
Because I think it would be a nightmare. 
My thoughts:
- (Please have a look to) C:\Windows\Fonts  (more than 200 fonts on my computer) or /usr/share/fonts/ and ~/.local/share/fonts on Linux
- Only ttf fonts are loadable by PIL. (freetype also but let's forget it)
- Some are static fonts, other dynamic fonts: it is not always possible to differentiate style: bold/italic/light/light by filename.
- A dropdown of 200 elements is not understandable, especially if there is no preview (like in ms word)
- I tried to use dynamic fonts but had a problem with the index no. (which has to be used to indicate to PIL which style you'd like to load)
- If you share your project across different OS, or if you have some sw which installs specific fonts (like adobe), you could have trouble with font resource/name/path

joeyballentine:
Ok yeah, you've convinced me

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -155,10 +155,10 @@ class TextAsImagePosition(Enum):
 }
 
 
-@compositing_group.register(
+@create_images_group.register(
     schema_id="chainner:image:text_as_image",
-    name="Text as image",
-    description="Converts a text to an image",
+    name="Text As Image",
+    description="Convert a text to an image.",
```
joeyballentine, line 161: /

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -0,0 +1,271 @@
...
+        "x": np.array([1, -0.5]),
+        "y": np.array([1, -0.5]),
+    },
+}
+
+
+@create_images_group.register(
+    schema_id="chainner:image:text_as_image",
+    name="Text As Image",
+    description="Convert a text to an image.",
```
joeyballentine, line 161: Maybe something more like this?
```suggestion
    description="Create an image using any text and a chosen font.",
```

RunDevelopment:
Hmmm, yeah, that certainly will be difficult. But having 3 font families to pick from and having those fonts be vendored just doesn't sit right with me. 

I mean, what's the point of having a font selector if there is no real choice? So how about we just remove the font selector for now and just use one. Font style (bold, italic, etc.) could be its own input.

adegerard:
> Hmmm, yeah, that certainly will be difficult. But having 3 font families to pick from and having those fonts be vendored just doesn't sit right with me.
> 
> I mean, what's the point of having a font selector if there is no real choice? So how about we just remove the font selector for now and just use one. Font style (bold, italic, etc.) could be its own input.

Thanks for your feedback.
It's ok with me. KISS principle.

adegerard:
I made a version without font selection.
There are now 2 checkboxes to select font style.

image

Like other parameters, it would be great to convert them to "buttons".

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -0,0 +1,212 @@
...
+            option_labels=TEXT_AS_IMAGE_POSITION_LABELS,
+            default=TextAsImagePosition.CENTERED,
+        ),
+    ],
+    outputs=[
+        ImageOutput(
+            image_type="""
+                Image {
+                    width: Input5 & uint,
+                    height: Input6 & uint,
```
RunDevelopment, line 145: These 2 `& uint` should be unnecessary. If not, then I have some work to do.

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -0,0 +1,212 @@
...
+                    channels: Input3.channels
```
RunDevelopment, line 146: I think this is wrong. The color is either grayscale or RGB, but the image is either grayscale or RGBA. So you probably meant to do `match Input3.channels { 1 => 1, 3 => 4 }`.

adegerard: Finally, decided to output a RGBA image only.
I see no reason to create a grayscale image without transparent layer. If needed, user can use the "change colorspace" node.

joeyballentine: that makes sense to me

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -0,0 +1,206 @@
...
+    ],
+    outputs=[
+        ImageOutput(
+            image_type="""
+                Image {
+                    width: Input5,
+                    height: Input6,
+                    channels: 4
+                }
+                """,
```
RunDevelopment, line 148: For a fixed number of channels, it's better to use the `channels` parameter on `ImageOutput`. This will still produce the same type, but chainner can also check at runtime that your node actually returns an RGBA image.

```suggestion
            image_type="""
                Image {
                    width: Input5,
                    height: Input6,
                }
                """,
            channels: 4,
```

backend/src/packages/chaiNNer_standard/image/create_images/text_as_image.py
```diff
@@ -0,0 +1,206 @@
...
+    w = int(width * 100.0 / w_ref)
+    h = int(height * 100.0 / (h_ref * line_count))
+    font_size = min(w, h)
+    font = ImageFont.truetype(font_path, size=font_size)
+    w_text, h_text = font.getsize(max_line)
+    h_text *= line_count
+
+    # Text color
+    ink = 255 * np.array(color.value)
+    ink = tuple(ink.astype("uint8"))
```
RunDevelopment, line 186: Please use `to_uint8` from `nodes.impl.image_utils`.

```suggestion
    ink = tuple(to_uint8(color.value))
```

RunDevelopment (changes requested):
2 other minor things.

adegerard:
> 2 other minor things.
Done with some corrections.

adegerard:
wrong manipulation, sorry.

joeyballentine:
I'll play around with this again today and then code review it, but I'm guessing it's probably good now 👍

RunDevelopment (approved):
Thanks @adegerard!

PR #2127, opened 2023-08-20
Add "Separate Alpha" checkbox to upscale nodes

joeyballentine:
I've finally decided that this is probably worth adding so people don't have to do the channel splitting manually. As much as I think it's good to have people do things themselves, sometimes convenience is warranted. I'll do a "Seamless Edges" checkbox next.

image

### DIFF
--- a/backend/src/nodes/impl/upscale/convenient_upscale.py
+++ b/backend/src/nodes/impl/upscale/convenient_upscale.py
@@ -25,6 +25,7 @@ def convenient_upscale(
     model_in_nc: int,
     model_out_nc: int,
     upscale: ImageOp,
+    separate_alpha: bool = False,
 ) -> np.ndarray:
     """
     Upscales the given image in an intuitive/convenient way.
@@ -56,23 +57,33 @@ def convenient_upscale(
             unique_alpha = np.full(rgb.shape[:-1], unique[0], np.float32)
             return np.dstack((rgb, unique_alpha))
 
-        # Transparency hack (white/black background difference alpha)
-        black, white = with_black_and_white_backgrounds(img)
-        black_up = as_target_channels(
-            upscale(as_target_channels(black, model_in_nc, True)), 3, True
-        )
-        white_up = as_target_channels(
-            upscale(as_target_channels(white, model_in_nc, True)), 3, True
-        )
+        if separate_alpha:
+            # Upscale the RGB channels and alpha channel separately
+            rgb = as_target_channels(
+                upscale(as_target_channels(img[:, :, :3], model_in_nc, True)), 3, True
+            )
+            alpha = as_target_channels(
+                upscale(as_target_channels(img[:, :, 3], model_in_nc, True)), 1, True
+            )
+            return np.dstack((rgb, alpha))
+        else:
+            # Transparency hack (white/black background difference alpha)
+            black, white = with_black_and_white_backgrounds(img)
+            black_up = as_target_channels(
+                upscale(as_target_channels(black, model_in_nc, True)), 3, True
+            )
+            white_up = as_target_channels(
+                upscale(as_target_channels(white, model_in_nc, True)), 3, True
+            )
 
-        # Interpolate between the alpha values to get a more defined alpha
-        alpha_candidates = 1 - (white_up - black_up)  #  type: ignore
-        alpha_min = np.min(alpha_candidates, axis=2)
-        alpha_max = np.max(alpha_candidates, axis=2)
-        alpha_mean = np.mean(alpha_candidates, axis=2)
-        alpha = alpha_max * alpha_mean + alpha_min * (1 - alpha_mean)
+            # Interpolate between the alpha values to get a more defined alpha
+            alpha_candidates = 1 - (white_up - black_up)  #  type: ignore
+            alpha_min = np.min(alpha_candidates, axis=2)
+            alpha_max = np.max(alpha_candidates, axis=2)
+            alpha_mean = np.mean(alpha_candidates, axis=2)
+            alpha = alpha_max * alpha_mean + alpha_min * (1 - alpha_mean)
 
-        return np.dstack((black_up, alpha))
+            return np.dstack((black_up, alpha))
 
     return as_target_channels(
         upscale(as_target_channels(img, model_in_nc, True)), in_img_c, True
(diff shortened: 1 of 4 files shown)


backend/src/packages/chaiNNer_onnx/onnx/processing/upscale_image.py
```diff
@@ -96,9 +111,12 @@ def upscale_image_node(
     h, w, c = get_h_w_c(img)
     logger.debug(f"Image is {h}x{w}x{c}")
 
+    should_separate_alpha = separate_alpha and c == 4
```
RunDevelopment, line 114: Why make this variable? `convenient_upscale` will ignore the parameter unless it's applicable anyway, so you don't need to make any checks yourself.

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
```diff
@@ -138,14 +144,25 @@ def estimate_cpu():
             "Generally it's recommended to use the largest tile size possible for best performance (with the ideal scenario being no tiling at all), but depending on the model and image size, this may not be possible.",
             "If you are having issues with the automatic mode, you can manually select a tile size. On certain machines, a very small tile size such as 256 or 128 might be required for it to work at all.",
         ),
+        if_group(
+            Condition.type(1, "Image { channels: 4 } ")
+            and Condition.type(0, "NcnnNetwork { outputChannels: 1 | 3 } ")
```
RunDevelopment, line 149: 1. Welcome to python. The `and` operator will check for truthy-ness and returns the last true-y value. You meant to use bitwise AND `&`.
2. The condition is not quite correct. Since we only use this inputs if `in_nc == out_nc`, we have to use something like this:

Same for the others as well.
```suggestion
            & (Condition.type(0, "NcnnNetwork { inputChannels: 1, outputChannels: 1 }") | Condition.type(0, "NcnnNetwork { inputChannels: 3, outputChannels: 3 }"))
```

backend/src/nodes/impl/upscale/convenient_upscale.py
```diff
@@ -56,23 +57,33 @@ def convenient_upscale(
...
-        alpha_max = np.max(alpha_candidates, axis=2)
-        alpha_mean = np.mean(alpha_candidates, axis=2)
-        alpha = alpha_max * alpha_mean + alpha_min * (1 - alpha_mean)
-
-        return np.dstack((black_up, alpha))
+        if separate_alpha:
+            # Upscale the RGB channels and alpha channel separately
+            rgb = as_target_channels(
+                upscale(as_target_channels(img[:, :, :3], model_in_nc, True)), 3, True
+            )
```
RunDevelopment, line 64: This has a problem. Since the RGB value classically isn't defined for alpha=0 (but typically set to 0), models will typically add a black edge to images with binary alpha. 

Solving this problem is the very purpose of my Fill Alpha node. By filling the transparent pixels with nearby colors, and therefor extending the image into the transparent pixels (inpainting them), models don't produce black edges anymore. The issue now is that the results depends on the inpaint algorithm. This is why the Fill Alpha node has multiple modes and the user has to choose one depending on the image.

So what do we do here? The current solution isn't really ideal, but the better solution requires additional user input and can already be achieved with today's nodes.

RunDevelopment (changes requested):

backend/src/nodes/impl/upscale/convenient_upscale.py
```diff
@@ -56,23 +57,33 @@ def convenient_upscale(
...
+            alpha = as_target_channels(
+                upscale(as_target_channels(img[:, :, 3], model_in_nc, True)), 1, True
+            )
```
RunDevelopment, line 67: Assuming that we are using an RGB model, we can do better. `as_target_channels` will use OpenCV's cvtColor under the hood, which basically just averages the RGB channels to get one grayscale image. This isn't ideal, as any noise will get averaged, which makes it less likely to output pure white or pure black. See #1942 for more details.

I would suggest using the same approach for denoising alpha as the BW method uses. Of course, the only applies to RGB models.

backend/src/nodes/impl/upscale/convenient_upscale.py
joeyballentine: > So what do we do here? The current solution isn't really ideal, but the better solution requires additional user input and can already be achieved with today's nodes.

I don't want this to scope creep into making the upscale nodes one monolithic node of every possible option. If it doesn't even make sense to do this, then I should probably just scrap this PR.

RunDevelopment: True. Then let's not do any sort of inpainting right now. If black borders are an issue, then users can use our existing nodes to fix the issue.

RunDevelopment: So let's do this feature.

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -107,6 +112,24 @@ def estimate():
...
+                | Condition.type(
+                    0, "PyTorchModel { inputChannels: 3, outputChannels: 3 }"
+                )
+            )
+        )(
+            BoolInput("Separate Alpha", default=False).with_docs(
+                "Normally when dealing with an image with alpha, we take the difference between an"
+                " upscale with a black background and an upscale with a white background to get the"
+                " alpha channel. However, under certain circumstances it may be more desirable to"
+                " upscale the alpha channel separately from the RGB channels."
```
RunDevelopment, line 130: I think this description is too technical. If I didn't know how these techniques worked already, I would have no idea what this description was trying to communicate.

How about we describe the properties of the 2 techniques instead? Basically, we describe their tradeoffs and why you would use one over the other.

> Upscale alpha separately from color. Enabling this option will cause the alpha of the upscaled image to be less noisy and more accurate to the alpha of the original image, but the image may suffer from dark borders near transparency edges (transition from fully transparent to fully opaque).
> Whether enabling this option will improve the upscaled image depends on the original image. We generally recommend this option for images with smooth transitions between transparent and opaque regions.

RunDevelopment (approved):

PR #2129, opened 2023-08-21
Use grab cursor on node selector items

joeyballentine:
Small little UX thing to make it more intuitive that you can drag and drop selector panel items. Unfortunately, the way I had to add classes for this is really annoying.


RunDevelopment:
> make it more intuitive that you can drag and drop selector panel items

My intuition tells me that I can reorder the items by dragging them. E.g. to move Load Image below Save Image.

joeyballentine:
Hmmm that's a good point...

PR #2130, opened 2023-08-21
Dependency Manger - additional info for NCNN

stonerl:
Also: add an arrow and made the text justified.

CleanShot 2023-08-21 at 13 12 13@2x

### DIFF
--- a/backend/src/packages/chaiNNer_ncnn/__init__.py
+++ b/backend/src/packages/chaiNNer_ncnn/__init__.py
@@ -21,9 +21,10 @@ package = add_package(
     id="chaiNNer_ncnn",
     name="NCNN",
     description=(
-        f"{general} NCNN uses Vulkan for GPU"
-        " acceleration, meaning it supports any modern GPU. Models can be converted"
-        " from PyTorch to NCNN."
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_ncnn/__init__.py
```diff
@@ -21,9 +21,9 @@
     id="chaiNNer_ncnn",
     name="NCNN",
     description=(
-        f"{general} NCNN uses Vulkan for GPU"
-        " acceleration, meaning it supports any modern GPU. Models can be converted"
-        " from PyTorch to NCNN."
+        f"{general} Models can be converted from PyTorch to NCNN.\n\nNCNN uses Vulkan"
```
RunDevelopment, line 24: > Models can be converted from PyTorch to NCNN.

Only if ONNX is installed as well. Could you please mention that?

stonerl: Of course

stonerl: done.

RunDevelopment (approved):

PR #2132, opened 2023-08-22
handle unsaved changes on reboot

### LINKED ISSUE #1338: The "Restart chaiNer" button skips the unsaved data dialog
and results in loss of data.

stonerl:
fixes #1338

### DIFF
--- a/src/main/gui/main-window.ts
+++ b/src/main/gui/main-window.ts
@@ -196,8 +196,8 @@ const registerEventHandlerPostSetup = (
 
     const handleUnsavedChanges = (
         event: Electron.Event,
-        callback: () => void,
-        restartCallback?: () => void
+        onSave: () => void,
+        onDontSave?: () => void
     ) => {
         const choice = dialog.showMessageBoxSync(mainWindow, {
             type: 'question',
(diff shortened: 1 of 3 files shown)


stonerl:
CleanShot 2023-08-23 at 00 11 52

src/main/gui/main-window.ts
```diff
@@ -190,38 +182,78 @@ const registerEventHandlerPostSetup = (
...
 
+    ipcMain.on('reboot-after-save', () => {
+        app.relaunch();
+        app.exit();
+    });
+
+    const handleUnsavedChanges = (
+        event: Electron.Event,
+        callback: () => void,
+        restartCallback?: () => void
```
joeyballentine, line 200: I feel like this should be more generic, like `onSave` and `onDontSave`

stonerl: done

src/main/gui/main-window.ts
```diff
@@ -190,38 +182,78 @@ const registerEventHandlerPostSetup = (
...
+            // Save
+            event.preventDefault();
+            onSave();
+        }
+    };
+
+    ipcMain.handle('relaunch-application', (event) => {
+        if (backend.owned) {
+            backend.tryKill();
+        }
```
RunDevelopment, line 230: So when the user selects "Cancel", the backend will be killed the UI will stay open?

stonerl: Oh, thank you for pointing this out.

RunDevelopment (approved):

Issue #2133, opened 2023-08-22
Crop node: input are in the wrong order.


adegerard:
**Information:**
- Chainner version: 0.19.3
- OS: Windows 11

**Description**
regression after having unified the crop nodes.
**0.19.1**
![crop_0 19 1](https://github.com/chaiNNer-org/chaiNNer/assets/137266603/a8f27129-da73-4772-beea-6c72c611fc50)

**0.19.3**
![crop_0 19 3](https://github.com/chaiNNer-org/chaiNNer/assets/137266603/cf8e19fd-33d9-432e-99b4-c4d1265bfe56)

The order of the arguments is not correct.

**Correction**
I made a correction, I create a PR with correction and test project
![crop_0 19 3_corrected](https://github.com/chaiNNer-org/chaiNNer/assets/137266603/9a79927b-10d3-4a3e-a2e9-06d9553a3265)

PR #2134, opened 2023-08-22
Unified crop node

adegerard:
fixe for issue  #2133 
- corrected order of arguments
- added test project
- Checked that migration is correct (before/after unified crop node)

crop_0 19 3_corrected_validated


joeyballentine:
Whoops, thanks for catching this

joeyballentine (approved):

PR #2136, opened 2023-08-23
enable signing on macOS

### LINKED ISSUE #761: “chainner.app” cannot be opened because the developer cannot be verified
Hi, just wondering if you're planning on registering as a developer so we can install it easier on macs. I can't override this settings on my macbook since it's from work. Thank you!

stonerl:
fixes #761


joeyballentine (approved):
Thanks!

PR #2137, opened 2023-08-23
More settings

RunDevelopment:
Main changes:

- Added `SettingsParser` to abstract over different setting representations.
- Use the same structure as for inputs with setting components.
- Use setting components everywhere.

### DIFF
--- a/backend/src/nodes/impl/onnx/session.py
+++ b/backend/src/nodes/impl/onnx/session.py
@@ -11,7 +11,6 @@ def create_inference_session(
     model: OnnxModel,
     gpu_index: int,
     execution_provider: str,
-    should_tensorrt_cache: bool = False,
     should_tensorrt_fp16: bool = False,
     tensorrt_cache_path: str | None = None,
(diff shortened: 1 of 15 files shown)


backend/src/nodes/impl/onnx/session.py
```diff
@@ -21,7 +20,7 @@ def create_inference_session(
                 "TensorrtExecutionProvider",
                 {
                     "device_id": gpu_index,
-                    "trt_engine_cache_enable": should_tensorrt_cache,
+                    "trt_engine_cache_enable": tensorrt_cache_path is not None,
```
joeyballentine, line 23: you made the disabled case empty string, didn't you? How would this be None ever?

RunDevelopment: See `SettingParser#get_cache_location`. That one maps the empty string to None for a cleaner API on the python side.

joeyballentine (approved):

PR #2139, opened 2023-08-24
update pytorch on apple silicon to 2.0.1

### LINKED ISSUE #2138: Black output image using  4x_NMKD-Superscale-SP_178000_G
**Information:**

- Chainner version: chaiNNer 0.19.3
- OS: macOS 12.6.7 (M1)

**Description**
Upscaling images with 4x_NMKD-Superscale-SP_178000_G.pth or 4x_NMKD-SuperscaleV2_46k.pth result in a black output image.
Update1: I tested with both fp16 mode enabled and disabled, same result. I ran the chain in CPU mode and it properly produced an image. So it is probably acceleration related. I re-installed pytorch, but still black images outside of CPU mode. Is there an other upscaler system I should use instead of pytorch?

**Logs**
https://www.dropbox.com/scl/fi/ksm8lhyxvik0pmumjc7a1/chainner_logs.zip?rlkey=vj9fategc42z1uu36miul0j1w&dl=0

Project: 
[upscale.chn.zip](https://github.com/chaiNNer-org/chaiNNer/files/12423991/upscale.chn.zip)

Models:
https://nmkd.de/shared/ESRGAN/Models/Realistic%2C%20Multipurpose/4x_NMKD-Superscale-SP_178000_G.pth
https://nmkd.de/shared/ESRGAN/Models/Realistic%2C%20Multipurpose/4x_NMKD-SuperscaleV2_46k.pth

stonerl:
fixes #2138

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/__init__.py
+++ b/backend/src/packages/chaiNNer_pytorch/__init__.py
@@ -25,20 +25,20 @@ else:
 
 
 def get_pytorch():
-    # 1.13.1 can take advantage of MPS
+    # 2.0.1 can take advantage of MPS and solves some issues with black output
     if is_arm_mac:
         return [
             Dependency(
                 display_name="PyTorch",
                 pypi_name="torch",
-                version="1.13.1",
-                size_estimate=140 * MB,
+                version="2.0.1",
+                size_estimate=55.8 * MB,
             ),
             Dependency(
                 display_name="TorchVision",
                 pypi_name="torchvision",
-                version="0.14.1",
-                size_estimate=1.3 * MB,
+                version="0.15.2",
+                size_estimate=1.4 * MB,
             ),
         ]
     if python_version.minor < 10:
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I think we should update everything to PyTorch 2.0 all at once. Right now, PyTorch could be completely broken on mac and we wouldn't know, because our CI doesn't test this.

stonerl:
Shall I close it then?

stonerl:
I did some basic testing, and it seems to be working fine. That doesn't replace CI, though.

image

RunDevelopment:
> Shall I close it then?

Not sure.

@joeyballentine What's the state of PyTorch 2.0 again? Did we want to wait for 2.1? 

Also, if we do go for 2.0 (or 2.1), we should probably add support for py 3.11 as well since that requires >=2.0.

joeyballentine:
I want to update to 3.11 and 2.0 at the same time. Ideally we'd wait until 2.1 since that comes out in October but I suppose we could also just update to that when it comes out instead of waiting. 

Just give me a little time and I'll work on what's necessary for 3.11. The reason we need to do both is that 2.0 is slower than the current version we use. For Mac that might not matter, but I don't want to have a major version difference between platforms and accidentally cause issues from that

joeyballentine:
Closing in favor of https://github.com/chaiNNer-org/chaiNNer/pull/2143

PR #2144, opened 2023-08-26
Use Python 3.11 for integrated Python

joeyballentine:
It's a bit hardcodey, but it works. Basically, I just check the version of the existing integrated python during setup, and if it's a version between 3.9 and 3.10 we get rid of it and do everything else like normal.

Everything seems to just work on 3.11 (except pytorch obviously, but that's been resolved by #2143).

### DIFF
--- a/src/main/python/integratedPython.ts
+++ b/src/main/python/integratedPython.ts
@@ -18,20 +18,20 @@ interface PythonDownload {
 
 const downloads: Record<SupportedPlatform, PythonDownload> = {
     linux: {
-        url: 'https://github.com/indygreg/python-build-standalone/releases/download/20230726/cpython-3.11.4+20230726-x86_64-unknown-linux-gnu-install_only.tar.gz',
-        version: '3.11.4',
+        url: 'https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-install_only.tar.gz',
+        version: '3.11.5',
         path: 'python/bin/python3.11',
     },
     darwin: {
         url: isArmMac
-            ? 'https://github.com/indygreg/python-build-standalone/releases/download/20230726/cpython-3.11.4+20230726-aarch64-apple-darwin-install_only.tar.gz'
-            : 'https://github.com/indygreg/python-build-standalone/releases/download/20230726/cpython-3.11.4+20230726-x86_64-apple-darwin-install_only.tar.gz',
-        version: '3.11.4',
+            ? 'https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-aarch64-apple-darwin-install_only.tar.gz'
+            : 'https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-x86_64-apple-darwin-install_only.tar.gz',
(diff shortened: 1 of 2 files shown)


src/main/python/integratedPython.ts
```diff
@@ -61,6 +76,21 @@ export const getIntegratedPython = async (
...
+    const legacyPythonBinExists = await checkFileExists(legacyPythonPath);
+
+    if (legacyPythonBinExists) {
+        const confirmLegacyPythonVersion = await getPythonVersion(legacyPythonPath);
+        if (gt(confirmLegacyPythonVersion, '3.9.0') && lt(confirmLegacyPythonVersion, '3.10.0')) {
+            const legacyPythonFolder = path.resolve(path.join(directory, '/python'));
+            // Remove legacy integrated python
+            await fs.rm(legacyPythonFolder, { recursive: true, force: true });
+        }
+    }
```
RunDevelopment, line 92: This check is kinda dumb.

1. On linux and mac, `checkFileExists` is enough to verify that it is an old version.
2. On windows, we just throw away the python version, and do all these checks (file exists, get version) again.

---

Frankly, I don't think that this approach is all that good in general. I doubt that this will be the only python migration we'll ever do, so we might as well set up a system that makes it easy for us to migrate. How about this:

1. Since the download links are hard-coded, hard-code the python version as well.
2. Optimistically check whether a python exe exists and whether its version is the hard-coded version.
    1. If yes, return the python info.
    2. If no, go to step 3.
3. Delete `<app dir>/python/python` and download the integrated python.
4. Check for the python exe again.

The main advantages of this approach are that 1) we can completely ignore the file structure of old python installations and 2) we do the minimal amount of work if the correct python version is installed.

src/main/python/integratedPython.ts
```diff
@@ -61,40 +67,49 @@ export const getIntegratedPython = async (
...
-        log.info(`Integrated Python not found at ${pythonPath}`);
+    if (pythonBinExists) {
+        const pythonInfo = await checkPythonPaths([pythonPath]);
+        if (eq(pythonInfo.version, version)) {
+            return pythonInfo;
+        }
+        // Invalid version, remove legacy integrated python
+        const legacyPythonFolder = path.resolve(path.join(directory, '/python'));
+        await fs.rm(legacyPythonFolder, { recursive: true, force: true });
+    }
```
RunDevelopment, line 83: The `/python` folder might exist even if the python bin doesn't. So you always need to do the `rm -rf`.

src/main/python/integratedPython.ts
```diff
@@ -61,40 +67,49 @@ export const getIntegratedPython = async (
...
+    const { url, version, path: relativePath } = downloads[platform];
 
+    const pythonPath = path.resolve(path.join(directory, relativePath));
     const pythonBinExists = await checkFileExists(pythonPath);
 
-    if (!pythonBinExists) {
-        log.info(`Integrated Python not found at ${pythonPath}`);
+    if (pythonBinExists) {
+        const pythonInfo = await checkPythonPaths([pythonPath]);
+        if (eq(pythonInfo.version, version)) {
```
RunDevelopment, line 77: Minor thing: Could you `import semver from 'semver';`, so that this line reads `semver.eq`? I think that would be more readable.

RunDevelopment (approved):

PR #2147, opened 2023-08-26
Update opencv and numpy 

joeyballentine:
This version of numpy is basically the highest and lowest we can go without causing conflicts. Requires numba to be updated, which is included in #2146. Also requires me to fix a bunch of type issues


RunDevelopment:
Are we going with a conversion function or not? I thought we agreed on that yesterday.

joeyballentine:
Oh yeah, I'll do that today

joeyballentine:
Btw in case you didn't see on discord, we don't need a conversion function. Mats are just ndarrays.

backend/src/nodes/impl/blend.py
```diff
@@ -135,10 +135,10 @@ def __overlay(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
         return np.where(b < 0.5, 2 * b * a, 1 - 2 * (1 - b) * (1 - a))
 
     def __difference(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
-        return cv2.absdiff(a, b)
+        return np.asarray(cv2.absdiff(a, b))
 
     def __negation(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
-        return 1 - cv2.absdiff(1 - b, a)
+        return np.subtract(1, cv2.absdiff(1 - b, a))
```
RunDevelopment, line 141: Why is this different from the rest?

joeyballentine, line 141:
```suggestion
        return 1 - cv2.absdiff(1 - b, a))  # type: ignore
```

backend/src/nodes/impl/blend.py
```diff
@@ -135,10 +135,10 @@ def __overlay(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
         return np.where(b < 0.5, 2 * b * a, 1 - 2 * (1 - b) * (1 - a))
 
     def __difference(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
-        return cv2.absdiff(a, b)
+        return np.asarray(cv2.absdiff(a, b))
 
     def __negation(self, a: np.ndarray, b: np.ndarray) -> np.ndarray:
-        return 1 - cv2.absdiff(1 - b, a)
+        return 1 - cv2.absdiff(1 - b, a))  # type: ignore
```
joeyballentine, line 141:
```suggestion
        return 1 - cv2.absdiff(1 - b, a)  # type: ignore
```

PR #2150, opened 2023-08-26
open .chn files on macOS after first start

### LINKED ISSUE #2131: macOS: Open .chn files on cold start
Now that .chn open on double-click and are associated on macOS we need to make sure that they also open when chaiNNer is still closed. Currently it only starts chaiNNer but doesn't load the file.

It seems the `open-file` event needs to be invoked earlier.

stonerl:
fixes #2131

### DIFF
--- a/src/main/gui/main-window.ts
+++ b/src/main/gui/main-window.ts
@@ -127,11 +127,11 @@ const registerEventHandlerPreSetup = (
     });
 
     if (isMac) {
-        if (process.env.STARTUPFILE) {
+        if (globalThis.startupFile) {
             // Open file with chaiNNer on other platforms
-            const result = openSaveFile(process.env.STARTUPFILE);
+            const result = openSaveFile(globalThis.startupFile);
             ipcMain.handle('get-cli-open', () => result);
-            delete process.env.STARTUPFILE;
+            globalThis.startupFile = null;
         } else {
             ipcMain.handle('get-cli-open', () => undefined);
         }
--- a/src/main/gui/main-window.ts
+++ b/src/main/gui/main-window.ts
@@ -126,20 +126,9 @@ const registerEventHandlerPreSetup = (
         mainWindow.webContents.send('window-blur');
     });
 
-    if (isMac) {
-        // Open file with chaiNNer on macOS
-        app.on('open-file', (event, filePath) => {
-            (async () => {
-                const file = filePath;
-                if (file) {
-                    const result = await openSaveFile(file);
-                    mainWindow.webContents.send('file-open', result);
-                }
-            })().catch(log.error);
-        });
-    } else {
-        // Opening file with chaiNNer on other platforms
+    if (!isMac) {
         if (args.file) {
+            // Open file with chaiNNer on other platforms
             const result = openSaveFile(args.file);
             ipcMain.handle('get-cli-open', () => result);
         } else {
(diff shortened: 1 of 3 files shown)


src/main/gui/main-window.ts
```diff
@@ -126,39 +125,39 @@ const registerEventHandlerPreSetup = (
...
-            })().catch(log.error);
-        });
+    if (args.file) {
+        // Open file with chaiNNer on other platforms
+        const result = openSaveFile(args.file);
+        ipcMain.handle('get-cli-open', () => result);
+    } else if (process.env.STARTUPFILE) {
+        // Open file with chaiNNer on macOS when app was closed
+        const result = openSaveFile(process.env.STARTUPFILE);
+        ipcMain.handle('get-cli-open', () => result);
```
RunDevelopment, line 144: This is a race condition. You are racing the app loading enough to register events and the `app.on('open-file'` in `src/main/main.ts`.

stonerl: How do you mean? The `app.on` event in main gets always executed first and just sets the environment variable.

I tried different approaches that is the only one that works.

RunDevelopment: > The `app.on` event in main gets always executed first

There is no guarantee that this happens. `app.on` is for asynchronous events, so `open-file` may be fired whenever. Of course, it will likely be fired near the beginning of the application start up, but that doesn't solve anything. When the main window loads quickly enough, when we speed up the start-up process in the future, or when we register other events sooner, this will stop working.

You are relying on the order of 2 asynchronous events without any synchronization mechanism between them. That's a race condition.

stonerl: Thanks for the feedback. But how should this be done, then? The documentation says on macOS, one needs to listen for this event before the `app.on('ready'`.
[https://www.electronjs.org/docs/latest/api/app#event-open-file-macos](https://www.electronjs.org/docs/latest/api/app#event-open-file-macos)

There are only two spots where I could do this. The first is in the `main.ts` (as per this PR) and the second one is `create.ts`. If I listen to this event later (like we do currently), then the event has already disappeared.

stonerl:
Made some changes.

To receive the file path on startup, `app.on('open-file'` gets registered in `main.ts`, but the listener will be removed immediately. Otherwise, there would be two listeners for the `app.on('open-file'`, later on.

Furthermore, `env.STARTUPFILE` gets deleted, so we don't have a dangling env variable, and the `app.on('open-file'` listener for runtime gets registered afterward.

stonerl:
@RunDevelopment 

I revised it. We basically wait until the main window gets opened. If the `open-file`event hasn't fired until then. The first listener is removed, and the new listener takes over. Is this sensible?

src/main/gui/main-window.ts
```diff
@@ -379,6 +368,28 @@ export const createMainWindow = async (args: OpenArguments) => {
             return;
         }
 
+        if (isMac) {
+            if (process.env.STARTUPFILE) {
+                // Open file with chaiNNer on other platforms
+                const result = openSaveFile(process.env.STARTUPFILE);
+                ipcMain.handle('get-cli-open', () => result);
```
RunDevelopment, line 375: Would it be possible to move the whole `if (isMac)` block back into `registerEventHandlerPreSetup`?

These ICP handlers must be registered for the frontend to be able to call them, and right now, it is not clear that `get-cli-open` is always handled.

src/main/gui/main-window.ts
```diff
@@ -379,6 +368,28 @@ export const createMainWindow = async (args: OpenArguments) => {
             return;
         }
 
+        if (isMac) {
+            if (process.env.STARTUPFILE) {
```
RunDevelopment, line 372: Could we use a global variable instead of an untyped process env variable?

RunDevelopment (changes requested):

src/main/gui/main-window.ts
stonerl: Done.

src/main/gui/main-window.ts
stonerl: done

joeyballentine (approved):
This is probably good now. Thanks!

PR #2155, opened 2023-08-27
Add DAT support

joeyballentine:
Not perfect param detection, but close.


joeyballentine (commented):
doesn't need a review

PR #2166, opened 2023-08-30
Fix macOS universal building

joeyballentine:
This isn't exactly a fix, more of a workaround. The universal build both builds and signs in CI when using debug logging. Nobody can explain why, but it works. You probably think "this can't possibly be the case" except it is.


stonerl (approved):
If it works, let's do it. I have no idea why setting the DEBUG env works or what changes, but so be it.

PR #2167, opened 2023-08-30
Resizable Image Previews

### LINKED ISSUE #782: Custom size node (for View Image)

joeyballentine:
Self explanatory. Closes #782

Decisions I had to make:
- It now only encodes the largest image preview, up to 2048x resolution. Maybe in the future we could make this value customizable for performance or something.
- Outputs can now save their size in nodeState

image
image
image


src/common/common-types.ts
```diff
@@ -278,6 +279,7 @@ export interface NodeData {
     readonly inputData: InputData;
     readonly groupState?: GroupState;
     readonly inputSize?: InputSize;
+    readonly outputSize?: OutputSize;
```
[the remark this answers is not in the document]
joeyballentine: Do I have to do that in this PR? Afaik there aren't any nodes that have both a textarea input and an image preview output so the syncing is irrelevant right now, and that feels like a bigger overall change/refactor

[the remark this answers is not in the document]
RunDevelopment: Alrighty.

joeyballentine:
> Why did you remove that anyway?

Because the size of the image itself is now more than just a display thing. If you make the node full res and the UI switches it to the smaller image, it'll shrink the image being displayed. 

I'm not sure how to get around that besides always stretching the image to the bounds of the node, which I'm not sure could be made to retain aspect ratio. I'm open to suggestions

joeyballentine:
I did figure out how to get around that btw. It works pretty well now

src/renderer/components/outputs/LargeImageOutput.tsx
```diff
@@ -1,14 +1,19 @@
...
+import { Size } from '../../../common/common-types';
 import { GlobalVolatileContext } from '../../contexts/GlobalNodeState';
+import { SettingsContext } from '../../contexts/SettingsContext';
 import { useDevicePixelRatio } from '../../hooks/useDevicePixelRatio';
+import { useMemoArray } from '../../hooks/useMemo';
+import { DragHandleSVG } from '../CustomIcons';
 import { OutputProps } from './props';
 
-const IMAGE_PREVIEW_SIZE = 200;
+const IMAGE_PREVIEW_SIZE = 216;
```
RunDevelopment, line 16: Why this specific number?

src/renderer/components/outputs/LargeImageOutput.tsx
```diff
@@ -30,120 +35,251 @@ const pickImage = (previews: PreviewImage[], realSize: number) => {
...
+        useEffect(() => {
+            if (last?.previews) {
+                const biggestImage = last.previews[last.previews.length - 1];
+                const img = new window.Image();
+                img.src = biggestImage.url;
+                img.onload = () => {
+                    setMaxSize({
+                        width: img.width,
+                        height: img.height,
+                    });
```
RunDevelopment, line 72: Q: Does this make sense? Why should users be allowed to make a comically large View Image node?

Practically speaking, it's also not very user-friendly. Image setting up a chain to your liking, including a large View Image node, and then chainner just auto-shrinks the node when you run the chain with a small image. After all, people that work on small pixel art sprites would probably like to see their sprite in big.

RunDevelopment (commented):
Code looks good. Just a few questions.

src/renderer/components/outputs/LargeImageOutput.tsx
joeyballentine: I removed the 8px of padding from both sides, and added it to the image preview size. I noticed there was a bit you could expand the box without making the node bigger, which was a bit odd. So I just made it bigger by default instead.

src/renderer/components/outputs/LargeImageOutput.tsx
joeyballentine: So you're suggesting we scale the image to the size (maintaining aspect ratio) no matter what?

joeyballentine: And keeping the max size something like 1024x instead?

RunDevelopment: > So you're suggesting we scale the image to the size (maintaining aspect ratio) no matter what?

Yes, but I just realized that we didn't do that before. We just centered small images before. Just centering small images would be fine too.

But I was mainly saying that resizing the LargeImageOutput based on the preview image size is bad.

joeyballentine: Gotcha. Well, I could certainly make it so it's always something like a 216-1024 range and then we just center anything smaller

RunDevelopment: Sounds good.

RunDevelopment:
Found some bugs:

1. Resizing when snap to grid is enabled: With 16px grid size, you cannot resize the image preview back to its smallest size.
2. Output sizes are not saved.

joeyballentine:
> Resizing when snap to grid is enabled: With 16px grid size, you cannot resize the image preview back to its smallest size.

I don't think this is something I can really fix. There's actually a bunch of issues with snap to grid all around chaiNNer, like iterator child nodes popping out of the iterator container. IMO this is a non-issue and the real issue is we need to better support snap to grid somehow

RunDevelopment (approved):

PR #2174, opened 2023-08-31
Separate input height from node width

joeyballentine:
Decided to do this before continuing #2167. I separated input height and node width into separate state values.

I will make all nodes width-resizable in a future PR.


joeyballentine:
I'm guessing I need to make a migration for this? I will do that tomorrow.

joeyballentine:
Gonna merge this so I can work on #2167 with this. I tested the migration and it works fine.

Issue #1527, opened 2023-02-01
Fails to launch using Wayland on Linux


illode:
**Information:**

- Chainner version: any
- OS: Arch Linux

**Description**
Attempting to launch chaiNNer without X11 fails with:
```
❯ chainner
[647261:0201/095909.648082:ERROR:ozone_platform_x11.cc(240)] Missing X server or $DISPLAY
[647261:0201/095909.648114:ERROR:env.cc(255)] The platform failed to initialize.  Exiting.
The futex facility returned an unexpected error code.
[1]    647261 IOT instruction (core dumped)  chainner
```

Electron has supported wayland with the flags `--enable-features=UseOzonePlatform --ozone-platform=wayland` since ~v12 and `--ozone-platform-hint=auto` since ~v18. I also can't test to see if it works since there's no way to pass the arguments to electron.

joeyballentine:
Someone with more Linux experience is going to need to fix this for us

illode:
I don't know anything about packaging electron so I can't experiment much myself, but as far as I understand it should just work if we could pass the arguments electron itself.

Chromium, for instance, works if I just run it with `chromium --ozone-platform-hint=auto` or  `chromium --enable-features=UseOzonePlatform --ozone-platform=wayland`. Discord Canary launches (then crashes) with `discord-canary --enable-features=UseOzonePlatform --ozone-platform=wayland`

If you're able to add them, or can tell me how to add them, I can test it out. Unfortunately, I don't really know how electron works, so I'm not sure how.

[Here's](https://github.com/electron/electron/issues/10915#issuecomment-789068112) an electron issue with more info.

PR #2178, opened 2023-09-01
Make it possible to run under Wayland

illode:
Fixes [#1527](https://github.com/chaiNNer-org/chaiNNer/issues/1527) by adding the chromium flags needed to run under Wayland.

### Flags
Since chromium parses the flags regardless of chainner, preventing the arg parser from throwing an error is enough to make the flags work. Technically, this can be done in v0.19.4 already using `chainner open '' --ozone-platform-hint=auto`, but that's cumbersome and not obvious.

Ideally `--ozone-platform-hint` should be enabled by default, but I couldn't find a way to pass it to electron outside of the CLI. The arg parsing would probably have to be done in a wrapper that calls the real chainner binary for this to work.

**References**:
- [Reasons the flags are needed](https://wiki.archlinux.org/title/Wayland#Command_line_flags)
- [ozone-platform-hint docs](https://github.com/chromium/chromium/blob/main/docs/ozone_overview.md#linux-desktop---x11-waterfall-)

### desktopName package.json field
On Wayland, if the .desktop file name doesn't match the [app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id), the compositor won't be able to find .desktop file and thus can't find the icon. With the `desktopName` field unset, electron falls back to `chaiNNer.desktop` as the app_id, but the Linux makers all package the .desktop file as `chainner.desktop`. Since they don't match, the icon is missing. The `desktopName` field can be used to [manually specify the app_id](https://github.com/electron/electron/pull/34855) so everything works as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "chainner",
   "productName": "chaiNNer",
+  "desktopName": "chainner.desktop",
   "version": "0.19.4",
   "description": "A flowchart based image processing GUI",
   "main": ".webpack/main",
--- a/src/main/arguments.ts
+++ b/src/main/arguments.ts
@@ -75,6 +75,21 @@ export const parseArgs = (args: readonly string[]): ParsedArguments => {
                     'An internal developer option to use a different backend. Do not use this as this is not a stable option and may change or disappear at any time',
                 hidden: true,
             },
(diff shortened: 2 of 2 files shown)


joeyballentine (approved):
I'll trust you that this works. Thanks!

PR #2158, opened 2023-08-28
Improve Video Frame Iterator: codec/container/audio/additional parameters

adegerard:
**Motivation**
- Replace the videotype to let the user choose encoder/container
- Add a textfield to allow the user enter custom parameters #2151 
- The container dropdown is dynamically updated when selecting an encoder

video_frame_iterator

**Limitation and choices**
- No validation of the advanced parameters before execution. _Reason: too many options, parsing in Navi would be very complicated_

**Other**
- complexity of the implementation: it seems it is not currently possible to create a dropdown depending on the encoder value.

**This is a draft, please give feedback (here or on Discord)**

PR #2182, opened 2023-09-03
Throttle iterator progress updates

joeyballentine:
I realized while testing #2158 that the frontend displays a maximum update depth warning when you run an iterator that is so fast it sends constant updates. I used the existing useBatchedCallback hook that we use for regular node updates and applied it to the iterator one. This fixed the problem.

I'm guessing this might also help improve CPU load when running iterators, maybe.


RunDevelopment:
> I realized while testing https://github.com/chaiNNer-org/chaiNNer/pull/2158 that the frontend displays a maximum update depth warning when you run an iterator that is so fast it sends constant updates. I used the existing useBatchedCallback hook that we use for regular node updates and applied it to the iterator one. This fixed the problem.

This doesn't make any sense. `useBatchedCallback` does not reduce the number of calls, it just batches them (= waiting for a few to come in and then executing them all at once). So if too many calls were the problem, how does `useBatchedCallback` help?

joeyballentine:
I'm not really sure how, but it does. I could have made it actually throttle the updates, but this worked so I didn't question it. If you think an actual throttle would be better, I can switch it to that

RunDevelopment:
Yeah, throttle would probably be better. Make sure throttle throws away previous events though. I'm not sure whether it will use the first or last event.

joeyballentine:
done

RunDevelopment (approved):

PR #2185, opened 2023-09-04
change the name for maker-squirrel to app.chainner

stonerl:
- according to the documentation, this is the appId and not the app's name: https://js.electronforge.io/interfaces/_electron_forge_maker_squirrel.InternalOptions.Options.html#name
- since we already us app.chainner on macOS, we should use this across the board.
- also to avoid manually using typing the keys more than on time, we now us as many values from the package-json as possible

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,10 @@
 {
   "name": "chainner",
   "productName": "chaiNNer",
+  "appId": "app.chainner",
   "desktopName": "chainner.desktop",
   "version": "0.19.4",
-  "description": "A flowchart based image processing GUI",
+  "description": "A flowchart-based image processing GUI",
   "main": ".webpack/main",
   "scripts": {
     "start": "electron-forge start -- --devtools",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,162 +1,163 @@
...
-    "use-context-selector": "^1.4.0",
-    "use-debounce": "^8.0.1",
-    "uuid": "^8.3.2",
-    "wcag-contrast": "^3.0.0",
-    "yargs": "^17.5.1"
-  },
-  "optionalDependencies": {
-    "appdmg": "^0.6.6"
-  }
+    "name": "chainner",
```
joeyballentine, line 2: can you undo the formatting change? i think if you just do another `npm i` it should revert it to what it wants. if that doesn't do anything, let me know and we can just leave it

stonerl: Done. Had to adjust my JSON formatter.

package.json
```diff
@@ -27,7 +28,9 @@
     "test": "npm run test:js && npm run test:py",
     "postinstall": "patch-package"
   },
-  "keywords": [],
+  "keywords": [
+
+  ],
```
joeyballentine, line 33:
```suggestion
  "keywords": [],
```

joeyballentine (approved):

PR #2194, opened 2023-09-08
Visually show feature-disabled nodes as disabled

joeyballentine:
when a feature is unavailable:
- lower the opacity to 50%
- prevent drag & dropping the node

thoughts: we probably need to also filter out these nodes from the context menu selector, as well as maybe indicate in the docs that they are unavailable.


RunDevelopment:
1. Presenting an error in exactly the same way we normally present node descriptions is not a good idea. How about we show the node description + "\n\nThis node is currently disabled because the <feature name> feature is not enabled."
    image
1. "prevent drag & dropping the node" Is that actually a good idea? What if you don't have a feature enabled, but want to create a chain for a friend that does?

RunDevelopment (approved):

Issue #2195, opened 2023-09-09
Investigate BC5 DDS


RunDevelopment:
We received a bug report that saving an image with BC5 (UNORM) causes washed out colors. Investigate why this is the case.

Also, add BC5 SNORM.

PR #2198, opened 2023-09-10
Fixed BC5 DDS loading

RunDevelopment:
This fixes chainner decoding BC5 SNORM DDS files incorrectly.

The problem was https://github.com/python-pillow/Pillow/issues/7386, so the fix is to simply read DDS files with texconv. PIL was only the preferred DDS decoder because it's faster, but that doesn't matter when it's incorrect.


RunDevelopment:
Fixes part of #2195.

joeyballentine (approved):

PR #2203, opened 2023-09-11
Set fp16 as default behavior if user has supported Nvidia GPU (or arm mac)

joeyballentine:
Since we're resetting settings, I thought this would be pretty important to do. Whether or not a GPU can fp16 is determined by the arch enum kindly provided by the pynvml package. Now, users won't accidentally have fp16 off.

In the future, we could also do the same kind of thing for bf16 and implement support for that to speed up things like swinir on supported gpus.

### DIFF
--- a/backend/src/gpu.py
+++ b/backend/src/gpu.py
@@ -15,17 +15,6 @@ except nv.NVMLError as e:
 except Exception as e:
     logger.info(f"Unknown error occurred when trying to initialize Nvidia GPU: {e}")
 
-FP16_ABILITY_MAP = {
-    nv.NVML_DEVICE_ARCH_KEPLER: False,
-    nv.NVML_DEVICE_ARCH_MAXWELL: False,
-    nv.NVML_DEVICE_ARCH_PASCAL: False,
-    nv.NVML_DEVICE_ARCH_VOLTA: True,
-    nv.NVML_DEVICE_ARCH_TURING: True,
-    nv.NVML_DEVICE_ARCH_AMPERE: True,
-    nv.NVML_DEVICE_ARCH_ADA: True,
-    nv.NVML_DEVICE_ARCH_HOPPER: True,
-}
-
 
 @dataclass
 class _GPU:
@@ -34,6 +23,30 @@ class _GPU:
     index: int
     handle: int
     arch: int
+
+
+FP16_ARCH_ABILITY_MAP = {
+    nv.NVML_DEVICE_ARCH_KEPLER: False,
+    nv.NVML_DEVICE_ARCH_MAXWELL: False,
+    nv.NVML_DEVICE_ARCH_PASCAL: False,
+    nv.NVML_DEVICE_ARCH_VOLTA: True,
+    nv.NVML_DEVICE_ARCH_TURING: True,
+    nv.NVML_DEVICE_ARCH_AMPERE: True,
+    nv.NVML_DEVICE_ARCH_ADA: True,
+    nv.NVML_DEVICE_ARCH_HOPPER: True,
+    nv.NVML_DEVICE_ARCH_UNKNOWN: False,
+}
+
+
+def can_gpu_fp16(gpu: _GPU):
+    # This generation also contains the GTX 1600 cards, which do not support FP16.
+    if gpu.arch == nv.NVML_DEVICE_ARCH_TURING:
+        # There may be a more robust way to check this, but for now I think this will do.
+        return "RTX" in gpu.name
+    if gpu.arch not in FP16_ARCH_ABILITY_MAP and gpu.arch > nv.NVML_DEVICE_ARCH_HOPPER:
+        # Future proofing. We can be reasonably sure that future architectures will support FP16.
+        return True
+    return FP16_ARCH_ABILITY_MAP[gpu.arch]
 
 
 class NvidiaHelper:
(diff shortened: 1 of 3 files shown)


backend/src/gpu.py
```diff
@@ -22,6 +22,31 @@ class _GPU:
...
+
+def can_gpu_fp16(gpu: _GPU):
+    # This generation also contains the GTX 1600 cards, which do not support FP16.
+    if gpu.arch == nv.NVML_DEVICE_ARCH_TURING:
+        # There may be a more robust way to check this, but for now I think this will do.
+        return "RTX" in gpu.name
+    if gpu.arch not in FP16_ARCH_ABILITY_MAP and gpu.arch > nv.NVML_DEVICE_ARCH_HOPPER:
+        # Future proofing. We can be reasonably sure that future architectures will support FP16.
+        return True
+    return FP16_ARCH_ABILITY_MAP[gpu.arch]
```
RunDevelopment, line 49: I don't like that you assume that `FP16_ARCH_ABILITY_MAP` contains all older archs. It probably does, but why make that assumption at all?

```suggestion
    # Future proofing. We can be reasonably sure that future architectures will support FP16.
    return FP16_ARCH_ABILITY_MAP.get(gpu.arch, gpu.arch > nv.NVML_DEVICE_ARCH_HOPPER)
```

backend/src/gpu.py
```diff
@@ -22,6 +22,31 @@ class _GPU:
...
+    nv.NVML_DEVICE_ARCH_VOLTA: True,
+    nv.NVML_DEVICE_ARCH_TURING: True,
+    nv.NVML_DEVICE_ARCH_AMPERE: True,
+    nv.NVML_DEVICE_ARCH_ADA: True,
+    nv.NVML_DEVICE_ARCH_HOPPER: True,
+    nv.NVML_DEVICE_ARCH_UNKNOWN: False,
+}
+
+
+def can_gpu_fp16(gpu: _GPU):
```
RunDevelopment, line 41: Same here with the name.

backend/src/packages/chaiNNer_onnx/settings.py
```diff
@@ -66,12 +67,16 @@ def get_providers():
         )
     )
 
+    should_fp16 = False
+    if nv is not None:
+        should_fp16 = nv.get_can_fp16()
```
RunDevelopment, line 72: `nv?.get_can_fp16() ?? False`

One can only dream 😢

backend/src/packages/chaiNNer_pytorch/settings.py
```diff
@@ -41,7 +48,7 @@
             if is_arm_mac
             else "Runs PyTorch in half-precision (FP16) mode for less VRAM usage. RTX GPUs also get a speedup."
         ),
-        default=False,
+        default=should_fp16 or is_arm_mac,
```
RunDevelopment, line 51: Why is the `or is_arm_mac` here?

backend/src/packages/chaiNNer_onnx/settings.py
joeyballentine: 😭

backend/src/packages/chaiNNer_pytorch/settings.py
joeyballentine: Because arm Macs support fp16

joeyballentine: I guess I could move this over to the if condition

backend/src/gpu.py
```diff
@@ -57,6 +81,13 @@ def get_current_vram_usage(self, gpu_index=0) -> Tuple[int, int, int]:
 
         return info.total, info.used, info.free
 
+    @property
+    def supports_fp16(self, gpu_index: Union[int, None] = None) -> bool:
```
RunDevelopment, line 85: A property with an optional argument?

backend/src/packages/chaiNNer_pytorch/settings.py
RunDevelopment: Oh, okay.

backend/src/gpu.py
joeyballentine: Oh yeah, oops

backend/src/gpu.py
```diff
@@ -57,6 +81,13 @@ def get_current_vram_usage(self, gpu_index=0) -> Tuple[int, int, int]:
 
         return info.total, info.used, info.free
 
+    @property
```
joeyballentine, line 84:
```suggestion
```

backend/src/packages/chaiNNer_onnx/settings.py
```diff
@@ -66,12 +67,16 @@ def get_providers():
         )
     )
 
+    should_fp16 = False
+    if nv is not None:
+        should_fp16 = nv.supports_fp16
```
joeyballentine, line 72:
```suggestion
        should_fp16 = nv.supports_fp16()
```

backend/src/packages/chaiNNer_pytorch/settings.py
```diff
@@ -32,6 +35,12 @@
     ),
 )
 
+should_fp16 = False
+if nv is not None:
+    should_fp16 = nv.supports_fp16
```
joeyballentine, line 40:
```suggestion
    should_fp16 = nv.supports_fp16()
```

RunDevelopment (approved):

PR #2204, opened 2023-09-11
Add "Overwrite Files" bool to all save nodes

joeyballentine:
New users keep connecting their load image inputs to their outputs and accidentally overwriting their images. With this, you'd now have to explicitly allow that.

### DIFF
--- a/backend/src/packages/chaiNNer_ncnn/ncnn/io/save_model.py
+++ b/backend/src/packages/chaiNNer_ncnn/ncnn/io/save_model.py
@@ -3,7 +3,7 @@ import os
 from sanic.log import logger
 
 from nodes.impl.ncnn.model import NcnnModelWrapper
-from nodes.properties.inputs import DirectoryInput, NcnnModelInput, TextInput
+from nodes.properties.inputs import BoolInput, DirectoryInput, NcnnModelInput, TextInput
 
 from .. import io_group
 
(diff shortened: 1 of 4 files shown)


RunDevelopment:
This is a breaking change. You just broke all chains that rely on files being overwritten (e.g. chains that are part of a user workflow; I have several of those). Even worse, you silently broke them. Chains that now behave differently will not show any errors or warnings (static or during runtime).

BC aside, it's also unusual. Previously, side effect nodes *always* executed their side effects. You changed that. This has the effect (hehe) that chainner will run chains that don't have any effect. E.g. running a simple upscaling chain (Load Image -> Upscale -> Save Image with fixed name) twice will not have any effect on the second run, but the user still has to wait for the image to upscale. Imagine waiting minutes for an upscale to finish only for chainner to throw away the result.

So, yeah. I think this isn't a good idea. I understand that you want to protect new users, but I think that this is the wrong way to do it. 

I think what you really want is a linter. Just like linters for other programming languages, we could have a linter for chainner. A linter is essentially just a program that detect patterns in code (/chains). In this case, you wanted to prevent the pattern of a Save X node being connected to the directory + name of a Load X node.

joeyballentine:
I could write a migration that sets this to true by default for all old chains with these nodes, would that be acceptable?

RunDevelopment:
Doesn't solve the other issue.

joeyballentine:
Ok

PR #2214, opened 2023-09-14
Move "nodrag" on LargeImageOutput to just handle

joeyballentine:


RunDevelopment:
You forgot the resizing handles on the sides (marked in yellow):
image

joeyballentine:
I have no idea how I'm gonna handle that (hehe) so maybe I'll just disable resizing from the side? Otherwise I'll have to absolutely position some empty divs or something

joeyballentine:
I tried, but i couldn't get it to ignore drags on those edges using separate divs. So, i just made it only resize from the corner handle

PR #2217, opened 2023-09-16
PyTorch: fall back to FP32 on macOS when in CPU mode

### LINKED ISSUE #2216: macOS > FP16 Mode > Errors
Using chaiNNer 0.20.0 on a M1 Macbook with the latest Monterey, every upscale attempt (most basic settings/nodes/models) produces an error until 'Use FP16 Mode' (the default) is disabled. 

**Error with all default settings:** 
An error occurred in a PyTorch Upscale Image node:
"compute indices weights _nearest" not implemented for 'Half'

**Error with 'Use CPU Mode' enabled:** 
"slow conv2d cpu" not implemented for 'Half"

> Input values:
> • Image: RGB Image 600x759
> • Model: Value of type
> 'nodes.impl.pytorch.architecture.RRDB.RRDBNet'
> • Tile Size: Auto (estimate)
> • Separate Alpha: No

Just to let you know, I hope it's not somehow my own fault; thanks for everything.

stonerl:
fixes #2216

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/settings.py
+++ b/backend/src/packages/chaiNNer_pytorch/settings.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 
 import torch
+from sanic.log import logger
 
 from api import DropdownSetting, ToggleSetting
 from gpu import get_nvidia_helper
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_pytorch/settings.py
```diff
@@ -61,6 +72,12 @@ class PyTorchSettings:
     use_fp16: bool
     gpu_index: int
 
+    # PyTorch 2.0 on macOS does not support FP16 when using CPU
+    def __post_init__(self):
+        if self.use_cpu and self.use_fp16 and is_arm_mac:
```
joeyballentine, line 77: This actually would apply to anything, not just arm macs
```suggestion
        if self.use_cpu and self.use_fp16
```

stonerl: done

joeyballentine (approved):

PR #2218, opened 2023-09-16
Add ability to limit length of certain iterators

joeyballentine:
Applies to:
- Image file iterator
- Image pairs iterator
- Video frame iterator


RunDevelopment (approved):

RunDevelopment:
Man, I'm looking forward to having full iterators. Then stuff like this can just be [`take`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take).

PR #2225, opened 2023-09-17
Fixed RGBA preview image background

RunDevelopment:
As discussed on discord. This fixes the transparent background issue.

Other changes:
- I also changed the data the backend returns. I removed `size` from previews and replaced them with `width` and `height`.
- Since I keep track of the current size anyway, I made it pick the LOD level based on the current size. So when users increase the size of small previews, they won't see a blury image anymore.


RunDevelopment:
Btw, the use of `aspectRatio` also has another advantage: I noticed that the preview image's size would sometimes change when zooming in or out. This happened because the LOD's size was rounded down, and so it didn't fit perfectly anymore. This isn't an issue anymore.

joeyballentine (approved):
This works really well, thanks for figuring it out 👍

PR #2237, opened 2023-09-21
ncnn: Allow disabling Winograd/SGEMM

JeremyRand:
Should be helpful on memory-constrained systems.

In my test run, disabling both Winograd and SGEMM decreased memory usage for a CPU upscale from 10.3 GiB to 4.9 GiB, at the cost of increasing inference time from 58s to 3m0s.

I'm a little uneasy with the parameter count of `create_ncnn_net`; I think it might be better if it accepted a single `NcnnSettings` object instead of each setting being a separate parameter. Happy to make that refactor if you concur it would be better.

### DIFF
--- a/backend/src/nodes/impl/ncnn/session.py
+++ b/backend/src/nodes/impl/ncnn/session.py
@@ -12,12 +12,12 @@ except ImportError:
 
     use_gpu = False
 
+from packages.chaiNNer_ncnn.settings import NcnnSettings
+
 from .model import NcnnModelWrapper
 
 
-def create_ncnn_net(
-    model: NcnnModelWrapper, gpu_index: int, winograd: bool, sgemm: bool
-) -> ncnn.Net:
+def create_ncnn_net(model: NcnnModelWrapper, settings: NcnnSettings) -> ncnn.Net:
     net = ncnn.Net()
 
     if model.fp == "fp16":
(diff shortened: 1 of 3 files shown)


RunDevelopment:
> I'm a little uneasy with the parameter count of `create_ncnn_net`; I think it might be better if it accepted a single `NcnnSettings` object instead of each setting being a separate parameter. Happy to make that refactor if you concur it would be better.

Agreed. That would be nice.

This will also help with a follow-up PR we'll have to do. I just noticed that `get_ncnn_net` is incorrect (caching doesn't account for possibly different settings), and having a single settings object would make the fix easier. Don't worry, that will be its own PR.

JeremyRand:
> Agreed. That would be nice.

Done; hopefully this looks mostly OK.

> This will also help with a follow-up PR we'll have to do. I just noticed that `get_ncnn_net` is incorrect (caching doesn't account for possibly different settings), and having a single settings object would make the fix easier. Don't worry, that will be its own PR.

Indeed, I was going to ask if that was a bug, as this also occurred to me when I was looking at that area of the code. Agreed that the refactor should make fixing that less bothersome.

RunDevelopment (approved):
Code looks good. 

@joeyballentine @theflyingzamboni could you please see if this works correctly?

JeremyRand:
@joeyballentine Anything I can do to help speed up review?

JeremyRand:
No worries, I'm pretty sure this isn't the most important PR in the last few weeks. I'm fine with waiting.

joeyballentine (approved):
Looks like it works fine

joeyballentine:
Sorry it took so long to merge

JeremyRand:
Not a problem, thanks for the merge!

PR #2239, opened 2023-09-22
Fix bug from read-only image in CAS

joeyballentine:
Fix this issue: https://discord.com/channels/930865462852591648/930875396277280788/1154565428547424387

### DIFF
--- a/backend/src/nodes/impl/cas.py
+++ b/backend/src/nodes/impl/cas.py
@@ -37,7 +37,7 @@ def create_cas_mask(img: np.ndarray, kernel, bias: float = 2) -> np.ndarray:
     min_l = cv2.erode(l, kernel)
     max_l = cv2.dilate(l, kernel)
     min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
-    max_l += 1e-8 # type: ignore
+    max_l += 1e-8  # type: ignore
     min_d /= max_l
     mask = min_d
     if bias != 1:
--- a/backend/src/nodes/impl/cas.py
+++ b/backend/src/nodes/impl/cas.py
@@ -37,7 +37,7 @@ def create_cas_mask(img: np.ndarray, kernel, bias: float = 2) -> np.ndarray:
     min_l = cv2.erode(l, kernel)
     max_l = cv2.dilate(l, kernel, dst=l)
     min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
-    max_l += 1e-8  # type: ignore
+    max_l = np.add(max_l, 1e-8)
     min_d /= max_l
     mask = min_d
     if bias != 1:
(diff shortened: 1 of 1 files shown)


backend/src/nodes/impl/cas.py
```diff
@@ -37,7 +37,7 @@ def create_cas_mask(img: np.ndarray, kernel, bias: float = 2) -> np.ndarray:
     min_l = cv2.erode(l, kernel)
     max_l = cv2.dilate(l, kernel, dst=l)
     min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
-    max_l += 1e-8  # type: ignore
+    max_l = np.add(max_l, 1e-8)
```
RunDevelopment, line 40: This fix is not correct. For context:

```py
    l = _luminance(img)
    min_l = cv2.erode(l, kernel)
    max_l = cv2.dilate(l, kernel, dst=l)
    min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
    max_l = np.add(max_l, 1e-8)
```

The issue is that this code assumes that `l` is its own array (and not just a slice of the (readonly) input image). The fix is incorrect because `cv2.dilate(l, kernel, dst=l)` apparently just ignores that `l`'s underlying array is readonly, so we still modify the input image.

The correct fix is:

```diff
-max_l = cv2.dilate(l, kernel, dst=l)
+max_l = cv2.dilate(l, kernel)
```

RunDevelopment (changes requested):

backend/src/nodes/impl/cas.py
```diff
@@ -35,9 +35,9 @@ def create_cas_mask(img: np.ndarray, kernel, bias: float = 2) -> np.ndarray:
 
     l = _luminance(img)
     min_l = cv2.erode(l, kernel)
-    max_l = cv2.dilate(l, kernel, dst=l)
+    max_l = cv2.dilate(l, kernel)
     min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
-    max_l += 1e-8  # type: ignore
+    max_l = np.add(max_l, 1e-8)
```
RunDevelopment, line 40: This can be reverted.

```suggestion
    max_l += 1e-8 # type: ignore
```

backend/src/nodes/impl/cas.py
```diff
@@ -35,9 +35,9 @@ def create_cas_mask(img: np.ndarray, kernel, bias: float = 2) -> np.ndarray:
 
     l = _luminance(img)
     min_l = cv2.erode(l, kernel)
-    max_l = cv2.dilate(l, kernel, dst=l)
+    max_l = cv2.dilate(l, kernel)
     min_d = np.minimum(1.0 - max_l, min_l, out=min_l)  # type: ignore
-    max_l += 1e-8  # type: ignore
+    max_l += 1e-8 # type: ignore
```
joeyballentine, line 40:
```suggestion
    max_l += 1e-8  # type: ignore
```

RunDevelopment (approved):

PR #2241, opened 2023-09-26
Fixed CAS for grayscale images (again)

RunDevelopment:
This fixes the bug in High Boost Filter reported to us on discord.

image


joeyballentine (approved):
Thanks for fixing this

PR #2254, opened 2023-10-12
Iterator Rewrite: Part 1

### LINKED ISSUE #564: Sprite sheet iterator and generator

### LINKED ISSUE #1298: Iterator resizing is not recorded in history
You can't undo resizing an iterator.

Steps to reproduce:
1. Create an iterator node.
2. Fill it with nodes.
3. Resize the iterator such that it is smaller and has to rearrange its children to fit.
4. Ctrl+Z.

image

### LINKED ISSUE #1355: Unable to paste a copied/cut node(s) from the *iterator space outside of it
It keeps inserting inside the iterator no matter how hard I try.

### LINKED ISSUE #1682: Pass a directory list to the image file iterator
**Motivation**
I have several folders that contain image sequences that I want to run through an upscaler. This would be a great job to fire off overnight, but I can only do one directory at a time.

**Description**
I would like a directory list node to allow me to select multiple directories. This node would pass the directory name into the Image File Iterator, which would then process the files in those directories. The Directory list node could also expose a folder name that could be passed into the save file path, for example.

### LINKED ISSUE #1731: 'Save Image' node can overlap Image Directory Iterator
**Information:**

- Chainner version: 0.18.5
- OS: Windows 11
**Description**
It is possible for the save image node to extend outside the bounds of a directory iterator node, overlapping it.
As seen here:

image

### LINKED ISSUE #2120: While you change of Chain using a Model File Iterator node, size of previous node is kept in memory
**Information:**

- Chainner version: 0.19.3
- OS: Windows 11 Pro 22H02

**Description**
Using two Chains having a Model File Iterator node

If I close ChaiNNer and load chain A it looks like this
A

If I close ChaiNNer and load chain B it looks like this
B

If I close ChaiNNer and load chain A then next chain B, size of model file iterator node is messed up (keeping A size)
A Then B

If I close ChaiNNer and load chain B then next chain A, size of model file iterator node is messed up (keeping B size)
B then A

Probably some variable not cleaned correctly.

The two models (rename .chn.txt to .chn)
[A.chn.txt](https://github.com/chaiNNer-org/chaiNNer/files/12383423/A.chn.txt)
[B.chn.txt](https://github.com/chaiNNer-org/chaiNNer/files/12383424/B.chn.txt)

joeyballentine:
Part 1 of the Iterator Rewrite ™️. While not everything we want to be able to accomplish out of iterators, it works so much better than the current system.

Improvements:
- No more giant sub-editor behemoth nodes getting in your way
- Iterators and collectors can now be mixed and matched, for example to convert an image sequence to video
- Visual indicator on handles when something is iterable or collects an iterable sequence

Drawbacks:
- While you can mix and match iterators with collectors, you can only use one iterator per "lineage"
- no explicit mapping operations, or fancy things like take/every or filtering
- No type system information for iterators
- Still doesn't allow nodes to expand or contract the amount of items in an iterable, so frame interpolation is still not viable

Open Questions:
- Are the names of the nodes good?

Things that should be extensively tested:
- Regression of functionality. These should be identical to existing iterators
- Migration of old iterators to new
- The video iterator is a bit different, so we need to be sure I didn't break anything

Future work:
- There's a lot of old iterator code that I left in. I kinda don't wanna bog this PR down with removing all that, so I plan on doing that in a separate PR
- Expanding this system to fix the drawbacks

Issues that get closed as a result of this:
- Closes #2120
- Closes #1731
- Closes #1682
- Closes #1298
- Closes #1355
- Closes #564

image


src/renderer/components/Handle.tsx
```diff
@@ -14,6 +15,7 @@ interface HandleElementProps {
     isValidConnection: (connection: Readonly<Connection>) => boolean;
     validity: Validity;
     id: string;
+    nodeType: NodeType;
```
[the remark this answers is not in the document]
joeyballentine: I actually think it would be better for it to be derived from the Navi type eventually. So does this actually matter?

RunDevelopment:
Found some bugs.

1. You cannot connect to an already-connected input that is downstream of an iterator. Example:
image

2. Non-iterator outputs can be connected to iterator inputs.
image
image

joeyballentine:
> Non-iterator outputs can be connected to iterator inputs.

Without type information I dont think that can be avoided. I could just automatically convert those to arrays I guess?

RunDevelopment:
> I could just automatically convert those to arrays I guess?

Let's not make a bug a feature, okay? Since we should make iterators a type in the type system anyway, we can fix it in later PRs I guess.

On that note: we should make issues for the iterator things we want to change/fix.

src/renderer/components/Handle.tsx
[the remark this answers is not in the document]
RunDevelopment: Yeah, you're right.

backend/src/process.py
```diff
@@ -98,20 +116,41 @@ def enforce_output(raw_output: object, node: NodeData) -> Output:
     return output
 
 
-def run_node(node: NodeData, inputs: Iterable[object], node_id: NodeId) -> Output:
-    assert node.type == "regularNode" or node.type == "iteratorHelper"
+def run_node(
+    node: NodeData, inputs: Iterable[object], node_id: NodeId
```
RunDevelopment, line 120: `inputs` can't be an `Iterable`. `Iterable`s do not guarantee that they can be iterated multiple times, which we need to do for error messages.

backend/src/process.py
```diff
@@ -440,6 +486,33 @@ async def __process(self, node: Node) -> Output:
...
             await self.__broadcast_data(node_instance, node.id, execution_time, output)
+        elif node_instance.type == "newIterator":
+            output, execution_time = await self.loop.run_in_executor(
+                self.pool,
+                timed_supplier(
+                    functools.partial(run_node, node_instance, inputs, node.id)
+                ),
+            )
+            assert isinstance(output, Iterator)
+            # output = next(output_gen)
```
RunDevelopment, line 497: Lots of commented-out code everywhere.

backend/src/process.py
```diff
@@ -538,30 +609,218 @@ def __create_node_finish(self, node_id: NodeId) -> Event:
...
         output_nodes: List[NodeId] = []
         for node in self.chain.nodes.values():
             # we assume that iterator node always have side effects
-            side_effects = isinstance(node, IteratorNode) or node.has_side_effects()
+            side_effects = (
+                isinstance(node, IteratorNode)
+                or isinstance(node, NewIteratorNode)
+                or isinstance(node, CollectorNode)
+                or node.has_side_effects()
+            )
```
RunDevelopment, line 629: Please don't assume that the new iterators and collectors have side effects and declare side effects for them just like for regular nodes. Assuming that (old) iterator nodes have side effects was just a hack to make the old system work, let's not carry it over.

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -937,6 +944,63 @@ export const GlobalProvider = memo(
...
+                    targetDownstreamIterNodes.has(sourceNode.id) ||
+                    targetUpstreamIterNodes.has(sourceNode.id);
+
+                if (
+                    sourceHasIteratorLineage &&
+                    targetHasIteratorLineage &&
+                    !sourceAndTargetShareSameLineage
+                ) {
+                    return invalid('Cannot connect two nodes with unrelated iterator lineage.');
+                }
```
RunDevelopment, line 1002: This is just wrong. Downstream nodes have nothing to do with the lineage of a node.

Right now (with our limited iterator system), it's very easy to figure out the lineage of a node. Let's represent lineage as the node id of an iterator node. Then the lineage of any node can calculate as follows:
```js
function getLineage(node): NodeId | undefined {
  if (isIterator(node)) return node.id;
  if (isCollector(node)) return undefined;
  assert(isRegularNode(node));
  return node.inputNodes.find(n => getLineage(n))
}
```

backend/src/nodes/properties/inputs/file_inputs.py
```diff
@@ -112,8 +112,8 @@ class DirectoryInput(BaseInput):
 
     def __init__(
         self,
-        label: str = "Base Directory",
-        has_handle: bool = False,
+        label: str = "Directory",
+        has_handle: bool = True,
```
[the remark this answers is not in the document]
joeyballentine: The only nodes that use it are the iterators and the save nodes. All of these are now fine to have directories passed in.

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: I don't think this is correct. If you don't include downstream nodes, then you would be able to attach an iterator node to a node that is attached to an iterator node

backend/src/process.py
joeyballentine: What should I do instead then?

joeyballentine: nvm, i used List

RunDevelopment (approved):
LGTM; We still need to polish iterators a lot, but this is a solid foundation. Let's merge it!

PR #2255, opened 2023-10-13
Add Multi Gaussian filter to Normal Map Generator

RunDevelopment:
This adds a new filter to our Normal Map Generator node: Multi Gaussian. This filter is inspired by the frequency equalizer from [Materialize](https://boundingboxsoftware.com/materialize/). The basic idea is that Multi Gaussian is multiple Gauss kernels combined. The user can adjust the strength of each kernel, which allows them to amplify certain frequencies. Multi Gaussian is essentially a generalization of the Sobel filter to different scales. I went with a gauss kernel, because they are easy to generate, and because my intuition told me that they would work well (which they do).

I also added more documentation to the node.

<details>
<summary>Screenshot</summary>

image

</details>

The scale sliders don't have handles, because I might want to remove them with a custom component in the future, similar to what Materialize has.
image

Here are some results:

| Sobel (3x3) | Multi Gaussian |
| --- | --- |
image | image
image | image
image | image
n out | n out
image | image

### DIFF
--- a/backend/src/nodes/impl/normals/edge_filter.py
+++ b/backend/src/nodes/impl/normals/edge_filter.py
@@ -145,8 +145,6 @@ def create_gauss_kernel(parameters: List[Tuple[float, float]]) -> np.ndarray:
                 s / len(x_offsets) * -np.sign(x)
             )
 
-    print(kernel)
-
     return kernel
 
 
--- a/backend/src/nodes/impl/normals/edge_filter.py
+++ b/backend/src/nodes/impl/normals/edge_filter.py
@@ -1,7 +1,23 @@
+import math
 from enum import Enum
-from typing import Dict, Tuple
+from typing import Dict, List, Tuple
 
 import numpy as np
+
+a = [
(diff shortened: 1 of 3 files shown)


backend/src/nodes/impl/normals/edge_filter.py
```diff
@@ -77,18 +95,77 @@ class EdgeFilter(Enum):
...
+def create_gauss_kernel(parameters: List[Tuple[float, float]]) -> np.ndarray:
+    """
+    Parameters is a list of tuples (sigma, weight).
+    """
+
+    # We will use 2D gauss functions normalized to a volume of 1. Wikipedia
+    # has a nice article about this, so look it up if you want to know more:
+    # https://en.wikipedia.org/wiki/Gaussian_function#Two-dimensional_Gaussian_function
+    #
+    # We will one scaled gauss function for each parameter scaled by its weight.
```
joeyballentine, line 107: I think you missed a word in this

joeyballentine (approved):

PR #2267, opened 2023-10-14
Remove code pertaining to old iterators

joeyballentine:


RunDevelopment (commented):
1. This might be my favorite PR.
2. You were right. Z-ordering still works perfectly even without our custom code.
3. Could you add a migration that removes all the old iterator junk from save data? So all the iterator-related props and our manually-set z-indexes.

joeyballentine:
The iterator-related props (i assume you mean the iterator size and whatnot) is already gone because the old-to-new migration made new nodes rather than updating the existing ones (and i left that out for obvious reasons).

Removing the `zindex` fields was a good call though.

I looked through an old chain and i'm pretty sure that's all we need to remove.

PR #2270, opened 2023-10-16
Fix video iterator

joeyballentine:
An unfortunate flaw i didnt catch with the collector system not having separated inputs, is that connections from iterators aren't evaluated when collectors might need them. Therefore, I had to move the inputs that have possible connections into the on_iterate method where the writer gets initialized.

We really badly need that separate input thing we talked about


RunDevelopment:
Instead of hackily patching Save Video, I think we should just fix collectors. I'll give it a try.

PR #2280, opened 2023-10-18
Mixed iterator and non-iterator inputs/outputs

### LINKED ISSUE #1886: Add general loop
**Motivation**
I want to build a zoom animation on a gigapixel image. In order to do so, I need to scale crop a single image depending on the frame index of the output video. However, I don't want to create a dummy video to do this with the video frame iterator.

**Description**
If I use the gigapixel PNG image as video source, it does not seem to work as the video has no actual length.

Instead, I would want a custom loop, where I could probably define the start and the end number ([frame] index).

**Alternatives**
If it's easier, I wouldn't mind if the 'Video Frame Iterator' would be expanded to handle this scenario as well.

### LINKED ISSUE #2258: Support mixed iterator-non-iterator inputs and outputs

RunDevelopment:
Starts with #2258.
Fixes #1886

### DIFF
--- a/backend/src/api.py
+++ b/backend/src/api.py
@@ -224,10 +224,11 @@ class NodeGroup:
             p_inputs, group_layout = _process_inputs(inputs)
             p_outputs = _process_outputs(outputs)
 
-            run_check(
-                TYPE_CHECK_LEVEL,
-                lambda _: check_schema_types(wrapped_func, p_inputs, p_outputs),
-            )
+            if node_type == "regularNode":
+                run_check(
+                    TYPE_CHECK_LEVEL,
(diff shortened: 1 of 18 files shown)


joeyballentine:
Can you fix the type checking on this?

RunDevelopment:
I disabled the type checking for now. Fixing it will be a pain.

RunDevelopment:
Note: this currently only implements the backend side of things. I haven't touched the frontend at all yet. The backend changes are already complex enough as is, so we can do the frontend in a follow-up PR.

joeyballentine (approved):

PR #2284, opened 2023-10-24
Add type tags for number ranges

RunDevelopment:
image

I added type tags for number ranges. The tags only approximate the actual set of numbers. E.g. the types `0..100` and `0..5 | 95..100` will both be displayed as `0 ~ 100`. This simplification is necessary to display more complex number types, and I also think that the loss in information is okay. The bounds are the most important information (IMO) and those are still displayed correctly. 

Small sets of numbers will be displayed as is. E.g. `1 | int(98..100)` will be displayed as `1 | 98 | 99 | 100` instead of as `1 ~ 100`. This will only work for set with 4 or fewer numbers. So e.g. `int(1..4)` will be displayed as `1 | 2 | 3 | 4`, but `int(1..5)` will be displayed as `1 ~ 5`.

**Open questions:**

- Notation. How should we denote this information? I used type notation for small sets (e.g. `1 | 3 | 4`) and a new range notation for ranges with tilde `~`. Maybe we should use the type notations there as well? E.g. `1..100` vs `1 ~ 100`.

### DIFF
--- a/src/renderer/components/TypeTag.tsx
+++ b/src/renderer/components/TypeTag.tsx
@@ -1,13 +1,20 @@
 import {
+    Bounds,
+    IntIntervalType,
+    IntervalType,
     NeverType,
     Type,
+    intInterval,
     isNumericLiteral,
     isStringLiteral,
     isStructInstance,
+    isSubsetOf,
+    literal,
 } from '@chainner/navi';
 import { Tag, Tooltip, forwardRef } from '@chakra-ui/react';
-import React, { memo } from 'react';
+import React, { ReactNode, memo } from 'react';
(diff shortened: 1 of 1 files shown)


joeyballentine:
> How should we denote this information?

I'm not sure what way would be easiest for users to understand, though I certainly think 0..100 is a bit more intuitive than 0 ~ 100. Honestly either would be fine with me

RunDevelopment:
I changed it to `..` and improved the formatting in general. Numbers with a lot of decimal places now only show a few digits. Tooltips still show the full numbers.

image
image

joeyballentine (approved):
LGTM

PR #2287, opened 2023-10-27
Use ruff for linting & formatting python

joeyballentine:
I probably should have done this after #2285 ... oh well.

This switches us over to ruff. I still need to make the necessary changes to CI, but I at least have most of the work complete


backend/src/packages/chaiNNer_standard/material_textures/normal_map/normalize_normal_map.py
```diff
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-import numpy as np
-
 import navi
+import numpy as np
```
RunDevelopment, line 4: Seems like the isort rules in ruff aren't setup correctly. It seems to treat `navi` and `nodes` as third-party imports.

PR #2292, opened 2023-10-30
Add alpha matting node

RunDevelopment:
image
image

---

Changes:
- Added alpha matting node.
- Removed vendored pymatting and added PyMatting and numba as a dependency of chainner_standard. 

Options questions:
- We don't need numba as a direct dependency anymore. Should I remove it, which will make a transitive dependency (pymatting -> numba)?


joeyballentine:
We should keep numba because we've had issues when it tries to install specific versions we aren't expecting. I'd prefer to have its version pinned

RunDevelopment:
Sounds good.

joeyballentine (approved):

PR #2305, opened 2023-11-03
Enable ruff Flake, Naming, RUF, & B rules

joeyballentine:
The flake rules didn't do anything (we probably need to enable the individual ones, but our code seems to just be good enough to have not needed those rules) but the naming ones caught a lot of stuff that violated PEP naming convention.

### DIFF
--- a/backend/src/navi.py
+++ b/backend/src/navi.py
@@ -175,7 +175,7 @@ def match(
     return {"type": "match", "of": of, "arms": arms}
 
 
-def Image(
+def image(
     width: ExpressionJson | None = None,
     height: ExpressionJson | None = None,
     channels: ExpressionJson | None = None,
@@ -203,7 +203,7 @@ def Image(
     return named("Image", fields)
 
 
-def Color(
+def color(
     channels: ExpressionJson | None = None,
     channels_as: ExpressionJson | None = None,
 ) -> ExpressionJson:
--- a/backend/src/navi.py
+++ b/backend/src/navi.py
@@ -175,7 +175,7 @@ def match(
     return {"type": "match", "of": of, "arms": arms}
 
 
-def image(
+def Image(  # noqa: N802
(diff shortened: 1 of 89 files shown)


backend/src/progress_controller.py
```diff
@@ -2,7 +2,7 @@
 from abc import ABC, abstractmethod
 
 
-class Aborted(Exception):
+class AbortedError(Exception):
```
RunDevelopment, line 5: Can we remove whatever rule wants us to add `Error` or `Exception` at the end of all exception classes? I am not a fan of this naming convention because it is useless. Anything that inherits from `Exception` is an error, so why would I need to spell that out again?

backend/src/nodes/impl/upscale/auto_split.py
```diff
@@ -69,10 +71,9 @@ def _exact_split(
     def no_split_upscale(i: np.ndarray, r: Region) -> np.ndarray:
         result = upscale(i, r)
         if isinstance(result, Split):
-            raise _SplitEx
+            raise _SplitExError
         return result
 
-    MAX_ITER = 20
     for _ in range(MAX_ITER):
```
RunDevelopment, line 77: Not a good change. `MAX_ITER` is specific to this one loop, so it should be as close as possible to that loop. I only used the naming conventions for constants, because there is no `const` in python, so there was no way for me to mark this variable as immutable.

backend/src/navi.py
```diff
@@ -175,7 +175,7 @@ def match(
     return {"type": "match", "of": of, "arms": arms}
 
 
-def Image(
+def image(
```
RunDevelopment, line 178: I don't like this. I think functions for specific navi types should mirror the type name. Could we ignore the lint for `image` and `color`?

joeyballentine:
Going to assume this is good now. We can always undo stuff in separate PRs

PR #2309, opened 2023-11-04
Move type-only imports into type checking block (ruff TCH rule)

joeyballentine:
Another draft until #2308 is merged and i can rebase this. It changed a lot of files so i want it to be separate. 

This is something really cool that apparently you can do, and this rule handles it automatically

image

### DIFF
--- a/backend/src/api/api.py
+++ b/backend/src/api/api.py
@@ -4,6 +4,7 @@ import importlib
 import os
 from dataclasses import dataclass, field
 from typing import (
+    TYPE_CHECKING,
     Awaitable,
     Callable,
     Generic,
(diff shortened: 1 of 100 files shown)


joeyballentine:
Huh... I'm not sure exactly what's causing it to fail now, but i guess its good we have CI

RunDevelopment:
What is the benefit of doing this? Cause the code looks kinda horrible... 

Reminds me of how much of a pain ESLint rules for separating type-only imports are for TS. They are annoying because they lock you into a particular usage of an import. E.g. if you import `Foo` and only need it as a type annotation at first, the plugin will make that an `import type { Foo }`. But if you then need `Foo` as a class, you get a compiler error that the plugin doesn't fix.

joeyballentine:
> What is the benefit of doing this? Cause the code looks kinda horrible...

I'm not 100% certain but i think there's performance implications for importing things unnecessarily. It's probably small enough where it doesn't matter, but if we can avoid importing something that imports a bunch of other stuff, that generally seems like a win. It's not like we have a bundler with tree shaking or anything (like TS would have), so ignoring it here isn't exactly equivalent

joeyballentine:
I'll close this for now, we can always come back to it

PR #2313, opened 2023-11-04
Enable Ruff SIM and PERF rules

joeyballentine:
some explanations:
1. String literal unions can be combined into one literal expression
2. list comprehensions are preferred over appending to a list

### DIFF
--- a/backend/src/nodes/impl/color/convert_model.py
+++ b/backend/src/nodes/impl/color/convert_model.py
@@ -10,7 +10,7 @@ from ...utils.utils import get_h_w_c
 
 class ColorSpace:
     def __init__(self, id_: int, name: str, channels: int):
-        assert 0 <= id_ and id_ < 256
+        assert id_ >= 0 and id_ < 256
         self.id = id_
         self.name = name
         self.channels = channels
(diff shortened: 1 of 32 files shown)


joeyballentine:
maybe my find + replace was a bad idea lol

backend/src/nodes/impl/color/convert_model.py
```diff
@@ -10,15 +10,15 @@
 
 class ColorSpace:
     def __init__(self, id_: int, name: str, channels: int):
-        assert 0 <= id_ and id_ < 256
+        assert id_ >= 0 and id_ < 256
```
RunDevelopment, line 13: This is one of those yoda rules, right? Could we disable that one? `0 <= x && x <= 100` and similar is a common pattern to check whether a value is within a range. This code change just makes that pattern harder to recognize.

And no, I don't like the `0 <= x <= 100` python way, because it's python specific and requires that you know python's rules for auto inserting `and`.

backend/src/packages/chaiNNer_standard/material_textures/conversion/metal_to_specular.py
```diff
@@ -43,10 +43,7 @@ def metal_to_spec(
         scaled_albedo = resize(albedo, metal_size, InterpolationMethod.LANCZOS)
     spec = metal3 * scaled_albedo + metal3_inv * 0.22
 
-    if roughness is None:
-        gloss = np.zeros((1, 1), np.float32) + 0.5
-    else:
-        gloss = 1 - roughness
+    gloss = np.zeros((1, 1), np.float32) + 0.5 if roughness is None else 1 - roughness
```
RunDevelopment, line 46: I hate this. Python's inline ifs are already bad enough as is, so please let's not use them everywhere.

PR #2316, opened 2023-11-05
Fix execution buttons disabled state

### LINKED ISSUE #2307: Interface window buttons cannot enter disabled state
**Information:**

- Chainner version: 0.20.2 (20231104 source code)
- OS: Windows 11 pro

**Description**
I encountered a bug when using the chainner project, which caused the interface window buttons to not enter the disabled state.Specifically, when I run a task, the buttons (Run, Pause, Stop) on the interface look all available, but clicking on them will produce an error. I expect that, when a task is running, the Run button should become disabled, and the Pause and Stop buttons should become enabled. When I pause or stop a task, the Run button should become enabled, and the Pause and Stop buttons should become disabled.

**Logs**
main.log
[2023-11-04 12:10:01.241] [info]  Cleaning up temp folders...
[2023-11-04 12:10:15.227] [info]  chaiNNer Version: 0.20.2
[2023-11-04 12:10:15.234] [info]  Attempting to setup backend from http://127.0.0.1:8000...
[2023-11-04 12:10:15.402] [info]  Successfully connected to setup SSE.
[2023-11-04 12:11:40.808] [info]  Cleaning up temp folders...
render.log: N/A

This bug was caused by a breaking change introduced by the chakra ui upgrade .I checked the chakra ui documentation and found that they made some changes to the Button component recently, one of which is changing the disabled property to the isDisabled property.
[https://github.com/chakra-ui/chakra-ui/issues/7269](url)

joeyballentine:
Fixes #2307 

Thanks to hanrai for noticing this in the first place.

And sarcastic thanks to chakra ui for making a breaking change to their props but not their associating TS types 🙄

### DIFF
--- a/src/renderer/components/Header/ExecutionButtons.tsx
+++ b/src/renderer/components/Header/ExecutionButtons.tsx
@@ -27,7 +27,7 @@ export const ExecutionButtons = memo(() => {
                 <IconButton
                     aria-label={t('header.runButton', 'Run button')}
                     colorScheme="green"
-                    disabled={
+                    isDisabled={
                         !(status === ExecutionStatus.READY || status === ExecutionStatus.PAUSED)
                     }
                     icon={<IoPlay />}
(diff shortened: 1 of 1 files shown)


joeyballentine (commented):
doesnt need a review

RunDevelopment:
> And sarcastic thanks to chakra ui for making a breaking change to their props but not their associating TS types 🙄

Tag them coward

joeyballentine:
So it shows up on that issue: https://github.com/chakra-ui/chakra-ui/issues/7269

PR #2321, opened 2023-11-07
Fixed node_check for 3.8

RunDevelopment:
Turned out that *evaluating* `tuple[X, Y]` type hints did not work with Python 3.8. So I did a little AST rewriting to make it work similar to what we do for unions.


joeyballentine:
Or we could drop support for 3.8 😉

RunDevelopment:
Naaaaahhhhhhh

joeyballentine (approved):

Issue #1979, opened 2023-07-23
Fused input-output elements for pass through


RunDevelopment:
This is a design idea that could be used to visually communicate pass-through-enabled nodes (#373).

![image](https://github.com/chaiNNer-org/chaiNNer/assets/20878432/ebe4f34f-6ac7-4108-b28f-b7eeb1b9d645)

<details>
<summary>Current design</summary>

![image](https://github.com/chaiNNer-org/chaiNNer/assets/20878432/70985401-4e57-4e1f-878c-40733885fd58)

</details>

The basic idea is that pass-through input-output pairs should have the same name (e.g. Image) and the same type (e.g. here `ImageOutput(output_type="Input0")`). This allows us to fuse the input and output, since the displayed information for both is practically the same.

RunDevelopment:
We talked on Discord and decided that this visual representation is too confusing.

PR #2338, opened 2023-11-27
Simple fused input-output elements

RunDevelopment:
Progress towards #1979.

This adds support for a very simple kind of fused input-output elements. All it does is to move the output handle into the input and not display the output. Very simple, but even this is useful for nodes that just output their input value. Unfortunately, this does not include the Pass Through node because of the type tag. I plan to change this in follow-up PRs.

It should be noted this is PR only changes visuals. Nothing about execution or what inputs/outputs are changed.

image

### DIFF
--- a/backend/src/api/input.py
+++ b/backend/src/api/input.py
@@ -6,7 +6,7 @@ from typing import Any, Literal, Optional, TypedDict, Union
 
 import navi
 
-from .types import InputId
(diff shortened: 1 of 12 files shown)


joeyballentine (approved):
Very smart implementation just having it be a visual thing. Seems like the perfect solution for it

PR #2342, opened 2023-11-27
ncnn: Allow configuring multithreading

JeremyRand:
Informal test results of tweaking these settings on my POWER9 system:

* 31 threads, 20ms block time (ncnn default): 43% CPU usage, 60s to upscale.
* 4 threads, 20ms block time: 9% CPU usage, 72s to upscale.
* 31 threads, 0ms block time: 24% CPU usage, 57s to upscale. (The 3s improvement is within margin of error, I don't think it actually resulted in a speedup, it's just a CPU usage improvement with no significant effect on performance on my system.)

### DIFF
--- a/backend/src/packages/chaiNNer_ncnn/settings.py
+++ b/backend/src/packages/chaiNNer_ncnn/settings.py
@@ -9,7 +9,7 @@ except ImportError:
 
     use_gpu = False
 
-from api import DropdownSetting, ToggleSetting
+from api import DropdownSetting, NumberSetting, ToggleSetting
 from system import is_arm_mac
 
 from . import package
@@ -34,7 +34,8 @@ if not is_arm_mac and use_gpu:
 
 # Haven't tested disabling Winograd/SGEMM in the ncnn_vulkan fork, so only
 # allow it with upstream ncnn for now. It should work fine regardless of
-# CPU/GPU, but I only tested with CPU.
+# CPU/GPU, but I only tested with CPU. Ditto for multithreading, except it
+# only makes sense for CPU.
 if not use_gpu:
     package.add_setting(
         ToggleSetting(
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_ncnn/settings.py
```diff
@@ -53,12 +54,35 @@
...
     )
 
+    package.add_setting(
+        NumberSetting(
+            label="Thread Count",
+            key="threads",
+            description="Number of threads for NCNN. Only affects CPU mode.",
+            default=ncnn.Net().opt.num_threads,
+            min=1,
+            max=ncnn.Net().opt.num_threads,
```
RunDevelopment, line 64: Could these values be cached or promoted to constants? I don't know how expensive `ncnn.Net()`, so maybe it's not necessary. It would be nice to avoid the code duplication though.

JeremyRand: Good point. Yeah I can change that, hopefully by tomorrow.

JeremyRand: Done. I currently cache these settings regardless of whether upstream ncnn or chaiNNer's Vulkan fork is used; I don't think that will cause any problems (the settings exist in the Vulkan fork, and we don't actually change them when using the Vulkan fork), but if I've missed something and this causes breakage when using the Vulkan fork, let me know.

Also switched the Winograd/SGEMM settings to use the same approach, since there's no reason to duplicate the default settings in chaiNNer's codebase.

RunDevelopment (approved):
Code looks good, but I can't judge whether these changes make sense since I have no idea about NCNN. 

@joeyballentine Could you please take a look at this?

joeyballentine:
It only affects CPU mode which (afaik) is only used by Jeremy so I think it's fine

PR #2354, opened 2023-11-30
Update Spandrel to 0.1.0

joeyballentine:


joeyballentine:
Of course I forgot to actually bump the version number in the package. Oops. Will fix that ASAP

backend/src/nodes/properties/inputs/pytorch_inputs.py
```diff
@@ -57,12 +50,16 @@ def __init__(
             ),
         )
         if torch is not None:
-            self.associated_type = Union[SRModelDescriptor, RestorationModelDescriptor]
+            self.associated_type = ImageModelDescriptor
 
-    def enforce(self, value: object):
+    def enforce(self, value: ModelDescriptor):
```
RunDevelopment, line 55: This is not how `enforce` works. The whole point of `enforce` is to verify that a value is of a certain type, so you can't just declare that it already is that type.

joeyballentine: Whoops. I'll fix that.

PR #2357, opened 2023-11-30
Enforce objects

RunDevelopment:
Addresses [my comment here](https://github.com/chaiNNer-org/chaiNNer/pull/2354#discussion_r1411066934).


RunDevelopment:
Great, another pyright bug.

PR #2359, opened 2023-12-01
Update Spandrel again + fixes

joeyballentine:
My last PR didn't switch to using the new call API, for models like MM-RealSR weren't working. Also, since the call API is type-safe, I was able to see that inpainting models would never be able to be interpolated since we weren't handling their two-image parameter requirement.

I'm unable to use the call API for face inference though, since codeformer has no option for its weight through that. That's not too much of an issue though. 

Side note: We'll probably want to add a method to go alongside the call api that does the same thing. I realized when doing this that when you use the call API, you can't hover over anything to have vscode show you the args of the function.

### DIFF
--- a/backend/src/nodes/impl/pytorch/auto_split.py
+++ b/backend/src/nodes/impl/pytorch/auto_split.py
@@ -4,6 +4,8 @@ import gc
 
 import numpy as np
 import torch
+from sanic.log import logger
+from spandrel import ImageModelDescriptor
 
 from ..upscale.auto_split import Split, Tiler, auto_split
 from .utils import np2tensor, safe_cuda_cache_empty, tensor2np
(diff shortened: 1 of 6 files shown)


backend/src/nodes/impl/pytorch/auto_split.py
```diff
@@ -12,13 +14,13 @@
...
     img: np.ndarray,
-    model: torch.nn.Module,
+    model: ImageModelDescriptor,
     device: torch.device,
     use_fp16: bool,
     tiler: Tiler,
 ) -> np.ndarray:
     model = model.to(device)
-    model = model.half() if use_fp16 else model.float()
+    model.model.half() if use_fp16 else model.model.float()
```
RunDevelopment, line 23: That's kinda cursed.

```suggestion
    if use_fp16:
        model.model.half()
    else:
        model.model.float()
```

joeyballentine:
> Can't you just do __call__() to see that?

Yes. But that's bad python code to write it that way and it's kinds inconvenient to have to temporarily swap to that

RunDevelopment (approved):

PR #2360, opened 2023-12-01
Refactor out onnx conversion + use data for ncnn input name

joeyballentine:
Long overdue. This should make it so that nobody has to [manually edit a param file](https://github.com/upscayl/upscayl/wiki/%F0%9F%96%A5%EF%B8%8F-Model-Conversion%E2%80%89%E2%80%93%E2%80%89Get-more-models!#:~:text=Open%20the%20.param%20file%20and%20change%20the%20first%20%22input%22%20on%20the%20second%20column%20and%20all%20%22input%22s%20(usually%202)%20in%20the%20third%20column%20to%20%22data%22.) to get their models to work with upscayl or the realesrgan binary

### DIFF
--- a/backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
+++ b/backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
@@ -1,0 +1,42 @@
+from io import BytesIO
+
+import torch
+from spandrel import ModelDescriptor
+
+
+def convert_to_onnx_impl(
+    model: ModelDescriptor,
+    device: torch.device,
+    use_half: bool = False,
+    input_name: str = "input",
+    output_name: str = "output",
+) -> bytes:
+    # https://github.com/onnx/onnx/issues/654
+    dynamic_axes = {
+        input_name: {0: "batch_size", 2: "height", 3: "width"},
+        output_name: {0: "batch_size", 2: "height", 3: "width"},
+    }
+    dummy_input = torch.rand(1, model.input_channels, 64, 64)
+    dummy_input = dummy_input.to(device)
+
+    if use_half:
+        model.model.half()
+        dummy_input = dummy_input.half()
+    else:
+        model.model.float()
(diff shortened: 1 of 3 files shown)


backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
```diff
@@ -0,0 +1,42 @@
...
+    # https://github.com/onnx/onnx/issues/654
+    dynamic_axes = {
+        input_name: {0: "batch_size", 2: "height", 3: "width"},
+        output_name: {0: "batch_size", 2: "height", 3: "width"},
+    }
+    dummy_input = torch.rand(1, model.input_channels, 64, 64)
+    dummy_input = dummy_input.to(device)
+
+    if use_half:
+        model.model.half()
```
RunDevelopment, line 23: We should probably throw an error here if the model doesn't support FP16. I know that you check this before calling this function in the nodes somewhere, but that's global reasoning. So let's ensure correctness locally as well.

backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
```diff
@@ -0,0 +1,42 @@
...
+    use_half: bool = False,
+    input_name: str = "input",
+    output_name: str = "output",
+) -> bytes:
+    # https://github.com/onnx/onnx/issues/654
+    dynamic_axes = {
+        input_name: {0: "batch_size", 2: "height", 3: "width"},
+        output_name: {0: "batch_size", 2: "height", 3: "width"},
+    }
+    dummy_input = torch.rand(1, model.input_channels, 64, 64)
```
RunDevelopment, line 19: 64 as to be a valid input size for the model, right? If so, then we might want to round this up to the nearest multiple of `model.size_requirements.multiple_of`.

RunDevelopment (approved):

PR #2363, opened 2023-12-01
Added overlap blending for auto split

RunDevelopment:
This adds smooth blending between neighboring tiles for `auto_split`. This should eliminate many obvious tiling artifacts.

Note that blending is not done as simple linear gradient. It uses a sin function internally to get smoother blending. This function **not** hard-coded. We can use arbitrary functions to get the blending gradient we want. 

The custom blend function we use isn't just a sin. It also discards the outer 50% of the overlap. I.e. for an overlap of 16px, the outer 8px of that overlap are discarded as they are more dissimilar to the neighboring tile than the inner 8px. Doing this virtually eliminates the slight blur blending introduces.

### Demonstration

This is how tiles are blending for a 256x256 image that is being upscaled 4x with a tile size of 64 and overlap of 16. We see that the blending is smooth across all axes.

image

While the above image makes the blending easy to see, this configuration is very unlikely in practice. So are the tiles of a 512x512px image upscaled 4x with a tile size of 256 and an overlap of 16:

image

### Implementation details

The blended final image is iteratively constructed in place. Only the final and the current tile are held in memory. Tiles that have already been added are freed. The final buffer is allocated after we get the first tile in `auto_split`, just like before. 

The only difference is that there are 2 passes. We first blend the tiles of each row of tiles, and then blend the rows. This means that we allocate more memory than before, because we need to keep one whole row in memory.

### Performance

The implementation is about 2x **slower** than the old implementation. 

I measured the overhead of `auto_aplit` (=splitting the image into tiles and then assembling the tiles back together) for 4x upscaling an 1024x1024 image, using a tile size of 256 and an overlap of 16px. The overhead of ~250ms before this change and ~400ms with blending.

While this is quite a bit slower, it also shouldn't matter too much. The overhead of `auto_split` grows with the number of tiles. Since very fast models are typically also require less VRAM, larger tile sizes are viable which reduces the overhead of `auto_split`. Large and slow model typically require more VRAM and therefore require more tiles.

I also optimized a little. The blending itself uses a faster mixing function that avoid temporary arrays. The temporary blend array is reused whenever possible. The result array of rows is also reused whenever possible.

### DIFF
--- a/backend/src/nodes/impl/upscale/tile_blending.py
+++ b/backend/src/nodes/impl/upscale/tile_blending.py
@@ -55,13 +55,28 @@ class TileBlender:
         channels: int,
         direction: BlendDirection,
         blend_fn: Callable[[np.ndarray], np.ndarray] = sin_blend_fn,
+        _prev: TileBlender | None = None,
     ) -> None:
         self.direction: BlendDirection = direction
-        self.result: np.ndarray = np.zeros((height, width, channels), dtype=np.float32)
         self.blend_fn: Callable[[np.ndarray], np.ndarray] = blend_fn
         self.offset: int = 0
         self.last_end_overlap: int = 0
         self._last_blend: np.ndarray | None = None
+
+        if (
+            _prev is not None
+            and _prev.direction == direction
+            and _prev.width == width
+            and _prev.height == height
+            and _prev.channels == channels
+        ):
+            if _prev.blend_fn == blend_fn:
+                # reuse blend
+                self._last_blend = _prev._last_blend  # noqa: SLF001
+            result = _prev.result
+        else:
+            result = np.zeros((height, width, channels), dtype=np.float32)
+        self.result: np.ndarray = result
 
     @property
     def width(self) -> int:
--- a/backend/src/nodes/impl/upscale/tile_blending.py
+++ b/backend/src/nodes/impl/upscale/tile_blending.py
@@ -35,6 +35,18 @@ class TileOverlap:
         return self.start + self.end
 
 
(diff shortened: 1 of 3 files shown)


backend/src/nodes/impl/upscale/tile_blending.py
```diff
@@ -0,0 +1,214 @@
...
+                ] = _fast_mix(left, right, blend)
+
+                self.offset += h - o.total
+                self.last_end_overlap = o.end
+
+    def get_result(self) -> np.ndarray:
+        # if self.direction == BlendDirection.X:
+        #     assert self.offset == self.width
+        # else:
+        #     assert self.offset == self.height
```
joeyballentine, line 212: 🤔

joeyballentine (commented):
I tried this out with some models i know tend to create seams (SCUNet and SwinIR) and i can tell that this is working super well. I'm extremely impressed by how well this works, and it's totally worth the slowdown.

I do wonder if we could speed some of this up with numba though

backend/src/nodes/impl/upscale/tile_blending.py
RunDevelopment: Turns out, assertions that ensure correctness can get in the way when you're trying to see the incorrect result to understand what is wrong. Well, it works now, so I just forgot to uncomment them.

RunDevelopment:
> I do wonder if we could speed some of this up with numba though

I'm not sure. Most of the overhead seems to come from the actual blending (which is good because it means that the rest doesn't add much if any overhead) and I'm already using numpy quite efficiently (I think), so I don't think there's not much room for numba to optimize.

joeyballentine (approved):

PR #2365, opened 2023-12-02
Updated to latest version of chakra ui

RunDevelopment:
I also had to fix 2 minor thing:

- They seem to have fixed `HStack`. It previously applied its `spacing` to `position: absolute` children except for the last one. It not doesn't apply `spacing` to those elements at all. Unfortunately, this also broke the positioning of inputs with handles, so I fixed that.
- They changed the size of number input stepper icons (the up down buttons). They were way too big, so I made them smaller in CSS. Using chakra props like usual wasn't possible.

With those things fixed, it looks just like before.


joeyballentine (approved):
👍

PR #2370, opened 2023-12-02
Fix pointer cursor on NodeSelectorPanel accordion titles

joeyballentine:


RunDevelopment (approved):
I'm just gonna assume this does whatever it is supposed to do, because I don't know what issue this is trying to fix... Adding a few `cursor="pointer"` here and there shouldn't hurt, so sure.

joeyballentine:
@RunDevelopment the part of the node selector panel with the category name + icon would change bg color on hover but not change the cursor unless over the caret part on the right, or on the 1px border around the component. This makes it so the whole thing acts like a button and changes the cursor, since the whole thing is clickable

PR #2371, opened 2023-12-03
Improved normalize normals node

RunDevelopment:
This adds an option to control the B channel of the normalized normals. The 5 options should cover all commonly used values for B channels. What each option does is document in the node docs.

I also renamed the node to be consistent with "Add Normals" and "Convert Normals".

image


joeyballentine:
whoops, didnt realize the schema id stayed the same

joeyballentine (approved):

PR #2377, opened 2023-12-04
Auto-layouting

joeyballentine:
I realized as I was using chaiNNer for cleaning up a family photo that it would be really nice to auto-layout my messy chain, so I implemented it real quick.

This implementation is based on the [elkjs react-flow example](https://reactflow.dev/examples/layout/elkjs). The resulting code isn't the greatest, but it works.

I'd like this to also be a button somewhere or a menu item, but for now it's just ctrl+f (for format) as a proof of concept. I don't know if there would be a better hotkey to set it to if we want to keep a hotkey, otherwise it should be a button trigger.

The original implementation ran FitView after setting the nodes, but that doesn't really work since you can't set the nodes and do a .then after, so I just excluded it. I'd love to do that, but only if we can get it to actually work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "decompress": "^4.2.1",
     "electron-log": "^4.4.8",
     "electron-squirrel-startup": "^1.0.0",
+    "elkjs": "^0.8.2",
     "eventsource": "^2.0.2",
     "fast-deep-equal": "^3.1.3",
     "framer-motion": "^6.3.6",
@@ -155,6 +156,7 @@
     "use-debounce": "^8.0.1",
     "uuid": "^8.3.2",
     "wcag-contrast": "^3.0.0",
+    "web-worker": "^1.2.0",
     "yargs": "^17.5.1"
   },
   "optionalDependencies": {
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -475,29 +475,29 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
         getLayoutedElements(nodes, edges)
             .then((response) => {
                 const { nodes: layoutedNodes } = response;
-
                 if (layoutedNodes) {
-                    const nodesWithTheirPositionsModified = layoutedNodes.map((node) => {
-                        const originalNode = nodes.find((n) => n.id === node.id);
-                        if (!originalNode) {
-                            return undefined;
-                        }
-                        return {
-                            ...originalNode,
-                            position: node.position,
-                        };
+                    changeNodes((nds) => {
+                        const nodesWithTheirPositionsModified = layoutedNodes.map((node) => {
+                            const originalNode = nds.find((n) => n.id === node.id);
+                            if (!originalNode) {
+                                return undefined;
+                            }
+                            return {
+                                ...originalNode,
+                                position: node.position,
+                            };
+                        });
+                        const notUndefined = nodesWithTheirPositionsModified.filter(
+                            (n) => n !== undefined
+                        ) as Node<NodeData>[];
+                        return notUndefined;
                     });
-                    const notUndefined = nodesWithTheirPositionsModified.filter(
-                        (n) => n !== undefined
-                    ) as Node<NodeData>[];
-
-                    setNodes(notUndefined);
                 }
             })
             .catch((error) => {
                 log.error(error);
             });
-    }, [nodes, edges, setNodes]);
+    }, [nodes, edges, changeNodes]);
 
     useHotkeys('ctrl+f, cmd+f', onLayout);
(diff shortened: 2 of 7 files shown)


joeyballentine:
Going with ctrl+shift+f for formatting + menu item, no button. And, I got something acceptable for repositioning the viewport after formatting.

package.json
```diff
@@ -155,6 +156,7 @@
     "use-debounce": "^8.0.1",
     "uuid": "^8.3.2",
     "wcag-contrast": "^3.0.0",
+    "web-worker": "^1.2.0",
```
RunDevelopment, line 159: What do you need web workers for?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -42,10 +43,59 @@ import { SettingsContext } from '../contexts/SettingsContext';
...
+    const layoutedGraph = await elk.layout(graph);
+    return {
+        nodes: layoutedGraph.children?.map((node) => ({
+            ...node,
+            // React Flow expects a position property on the node instead of `x`
+            // and `y` fields.
+            position: { x: node.x, y: node.y },
+        })),
+
+        edges: layoutedGraph.edges,
```
RunDevelopment, line 95: The layout of our edges is determined solely by the connected nodes. So why do you return this here?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -422,6 +472,45 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
         []
     );
 
+    const onLayout = useCallback(() => {
+        getLayoutedElements(nodes, edges)
+            .then((response) => {
+                const { nodes: layoutedNodes } = response;
+                if (layoutedNodes) {
```
RunDevelopment, line 479: The API of `getLayoutedElements` doesn't make sense to me. What does `nodes: undefined` even mean? If it means "error", then why not return a rejecting promise?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -42,10 +43,59 @@ import { SettingsContext } from '../contexts/SettingsContext';
...
+    };
+
+    const layoutedGraph = await elk.layout(graph);
+    return {
+        nodes: layoutedGraph.children?.map((node) => ({
+            ...node,
+            // React Flow expects a position property on the node instead of `x`
+            // and `y` fields.
+            position: { x: node.x, y: node.y },
+        })),
```
RunDevelopment, line 93: I don't like this. You don't clean up the `x` and `y` (which will be saved...) as well as any other props you added for `elk` (or `elk` may add on its own). 

Since you are already passing different objects to `elk`, just give those objects an ID, width, height, and props required by elk. Nothing else. Once the graph is layouted, just copy the positions from the layouted nodes over into the original graph.

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -422,6 +472,45 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
+                            return {
+                                ...originalNode,
+                                position: node.position,
+                            };
+                        });
+                        const notUndefined = nodesWithTheirPositionsModified.filter(
+                            (n) => n !== undefined
+                        ) as Node<NodeData>[];
+                        return notUndefined;
+                    });
```
RunDevelopment, line 495: What is this?

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -422,6 +472,45 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
+                    const minX = Math.min(...layoutedNodes.map((n) => n.position.x || Infinity));
+                    const minY = Math.min(...layoutedNodes.map((n) => n.position.y || Infinity));
```
RunDevelopment, line 497: So nodes are x=0 or y=0 don't count?

package.json
joeyballentine: elkjs throws an error if this isn't installed. I think they forgot to include it as a dependency or something

src/renderer/components/ReactFlowBox.tsx
joeyballentine: Copypasta boilerplate

src/renderer/components/ReactFlowBox.tsx
joeyballentine: Typescript was yelling at me

src/renderer/components/ReactFlowBox.tsx
joeyballentine: Oops

RunDevelopment (approved):

PR #2381, opened 2023-12-05
Add Chroma Key node

### LINKED ISSUE #1786: [ Feature Request ] Color mask node
I needed a way to take a hand created [id matte](https://en.wikipedia.org/wiki/Cryptomatte) and generate different masks from each of the colors.  I would use this mask to apply a particular operation.

Once I have the mask, I think it's easy enough to apply the operation using it but creating the mask from a particular color in a source image is not possible in chaiNNer.  This feature request is to add a node in case someone would like to do something similar!

### LINKED ISSUE #2226: Add a chroma key node
We had people ask before how to remove a constant BG color.

The algorithms the node implements doesn't need to all that good. It just needs to be fast and good enough.

Reference:
- https://discord.com/channels/930865462852591648/1045361136196919366/1149980638477295616

RunDevelopment:
Fixes #1786
Fixes #2226

This adds a Chroma key node. 

image

The node as 2 key modes: binary and trimap matting. While the binary method is just a threshold to decide a binary transparent yes/no, the trimap option will generate a trimap and then perform alpha matting. As such, the trimap option is substantially slower, but can produce pretty good results. Here are some examples:

| Before | After | Settings |
| --- | --- | --- |
|  green coco 5 | image | image
| green stop | image | image
| green alice 1 clean |image |image
| white toki 1 |image |image
| gray original 3 |image |image

Not perfect, but pretty good.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_channel/misc/chroma_key.py
+++ b/backend/src/packages/chaiNNer_standard/image_channel/misc/chroma_key.py
@@ -32,6 +32,7 @@ class KeyMethod(Enum):
     name="Chroma Key",
     description=[
         "Removes a color from an image and replaces it with transparency.",
+        "To set he key color, either define a constant color with the `chainner:utility:color` node or pick a color from the image with the `chainner:image:pick_color` node.",
         "This nodes offers multiple strategies to key the image:",
         "- **Binary**:",
         "    A simple binary thresholding method is used to determine the whether a pixel in the image is transparent or not. This method is mostly useful images *without anti-aliasing*.",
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_standard/image_channel/misc/chroma_key.py
```diff
@@ -0,0 +1,167 @@
...
+    BINARY = 1
+    TRIMAP_MATTING = 2
+
+
+@node_group.register(
+    schema_id="chainner:image:chroma_key",
+    name="Chroma Key",
+    description=[
+        "Removes a color from an image and replaces it with transparency.",
+        "To set he key color, either define a constant color with the `chainner:utility:color` node or pick a color from the image with the `chainner:image:pick_color` node.",
```
joeyballentine, line 35:
```suggestion
        "To set the key color, either define a constant color with the `chainner:utility:color` node or pick a color from the image with the `chainner:image:pick_color` node.",
```

Issue #1035, opened 2022-09-26
Add more resizing filters


emarron:
Add more resizing filters because more options = better.
These are the two types I think would be neat to add:

- general resizers which are used pre or post super-resolution model.
- pixel art scalers which are used independently of the super-resolution models. 

So these are the general resizers already in chaiNNer:

- Nearest Neighbor (point)
- Box
- Bilinear (triangle)
- Bicubic (cubic or catrom??)
- Lanczos

Already in Pillow:

- Hamming

Can be implemented pretty easily:
[Basically everything on this page](https://imagemagick.org/Usage/filter/)
[comparison of filters](https://imagemagick.org/Usage/filter/montage_enlarge.png)
Most desirable of these is probably:

- mitchell
- spline
- lagrange

These are all simple algo to implement as I can just cannibalize the image magic [ 'resize.c'](https://github.com/ImageMagick/ImageMagick/blob/8a1428072f5f24f83113a0b29a330941485cd541/MagickCore/resize.c). I'm guessing adding imagemagick with [wand](https://github.com/emcconville/wand) ctype binding is not desirable.

More complex:
Pixel art scaling algorithms. 
These are different from above, as the above can resize with n scalar, where n = (0 , infinity), these are usually restricted to a single scalar greater than 1. [Wikipedia has a good list](https://en.wikipedia.org/wiki/Pixel-art_scaling_algorithms). I'm guessing this might need its own node?

==

Let me know what you think.

Athari:
Any Hermite fans here? 😁 Blurry or not, it doesn't introduce ringing, and the only currently available option without ringing is bilinear. Not a massive difference in a lot of cases, but still.

RunDevelopment:
Resolved by #2398.

PR #2387, opened 2023-12-06
Use Rust implementation for image resizing

RunDevelopment:
This PR uses the new Rust implementation for resizing and removes the PIL-based `resize` method. Since the Rust version support full float32, I was also able to remove some `limited_to_8bpc=True`, which is nice.

The new `resize` method I made around the Rust version also has a few new features:
- An option for gamma correction.
- An option to separate alpha.
- An OOM safeguard before resizing.

There are no user-facing changes in this PR. I commented out the new interpolation methods for now. They will be enabled in a follow-up PR.

---

Relates to #1035.

### DIFF
--- a/backend/src/nodes/impl/resize.py
+++ b/backend/src/nodes/impl/resize.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+from enum import Enum
+
+import numpy as np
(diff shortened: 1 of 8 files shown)


joeyballentine (approved):
Awesome

PR #2389, opened 2023-12-06
color_transfer_node: Avoid RGB contamination by transparent pixels

Splendide-Imaginarius:
The Color Transfer node had the same bug as the Average Color Fix bug that was fixed in #1907.

This commit adds a similar fix for Color Transfer.


backend/src/packages/chaiNNer_standard/image_filter/correction/color_transfer.py
```diff
@@ -79,17 +79,45 @@ def color_transfer_node(
...
+    # such RGB data is indeterminate.
+    valid_rgb_indices = np.ones(img.shape[:-1], dtype=bool)
+    if alpha is not None:
+        valid_rgb_mask = alpha > 0
+        valid_rgb_indices = np.nonzero(valid_rgb_mask)
+
+    ref_valid_rgb_indices = np.ones(ref_img.shape[:-1], dtype=bool)
+    if ref_alpha is not None:
+        ref_valid_rgb_mask = ref_alpha > 0
+        ref_valid_rgb_indices = np.nonzero(ref_valid_rgb_mask)
```
RunDevelopment, line 100: Why the `np.nonzero`? It works the same without it.

This would also allow us to change the type hints from `np.ndarray | tuple[np.ndarray, ...]` to just `np.ndarray`.

backend/src/nodes/impl/color_transfer/mean_std.py
```diff
@@ -30,7 +30,7 @@ def image_stats(img: np.ndarray):
     """Get means and standard deviations of channels"""
 
     # Compute the mean and standard deviation of each channel
-    channel_a, channel_b, channel_c = cv2.split(img)
+    channel_a, channel_b, channel_c = np.split(img, 3, 1)
```
RunDevelopment, line 33: Why?

backend/src/nodes/impl/color_transfer/linear_histogram.py
```diff
@@ -32,13 +34,16 @@ def linear_histogram_transfer(
...
 
-    mu_content = np.mean(content, axis=0)
-    mu_reference = np.mean(reference, axis=0)
+    valid_content = img[valid_indices]
+    valid_reference = ref_img[ref_valid_indices]
 
-    cov_content = np.cov(content, rowvar=False)
-    cov_reference = np.cov(reference, rowvar=False)
+    mu_content = np.mean(valid_content, axis=0)
+    mu_reference = np.mean(valid_reference, axis=0)
```
RunDevelopment, line 43: You don't use the reshaped images here unlike before.

RunDevelopment (changes requested):

backend/src/nodes/impl/color_transfer/linear_histogram.py
Splendide-Imaginarius: You're suggesting that I use `content` and `reference` instead of `img` and `ref_img` when assigning to `valid_content` and `valid_reference`, correct? (Just want to make sure I correctly follow what you're suggesting before I actually do it.)

backend/src/nodes/impl/color_transfer/mean_std.py
Splendide-Imaginarius: If I recall correctly, `cv2.split` errored here because I already reshaped `img` into a 1D image (i.e. a list of pixels), and it expected a 2D image (i.e. a matrix of pixels). If there's some other way you think I should handle that, I'm open to suggestions.

backend/src/packages/chaiNNer_standard/image_filter/correction/color_transfer.py
Splendide-Imaginarius: Hmm, good question. I don't think I had a good reason for doing it this way. I'll try removing the `np.nonzero` and see if it still works OK on my end.

backend/src/nodes/impl/color_transfer/linear_histogram.py
RunDevelopment: Either that or remove them. The previously `reference` variable is entirely unused and `ref_img.reshape(-1, shape[-1])` is a pure function call.

backend/src/nodes/impl/color_transfer/mean_std.py
RunDevelopment: Why is it 1D in the first place? If what you said is correct (I haven't checked and don't know `cv2.split` by heart), then it wasn't 1D before.

Splendide-Imaginarius:
Very sorry for letting this PR bitrot for months, I will try to get the feedback resolved soon.

backend/src/nodes/impl/color_transfer/mean_std.py
Splendide-Imaginarius: I can't remember for sure why I did that (due to this PR being a few months old by now) but I think it was because filtering out the transparent pixels was more straightforward when in 1D format (I think because different lines will have different numbers of pixels filtered out).

Splendide-Imaginarius: Pretty sure this would be a result of changing `image_stats(img)` to `image_stats(img[valid_indices])`. @RunDevelopment do you want me to add a comment clarifying this?

backend/src/packages/chaiNNer_standard/image_filter/correction/color_transfer.py
Splendide-Imaginarius: Yes, seems to still work; thanks for the spot.

backend/src/nodes/impl/color_transfer/linear_histogram.py
Splendide-Imaginarius: I think it's fixed now?

Splendide-Imaginarius:
@RunDevelopment I think I've addressed all review, what's still needed to get this PR moving forward again?

Splendide-Imaginarius:
@joeyballentine Is there anything I can do to get this moving?

joeyballentine:
Sorry, will review soon

PR #2396, opened 2023-12-11
Add Opset output to Convert ONNX

### LINKED ISSUE #2393: Add opset output to "Convert ONNX"
**Description**
For the same reason the fp output was added, It'd be very nice if there was an output for the current opset type. This would enable easy text appending like "modelname-fp16-opset17"

(this is only relevant to the latest nightly)

RunDevelopment:
Resolves #2393.

image


joeyballentine:
I feel like this would be a good use case for fused inputs and outputs. But as a separate pr

joeyballentine (approved):

PR #2398, opened 2023-12-11
Add categories and dividers for dropdowns

RunDevelopment:
Changes:
- Added support for dropdown groups and dividers. They are implemented as disabled options. This is kinda hacky, but it works.
- Added previously disabled interpolation filters for Resize and Resize to Side.
- Added categories for DDS formats.
- Added dividers for blend modes.

image
image
image


joeyballentine (approved):
Awesome

Issue #838, opened 2022-08-26
Iterator broadcast persistence can break Fill Alpha


theflyingzamboni:
When you use an iterator, Load Image (Iterator) only updates the image info broadcast when you run the chain. Since it can't know image info ahead of time, the validity checker does not catch if you have an RGB image connected to Fill Alpha, which requires an RGBA image. It only errors when you attempt to run the chain and the channel assertion in Fill Alpha fails. The issue here is that once Load Image (Iterator) has been run, the broadcast for the last loaded image persists. So if you run a folder with an RGB image and it fails on Fill Alpha, the load node will continue to output RGB as the type, which causes Fill Alpha to register as invalid. Since it's invalid, you can't run the chain even if the directory is changed and all images in it are valid RGBA images.

To test:
1) Create a simple iterator with Fill Alpha connected to Load Image
2) Input a directory with RGB images and run
3) Change directory to anything, the chain will not run

PR #2402, opened 2023-12-13
Reworked backend events

### LINKED ISSUE #2319: the animation keeps playing
**Information:**

- Chainner version: latest [nightly](https://github.com/chaiNNer-org/chaiNNer-nightly/releases/tag/2023-11-04)
- OS:  Windows 10

**Description**

The work seems to be actually finished but the animation keeps playing.

Snipaste_2023-11-07_12-30-11

**Logs**

[main.log](https://github.com/chaiNNer-org/chaiNNer/files/13275712/main.log)
[renderer.log](https://github.com/chaiNNer-org/chaiNNer/files/13275713/renderer.log)

RunDevelopment:
Fixes #2319

This PR reworks the events the executor sends to the frontend. I changed both, what events are send, how they are sent, the data the events contains, and how and in which order the events are processed by the frontend. The frontend now also keeps track of the overall progress of the chain instead of the backend like before. 

Let's start with the backend changes:
- Added `chain-start` event. This event contains the list of nodes that will actually be executed. Due to caching and optimizations, this list is often different from the list of nodes the frontend sent to the backend. This list will be used for animations and progress by the frontend.
- Moved broadcasts into its own event. This is important for iterators where we do one broadcast for each iteration.
- Move all event code in the executor into `__send_*` methods. This makes the code pretty clean.
- Fixed a bug in the executor whether broadcasts would be sent out even after the execution as aborted.
- Other minor improvements to the executor.
- Removed useless `message` field in some responses of the backend API.

Frontend changes:
- Running individual nodes no longer uses the `animate`/`unAnimate` methods like regular executions. They now have their own state to make sure these two systems don't interfere with each other.
- Edges now derive their animation state from their connected source node. This is not completely correct yet (discussion on discord), but will be soon.
- Iterators now do type narrowing for the current iteration. This means that they display e.g. the image name of the current iteration. I was careful not to re-introduce #838 here. The type narrowing is always removed after the execution complete (successfully or otherwise).
- Overall chain progress is now tracked entirely by the frontend. I implemented the weighted node progress idea I proposed some time ago, but weights aren't assigned in a clever way yet. This should be easy to improve in follow-up PRs. That being said, it's still better that what we had before because it takes iterator progress into account.
- All node events are now processed sequentially. We previously batched events to process them more efficiently. However, the batching was done per event type, which caused some events to be executed out of order (e.g. a `node-finish` might be run *before* a `node-start`). This was the cause of #2319, and I fixed it by batching all node events together. I made the batch queue explicit and called it a backlog. It's necessary for the backlog to be explicit because we have to wait for all events to be fully processed before we can declare a run to be truly finished, or else some nasty race conditions occur. \
    The backlog also has the advantage that we can process the whole queue at once instead of individual events, which allows for some nice optimizations when there are many events (e.g. a fast iterator).

### DIFF
--- a/backend/src/events.py
+++ b/backend/src/events.py
@@ -7,11 +7,29 @@ from typing import Dict, Literal, TypedDict, Union
 
 from api import ErrorValue, InputId, NodeId, OutputId
 
-# Data of events
+# General events
 
 
-class FinishData(TypedDict):
+class BackendStatusData(TypedDict):
     message: str
+    progress: float
+    statusProgress: float | None
+
+
+class BackendStatusEvent(TypedDict):
+    event: Literal["backend-status"]
+    data: BackendStatusData
+
+
+class BackendStateEvent(TypedDict):
+    event: Literal["backend-ready"] | Literal["backend-started"]
+    data: None
+
+
+BackendEvent = Union[BackendStatusEvent, BackendStateEvent]
+
+
+# Execution events
 
 
 InputsDict = Dict[InputId, ErrorValue]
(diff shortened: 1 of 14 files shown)


joeyballentine (approved):
Looks good to me. 👍

PR #2403, opened 2023-12-13
Mark ignored inputs

### LINKED ISSUE #1737: Mark ignored inputs
**Motivation**
We have input that are only used when certain conditions are met. We typically hide those inputs when their conditions aren't met, but we have to keep those with connections visible. This can be confusing, but it's not clear that those inputs are totally useless.

Example: "Color" is not used.
image

**Description**
Mark ignored inputs in the UI. Maybe something like an "unused" or "ignored" tag or similar.

RunDevelopment:
Fixes #1737

image

Right now, I just added a few simple styles, but we can do more in the future.

This PR is implemented using a new context which holds additional information about inputs. Right now, this context only holds the information whether an input is rendered even though it is conditionally ignored. I want to also do per-input validity displays using this context in the future.


joeyballentine (approved):
Neat idea 👍

PR #2404, opened 2023-12-14
Support capturing groups in Regex Replace node types

RunDevelopment:
rregex recently added an API for capturing groups, so supporting capturing groups in the type system is now possible.

image


joeyballentine (approved):
Awesome

PR #2407, opened 2023-12-14
PyTorch upscale optimizations

RunDevelopment:
Optimizations for pytorch upscaling:
- Do the RGB<->BGR, transpositions, and clipping on the GPU.
- Removed useless `gc.collect()`. At the very least, it didn't do anything in my testing.
- Turn off clipping inside `convenient_upscale`.
- Add `assume_normalized=True` to turn off clipping by `ImageOutput`.

### DIFF
--- a/backend/src/nodes/impl/pytorch/auto_split.py
+++ b/backend/src/nodes/impl/pytorch/auto_split.py
@@ -7,13 +7,49 @@ import torch
 from spandrel import ImageModelDescriptor
 
 from ..upscale.auto_split import Split, Tiler, auto_split
-from .utils import np2tensor, safe_cuda_cache_empty, tensor2np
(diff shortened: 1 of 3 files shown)


joeyballentine (approved):
Heck yeah

akx:
Quoth the docs:

> `torch.flip` makes a copy of input’s data. This is different from NumPy’s `np.flip`, which returns a view in constant time. Since copying a tensor’s data is more work than viewing that data, `torch.flip` is expected to be slower than `np.flip`.

That might be worth looking into still.

RunDevelopment:
I can tell you that's not slower than doing it in numpy on the CPU, at least not on my system. It would be cool if we could do it inplace on the GPU though.

PR #2412, opened 2023-12-15
Update to spandrel 0.1.5 and enforce model tiling hint

RunDevelopment:
Changes:
- Update to spandrel 0.1.5.
- Always use no tiling mode if the model does tiling internally no matter what the user selects.

### DIFF
--- a/src/common/types/chainner-scope.ts
+++ b/src/common/types/chainner-scope.ts
@@ -51,7 +51,9 @@ struct PyTorchModel {
     arch: string,
     size: string,
     subType: string,
+    tiling: ModelTiling,
 }
+enum ModelTiling { Supported, Discouraged, Internal }
 
 struct NcnnBinFile { path: string }
 struct NcnnParamFile { path: string }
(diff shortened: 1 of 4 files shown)


joeyballentine:
Would we be able to hide the tiling option when a model doesn't support it?

joeyballentine:
> Would we be able to hide the tiling option when a model doesn't support it?

RunDevelopment:
It's possible, but it also means that the tile size dropdown won't appear until the user connects a model.
image

RunDevelopment:
Oh, my bad. I already added an option to solve this problem before.
image

src/common/types/chainner-scope.ts
```diff
@@ -51,7 +51,9 @@ struct PyTorchModel {
     arch: string,
     size: string,
     subType: string,
+    tiling: ModelTiling,
 }
+enum ModelTiling { Supported, Discouraged, Internal }
```
joeyballentine, line 56: Why does this have to exist when other enums don't require it?

RunDevelopment: What do you mean? Are you asking why I used an explicit enum here instead of a union of string literals? If so, because I wanted to closely match spandrel's `ModelTiling` enum. I could have used string literals, but I thought that an enum was the cleaner solution.

joeyballentine (approved):

PR #2414, opened 2023-12-17
Make Resize To Side input order consistent with Resize

RunDevelopment:
Very minor change. I swapped Resize Condition and Interpolation Method. This makes the order consistent with Resize (last), and it also groups together the 3 inputs responsible for determining the output size.

image


joeyballentine (dismissed):

joeyballentine:
Oh, the tests are failing

joeyballentine (approved):

PR #2417, opened 2023-12-18
ONNX upscale reworked

adegerard:
Draft PR for information

**What has been done**
- removed `change_shape` flag; assume all models are converted from pytorch (or other) but are in BCHW order
- reworked like it has been done for pytorch
- simple tests (CPU only)

**TODO**
- validate with cuda
- check some perf. optimizations
- remove `np2nptensor` and `nptensor2np`?


adegerard:
>     * removed `change_shape` flag; assume all models are converted from pytorch (or other) but are in BCHW order

Revert this change because rife models (even if not supported by chaiNNer) are using a different order...

joeyballentine:
@adegerard Do you plan to finish this?

adegerard:
Sorry, I forgot.

I ended some tests: I have to check another time grayscale images.
I wonder whether I have to put back the possibility to change from BCHW to BHWC (onnx_auto_split function)
The reason is that is not implemented in pyTorch.
Except for rife onnx models, I found no other models with this property.

I have an implementation which supports this (not commited)
WHat's your opinion?

joeyballentine:
> I found no other models with this property.

Sorry that I didn't give context to this. This is common in models exported from tensorflow

joeyballentine (approved):
Looks good. Thanks for this 👍

PR #2424, opened 2023-12-29
Change edge running animation + behavior

joeyballentine:
This changes a few things:
- The running edge animation is now just a standard crawling ants animation, but with rounded caps and a solid gray edge behind it. Ideally IMO it would be kinda like a loading animation instead but I'm not sure that's possible.
- Edges will now only be animated when the source node is running. To accomplish this, I had to change when the node start event gets sent out. Now, it sends that out after all the inputs are computed. Honestly, this just makes more sense in general, since the node really isn't running until after its inputs are done.
- Yet-to-run edges are now 50% opacity white.

With this, I think it's a bit more obvious what's actually in the process of running. Ideally, this also should reduce CPU usage due to the fewer animations, but I haven't actually checked that.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -429,11 +429,10 @@ class Executor:
         logger.debug(f"node: {node}")
         logger.debug(f"Running node {node.id}")
 
-        await self.__send_node_start(node)
-        await self.progress.suspend()
-
         inputs = await self.__gather_inputs(node)
 
+        await self.progress.suspend()
+        await self.__send_node_start(node)
         await self.progress.suspend()
 
         output, execution_time = await self.loop.run_in_executor(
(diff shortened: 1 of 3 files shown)


joeyballentine:
I originally submitted this with an error. It is now fixed.

akx:
> Ideally IMO it would be kinda like a loading animation instead but I'm not sure that's possible.

If you're OK with losing rounded corners, you can do a loading animation sort of thing with conic gradients... https://snz5t2.csb.app/

joeyballentine:
That's certainly an interesting idea and definitely worth try out, thanks for the suggestion

joeyballentine:
After thinking about it a bit more, i don't think i'll be able to do that, since the direction that any edge could be going in is arbitrary, i don't think you could adapt a conical gradient to work in every scenario

RunDevelopment (approved):

PR #2432, opened 2024-01-04
Allow glob negation in image iterator

joeyballentine:
Realized that without this flag, you could not do negative globs


RunDevelopment:
Does this work? Won't our extensions filter get negated as well?

joeyballentine:
I haven't actually tested this (but I will now that you've mentioned this).

As far as I was aware from looking at the docs, this flag just allows you to do negation in globs by adding ! Before the expression.

It sounds like you think it might be automatically negating the glob, which is not the case

RunDevelopment:
In that case, this should be fine.

RunDevelopment (approved):

Issue #1447, opened 2022-12-30
Add a node to get the current execution number for this session


RunDevelopment:
This node should return how many times the user ran the current chain. The user should be able to reset and change the counter at will too (e.g. to re-do certain runs).

PR #2433, opened 2024-01-04
Add support for static value inputs (set by frontend) with Execution Number node

joeyballentine:
Adds support for non-deterministic nodes... kinda.

The only node I've implemented with this is an execution number node. It's just a number that increments every time you press run. If this gets approved/merged, I plan on doing one for the current datetime as well. 

This is really, really jank - but basically, I hardcode a node on the frontend that only exists on the frontend. Then, in the code that converts the flow to the backend format, I get rid of that node and replace its connections with a static value, as if that value was typed into the input.

I'm sure this is still technically "wrong" and blah blah determinism, but I think this is a good compromise as it doesn't affect how we handle caching. It's just the same as if the user were to manually change a value every time. And since this has been requested many times, I think it's worth having, even if it's a bit jank.

This only partially implements #1447. I have not added a way to manipulate this counter yet, and probably won't in this PR.

This also does not display any Navi/type info about what the value will be before running. I'm not sure how we could implement that.


joeyballentine:
For documentation purposes: This now utilizes a new type of input for adapting static values from the frontend.

backend/src/packages/chaiNNer_standard/utility/value/get_execution_number.py
```diff
@@ -0,0 +1,32 @@
...
+    icon="MdNumbers",
+    inputs=[
+        StaticValueInput(
+            label="Execution Number",
+            value_of="execution_number",
+            navi_type=get_number_type(
+                1,
+                None,
+                0,
+            ),
```
RunDevelopment, line 23:
```suggestion
            navi_type="int(1..)",
```

src/common/common-types.ts
```diff
@@ -134,6 +134,12 @@ export interface ColorInput extends InputBase {
     readonly def: string;
     readonly channels?: readonly number[] | null;
 }
+
+export interface StaticValueInput extends InputBase {
+    readonly kind: 'static';
+    readonly valueOf: 'execution_number' | 'datetime' | 'auto';
```
RunDevelopment, line 140: What's `auto`?

joeyballentine: Hypothetical case where we determine the value based on the schemaId rather than the valueOf type. I could just get rid of this since it's not used anywhere right now, and it's probably not even a good idea to support.

RunDevelopment: Please do. I'm not a fan of not-implemented features.

backend/src/packages/chaiNNer_standard/utility/value/execution_number.py
```diff
@@ -0,0 +1,27 @@
...
+from nodes.properties.inputs.__system_inputs import StaticValueInput
+from nodes.properties.outputs import NumberOutput
+
+from .. import value_group
+
+
+@value_group.register(
+    schema_id="chainner:utility:execution_number",
+    name="Execution Number",
+    description="Get the current execution number of this session. Increments by 1 every time you press the play button.",
```
RunDevelopment, line 12:
```suggestion
    description="Get the current execution number of this session. Increments by 1 every time you press the run button.",
```

src/renderer/components/inputs/StaticValueInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+        const { conditionallyInactive } = useContext(InputContext);
+
+        useEffect(() => {
+            switch (valueOf) {
+                case 'execution_number':
+                    setValue(executionNumber);
+                    break;
+                default:
+                    setValue(0);
+                    break;
```
RunDevelopment, line 23: To ensure that we don't forget anything.

```suggestion
                    assertNever(valueOf);
```

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -476,7 +479,8 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
         } else {
             await runNodes();
         }
-    }, [resume, runNodes, status]);
+        setExecutionNumber((prev) => prev + 1);
```
RunDevelopment, line 482: So resuming counts as an execution?

RunDevelopment:
Other things I noticed:

1. Should the execution number reset? E.g. when the chain is reset (Ctrl+N) or a different chain is loaded? I'm not sure.
2. The current design is confusing. The output value is styled in the same way we usually style labels. E.g. this looks like an input called "19" that expects an integer.
    image
   I think something similar to how Pass Through does type tags would look better.
    image
    E.g. we could do:
    ```
     # Execution Number
     Current       [19]
    ```
    And use a big tag for the number, like model tags do.

joeyballentine:
> Should the execution number reset? E.g. when the chain is reset (Ctrl+N) or a different chain is loaded?

I don't personally think so. I plan on doing a follow-up to this where I add the ability to manually reset the number. I just haven't decided how I want to expose that yet

PR #2440, opened 2024-01-07
Add .safetensors as supported format for Load Models

joeyballentine:
discovered this while testing stuff

PR #2442, opened 2024-01-07
Add "Defer errors" option to applicable iterators

joeyballentine:
So it turns out the reason it wasn't working properly before was because of #2440... The models i was using just weren't being picked up by the node i was using to test. Therefore, I didn't actually need to do the manual iteration with the `while True` + `next()` nonsense.

This might not be the most glamorous solution, but it works and I'm happy enough with it.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -571,10 +571,10 @@ class Executor:
         await self.__send_node_progress(node, times, 0, expected_length)
 
         errors: list[str] = []
-        iterable = iterator_output.iterator.iter_supplier()
-        while True:
+        for values in iterator_output.iterator.iter_supplier():
             try:
-                values = next(iterable)  # type: ignore
+                if isinstance(values, Exception):
+                    raise values
 
                 # write current values to cache
                 iter_output = fill_partial_output(values)
@@ -603,10 +603,7 @@ class Executor:
                 await self.progress.suspend()
                 await update_progress()
                 await self.progress.suspend()
-            except StopIteration:
-                break
-            except StopAsyncIteration:
-                break
+
             except Exception as e:
                 if iterator_output.iterator.defer_errors:
                     errors.append(str(e))
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -570,32 +570,41 @@ class Executor:
         # iterate
         await self.__send_node_progress(node, times, 0, expected_length)
 
-        for values in iterator_output.iterator.iter_supplier():
-            # write current values to cache
-            iter_output = fill_partial_output(values)
-            self.cache.set(node.id, iter_output, StaticCaching)
+        error = None
+        try:
+            for values in iterator_output.iterator.iter_supplier():
+                # write current values to cache
+                iter_output = fill_partial_output(values)
+                self.cache.set(node.id, iter_output, StaticCaching)
 
-            # broadcast
-            await self.__send_node_broadcast(node, iter_output.output)
+                # broadcast
+                await self.__send_node_broadcast(node, iter_output.output)
 
-            # run each of the output nodes
-            for output_node in output_nodes:
-                await self.process_regular_node(output_node)
+                # run each of the output nodes
+                for output_node in output_nodes:
+                    await self.process_regular_node(output_node)
 
-            # run each of the collector nodes
-            for collector, timer, collector_node in collectors:
+                # run each of the collector nodes
+                for collector, timer, collector_node in collectors:
+                    await self.progress.suspend()
+                    iterate_inputs = await self.__gather_collector_inputs(
+                        collector_node
+                    )
+                    await self.progress.suspend()
+                    with timer.run():
+                        run_collector_iterate(collector_node, iterate_inputs, collector)
+
+                # clear cache for next iteration
+                self.cache.delete_many(all_iterated_nodes)
+
                 await self.progress.suspend()
-                iterate_inputs = await self.__gather_collector_inputs(collector_node)
+                await update_progress()
                 await self.progress.suspend()
-                with timer.run():
-                    run_collector_iterate(collector_node, iterate_inputs, collector)
-
-            # clear cache for next iteration
-            self.cache.delete_many(all_iterated_nodes)
-
-            await self.progress.suspend()
-            await update_progress()
-            await self.progress.suspend()
+        except Exception as e:
+            if iterator_output.iterator.defer_errors:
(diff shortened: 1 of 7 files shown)


RunDevelopment:
Instead of dealing with deferred errors in one place (where you actually do the iterating), you now handle it separately in 4 places. Iterator progress is also messed up because errors silently caught by the iterator itself can't be accounted for (the issue is that `expected_length` reliable at all anymore).

Yeah, I'm not a fan of this. Could we do the `while True` + `next()` nonsense instead?

joeyballentine:
sure

joeyballentine:
Doesn't work. I reverted the stuff in api.py back aside from setting `defer_errors` to true, and i replaced the `for values in ...` loop in process.py with the `while`. It doesn't work, it just errors at the first error and gives up.

Since I know you'll wanna test it yourself, I pushed my changes to a new branch called `defer-errors-nonworking`. Maybe I just did something wrong.

RunDevelopment:
> Doesn't work. 

That's my bad.

```py
        def supplier():
            for i, x in enumerate(l):
                yield map_fn(x, i)
```

Obviously the generator will stop when `map_fn` throws. That's my mistake.

So I read through python's docs, and it just doesn't tell you what will happen when [the `__next__` method](https://docs.python.org/3/library/stdtypes.html#typeiter) of a generator/iterator throws an exception beside `StopIteration`. Great: unspecified behavior. So I asked chatGPT, and it said that this is strongly discouraged. It told me that no one expects this and that a lot of code might just completely bug out.

So I guess the only solution compatible with python would be to roll our own `Result<T, E>` type. Rust had it all figured out after all.

So we need to change our API to `iter_supplier: Callable[[], Iterable[Result[I]]]` where `Result[I]` is either an item `I` or an `Exception`. We *could* assume that `I` will never be of type `Exception` and change the API to ``iter_supplier: Callable[[], Iterable[I | Exception]]`` instead. This would get rid of the `Result` type.

---

Also, first thing I noticed:

```py
        iterable = iterator_output.iterator.iter_supplier()
        while True:
            try:
                values = next(iterable)  # type: ignore
```

This is wrong and the type error even told you that. `next` expects an iterator, not an iterable. An iterator is the thing that has the `__next__` method, and an iterable is something that can give you can iterator via its `__iter__` method. E.g. a `list` is the iterable that will give you an iterator over its elements. The `list` iterator is essentially just an index that will be incremented whenever `__next__` is called.

So why did your previous code seem to work anyway? Luck. Under the hood, we use python generator functions (`yield`) and those return a `Generator` object. A `Generator` is both iterator and iterable, so it has both a `__next__` and `__iter__` method. The trick is that `__iter__` just returns the generator itself.

Here's what it needs to be:

```py
        iterator = iter(iterator_output.iterator.iter_supplier())
        while True:
            try:
                values = next(iterator)
```

joeyballentine:
> Here's what it needs to be:

I did try that but I got rid of it because it wasn't doing anything

joeyballentine:
Also, I really don't want to implement a whole result system and make all those changes right now. That's something I actually started doing back when I first tried this and it was gonna be a lot of work. 

Can we just use what this branch does for now? Even if it's not ideal?

RunDevelopment:
> I really don't want to implement a whole result system and make all those changes right now.

Checkout the `defer-errors-nonworking` branch. I made it work with my `iter_supplier: Callable[[], Iterable[I | Exception]]` suggestion :) https://github.com/chaiNNer-org/chaiNNer/commit/ba312c6ff7022a4bae65d456150c654b74d6fd09

> I did try that but I got rid of it because it wasn't doing anything (as in, it didn't fix my problem)

Sorry, I put this in a confusing way. Yes, this doesn't fix your problem. It's just also wrong.

backend/src/process.py
```diff
@@ -570,32 +570,45 @@ async def update_progress():
...
-                with timer.run():
-                    run_collector_iterate(collector_node, iterate_inputs, collector)
 
-            # clear cache for next iteration
-            self.cache.delete_many(all_iterated_nodes)
-
-            await self.progress.suspend()
-            await update_progress()
-            await self.progress.suspend()
+            except Exception as e:
```
RunDevelopment, line 607: I just noticed that we need to re-raise abort exceptions.

RunDevelopment: Otherwise, iterators are unstoppable :)

backend/src/process.py
```diff
@@ -628,6 +642,10 @@ async def update_progress():
                 self.cache_strategy[collector_node.id],
             )
 
+        if iterator_output.iterator.defer_errors and len(errors) > 0:
```
RunDevelopment, line 645: Last thing: instead of checking `iterator_output.iterator.defer_errors` again here, could we rename `errors` to `deferred_errors` and omit the `iterator_output.iterator.defer_errors` check?

joeyballentine: Sure

backend/src/process.py
```diff
@@ -570,32 +570,46 @@ async def update_progress():
         # iterate
         await self.__send_node_progress(node, times, 0, expected_length)
 
+        errors: list[str] = []
```
joeyballentine, line 573:
```suggestion
        deferred_errors: list[str] = []
```

backend/src/process.py
```diff
@@ -570,32 +570,46 @@ async def update_progress():
...
+                self.cache.delete_many(all_iterated_nodes)
+
+                await self.progress.suspend()
+                await update_progress()
+                await self.progress.suspend()
+            except Aborted:
+                raise
+            except Exception as e:
+                if iterator_output.iterator.defer_errors:
+                    errors.append(str(e))
```
joeyballentine, line 610:
```suggestion
                    deferred_errors.append(str(e))
```

backend/src/process.py
joeyballentine, line 645:
```suggestion
        if len(deferred_errors) > 0:
```

backend/src/process.py
```diff
@@ -628,6 +642,10 @@ async def update_progress():
                 self.cache_strategy[collector_node.id],
             )
 
+        if iterator_output.iterator.defer_errors and len(errors) > 0:
+            error_string = "- " + "\n- ".join(errors)
```
joeyballentine, line 646:
```suggestion
            error_string = "- " + "\n- ".join(deferred_errors)
```

RunDevelopment (approved):

PR #2443, opened 2024-01-07
Support spandrel guided SR models

RunDevelopment:
This adds support for guided SR models.

Right now, this is very much WIP and just makes it possible to run the models from https://github.com/chaiNNer-org/spandrel/pull/79. I'm still not sure about SGNet, so I'll have to think about this a bit more.


RunDevelopment:
See https://github.com/chaiNNer-org/spandrel/pull/79 as to why.

PR #2444, opened 2024-01-08
Replace global execution options with node context parameter

RunDevelopment:
This removes the global mutable execution options and replaces them with a node context parameter: `context: NodeContext`.

The new parameter is opt in and can be enabled in `@register` with the `node_context` property. This not only minimizes the diff (I didn't want to change all nodes), it might also enable optimizations in the future.

Right now, the `NodeContext` class has 2 aspects to it: settings and progress. Settings and other execution options are now accessed through the node context parameter. While only the settings aspect is currently used, I also laid the groundwork for per-node progress, which will be my next goal.


joeyballentine (approved):

joeyballentine:
Looks great 👍

PR #2445, opened 2024-01-09
Better node errors using error messages in output types

RunDevelopment:
This PR adds the ability for output types to return error messages. This is useful for complex types where `never_reason` alone isn't enough. Since `never_reason` is only one message, we previously weren't able to communicate the precise reason for the error, even though the type system knows the reason. I fixed this by allowing output types to return not only their type, but also an optional `Error { message: string }` type. This allows us to define relevant error messages, instead of just listing reasons why the node could be invalid.

In actual output type code, this looks like this (Alpha Matting node):
```
if fg <= bg {
    error("The foreground threshold must be greater than the background threshold.")
} else if bool::or(image.width != trimap.width, image.height != trimap.height) {
    error("The image and trimap must have the same size.")
} else {
    Image { width: image.width, height: image.height }
}
```

image

Note that this PR doesn't remove `never_reason`. `never_reason` is just simpler to use in many cases, and it still provides a useful fallback/default for error messages. I think of the new `error("...")` as an addition, not as a replacement.

### DIFF
--- a/backend/src/navi.py
+++ b/backend/src/navi.py
@@ -150,6 +150,10 @@ def intersect(*items: ExpressionJson) -> ExpressionJson:
     return {"type": "intersection", "items": list(items)}
 
 
+def intersect_with_error(*items: ExpressionJson) -> ExpressionJson:
+    return union(intersect(*items), *[intersect("Error", item) for item in items])
+
+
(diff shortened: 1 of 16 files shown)


joeyballentine (approved):

joeyballentine:
LGTM

PR #2449, opened 2024-01-11
Rename "Defer errors" to "Stop on first error"

joeyballentine:
As suggested by @adegerard on Discord.

I had to invert how this boolean behaves for it to make sense. I think this is a much better and more intuitive label for it.


RunDevelopment:
Does this need a migration?

RunDevelopment:
I like the new label for the input, but I'm not a fan of the internal `throw_early` variables. It's a weird name IMO. A common paradigm is called "fail fast", so maybe we could call the variables `fail_fast`?

joeyballentine:
Going to merge this so I can get it in today's nightly.

PR #2452, opened 2024-01-12
Update README

joeyballentine:


README.md
```diff
@@ -61,32 +61,30 @@ Don't forget, there are plenty of non-upscaling tasks you can do with chaiNNer a
 
 To select multiple nodes, hold down shift and drag around all the nodes you want to be selected. You can also select an individual node by just clicking on it. When nodes are selected, you can press backspace or delete to delete them from the editor.
 
-To batch upscale, create an Image Iterator node and drag the nodes you want to use into the iterator's editor area. You can expand the iterator by clicking and dragging the bottom right corner outwards (like you would a UI window). Simply wire up a chain in an iterator the same as you would normally, and when you click run it will run on every image in the folder you chose. You also can select an entire existing chain, and drag it into the iterator's editor area to essentially convert the entire thing into an iterable chain.
+To batch upscale, use a "Load Images" node. For video, a "Load Video" node. You can't combine "Load Images" and "Load Video" together, but you _can_ use "Save Image" with "Load Video", and "Save Video" with "Load Images".
```
Kim2091, line 64: This sentence flows very oddly. How about the following? I may have misinterpreted your explanation of "combining" these two nodes. Did you mean connecting them directly, or using them both in the chain at once?

> To perform a batch upscale of images, use the "Load Images" node. To upscale videos, use the "Load Video" node. It's important to note however that you cannot use both "Load Images" and "Load Video" nodes at once. In addition, the "Save Image" node is compatible with "Load Video", and "Save Video" can be used with "Load Images".

joeyballentine: You are correct in what I was trying to convey, but I just didn't want to give the impression you can't use them in the same chain. You can, you just can't have them share lineage. But i don't think a user would understand that wording

Kim2091 (approved):

PR #2471, opened 2024-01-16
Remove TODO comment

joeyballentine:
Since we have Spandrel now, this TODO is no longer to do, and is actually finished. 

(that comment was because at the time, models just arbitrarily had whatever properties were set on them, and thus having in_nc and out_nc was not a given. Now we have input_channels and output_channels guaranteed on the descriptor)


joeyballentine:
I don't think this needs a review lol

PR #2472, opened 2024-01-16
Add PyTorch memory budget limit setting

### LINKED ISSUE #2450: Set a limit for maximum VRAM that can be used on Windows Version
**Motivation**
When i upscale in background and my VRAM is fully used, i can't do sth video-like on forground (like watching a video). It's because of the VRAM, not GPU Usage.

**Description**
That's why it would be helpfull, if we could set a limit of how much can be used. For example if i have 12GB RAM, it would be nice, if the software (chainner) thinks, there is only 11GB. And then the rest 1GB is untouched for my system.

I know i could use tilesize, but auto-tile-size is for me a lot better, because otherwise i would have to test on each upscale, what tilesize is best. With this i just could use auto-tile-size but have still 1GB free.

joeyballentine:
Partially reimplements #2088. However, I don't understand what the Mac stuff is doing there, so someone else (@stonerl?) will have to finish that out. For just limiting VRAM, this works. This was mostly based on the existing NCNN setting.

image

closes #2450.

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/settings.py
+++ b/backend/src/packages/chaiNNer_pytorch/settings.py
@@ -70,7 +70,7 @@ package.add_setting(
     NumberSetting(
         label="Memory Budget Limit (GiB)",
         key="budget_limit",
-        description="Maximum memory to use for PyTorch inference. 0 means no limit. Memory usage measurement is not completely accurate yet; you may need to significantly adjust this budget limit via trial-and-error if it's not having the effect you want.",
+        description="Maximum memory (VRAM) to use for PyTorch inference. 0 means no limit. Memory usage measurement is not completely accurate yet; you may need to significantly adjust this budget limit via trial-and-error if it's not having the effect you want.",
         default=0,
         min=0,
         max=1024**2,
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_pytorch/settings.py
```diff
@@ -66,12 +66,24 @@
     ),
 )
 
+package.add_setting(
+    NumberSetting(
+        label="Memory Budget Limit (GiB)",
+        key="budget_limit",
```
RunDevelopment, line 72: A budget limit for what?

backend/src/packages/chaiNNer_pytorch/settings.py
```diff
@@ -66,12 +66,24 @@
...
+        description="Maximum memory (VRAM) to use for PyTorch inference. 0 means no limit. Memory usage measurement is not completely accurate yet; you may need to significantly adjust this budget limit via trial-and-error if it's not having the effect you want.",
```
RunDevelopment, line 73: 1. This limits RAM when using the CPU.
2. This currently only affects the Auto tile size mode.

backend/src/packages/chaiNNer_pytorch/settings.py
joeyballentine: This is the same name as the NCNN setting, for consistency

RunDevelopment (approved):

PR #2473, opened 2024-01-16
Add "Copy" to context menus

joeyballentine:
We've had duplicate in these menus before, but not copy.

I will add a paste button to the node selector menu (probably above the search bar) in a separate pr

### DIFF
--- a/src/renderer/hooks/useNodeMenu.tsx
+++ b/src/renderer/hooks/useNodeMenu.tsx
@@ -9,10 +9,12 @@ import {
 import { MenuDivider, MenuItem, MenuList } from '@chakra-ui/react';
 import { BsFillJournalBookmarkFill } from 'react-icons/bs';
 import { MdPlayArrow, MdPlayDisabled } from 'react-icons/md';
+import { useReactFlow } from 'reactflow';
 import { useContext } from 'use-context-selector';
-import { NodeData } from '../../common/common-types';
+import { EdgeData, NodeData } from '../../common/common-types';
 import { GlobalContext } from '../contexts/GlobalNodeState';
 import { NodeDocumentationContext } from '../contexts/NodeDocumentationContext';
+import { copyToClipboard } from '../helpers/copyAndPaste';
 import { UseContextMenu, useContextMenu } from './useContextMenu';
 import { UseDisabled } from './useDisabled';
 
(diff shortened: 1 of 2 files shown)


src/renderer/hooks/useNodeMenu.tsx
```diff
@@ -33,8 +35,24 @@ export const useNodeMenu = (
...
+            <MenuItem
+                icon={<CopyIcon />}
+                onClick={() => {
+                    const node = getNode(id);
+                    if (node && !node.selected) {
+                        const nodeCopy = { ...node, selected: true };
+                        copyToClipboard([nodeCopy], []);
+                    } else {
+                        copyToClipboard(getNodes(), getEdges());
+                    }
```
RunDevelopment, line 51: What is this? The `node &&` part of the condition is always true (otherwise, what are we right-clicking on?). And the rest of the condition... If the node is not selected, copy the node, otherwise copy the whole chain. How does that make sense? Right-clicking does not select a node, so just right-clicking + Copy any node will copy the whole chain.

src/renderer/hooks/useNodesMenu.tsx
```diff
@@ -1,18 +1,29 @@
...
     const nodeIds = nodes.map((n) => n.id);
 
+    const { getNodes, getEdges } = useReactFlow<NodeData, EdgeData>();
+
     return useContextMenu(() => (
         <MenuList className="nodrag">
+            <MenuItem
+                icon={<CopyIcon />}
+                onClick={() => {
+                    copyToClipboard(getNodes(), getEdges());
```
RunDevelopment, line 22: This context menu is for an area selection of nodes. Why are you copying the whole chain when this context menu explicitly refers to a selected set of nodes?

image

RunDevelopment (commented):
The behavior doesn't make sense...

src/renderer/hooks/useNodeMenu.tsx
joeyballentine: > The node && part of the condition is always true

False. The type from react flow for this helper is it returns a node or undefined. After all, I could be passing it a garbage id for all it cares. 

> And the rest of the condition... If the node is not selected, copy the node, otherwise copy the whole chain

I knew I should have left a comment explaining this... The copyToClipboard() function gets every selected node and connected edge between selected nodes. Under normal copying behavior, you are selecting nodes and ctrl+c'ing, so we pass in the entire chain in order to figure this out. 

When you just right click on a node to copy it, it's not necessarily selected, so what I do is if the node you're right clicking on isn't selected, then that means you just want to copy that one node. If it is selected, then other ones might be selected too, and therefore we'll pass the entire chain to it like normal ctrl+c behavior.

This makes it work exactly how you would expect it to in practice. Yeah this code is a bit weird but that's purely due to how that copy to clipboard function was implemented.

src/renderer/hooks/useNodesMenu.tsx
joeyballentine: Read my above comment and also take a look at the copy to clipboard function.

RunDevelopment:
I completely misunderstood what `copyToClipboard` does. Sorry about that.

Maybe we should rename `copyToClipboard` to `copySelectedToClipboard` to make its function clearer...

RunDevelopment (approved):

joeyballentine:
That's a good idea. I was also quite surprised by that behavior when I was working on this

PR #2477, opened 2024-01-17
Optimize static switches

### LINKED ISSUE #2262: Compiler optimization: "Execute" Switch node statically
image

The switch node is essentially a compile-time switch right now. Since we statically know which connection it will pass on, we can "exectue" the node during chain optimization before the actual execution. The main advantage to that is that **only the selected input will be computed**.

RunDevelopment:
Fixes #2262.

This adds a backend compiler optimization that statically executes (kinda) the Switch node if the selected input is statically known (it currently always is). Static execution is a strong word. In actuality, it just removes the edges from inputs that aren't selected. However, the effect is the same: only the selected input will be executed.

Of course, this new optimization works together with the already present `removed_dead_nodes` optimization.

Other changes:
- To implement this optimization, I also had to make some (long overdue) changes to the `Chain`. It now contains node input values instead of just edges. Since optimization may add, change, and remove nodes, they also need to be able to influence input values.
- Optimization now use mutation signal instead of returning a bool to say that they changed the chain. This just makes optimizations a little easier to implement.
- Use data classes in a few places.
- The executor now derives the input map it used from the given chain.

### DIFF
--- a/backend/src/chain/optimize.py
+++ b/backend/src/chain/optimize.py
@@ -3,23 +3,54 @@ from sanic.log import logger
 from .chain import Chain
 
 
-def __removed_dead_nodes(chain: Chain) -> bool:
+class _Mutation:
+    def __init__(self) -> None:
+        self.changed = False
+
+    def signal(self) -> None:
+        self.changed = True
+
+
+def __removed_dead_nodes(chain: Chain, mutation: _Mutation):
     """
     If a node does not have side effects and has no downstream nodes, then it can be removed.
     """
-    changed = False
 
     for node in list(chain.nodes.values()):
         is_dead = len(chain.edges_from(node.id)) == 0 and not node.has_side_effects()
         if is_dead:
             chain.remove_node(node.id)
-            changed = True
+            mutation.signal()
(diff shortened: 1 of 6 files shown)


backend/src/chain/optimize.py
```diff
@@ -3,23 +3,54 @@
...
 
-    return changed
+
+def __static_switch_trim(chain: Chain, mutation: _Mutation):
+    """
+    If the selected variant of the Switch node is statically known, then we can remove the input edges of all other variants.
+    """
+
+    for node in list(chain.nodes.values()):
+        if node.schema_id == "chainner:utility:switch":
```
joeyballentine, line 33: in the interest of protecting my sanity, is there a way you could place this functionality in a lambda function defined on `@register` or something rather than doing it this way?

RunDevelopment: Only if we make the backend representation of chains (+ the interface for optimizers) part of the API.

joeyballentine: Maybe then we just leave this as is for now and consider it in the future.

joeyballentine (approved):

PR #2478, opened 2024-01-18
Enabled Ruff S101 and replace all asserts with proper ifs + errors

joeyballentine:
> Assertions are removed when Python is run with optimization requested (i.e., when the -O flag is present), which is a common practice in production environments. As such, assertions should not be used for runtime validation of user input or to enforce interface constraints.

> Consider raising a meaningful error instead of using assert.

This PR removes (almost) all assertions and replaces them with equivalent checks + actual error messages where missing.

While there may not be a functional difference unless -O is used, it is still recommended to only use asserts for tests or debugging purposes and to use regular `raise` exceptions instead. 

This does leave 3 `assert isinstance(...` checks in place, because converting them to regular if statements gives me a nice pylance error about it being impossible for these to _not_ be an instance of that type. So, I just left them as is.

Note: While I was as careful as possible only manually flipping trivial checks like < into >= and just `if not (...)`ing the more complicated ones, there is still a chance that this introduces a bug from a flipped condition. However, I think this is unlikely.

### DIFF
--- a/backend/src/chain/cache.py
+++ b/backend/src/chain/cache.py
@@ -14,7 +14,8 @@ class CacheStrategy:
     STATIC_HITS_TO_LIVE = 1_000_000_000
 
     def __init__(self, hits_to_live: int) -> None:
-        assert hits_to_live >= 0
+        if hits_to_live <= 0:
+            raise ValueError("hits_to_live must be greater than 0")
         self.hits_to_live = hits_to_live
 
     @property
@@ -79,7 +80,8 @@ T = TypeVar("T")
 
 class _CacheEntry(Generic[T]):
     def __init__(self, value: T, hits_to_live: int):
-        assert hits_to_live > 0
+        if hits_to_live <= 0:
+            raise ValueError("hits_to_live must be greater than 0")
         self.value: T = value
         self.hits_to_live: int = hits_to_live
 
--- a/backend/src/nodes/impl/blend.py
+++ b/backend/src/nodes/impl/blend.py
@@ -196,13 +196,13 @@ def blend_images(overlay: np.ndarray, base: np.ndarray, blend_mode: BlendMode):
     o_shape = get_h_w_c(overlay)
     b_shape = get_h_w_c(base)
 
-    assert (
-        o_shape[:2] == b_shape[:2]
-    ), "The overlay and the base image must have the same size"
+    if not (o_shape[:2] == b_shape[:2]):
+        raise ValueError("The overlay and the base image must have the same size")
 
     def assert_sane(c: int, name: str):
         sane = c in (1, 3, 4)
-        assert sane, f"The {name} has to be a grayscale, RGB, or RGBA image"
+        if not sane:
+            raise ValueError(f"The {name} has to be a grayscale, RGB, or RGBA image")
 
     o_channels = o_shape[2]
     b_channels = b_shape[2]
--- a/backend/src/nodes/impl/blend.py
+++ b/backend/src/nodes/impl/blend.py
@@ -196,7 +196,7 @@ def blend_images(overlay: np.ndarray, base: np.ndarray, blend_mode: BlendMode):
     o_shape = get_h_w_c(overlay)
     b_shape = get_h_w_c(base)
 
-    if not (o_shape[:2] == b_shape[:2]):
+    if o_shape[:2] != b_shape[:2]:
         raise ValueError("The overlay and the base image must have the same size")
 
     def assert_sane(c: int, name: str):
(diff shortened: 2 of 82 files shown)


joeyballentine:
>  However, I think this is unlikely.

I spoke too soon. I naively thought that i could combine if statements together like that... I will review this PR myself to verify I didn't make any more changes like that

joeyballentine:
Ok, the rest of the PR looks fine to me

backend/src/nodes/impl/cas.py
```diff
@@ -31,7 +31,8 @@ def create_cas_mask(img: np.ndarray, kernel: np.ndarray, bias: float = 2) -> np.
     Lou Kramer, FidelityFX CAS, AMD Developer Day 2019, https://gpuopen.com/wp-content/uploads/2019/07/FidelityFX-CAS.pptx
     https://www.shadertoy.com/view/wtlSWB#
     """
-    assert bias > 0, "Bias must be greater than or equal to 0."
+    if bias <= 0:
+        raise ValueError("Bias must be greater than or equal to 0.")
```
RunDevelopment, line 35:
```suggestion
        raise ValueError("Bias must be >0.")
```

backend/src/nodes/impl/blend.py
```diff
@@ -196,13 +196,13 @@ def blend_images(overlay: np.ndarray, base: np.ndarray, blend_mode: BlendMode):
     o_shape = get_h_w_c(overlay)
     b_shape = get_h_w_c(base)
 
-    assert (
-        o_shape[:2] == b_shape[:2]
-    ), "The overlay and the base image must have the same size"
+    if not (o_shape[:2] == b_shape[:2]):
```
RunDevelopment, line 199:
```suggestion
    if o_shape[:2] != b_shape[:2]:
```

backend/src/packages/chaiNNer_onnx/onnx/io/load_model.py
```diff
@@ -36,9 +36,11 @@ def load_model_node(path: str) -> tuple[OnnxModel, str, str]:
     """Read a pth file from the specified path and return it as a state dict
     and loaded model after finding arch config"""
 
-    assert os.path.exists(path), f"Model file at location {path} does not exist"
+    if not os.path.exists(path):
+        raise FileNotFoundError(f"Model file at location {path} does not exist")
 
-    assert os.path.isfile(path), f"Path {path} is not a file"
+    if not os.path.isfile(path):
+        raise ValueError(f"Path {path} is not a file")
```
RunDevelopment, line 43: We might as well remove those checks, because `FileInput` already does them.

backend/src/packages/chaiNNer_pytorch/pytorch/io/load_model.py
```diff
@@ -68,9 +68,11 @@ def load_model_node(
     """Read a pth file from the specified path and return it as a state dict
     and loaded model after finding arch config"""
 
-    assert os.path.exists(path), f"Model file at location {path} does not exist"
+    if not os.path.exists(path):
+        raise FileNotFoundError(f"Model file at location {path} does not exist")
 
-    assert os.path.isfile(path), f"Path {path} is not a file"
+    if not os.path.isfile(path):
+        raise ValueError(f"Path {path} is not a file")
```
RunDevelopment, line 75: Same here.

backend/src/nodes/impl/cas.py
joeyballentine: Oops

RunDevelopment (approved):

backend/src/chain/cache.py
```diff
@@ -14,7 +14,8 @@ class CacheStrategy:
     STATIC_HITS_TO_LIVE = 1_000_000_000
 
     def __init__(self, hits_to_live: int) -> None:
-        assert hits_to_live >= 0
+        if hits_to_live <= 0:
```
RunDevelopment, line 17: Wrong.

backend/src/nodes/impl/upscale/tile_blending.py
```diff
@@ -206,9 +223,13 @@ def add_tile(self, tile: np.ndarray, overlap: TileOverlap) -> None:
...
-        else:
-            assert self.offset == self.height
+        if self.direction == BlendDirection.X and self.offset != self.width:
+            raise ValueError(
+                f"Expected {self.width} pixels, but got {self.offset} pixels"
+            )
+        elif self.offset != self.height:
+            raise ValueError(
+                f"Expected {self.height} pixels, but got {self.offset} pixels"
+            )
```
RunDevelopment, line 233: Wrong.

RunDevelopment:
Okay, that's it. I am not confident that those changes are correct anymore. Joey, you and I reviewed this and we both did not see those 2 obvious errors. We cannot be sure that more errors like this aren't hiding in there somewhere.

I'm sorry, but I want to revert this. The potential for new bugs in old code is too high with this PR.

joeyballentine:
Yep that's fine with me. let's revert this. Sorry about that

PR #2487, opened 2024-01-20
Update to spandrel v0.2.1

### LINKED ISSUE #2461: Pytorch Upscale Image node tiling issue
**Information:**

- Chainner version: [0.21.0]
- OS: [Windows 10]

**Description**
Trying to simple process images through model, however, with tile set to anything lower than 4096 it gives errors at some large images. Workflow preview attached.
image

**Logs**
Error log from interface:
```
Error

An error occurred in a pytorch Upscale Image node:



Input values:
• Image: RGB Image 3900x4680
• Model: Value of type 'spandrel.__helpers.model_descriptor.ImageModelDescriptor'
• Tile Size: Auto (estimate)
• Separate Alpha: No
```
[logs.zip](https://github.com/chaiNNer-org/chaiNNer/files/13928974/logs.zip)

RunDevelopment:
Fixes #2461.

This updates spandrel to v0.2.1 and changes our code to take advantage of the new API and guarantees. E.g. since spandrel now handles size requirements, I was able to remove our padding code.

### DIFF
--- a/backend/src/nodes/impl/pytorch/auto_split.py
+++ b/backend/src/nodes/impl/pytorch/auto_split.py
@@ -4,7 +4,7 @@ import gc
 
 import numpy as np
 import torch
-from spandrel import ImageModelDescriptor, SizeRequirements
+from spandrel import ImageModelDescriptor
 
 from ..upscale.auto_split import Split, Tiler, auto_split
 from .utils import safe_cuda_cache_empty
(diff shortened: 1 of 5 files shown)


joeyballentine:
When we update spandrel, we should also always update the link to the supported models section in the readme. I linked to the specific version we use so that it's always correct for whatever chaiNNer currently supports.

RunDevelopment:
Can't we reasonably expect that chainner will always be close to up to date? I don't think it's a huge problem when the spandrel README lists a few archs that chainner will only support in the next release.

joeyballentine:
That's a good point. In that case, can you update it to just point to the normal readme?

joeyballentine (approved):
actually ill just do it in a separate pr

Issue #2266, opened 2023-10-14
Fix lineage


RunDevelopment:
The current way the frontend handles lineage is incorrect. Fixing lineage will require #2258 because of collector.

How lineage affects connectability is described by this pseudocode:
```js
let sourceL = getLineage(sourceNode);
if (isIterator(targetNode)) {
  // the inputs of iterator nodes only allow non-iterator connections
  return sourceL === undefined ? CONNECTABLE : INVALID;
}
if (isRegularNode(targetNode)) {
  if (sourceL === undefined) return CONNECTABLE;
  let targetL = getLineage(targetNode, excluded=targetInput)
  return targetL === undefined || targetL === sourceL ? CONNECTABLE : INVALID;
}
assert(isCollector(targetNode));
if (isIteratorInput(targetInput)) {
  if (sourceL === undefined) return INVALID;
  let targetL = getLineage(targetNode.iteratorInputs, excluded=targetInput)
  return targetL === undefined || targetL === sourceL ? CONNECTABLE : INVALID;
} else {
  // the non-iterator inputs of collector nodes only allow non-iterator connections
  return sourceL === undefined ? CONNECTABLE : INVALID;
}
```

[Original discussion](https://discord.com/channels/930865462852591648/1078534152464384152/1162842110022668288).

PR #2491, opened 2024-01-21
Replace RF component selector with new type

joeyballentine:
Follow up to #2490 

Unfortunately, the RF node types aren't type-safe yet, but they will be in v12 (i think).

All this does is finish decoupling the RF node types from the node kinds, and replaces them with just a generic "backendNode" node type which uses our Node component.

### DIFF
--- a/src/common/common-types.ts
+++ b/src/common/common-types.ts
@@ -251,6 +251,7 @@ export type OfKind<T extends { readonly kind: string }, Kind extends T['kind']> 
     : never;
 
 export type NodeKind = 'regularNode' | 'newIterator' | 'collector';
+export type NodeType = 'backendNode';
 
 export type InputData = Readonly<Record<InputId, InputValue>>;
 export type InputHeight = Readonly<Record<InputId, number>>;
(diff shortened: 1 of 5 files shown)


joeyballentine:
This needs a migration... will do in a sec

RunDevelopment:
You just activated my trap card or something.

Anyway, we still have a bunch of code that uses `node.type` to get the node kind. You broke all of that, and since `node.type` is typed as `string` there aren't compiler errors. Fun. I would suggest using VSCode's "Find All References" to find all uses of `.type`.

I know this because I investigated this while doing #2490, but I didn't make this change in this PR, because it seemed like it should be its own PR. Well, I guess that PR is right now :)

joeyballentine:
Hmm, I think I'll just close this and wait til we update to RF v12 to make this change, so that it will actually be typed correctly

RunDevelopment:
You can also keep it open. I'm gonna work on #2266 tomorrow and I'll change most of the code using `node.type`. So you'll only have to fix maybe one or two occurrences after that.

PR #2494, opened 2024-01-22
Fix lineage checks in UI

### LINKED ISSUE #2266: Fix lineage

RunDevelopment:
Fixes #2266.

This PR fixes the lineage checks for nodes. They should just work now, even in more difficult cases. I also added lineage checks to `checkNodeValidity` which ensures that a chain with invalid lineage never makes it to the backend. This also means that we now see lineage errors in the footer.

Note: It is still possible to create invalid connections using the context node selector. I haven't touched that part of the code yet. However, invalid connections will be correctly detected as such. I'll fix the context selector in a follow-up PR.

Examples:
image
image
image

### DIFF
--- a/src/common/nodes/checkNodeValidity.ts
+++ b/src/common/nodes/checkNodeValidity.ts
@@ -159,7 +159,7 @@ export const checkAssignedLineage = (
     const input = schema.inputs.find((i) => i.id === inputId)!;
 
     const differentLineage = () =>
-        invalid('Cannot connect node to 2 different sequences of different origin.');
+        invalid("Cannot connect node to 2 sequences that don't share the same source node.");
     const nonIteratorInput = () =>
         invalid(`Input ${input.label} cannot be connected to a sequence.`);
 
--- a/src/common/nodes/checkNodeValidity.ts
+++ b/src/common/nodes/checkNodeValidity.ts
@@ -160,7 +160,8 @@ export const checkAssignedLineage = (
 
     const differentLineage = () =>
         invalid('Cannot connect node to 2 different sequences of different origin.');
-    const nonIteratorInput = () => invalid(`Input ${input.label} cannot be connected to a sequence.`);
+    const nonIteratorInput = () =>
+        invalid(`Input ${input.label} cannot be connected to a sequence.`);
 
     switch (schema.kind) {
         case 'regularNode': {
--- a/src/common/nodes/checkNodeValidity.ts
+++ b/src/common/nodes/checkNodeValidity.ts
@@ -158,6 +158,10 @@ export const checkAssignedLineage = (
 ): Validity => {
     const input = schema.inputs.find((i) => i.id === inputId)!;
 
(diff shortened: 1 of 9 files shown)


joeyballentine:
Maybe this would be better suited to a follow up PR, but I kinda wonder if we should change the verbiage we use to use a more user-friendly term rather than iterator. From a user perspective, any iterator/iterable is just a sequence, list, batch, etc rather than something that iterates over some data. Perhaps we can explain it that way? I'm not super sure about that though

src/common/nodes/checkNodeValidity.ts
```diff
@@ -118,3 +148,62 @@ export const checkRequiredInputs = (schema: NodeSchema, inputData: InputData): V
...
+    sourceLineage: Lineage | null,
+    nodeId: string,
+    inputId: InputId,
+    schema: NodeSchema,
+    chainLineage: ChainLineage
+): Validity => {
+    const input = schema.inputs.find((i) => i.id === inputId)!;
+
+    const differentLineage = () =>
+        invalid('Cannot connect node to 2 different sequence of different origin.');
```
joeyballentine, line 162:
```suggestion
        invalid('Cannot connect node to 2 different sequences of different origin.');
```

src/common/nodes/checkNodeValidity.ts
```diff
@@ -118,3 +148,62 @@ export const checkRequiredInputs = (schema: NodeSchema, inputData: InputData): V
...
+    const nonIteratorInput = () => invalid(`Input ${input.label} cannot be connect to a sequence.`);
```
joeyballentine, line 163:
```suggestion
    const nonIteratorInput = () => invalid(`Input ${input.label} cannot be connected to a sequence.`);
```

joeyballentine (approved):

PR #2513, opened 2024-01-25
Add support for conditional dropdown options

RunDevelopment:
This adds the ability to enable dropdown options only if certain conditions are met. It uses the same condition API as conditional groups, so it's quite powerful.

Changes:
- Moved `Condition` class and familiars into its own file.
- Add optional `condition` key to `DropDownOption`.
- Add `conditions: dict[T, Condition]` parameter to `EnumInput`. This makes it easy to define option conditions.
- Add `Dropdown.wrap_with_conditional_group()`. This function essentially adds a conditional group around the dropdown with its options are conditions. The basic idea is that if none of the options have their conditions met, then the dropdown is effectively disabled since you cannot input a value. `wrap_with_conditional_group` adds a conditional group that hides the dropdown if this is the case.
- Generally improved the condition API in the frontend.
- Changed when enum conditions are considered to be fulfilled. It now requires the value of the enum to match **and** the conditions of the enum input to match. The idea is that an enum value is only valid if the enum is visible/enabled.


joeyballentine (approved):
LGTM

PR #2519, opened 2024-01-27
Add some directory manipulation nodes

joeyballentine:
Making this draft for feedback on these nodes. AFAIK, doing the type for the back directory node is impossible right now.

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+from nodes.properties.inputs import DirectoryInput, TextInput
+from nodes.properties.outputs import DirectoryOutput
+
+from .. import value_group
+
+separator = "/" if sys.platform == "linux" else r"\\"
+
+
+@value_group.register(
+    schema_id="chainner:utility:into_directory",
+    name="Into Directory",
+    description="Goes forward into a directory.",
+    icon="BsFolder",
+    inputs=[
+        DirectoryInput(
+            "Directory", must_exist=False, label_style="hidden", has_handle=True
+        ),
+        TextInput("Folder", has_handle=True),
+    ],
+    outputs=[
+        DirectoryOutput(
+            "Directory",
+            output_type='Directory { path: string::concat(Input0.path, "'
+            + separator
+            + '", Input1) }',
+        ),
+    ],
+)
+def into_directory_node(directory: Path, folder: str) -> Path:
+    return directory / folder
--- a/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from pathlib import Path
 
-from nodes.properties.inputs import DirectoryInput, TextInput
+from nodes.properties.inputs import BoolInput, DirectoryInput, TextInput
 from nodes.properties.outputs import DirectoryOutput
 
 from .. import value_group
@@ -18,6 +18,7 @@ from .. import value_group
             "Directory", must_exist=False, label_style="hidden", has_handle=True
         ),
         TextInput("Folder", has_handle=True),
+        BoolInput("Create if not exists", default=True),
     ],
     outputs=[
         DirectoryOutput(
@@ -26,5 +27,10 @@ from .. import value_group
         ),
     ],
 )
-def into_directory_node(directory: Path, folder: str) -> Path:
-    return (directory / folder).resolve()
+def into_directory_node(directory: Path, folder: str, create_path: bool) -> Path:
+    resolved = (directory / folder).resolve()
+    if create_path and not resolved.exists():
+        resolved.mkdir(parents=True)
+    elif not resolved.exists():
+        raise FileNotFoundError(f"Directory does not exist: {resolved}")
+    return resolved
--- a/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
@@ -8,7 +8,7 @@ from nodes.properties.outputs import DirectoryOutput
 
 from .. import value_group
 
-separator = "/" if sys.platform == "linux" else r"\\"
+separator = r"\\" if sys.platform == "win32" else "/"
 
 
 @value_group.register(
--- a/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
@@ -27,4 +27,4 @@ from .. import value_group
     ],
 )
 def into_directory_node(directory: Path, folder: str) -> Path:
-    return directory / folder
+    return (directory / folder).resolve()
--- a/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
@@ -1,14 +1,11 @@
 from __future__ import annotations
 
-import sys
 from pathlib import Path
 
 from nodes.properties.inputs import DirectoryInput, TextInput
 from nodes.properties.outputs import DirectoryOutput
 
 from .. import value_group
-
-separator = r"\\" if sys.platform == "win32" else "/"
 
 
 @value_group.register(
@@ -25,9 +22,7 @@ separator = r"\\" if sys.platform == "win32" else "/"
     outputs=[
         DirectoryOutput(
             "Directory",
-            output_type='Directory { path: string::concat(Input0.path, "'
-            + separator
-            + '", Input1) }',
+            output_type="Directory { path: combinePath(Input0.path, Input1) }",
         ),
     ],
 )
--- a/backend/src/packages/chaiNNer_standard/utility/directories/back_directory.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directories/back_directory.py
@@ -1,0 +1,30 @@
+from __future__ import annotations
+
+from pathlib import Path
+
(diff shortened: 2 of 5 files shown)


backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
```diff
@@ -0,0 +1,35 @@
...
+        DirectoryOutput(
+            "Directory",
+            output_type='Directory { path: string::concat(Input0.path, "'
+            + separator
+            + '", Input1) }',
+        ),
+    ],
+)
+def into_directory_node(directory: Path, folder: str) -> Path:
+    return directory / folder
```
RunDevelopment, line 35: This has issues with path traversal. Example:
```py
>>> Path("C:/foo/bar") / "/bar"
WindowsPath('C:/bar')
>>> Path("C:/foo/bar") / "D:/bar"
WindowsPath('D:/bar')
>>> Path("C:/foo/bar") / "../"
WindowsPath('C:/foo/bar/..') 
# ".." is not a valid dir name, so some libraries may resolve the path before. 
# This could lead to parts of chainner interpreting paths differently, which can lead to potential security vulnerabilities.
>>> (Path("C:/foo/bar") / "../").parent
WindowsPath('C:/foo/bar')
```

We need to define proper behavior for all these cases. The first question is whether we even want to allow `..` and absolute paths.

RunDevelopment:
> AFAIK, doing the type for the back directory node is impossible right now.

Just add [an intrinsic function](https://github.com/chaiNNer-org/chaiNNer/blob/36c01c18dd8255cbd75918386d8aebe2efac4903/src/common/types/chainner-scope.ts#L136).

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
```diff
@@ -0,0 +1,35 @@
...
+
+import sys
+from pathlib import Path
+
+from nodes.properties.inputs import DirectoryInput, TextInput
+from nodes.properties.outputs import DirectoryOutput
+
+from .. import value_group
+
+separator = "/" if sys.platform == "linux" else r"\\"
```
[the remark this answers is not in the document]
joeyballentine: Oh right. Oops

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
```diff
@@ -0,0 +1,35 @@
...
+            "Directory", must_exist=False, label_style="hidden", has_handle=True
+        ),
+        TextInput("Folder", has_handle=True),
+    ],
+    outputs=[
+        DirectoryOutput(
+            "Directory",
+            output_type='Directory { path: string::concat(Input0.path, "'
+            + separator
+            + '", Input1) }',
```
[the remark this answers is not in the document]
joeyballentine: Then i don't think we can have a navi type for this

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
```diff
@@ -0,0 +1,35 @@
...
+from nodes.properties.outputs import DirectoryOutput
+
+from .. import value_group
+
+separator = "/" if sys.platform == "linux" else r"\\"
+
+
+@value_group.register(
+    schema_id="chainner:utility:into_directory",
+    name="Into Directory",
```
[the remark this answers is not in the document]
joeyballentine: What if we add a "make directory if not exists" toggle, that if unchecked requires the dir to exist, and if checked would make the dir?

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
joeyballentine: I believe with my most recent push all of these are now consistent between the python value and navi type and work as you would expect.

joeyballentine:
Question: do we need the back directory node if into directory supports `..`? I think we could just get rid of back directory entirely and rename into directory to change directory

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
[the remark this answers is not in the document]
RunDevelopment: I like that.

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
```diff
@@ -0,0 +1,36 @@
...
+            output_type="Directory { path: combinePath(Input0.path, Input1) }",
+        ),
+    ],
+)
+def into_directory_node(directory: Path, folder: str, create_path: bool) -> Path:
+    resolved = (directory / folder).resolve()
+    if create_path and not resolved.exists():
+        resolved.mkdir(parents=True)
+    elif not resolved.exists():
+        raise FileNotFoundError(f"Directory does not exist: {resolved}")
```
RunDevelopment, line 35: So the directory must always exist?

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
RunDevelopment: ... You solve one thing. Since you `.resolve()` now, it does the path traversal for `..`.

But my question was whether we should even allow that. Same for absolute paths.

RunDevelopment (commented):
>  I think we could just get rid of back directory entirely and rename into directory to change directory

User experience. Normal people don't know that they can use `..` in paths.

---

Also, about naming. I'm not sure whether "back" and "into" are the terms to communicate the operations. I dislike that there is no symmetry between "back" and "into" in their normal usage. "Back" also has the issue that is term is typically used in browsers and file explorers to refer to the history of the current tab/window.

The programming terms are "parent" and "join", but I don't think that we should necessarily copy that (UX again).

I was thinking that we could use the terms "up" and "down", but I'm not sure how to integrate that into the node names...

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
RunDevelopment: ... You solve one thing. Since you `.resolve()` now, it does the path traversal for `..`.

But my question was whether we should even allow that. Same for absolute paths.

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
joeyballentine: Yes. Do you not think so? I don't think we should be passing in invalid directories into nodes willy nilly

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
joeyballentine: I think so, because it mirrors the behavior of both javascript and python for how they combine paths. Why would we make this different if there's a "standard" for resolving a valid path out of combining path strings like this?

RunDevelopment: Because they are joining paths. They don't call these functions "go into directory".

joeyballentine: Sure yeah, but we also don't have to call it that. I would personally think it would be far more convenient for someone who does know how paths work to be able to pass in a directory, type `../../../whatever/whatever` and have it work, while also retaining the ability to go up _x_ directories in a dedicated node.

RunDevelopment: Sure. Then let's support `..`. But I don't see how joining with absolute paths would be useful.

joeyballentine: > But I don't see how joining with absolute paths would be useful.

These are not absolute paths

RunDevelopment: See the following examples:

```py
>>> Path("C:/foo/bar") / "/bar"
WindowsPath('C:/bar')
>>> Path("C:/foo/bar") / "D:/bar"
WindowsPath('D:/bar')
```

The string value this node gets can very well represent an absolute path.

joeyballentine: Ok, then what do you suggest we do? restrict it to just single folder names?

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
RunDevelopment: Yeah, I don't like it.

1. This means that the correctness of a chain now depends on the file system. If user A has folder `foo` but user B doesn't, it's not gonna work for user B unless they check that box. Depending on the file system is unavoidable for node like Load Image (singular), but it can be avoided here.
2. This leads to errors being ignored. Nodes like Load Images don't allow non-existent directories as input for obvious reasons. So if you always just create those directories, then Load Images can't error for incorrect chains.
3. We already allow passing non-existent directories. Save Image and the Directory node support it.
4. Side effects. You turned this node into a side effect node.

backend/src/packages/chaiNNer_standard/utility/directories/into_directory.py
RunDevelopment: > restrict it to just single folder names?

We can't do that since nodes like Load Images can return multiple levels as their subdirectory path.

I guess we should just disallow absolute paths as the join argument. Unfortunately, python doesn't consider `/bar` an absolute path on windows. This is consistent with how windows itself treats paths, but it doesn't make sense for a cross-platform API. So we have to handle this ourselves.

RunDevelopment:
Also, we have to handle invalid characters ourselves. Basically, names can't contain `<>:"|?*` and non-printable ASCII characters (0-31). Yes, linux allows more, but it's better to be more restrictive for this kind of stuff.

https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

joeyballentine:
> Also, we have to handle invalid characters ourselves

what would be your recommended approach for that?

RunDevelopment:
Idk. Use a regex or define a set of characters and check each one. Doesn't really matter.

PR #2525, opened 2024-01-27
Move model to device after pytorch interpolation

joeyballentine:
Fixes the bug mentioned by metin on discord

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/pytorch/utility/interpolate_models.py
+++ b/backend/src/packages/chaiNNer_pytorch/pytorch/utility/interpolate_models.py
@@ -12,9 +12,11 @@ from spandrel import (
     ModelLoader,
 )
 
+from api.node_context import NodeContext
 from nodes.impl.pytorch.utils import np2tensor, tensor2np
 from nodes.properties.inputs import ModelInput, SliderInput
 from nodes.properties.outputs import ModelOutput, NumberOutput
+from packages.chaiNNer_pytorch.settings import get_settings
 
 from .. import utility_group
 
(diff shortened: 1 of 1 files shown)


RunDevelopment:
... Why does this solve the bug?

joeyballentine:
Since the load model node moves the model to gpu, if you did an interp into another interp, where the second interp comes from a load model node (so model 1 and model 2 into interp 1, and interp 1 and model 3 into interp 2), then one model would be on the gpu and one would be on cpu. This fixes that and makes it consistent.

Alternatively, perhaps we could change the model input's enforce method to always move it to the gpu?

RunDevelopment:
Okay, I just looked at the discord messages. This is not a fix.

The issue is that the node currently can't interpolate 2 models when they are on different devices. So you tried to make sure that this node puts the new model it returns on the same device as other nodes.

Why is this not a fix? You didn't solve the issue that this node cannot interpolate 2 models when they are on different devices. The cause of the issue still remains and there will be a bug report again in the future. For example, like this:

1. Add a Load model node and select a model.
2. Go to settings and change PyTorch from GPU to CPU (or vise versa).
3. Add another Load model node.
4. Setup the chain to interpolate both models are run it.
5. Same error.

Instead of *assuming* that models will always be on the same device, just move both models to the same device.

RunDevelopment:
> Alternatively, perhaps we could change the model input's enforce method to always move it to the specified device?

This would require inputs to have access to the current node execution context. Not ideal, since this would prevent optimizations in the future.

RunDevelopment (approved):

PR #2528, opened 2024-01-28
Create "Add" node

mrjschulte:
Create a new chaiNNer node that "adds" values to an input image.


joeyballentine:
Is this functionality not already covered by the Brightness & Contrast node?

mrjschulte:
I would say that is a form of "combined functionality" in the Brightness/Contrast node. 

As I plan to add more atomic image processing operations to chaiNNer having individual Core Image processing operators available such as Add, Multiply, Divide, etc allows the users to be much more granular in their processing steps with the application.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
```diff
@@ -0,0 +1,55 @@
...
+    """Add values to an image"""
+
+    add /= 100
+
+    if add == 0:
+        return img
+
+    _, _, c = get_h_w_c(img)
+
+    add: float = add
```
RunDevelopment, line 42: Why?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
```diff
@@ -0,0 +1,55 @@
...
+            controls_step=1,
+        ),
+    ],
+    outputs=[ImageOutput(image_type="Input0")],
+)
+def add_node(
+    img: np.ndarray, add: float
+) -> np.ndarray:
+    """Add values to an image"""
+
```
RunDevelopment, line 34: We stopped documenting nodes like since the same information can be obtained from the node's description. I know that a few of the old nodes still do this, so sorry for misleading you.

```suggestion
```

RunDevelopment (commented):
A few more thoughts I had while reviewing:

1. Considering your plan, it might be good to make a new node group for these nodes. You can do that later of course.
2. Please note that chainner's images are always in the range 0-1. So if the Add node returns values outside that range, chainner will automatically clip them (with `np.clip(img, 0, 1)`).

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
mrjschulte: copied from other node I was playing with, will remove.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
mrjschulte: All good, will remove.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
mrjschulte: and gone.

mrjschulte:
(What's left to do here? @joeyballentine @RunDevelopment )

RunDevelopment:
Ci, so formatting.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/add.py
```diff
@@ -0,0 +1,35 @@
...
+            controls_step=1,
+        ),
+    ],
+    outputs=[ImageOutput(image_type="Input0")],
+)
+def add_node(img: np.ndarray, add: float) -> np.ndarray:
+    if add == 0:
+        return img
+
+    img = img + add
```
RunDevelopment, line 33: I think you forgot to divide `add` by 100.

mrjschulte: Actually no, I've put the processing for this node in an explicit float workflow. Do you think we need a toggle?

mrjschulte: This feels like a bigger conversation so I'll just add in the divide so we can merge the node and get it into the app toolset.

joeyballentine (approved):

PR #2532, opened 2024-01-28
Update load_video node for more accurate color conversion and chroma reconstruction

mrjschulte:
This PR updates the ffmpeg based load_video node to use a more accurate color decoding and chroma reconstruction path.  

This node now uses lanczos filtering by default and also skips the previously present extra conversion from rgb24 to bgr24 by having ffmpeg send a bgr24 rawvideo pipe natively to chaiNNer

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/video_frames/load_video.py
+++ b/backend/src/packages/chaiNNer_standard/image/video_frames/load_video.py
@@ -68,7 +68,7 @@ def load_video_node(
 
     ffmpeg_reader = (
         ffmpeg.input(path)
-        .output("pipe:", format="rawvideo", pix_fmt="rgb24")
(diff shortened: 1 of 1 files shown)


joeyballentine (dismissed):
Thanks! Videos process with a roughly 1.05x speed increase with this. That, combined with the better filtering, makes this a decent improvement.

joeyballentine:
We use a great package called `ruff` for formatting and linting, which found some small issues with this. This PR will be good to go once those detected issues are fixed. Thanks again!

mrjschulte:
Ok, removed the dead cv2 import so should be good now.

joeyballentine (approved):

adegerard:
question: sws flags are for scaling, why having added them?

PR #2544, opened 2024-02-03
Collapsing/expanding nodes

### LINKED ISSUE #2529: Allow for collapsing of nodes on the schematic.
It would be great to be able to collapse a node down to its title bar only. With this in place the Schematic becomes a lot more manageable and the user retains the overview of which nodes they are working on/with.

joeyballentine:
image

Closes #2529

### DIFF
--- a/src/renderer/components/Handle.tsx
+++ b/src/renderer/components/Handle.tsx
@@ -84,7 +84,7 @@ const HandleElement = memo(
     }
 );
 
-export const Div = chakra('div', {
+const Div = chakra('div', {
     baseStyle: {},
 });
 
--- a/src/renderer/components/Handle.tsx
+++ b/src/renderer/components/Handle.tsx
@@ -84,7 +84,7 @@ const HandleElement = memo(
     }
 );
 
-const Div = chakra('div', {
+export const Div = chakra('div', {
     baseStyle: {},
 });
 
@@ -98,7 +98,7 @@ export interface HandleProps {
     isIterated: boolean;
 }
 
-const getBackground = (colors: readonly string[]): string => {
+export const getBackground = (colors: readonly string[]): string => {
     if (colors.length === 1) return colors[0];
 
     const handleColorString = colors
--- a/src/renderer/components/NodeDocumentation/NodeExample.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeExample.tsx
@@ -177,6 +177,10 @@ export const NodeExample = memo(({ accentColor, selectedSchema }: NodeExamplePro
                                 icon={selectedSchema.icon}
                                 name={selectedSchema.name}
                                 selected={false}
+                                useCollapse={{
+                                    isCollapsed: false,
+                                    toggleCollapse: () => {},
+                                }}
                             />
                             <NodeBody
                                 animated={false}
--- a/src/renderer/components/NodeDocumentation/NodeExample.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeExample.tsx
@@ -19,6 +19,7 @@ import { EMPTY_ARRAY, EMPTY_MAP, EMPTY_OBJECT, EMPTY_SET } from '../../../common
 import { BackendContext } from '../../contexts/BackendContext';
 import { FakeNodeProvider } from '../../contexts/FakeExampleContext';
 import { NodeState, TypeInfo, testForInputConditionTypeInfo } from '../../helpers/nodeState';
+import { CollapsedHandles } from '../node/CollapsedHandles';
 import { NodeBody } from '../node/NodeBody';
 import { NodeFooter } from '../node/NodeFooter/NodeFooter';
 import { NodeHeader } from '../node/NodeHeader';
@@ -171,6 +172,8 @@ export const NodeExample = memo(({ accentColor, selectedSchema }: NodeExamplePro
         testCondition: testForInputConditionTypeInfo(inputData, selectedSchema, typeInfo),
     };
 
+    const [isCollapsed, setIsCollapsed] = useState(false);
+
     return (
         <Center key={selectedSchema.schemaId}>
             <FakeNodeProvider isFake>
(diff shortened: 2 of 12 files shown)


RunDevelopment:
I think we should largely copy blender for this feature. They do it like [this](https://docs.blender.org/manual/en/latest/interface/controls/nodes/parts.html):

image
image

The main aspects I would copy from this are:
1. The collapse/expand icons. I think their icons are a lot clearer.
2. Their input/output handles are usable even when collapsed. This would make some of our nodes fully functional even when minimized.

What do you think?

joeyballentine:
How does 2. look when there's a shit ton of handles?

RunDevelopment:
It increases the size of the header. Quite clever.

image

joeyballentine:
To be completely honest, that sounds like wayyyyy too much work to do, especially now that i've already done it this way. And not to mention our nodes are a different shape than blender's, and they have more information on them. Unless you think we should get rid of the validation marker and the disable toggle when a node is collapsed and just have the header?

RunDevelopment:
> To be completely honest, that sounds like wayyyyy too much work to do

It might not be. Since you brought up removing the footer, I tried doing that. (Didn't sound like such a bad idea.) Honestly, it looks pretty good with centered handles:

image

As for how this works with a lot of handles: it doesn't.

image

It can deal with up to 6 handles without overflowing, which should cover most nodes. And if it does overflow, it just looks a little bad, but it doesn't break anything.

The only thing I would still change is to show the status icon where there is currently empty space (where you have a placeholder icon button). But I would only show it when there is an error. No need to clutter the clean UI when we have nothing to report.

What do you think?

---

That aside, I found another minor issue: when I minimize a node with a multiline text box (e.g. text to image), the collapse toggle moves (since it is right aligned). Instead of making nodes keep the increased width, I would suggest moving the toggle to the left of the title. You can see what that looks like in the screenshots above.

RunDevelopment:
I made a PR for both: #2551

RunDevelopment:
@joeyballentine please don't merge after https://github.com/chaiNNer-org/chaiNNer/pull/2551. There are still 2 bugs:

1. The color of minimized input handles isn't always the same as that of the real ones.
2. Node documentation has the collapse toggle, but it does nothing.

joeyballentine:
Fixed the two bugs, should be good to go now

src/renderer/components/Handle.tsx
```diff
@@ -84,7 +84,7 @@ const HandleElement = memo(
     }
 );
 
-const Div = chakra('div', {
+export const Div = chakra('div', {
```
RunDevelopment, line 87: This element is never used outside this file.
```suggestion
const Div = chakra('div', {
```

src/renderer/components/node/CollapsedHandles.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const connectedEdge = useMemo(() => {
+        return getEdges().find(
+            (e) => e.target === nodeId && parseTargetHandle(e.targetHandle!).inputId === inputId
+        );
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [edgeChanges, getEdges, nodeId, inputId]);
+
+    const { functionDefinitions } = useContext(BackendContext);
+
+    const sourceTypeColor = useMemo(() => {
```
RunDevelopment, line 39: Joey... did you actually just copy that entire color logic? Please extract all of this out into a hook or similar to avoid code duplications between this `InputHandle` and the real one.

src/renderer/components/node/Node.tsx
```diff
@@ -75,9 +76,10 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
 
     const { sendToast } = useContext(AlertBoxContext);
     const { categories } = useContext(BackendContext);
+    const { setNodeCollapsed } = useContext(GlobalContext);
     const { getNodeProgress, getNodeStatus } = useContext(ExecutionContext);
 
-    const { id, inputData } = data;
+    const { id, inputData, isCollapsed } = data;
```
RunDevelopment, line 82: Instead of assuming that `undefined` is roughly the same `false` below, just set it to `false` here.

```suggestion
    const { id, inputData, isCollapsed = false } = data;
```

src/renderer/components/node/Node.tsx
```diff
@@ -192,6 +194,11 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
     const disabled = useDisabled(data);
     const menu = useNodeMenu(data, disabled, { reload: startingNode ? reload : undefined });
 
+    const toggleCollapse = useCallback(() => {
+        setNodeCollapsed(id, !isCollapsed);
+    }, [id, isCollapsed, setNodeCollapsed]);
+    const useCollapse = { isCollapsed: isCollapsed ?? false, toggleCollapse };
```
RunDevelopment, line 200: And with this, you effectively disabled memoization for `NodeHeader` and all of its parents...

Why do you use these fake hooks in props in the first place? I never understood that. Why not pass 2 props?

src/renderer/components/NodeDocumentation/NodeExample.tsx
```diff
@@ -150,6 +151,29 @@ export const NodeExample = memo(({ accentColor, selectedSchema }: NodeExamplePro
...
+        isLocked: false,
+        connectedInputs: EMPTY_SET,
+        connectedOutputs: EMPTY_SET,
+        iteratedInputs,
+        iteratedOutputs,
+        type: typeInfo,
+        testCondition: testForInputConditionTypeInfo(inputData, selectedSchema, typeInfo),
+    };
+
+    const [isCollapsed, setIsCollapsed] = useState(false);
```
RunDevelopment, line 175: What's the point of being able to minimize nodes in node documentation? I think example nodes shouldn't have the collapse toggle.

RunDevelopment (changes requested):

src/renderer/components/NodeDocumentation/NodeExample.tsx
joeyballentine: I thought we wanted the nodes in the docs to behave just like they do in the editor for demonstration purposes?

RunDevelopment: For inputs and outputs, yes. But I don't see any reason why nodes should be collapsible in our docs.

joeyballentine: Ok

src/renderer/components/node/Node.tsx
joeyballentine: Honestly, I have no idea

RunDevelopment (approved):

PR #2546, opened 2024-02-04
Create "Log2Lin" node

mrjschulte:
This node will apply the Kodak Cineon set of Logarithmic functions to an image.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
+++ b/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
@@ -46,13 +46,13 @@ from .. import adjustments_group
     ],
     outputs=[ImageOutput(image_type="Input0")],
 )
-def log2lin_node(
+def log_to_linear_node(
     img: np.ndarray, black: float, white: float, gamma: float, invert_log2lin: bool
 ) -> np.ndarray:
     offset = pow(10.0, (black - white) * 0.002 / gamma)
     gain = 1.0 / (1.0 - offset)
 
-    if not invert_log2lin:
+    if not invert_log_to_linear:
         img = gain * (pow(10.0, (1023.0 * img - white) * 0.002 / gamma) - offset)
     else:
         img = (np.log10(img / gain + offset) / (0.002 / gamma) + white) / 1023.0
--- a/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
+++ b/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
@@ -47,8 +47,7 @@ from .. import adjustments_group
     outputs=[ImageOutput(image_type="Input0")],
 )
 def log2lin_node(
-    img: np.ndarray, black: float, white: float, gamma: float, invert_log2lin: bool
-) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
```diff
@@ -0,0 +1,59 @@
...
+
+from nodes.properties.inputs import BoolInput, ImageInput, SliderInput
+from nodes.properties.outputs import ImageOutput
+
+from .. import adjustments_group
+
+
+@adjustments_group.register(
+    schema_id="chainner:image:log2lin",
+    description="Convert all channels in an image to a scene linear encoding using the Cineon logarthmic function set.",
```
RunDevelopment, line 13: Typo and could you explain this a bit more? I have no idea what this is for from just reading the description. (Descriptions are markdown, so you can also link to websites to do the explaining for you.)

```suggestion
    description="Convert all channels in an image to a scene linear encoding using the Cineon logarithmic function set.",
```

mrjschulte:
The mentioned Kodak Cineon functions are a means to help convert the logarithmic densities encoded on a 0-1 real world medium such as Film Negative into a linear floating point encoding for use in imaging workflows , and vice versa.

RunDevelopment:
Then could you please mention this in the node's description?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
```diff
@@ -0,0 +1,59 @@
...
+from nodes.properties.inputs import BoolInput, ImageInput, SliderInput
+from nodes.properties.outputs import ImageOutput
+
+from .. import adjustments_group
+
+
+@adjustments_group.register(
+    schema_id="chainner:image:log2lin",
+    description="Convert all channels in an image to a scene linear encoding using the Cineon logarithmic function set.",
+    name="Log2Lin",
```
RunDevelopment, line 14: I think we should call this node "Log To Linear" to be consistent with names of other nodes (e.g. "Metal To Specular", "ONNX To NCNN", "Text As Image").

mrjschulte: Elaborating on this note, I have a more general question about the naming schema for nodes: I noticed a lot of the nodes are also using additional verbs such as "Change", "Create", "Add", etc. Is there some sort of specific thinking around this approach so far? Or is it solely based on providing clarity to the Nodes function?

joeyballentine: We don't have any kind of consistent system right now, we just do whatever seems right at the time. We've never really gone back and reviewed anything or tried to make it consistent

joeyballentine:
Just FYI, we require all node names to match their function name and file name for ease of searchability. This will actually automatically be changed for you if you run with `npm run dev` after/while you update it

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/log2lin.py
mrjschulte: Node renamed and updated explanations and other items. Relinted. Gtg from my pov.

joeyballentine (approved):

PR #2548, opened 2024-02-05
Add "Breakpoints" feature to edges for custom edge routing

### LINKED ISSUE #2420: Reroute edges for organizing nodes
**Motivation**
I often have noodles from nodes crossing other nodes and it makes it hard to organize things and often I lose track of which noodle is connected to which node.

**Description**
Just like in **Blender** with the *Reroute* function.
## without Reroute
noReroute

## with Reroute
with reroute I can simply *route* my nodes for better organization
Reroute

**Alternatives**
some kind of node that can be used as a portal, for example:
I input a node into *Portal A* and i can get the node from another *Portal A* at the location i need my node.
Basically like the game Portal

joeyballentine:
image

The implementation actually ended up being quite simple. I added a new node type that is secretly just a passthrough node with special styling. This node type is not manually connectable but rather is entirely controlled by us managing the edges going between it and the actual nodes.

This means that everything just works with it. There's a tad bit of jankiness with the node-merge (hold alt and connect) feature since these lines aren't bezier -- but I would prefer to just fix that in a separate PR as it works well enough still.

The breakpoints can be added or removed via context menu:
image
image

When a breakpoint is removed, an edge fills its place. When a breakpoint is added, it creates two new edges.

If an edge is deleted, all edges and breakpoints along what visually appears as the single edge will be removed. The way I made this happen was I jut made the BreakPoint component handle deleting itself if at any point in time its missing an edge, as a breakpoint can't just exist on its own. Perhaps in the future we could make it so breakpoints _can_ be on their own, but I think this is good enough for a V1, and it can be improved upon later.

Oh yeah, and since they're just nodes, it works with the auto-layout feature as well.

image

## Fast follows:
- Improving the alt-drag-to-connect feature with this
- Making the edges out and in to actual nodes be half-bezier (if that's even possible)

## Longer-term followups
- allowing breakpoints to stand on their own and be connectable from them

Closes #2420


RunDevelopment:
Very cool. It's very clever to use pass through nodes. I haven't look at the code too deeply yet, so I'll start with bugs and suggestions:

1. Added and removing breakpoints is a pain. I would suggest that Alt+Click (or Ctrl+Click or similar) on an edge adds a breakpoint there. Double-clicking on a breakpoint could remove it, just like double-clicking an edge removes it.
2. Using straight lines looks kinda bad compared to our smooth edges. 
3. The hit box for breakpoints is small and inconsistent. It's sometimes very difficult to select breakpoints because their hitbox is *under* the hitbox of connected edges. So I would suggest increasing the size of the hitbox of breakpoints and keeping breakpoints above connected edges at all times.
4. When the size of a breakpoint increases on hover, the center of the breakpoint slightly shifts, which causes connected edges to move slightly.

joeyballentine:
> I would suggest that Alt+Click (or Ctrl+Click or similar)

Sure, we should just still leave the context menu though so this doesn't become a feature nobody is ever aware of like the Alt drag one

> Using straight lines looks kinda bad compared to our smooth edges.

You should have seen it with the smooth edges, it looked a whole lot worse. I think the ideal thing would be for it to be curved coming in and out of the handles still, but straight to and a from breakpoints. I'm just not sure how to make it so that an edge is half bezier half straight 

> keeping breakpoints above connected edges at all times.

Honestly this is just kinda an issue with react flow that I have no idea how to solve without re-implementong that complex custom sorting algo we had, and even then we still had this exact issue where edges would be over the nodes 

> the center of the breakpoint slightly shifts,

I'll try to fix this 👍

RunDevelopment:
> Sure, we should just still leave the context menu

Definetly. On that note: we should probably show the short cuts for options in our context menus.

> Honestly this is just kinda an issue with react flow that I have no idea how to solve without re-implementong that complex custom sorting algo we had, and even then we still had this exact issue where edges would be over the nodes

Edges over breakpoints is a huge issue though. Two unfortunately positioned edges can cover almost the entire hitbox of a breakpoint, making it almost unusable. We need a solution for this.

joeyballentine:
> We need a solution for this.

In my opinion to fix this we need to PR react flow and adjust the default sorting logic so that edges don't ever get hoisted over the nodes they're connecting to. Like clamp them at the node's zIndex - 1 max.

That doesn't fix the issue of overlapping separate edges, but that's an entirely different problem that I don't actually think is solvable personally

joeyballentine:
I believe I have fixed all of your concerns. Please try it out again and let me know if there's anything else I should change

joeyballentine:
Fixed both things 
image
image

It still overlaps a tiny amount, but it doesnt seem to interfere with use.

RunDevelopment:
Very nice!

The only remaining issue is the weird alignment of breakpoints when snapped to grid.

joeyballentine:
I think I'd be able to make it so breakpoints are snapped at their position minus their radius, but I fear that that will cause some compounding issue where each render will just offset them more and more... 

As for being "weirdly off grid", snapping to grid does not guarantee that the dots in the background align with the nodes at all. Never has, unless you pick a specific value that happens to line up with the dots.

RunDevelopment:
> I fear that that will cause some compounding issue where each render will just offset them more and more...

This shouldn't be an issue.

> Never has, unless you pick a specific value that happens to line up with the dots.

That's why I changed the default snap to grid amount to 16 in #1445. This ensures that things line up with the grid. So things line up by default.

joeyballentine:
Ok, I will try to align those to center

joeyballentine:
image

src/renderer/components/ReactFlowBox.tsx
```diff
@@ -119,7 +119,14 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
         if (isSnapToGrid) {
             for (const n of displayNodes) {
-                if (!isSnappedToGrid(n.position, snapToGridAmount)) {
+                if (n.type === 'breakPoint') {
+                    const snapped = snapToGrid(n.position, snapToGridAmount);
+                    const pos = {
+                        x: snapped.x - (n.width ?? 0) / 2,
+                        y: snapped.y - (n.height ?? 0) / 2,
+                    };
+                    n.position = pos;
```
RunDevelopment, line 128: You just changed what it means for a node to be snapped to grid. This will cause problems. E.g. the node formatting already makes assumptions about grid alignment, and this change will likely interfere with it.

Why not just move the visuals of the node with CSS?

joeyballentine: that sounds infinitely more complicated. If i do it that way, i then have to go and change the logic for calculating the edge lines and i would have to offset all those by however much we offset the breakpoints in CSS. I suppose i could try it... but that doesnt seem like an elegant solution to this

joeyballentine: image

This is after auto-formatting. I honestly don't see an issue with this

RunDevelopment: See #2577

RunDevelopment (approved):

PR #2549, opened 2024-02-05
Add a "Hub" node

joeyballentine:
It's like passthrough, but where you can add many inputs/outputs to it. Suggested by @mrjschulte 

Part of me does wonder if we should just put this functionality into passthrough though


RunDevelopment:
What's the motivation for this?

backend/src/api/output.py
```diff
@@ -66,5 +68,6 @@ def get_broadcast_type(self, _value: object) -> navi.ExpressionJson | None:
         return None
 
     def enforce(self, value: object) -> object:
-        assert value is not None
+        if not self.allow_none:
+            assert value is not None
```
RunDevelopment, line 72: This method is overridden by derived classes, so all derived classes now have a useless `allow_none` property that is not checked.

I would suggest the following: make an `AnyOutput` class where you override `enforce`.

mrjschulte:
Sometimes you want to be able to design a node flow that doesn't have to source all of its originating connections from the other end of the schematic. A Passthrough/Hub such as this is a great addition to the layout and flow of the processes and helps the User visually "lay out" where things need to go. Thanks for the work here @joeyballentine !

backend/src/api/output.py
```diff
@@ -66,5 +66,4 @@ def get_broadcast_type(self, _value: object) -> navi.ExpressionJson | None:
         return None
 
     def enforce(self, value: object) -> object:
-        assert value is not None
```
RunDevelopment, line 69: Why?

joeyballentine: Because I wasn't careful, oops

RunDevelopment:
I just noticed that this node will interfere with our [dead node elimination](https://github.com/chaiNNer-org/chaiNNer/blob/df46926e55f52c618a1d9240175e4e2e496817ef/backend/src/chain/optimize.py#L14) optimization. Since that optimization doesn't know which inputs are actually used, it can't remove unused ones. So we'll have to add another optimization to do that for us. That can be its own PR though.

RunDevelopment (approved):

PR #2551, opened 2024-02-06
More compact collapsed node design

RunDevelopment:
This implements the design for collapsed nodes that I mentioned.

image
image
image


joeyballentine:
I like this a lot. The only thing I'm not a fan of is when it overflows (like what you sent in the other PR). I'm wondering if we can come up with a good solution for that

RunDevelopment:
Ugly tall nodes:

image

Not a nice design, but it works.

mrjschulte:
I'd be interested in seeing something proposed with the connection density on the side of the nodes compressing as shown but with the height of the collapsed node limited something to something that is a max 1.5x factor of a "plain" collapsed node such as the TEXT node on the left of the above image.

joeyballentine:
@mrjschulte So you're suggesting changing the size of the handles instead? I could try that out, but first I'm gonna merge this

joeyballentine (approved):

PR #2563, opened 2024-02-08
Add `NodeView` component

RunDevelopment:
This adds a `NodeView` component so `NodeExample` doesn't have to piece together head, body, and footer anymore. This component decouples the effects of `NodeInner` from its view.


joeyballentine (approved):
fantastic, i was just thinking yesterday that we should do this

PR #2565, opened 2024-02-09
Fix react ref warning

joeyballentine:
image

React apparently doesn't like when you have a custom prop called `ref` because it thinks you're trying to set a ref on the component rather than pass one in. So, I just renamed it.


joeyballentine:
I'm getting annoyed by this warning so I'm just going to merge this. If you dislike the name, just make another Pr to change it

PR #2566, opened 2024-02-09
Improved node resetting

joeyballentine:
image

### DIFF
--- a/src/renderer/components/node/Node.tsx
+++ b/src/renderer/components/node/Node.tsx
@@ -280,12 +280,12 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
     useWatchFiles(filesToWatch, reload);
 
     const disabled = useDisabled(data);
-    const [forceReloadNumber, setForceReloadNumber] = useState(0);
+    const [memoBustNumber, setMemoBustNumber] = useState(0);
     const menu = useNodeMenu(
         data,
         disabled,
         { reload: startingNode ? reload : undefined },
-        forceReloadNumber
+        memoBustNumber
     );
 
     const toggleCollapse = useCallback(() => {
@@ -305,8 +305,8 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
             toggleCollapse={toggleCollapse}
             validity={validity}
             onContextMenu={(e) => {
+                setMemoBustNumber((n) => n + 1);
                 menu.onContextMenu(e);
-                setForceReloadNumber((n) => n + 1);
             }}
             onDragOver={onDragOver}
             onDrop={onDrop}
--- a/src/renderer/components/node/Node.tsx
+++ b/src/renderer/components/node/Node.tsx
@@ -9,7 +9,6 @@ import {
     useCallback,
     useMemo,
     useRef,
-    useState,
 } from 'react';
 import { useReactFlow } from 'reactflow';
 import { useContext, useContextSelector } from 'use-context-selector';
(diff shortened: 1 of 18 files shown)


RunDevelopment:
image

Not a fan. I always thought that the Reset option was somewhat useless, and you went and made 3 versions of it. So I'm not too keen on dedicating 25% of the space in the likely most important context menu of chainner to something like that.

So could you please explain your motivation for adding these options?

That said, I'm not strictly against adding these options (especially if there is a good use case), I'm only very against the current presentation. Assuming we keep the current options as is, I think putting them into a sub menu would be the best option.

image

joeyballentine:
So i looked into creating a submenu. It's not something chakra supports natively, and as far as i can tell would be somewhat complicated to do, so i would prefer to just not do that now if that's ok with you.

joeyballentine:
Maybe i could re-use the context menu stuff and manually open the context menu on hover? i was thinking we'd always want the submenu aligned with the parent menu, but if we're ok just sticking to the mouse that could work

joeyballentine:
The code is super gross now, but it's in a submenu and it works as you would expect

joeyballentine:
image

RunDevelopment:
Clicking on "Reset node" itself just closes the context menu. Clicking on it should do nothing.

joeyballentine:
@RunDevelopment I tried to prevent that, but doing e.preventdefault and e.stoppropagation didn't do anything

src/renderer/hooks/useNodeMenu.tsx
```diff
@@ -64,12 +75,90 @@ export const useNodeMenu = (
...
+                    setShowResetSubMenu(false);
+                }}
+                onMouseEnter={() => {
+                    setShowResetSubMenu(true);
+                }}
+                onMouseLeave={() => {
+                    setShowResetSubMenu(false);
+                }}
+                onMouseOver={() => {
+                    setShowResetSubMenu(true);
```
RunDevelopment, line 94: https://www.youtube.com/watch?v=Xn4yAbA9cZw

Please use CSS. A simple `.container:not(:hover) > .child { display: none; }` could go so far.

joeyballentine: This either doesn't work the way you think it does or I'm doing something wrong

joeyballentine: Ok, i had to put the child classname on a div wrapping the submenu list... wtf

src/renderer/components/node/Node.tsx
```diff
@@ -297,7 +297,9 @@ const NodeInner = memo(({ data, selected }: NodeProps) => {
             targetRef={targetRef}
             toggleCollapse={toggleCollapse}
             validity={validity}
-            onContextMenu={menu.onContextMenu}
+            onContextMenu={(e) => {
+                menu.onContextMenu(e);
+            }}
```
RunDevelopment, line 302: I think we can revert this.

```suggestion
            onContextMenu={menu.onContextMenu}
```

src/renderer/hooks/useContextMenu.ts
```diff
@@ -18,6 +19,8 @@ export const useContextMenu = (render: () => JSX.Element): UseContextMenu => {
     // eslint-disable-next-line react/hook-use-state
     const [id] = useState(createUniqueId);
 
+    const [position, setPosition] = useState({ x: 0, y: 0 });
```
RunDevelopment, line 22: Are you even using this value? I know it's returned by the hook, but I don't see anyone using the returned value.

joeyballentine: I explained this on discord. Unless I just missed something, the menu doesn't get rerendered when the position changes, causing the ref to not get updated and therefore the submenu to not be positioned properly. This was a way to get this hook to rerender when the position changes. 

If you can find a better solution, please make a PR. But this was the best I could do

RunDevelopment:
The first item of the submenu isn't aligned with the parent item.

image

src/renderer/hooks/useContextMenu.ts
joeyballentine: turns out, since we changed this file in the breakpoints PR to save the event as state, this is no longer necessary

joeyballentine:
image

RunDevelopment (approved):

PR #2570, opened 2024-02-09
Avoid undefined behavior in PyTorch

RunDevelopment:
See https://pytorch.org/docs/stable/generated/torch.from_numpy.html

I think this fixes the UB. Could you please test this @joeyballentine?


joeyballentine:
Will test. But I was having trouble recreating the problem yesterday

joeyballentine:
I was able to recreate the error under a "no tiling" case. Will see if this avoids that.

joeyballentine:
Confirmed this fixes that.

joeyballentine (approved):

PR #2573, opened 2024-02-10
Create "Clamp" Node

mrjschulte:
Add a new node to the chaiNNer toolset that will allow the user to Clamp the pixel ranges in an image to a user defined Minimum and Maximum value.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
+++ b/backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+import numpy as np
+
+from nodes.properties.inputs import ImageInput, SliderInput
+from nodes.properties.outputs import ImageOutput
+
+from .. import adjustments_group
+
+
+@adjustments_group.register(
+    schema_id="chainner:image:clamp",
+    name="Clamp",
+    description="Clamps the values of an image.",
+    icon="ImContrast",
+    inputs=[
+        ImageInput(),
+        SliderInput(
+            "Minimum",
+            minimum=0.0,
+            maximum=1.0,
+            default=0.0,
+            precision=4,
+            controls_step=0.001,
+            scale="log",
+        ),
+        SliderInput(
+            "Maximum",
+            minimum=0.0,
+            maximum=1.0,
+            default=1.0,
+            precision=4,
+            controls_step=0.001,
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
```diff
@@ -0,0 +1,44 @@
...
+    inputs=[
+        ImageInput(),
+        SliderInput(
+            "Minimum",
+            minimum=0.0,
+            maximum=1.0,
+            default=0.0,
+            precision=4,
+            controls_step=0.001,
+            scale="log",
```
RunDevelopment, line 25: Does `log` make sense here? Wouldn't a linear scale better?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
```diff
@@ -0,0 +1,44 @@
...
+    schema_id="chainner:image:clamp",
+    name="Clamp",
+    description="Clamps the values of an image.",
+    icon="ImContrast",
+    inputs=[
+        ImageInput(),
+        SliderInput(
+            "Minimum",
+            minimum=0.0,
+            maximum=1.0,
```
RunDevelopment, line 21: Same old: 0-1, 0-100, or 0-255? What do you think would be best here?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
mrjschulte: I generally like the log scale for my sliders as it feels more exact at the lower end.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/clamp.py
mrjschulte: I think we should slowly consider a move to a normalized / float (0.0 - 1.0 in this case) Ui paradigm for denoting image bit-depth and applicable and relevant operational value ranges.

RunDevelopment: Man, we really need a way for users to select the unit they want, and we always get 0-1. 

Let's keep it as is for now.

RunDevelopment (approved):

PR #2574, opened 2024-02-10
Rename "Change Colorspace" node to "Change Color Model" 

mrjschulte:
Renames the "Change Colorspace" node to the more correct "Change Colormodel" name which now accurately reflects what the node actually does.


backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/change_colormodel.py
```diff
@@ -26,10 +26,10 @@
 
 
 @miscellaneous_group.register(
-    schema_id="chainner:image:change_colorspace",
-    name="Change Colorspace",
+    schema_id="chainner:image:change_colormodel",
```
joeyballentine, line 29: You can't just change a schema id without also creating a migration. Changing this will break all existing chains that use this node.

backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/change_colormodel.py
```diff
@@ -26,10 +26,10 @@
...
+    name="Change Colormodel",
```
joeyballentine, line 30: Maybe this was a mistake and we should have previously separated "color" and "space", but i think here it should for sure be "color model" and not "colormodel". I double checked wikipedia and it has it as two words

backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/change_colormodel.py
RunDevelopment: Also, there's no need to change the schema ID, because it is purely internal.

mrjschulte: left Schema ID as is

backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/change_colormodel.py
mrjschulte: Updated name to "Change Color Model"

joeyballentine (approved):

PR #2575, opened 2024-02-11
Arrow-key navigation in node search panel

### LINKED ISSUE #2526: Allow for (expanded) Keyboard Arrow support in the Node Search UI
When dragging in a new/empty "noodle" to the schematic, or just right clicking on the schematic, causes the node Search UI dialog to pop up where we can then search for a node. 

This operation forces the user to remove their hand from the mouse and type with both hands on the keyboard to have the appropriate nodes/options to display in the Search UI. Being able to the subsequently also use the arrow keys to quickly move through the list of presented/filtered nodes and confirm a selection with a quick "Enter" press saves much time when quickly laying down nodes.
 
Currently having to move your hands off of the keyboard and back to the mouse to then select the node among the Search results for what you've just typed (ie. color) just feels like a Ux disconnect that can be made to work smoother.

joeyballentine:
image

Closes #2526

### DIFF
--- a/src/renderer/hooks/usePaneNodeSearchMenu.tsx
+++ b/src/renderer/hooks/usePaneNodeSearchMenu.tsx
@@ -195,7 +195,6 @@ const Menu = memo(({ onSelect, schemata, favorites, categories }: MenuProps) => 
     const scrollRef = useRef<HTMLDivElement>(null);
     useEffect(() => {
         scrollRef.current?.scrollIntoView({
-            behavior: 'instant',
             block: 'center',
             inline: 'nearest',
         });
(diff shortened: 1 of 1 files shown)


mrjschulte:
This would close #2526 ! Thanks @joeyballentine

RunDevelopment:
This is great!

I only have one request: please wrap around. When the last element and down is pressed, it should wrap around to the first element. Same for the first element + up.

joeyballentine:
wtf
image
image
image

mrjschulte:
Works great! Nice addition to the Ux!

RunDevelopment (approved):

PR #2576, opened 2024-02-11
Create "Create Colorwheel" node

mrjschulte:
Add a new node to the chaiNNer toolset that will allow the user to create a Colorwheel tile at a user defined size.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/create_images/create_colorwheel.py
+++ b/backend/src/packages/chaiNNer_standard/image/create_images/create_colorwheel.py
@@ -1,13 +1,12 @@
 from __future__ import annotations
 
-import numpy as np
 import cv2
+import numpy as np
 
 import navi
 from nodes.impl.gradients import (
     conic_gradient,
 )
-
 from nodes.properties.inputs import (
     NumberInput,
 )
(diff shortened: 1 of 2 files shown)


backend/src/packages/chaiNNer_standard/image/create_images/create_colorwheel.py
```diff
@@ -0,0 +1,52 @@
...
+    icon="MdLens",
+    inputs=[
+        NumberInput("Size", minimum=1, unit="px", default=512),
+    ],
+    outputs=[
+        ImageOutput(
+            image_type=navi.Image(  # Creates a square output buffer from scratch and sets it to the width and height defined by the (Number)Input0 defined earlier
+                width="Input0",
+                height="Input0",
+            )
```
RunDevelopment, line 31: This node always returns RGB images, so let's annotate that.
```suggestion
            ),
            channels=3,
```

RunDevelopment (commented):
The color wheels produced by this node are all slightly off-center because `conic_gradient` is implemented incorrectly.

image

Luckily, the fix is pretty simple:

```diff
@@ -50,7 +50,7 @@ def conic_gradient(img: np.ndarray, rotation: float = 0):
     if rotation < -np.pi:
         rotation += 2 * np.pi

-    center = np.array(img.shape[:2], dtype="float32") / 2
+    center = (np.array(img.shape[:2], dtype="float32") - 1) / 2
     pixels = np.array(
         [(r, c) for r in range(img.shape[0]) for c in range(img.shape[1])]
     )
```

image

Could you please add this to your PR?

backend/src/packages/chaiNNer_standard/image/create_images/create_colorwheel.py
mrjschulte: Items add to PR. Thanks for the review!

mrjschulte:
Requested Fixes have been made and the PR has been updated.

RunDevelopment (approved):

PR #2577, opened 2024-02-11
Improved snapping for breakpoints

RunDevelopment:
This implements the CSS positions I talked about [here](https://github.com/chaiNNer-org/chaiNNer/pull/2548#discussion_r1485105525). The trick is simple: we have a 1x1 box that is the node, and then we use CSS to absolute position the actual handles.

Other changes:
- I used a `Center` to center the handles instead of 2 `Flex` components.
- I removed the unused `breakpoints` property.

### DIFF
--- a/src/common/common-types.ts
+++ b/src/common/common-types.ts
@@ -324,7 +324,6 @@ export interface EdgeData {
     sourceY?: number;
     targetX?: number;
     targetY?: number;
-    breakpoints?: readonly [number, number][];
 }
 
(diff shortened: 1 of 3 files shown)


joeyballentine:
image

Something with the way you are positioning these now causes the edge to not be perfectly around the circumference

joeyballentine:
oh never mind, this is an issue even on my branch and i just hadn't noticed

joeyballentine (approved):

PR #2578, opened 2024-02-11
Create "Premultiplied Alpha" Node

mrjschulte:
This node premultiplies the RGB channels with the Alpha channel and produces an image with an "Associated pixel state". 

A lot of the graphics/workflows out there "assume" premultiplied RGBA channels so having this node available in chaiNNer's toolset will help it do more conversions that previously weren't possible in the application.


mrjschulte:
If I was to change this PR into a single node that would allow the user to select between unpremultiplying and premultiplying with an enum, would that help clear-up any concerns?

mrjschulte:
Hmmm looks like this latest update doesn't pass the backend lint even though the actual channel pad call was copied from an already existing chaiNNer node. Thoughts @joeyballentine ?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/premultiplied_alpha.py
```diff
@@ -0,0 +1,48 @@
...
+        ImageInput(),
+        EnumInput(AlphaAssociation, label="Alpha Association"),
+    ],
+    outputs=[ImageOutput(image_type="Input0")],
+)
+def premultiplied_alpha_node(
+    img: np.ndarray, alpha_association: AlphaAssociation
+) -> np.ndarray:
+    if len(img.shape) != 3:
+        return img, np.ones(img.shape[:2], dtype="float32")
```
RunDevelopment, line 34: This is the reason for the type error. What is this even supposed to do?

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/premultiplied_alpha.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@adjustments_group.register(
+    schema_id="chainner:image:premultiplied_alpha",
+    description="Divide (or multiply) the RGB channels with the Alpha channel values of an image to go from a Straight Alpha to a Premultiplied Alpha image state.",
+    name="Premultiplied Alpha",
+    icon="CgMathDivide",
+    inputs=[
+        ImageInput(),
+        EnumInput(AlphaAssociation, label="Alpha Association"),
```
RunDevelopment, line 26: This is confusing. It's not clear whether this input is supposed to set the alpha association of the input image or determine the alpha association of the output image.

I think it would be better to call the option "Straight -> Premultiplied" and "Premultiplied -> Straight" to make the operation clear.

RunDevelopment (commented):
Sorry for the delayed response @mrjschulte!

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/premultiplied_alpha.py
mrjschulte: Originally the thought was that if the incoming image only had 3 channels we would pad to 4 channels with a solid channel of 1's. I'll just always force a 4 channel input and this can go away.

backend/src/packages/chaiNNer_standard/image_adjustment/adjustments/premultiplied_alpha.py
mrjschulte: Done, update inbound!

joeyballentine (approved):

PR #2579, opened 2024-02-11
Create "Unpremultiply" node

mrjschulte:
This node unpremultiplies the RGB channels with the Alpha channel and produces an image with a "Disassociated pixel state".

Having this node together with the premultiply node will allow chaiNNer's toolset to manage the Associated pixel state for some operations and allow it do more conversions that previously weren't possible in the application.


mrjschulte:
Closing this PR as the functionality will be merged into a new single unpremult/premult node.

PR #2580, opened 2024-02-11
Create "Mix" node

mrjschulte:
This node allow let the user do a quick and simple mix (otherwise known as a dissolve/fade) between two input images.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_utility/compositing/mix.py
+++ b/backend/src/packages/chaiNNer_standard/image_utility/compositing/mix.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+import numpy as np
+
+from nodes.properties.inputs import ImageInput, SliderInput
+from nodes.properties.outputs import ImageOutput
+
+from .. import compositing_group
+
+
+@compositing_group.register(
+    schema_id="chainner:image:mix",
+    name="Mix",
+    description="Mixes 2 images together.",
+    icon="BsLayersHalf",
+    inputs=[
+        ImageInput("Image A"),
+        ImageInput("Image B"),
+        SliderInput(
+            "Mix",
+            minimum=0.0,
+            maximum=1.0,
+            default=0.5,
+            precision=4,
+            controls_step=0.001,
+            scale="linear",
+        ),
+    ],
+    outputs=[
+        ImageOutput(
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_standard/image_utility/compositing/mix.py
```diff
@@ -0,0 +1,37 @@
...
+            default=0.5,
+            precision=4,
+            controls_step=0.001,
+            scale="linear",
+        ),
+    ],
+    outputs=[
+        ImageOutput(
+            image_type="Input0",
+        )
```
RunDevelopment, line 32: This won't work if the images are a different size or have a different number of channels. 

For the number of channels, we could probably automatically add the missing channels. So if you give the node a grayscale and an RGBA image, it will automatically convert the grayscale image to RGBA to make this operation possible. Please use the `as_target_channels` functions to do the conversion.

As for size, we can check that in the image output type:
```suggestion
        ImageOutput(
            image_type="""
                Image {
                    width: Input0.width & Input1.width,
                    height: Input0.height & Input1.height,
                    channels: max(Input0.channels, Input1.channels),
                }
            """,
        ).with_never_reason("Both images must have the same size."),
```

RunDevelopment (commented):
About the mixing itself: it's only correct for images with premultiplied alpha. This is a problem because we use straight alpha everywhere.

I think the best solution would be to have a "Separate Alpha" setting like for Resize. Basically, if "Separate Alpha" is checked, we assume that the alpha channel is independent of the color channels. This allows us to use `(1 - mix) * input1 + mix * input2` directly if checked. Otherwise, we have to multiply by alpha.

What do you think?

mrjschulte:
I can see your proposal working and being useful in the current Ux for a lot of the chaiNNer workflows out there. I need to look at it from typical user's pov where they don't really want/need to have to deal with anything at that granular of a level besides just getting their chain to run and for the output to be predictable. 

Quite the opposite to a more informed, technical and also usually much more complex chain structure where the user would need and want to be aware of premultiplied RGB and other items.

mrjschulte:
Straight alpha really needs to go :)

Sopel97:
How is this basic feature still not included? I was splitting my hair on this yesterday trying out this software. Unbelievable.

PR #2581, opened 2024-02-11
Refactored menu

RunDevelopment:
The split between favorites and category groups caused a lot of code duplication in #2575, so I refactored it. The basic idea is that the menu now displays node groups. This makes a few things easier. The abstraction of node categories also means that we have a lot more freedom to restructure the list as much as we want now.

Other changes:
- The first element is now selected default, even if no query was entered. This mirrors the behavior of VSCode.


joeyballentine (approved):

joeyballentine:
👍 awesome

PR #2582, opened 2024-02-12
Suggested connections

joeyballentine:
This adds a simple "suggested connections" feature, which will display recommended nodes to be connected to in the node search panel when connecting from another node. This is granular to the input/output level, so different inputs and outputs can have different recommended nodes to connect to per node.

I figure this feature (if used correctly) will help newcomers who don't necessarily know what they should connect to. In the future, these same suggestions could be used to highlight the handles of suggested nodes in a special way to encourage connection as well, but for now it's just in the node search menu.

image

Currently I only added a few suggestions to and from Upscale Image. We'll probably want to think about what suggestions would be useful to add where.


RunDevelopment:
I like the idea of suggestions a lot. As we get more and more nodes, we will basically need to make a search engine kind of thing in the future anyway. So these suggestions seems a like a good step towards that.

The only thing I don't like is that suggestions based on the specific inputs/outputs of specific nodes. This just doesn't scale. Imagine if we went all in with adding suggestions. We'd have to maintain multiple suggestions for each node.

I'd make suggestions based on data type + direction (input vs output). E.g. Load Image is a reasonable suggestion for pretty much all nodes that take an image, and Upscale Image is a good suggestion for pretty much all nodes that output a PyTorch model.

---

Future talk: To further improve suggestions, we need user data. Search engines give good results, because they use metrics (e.g. how many sites you visited from a search, how long you stayed, whether you changed the search query) to score their search results, and then use those score to adjust the rank of results.

With enough data, we could provide suggestions not just based on data type, or node, or node output/input, but even based on node + input data + connected nodes. That's the power of data.

That said, I'm not suggesting (hehe) that we should spy on our users. We basically just need a large collection of chains to use as data to get rankings. So we should probably encourage users to share more chains and make it easier to import shared chains. I imagine users would probably share a lot more chains/sub-chains if there was an easy way to copy'n'paste them directly from discord or elsewhere. We could then use that as a data source.

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -548,12 +582,59 @@ export const usePaneNodeSearchMenu = (): UsePaneNodeSearchMenuValue => {
...
+                            )
+                        )
+                        .map(([schema]) => schema.schemaId)
+                );
+            }
+            default:
+                assertNever(connection);
+        }
+        return new Set();
+    }, [connectingFrom, getNode, matchingEnds]);
```
RunDevelopment, line 629: This is unnecessarily complex and the `if (!node) return new Set()` doesn't do anything.

```suggestion
    const suggestions: ReadonlySet<SchemaId> = useMemo(() => {
        const connection = parseConnectStartParams(connectingFrom);
        if (!connection) return EMPTY_SET;

        return new Set(
            [...matchingEnds.entries()]
                .filter(([schema, end]) => {
                    if (!end) return false;
                    return end.type === 'target'
                        ? schema.outputs.some((o) => o.suggest && o.id === end.output)
                        : schema.inputs.some((i) => i.suggest && i.id === end.input);
                })
                .map(([schema]) => schema.schemaId)
        );
    }, [connectingFrom, matchingEnds]);
```

src/renderer/hooks/usePaneNodeSearchMenu.tsx
```diff
@@ -165,21 +174,58 @@ const groupSchemata = (
...
+        case 'category':
+            return (
+                <IconFactory
+                    accentColor={getCategoryAccentColor(categories, group.categoryId)}
+                    boxSize={3}
+                    icon={group.category?.icon}
+                />
+            );
+        default:
+            assertNever(group);
```
RunDevelopment, line 216: The exhaustive switch pattern needs a `return` here.

```suggestion
            return assertNever(group);
```

RunDevelopment (changes requested):

RunDevelopment (approved):

PR #2586, opened 2024-02-15
Fix the hold-alt-to-intersect feature when using breakpoints

joeyballentine:
This mostly fixes using the alt-intersect feature when also using breakpoints. Now, it properly checks straight lines between breakpoints and half-curves for the other edges that are attached to breakpoints but also nodes.

The only issue is that it seems Bezier.js doesn't take curvature into account the same way an SVG does, so the "sticking out" part of a curve won't get collision detected, since the curve along that point (close to the handle) is different in bezier.js. I don't think this is something we can fix unless we fork bezierjs or write our own bezier code.


src/renderer/components/ReactFlowBox.tsx
```diff
@@ -192,21 +199,91 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
...
-                        targetPosition: Position.Left,
-                    });
+                    const leftNode = getNode(sourceHandle.nodeId);
+                    const rightNode = getNode(e.target);
 
-                    // Here we use Bezier-js to determine if any of the node's sides intersect with the curve
-                    const curve = new Bezier(bezierPathCoordinates);
-                    if (!nodeBB.intersectsCurve(curve)) {
-                        return EMPTY_ARRAY;
+                    const leftNodeIsBreakpoint = leftNode?.type === 'breakPoint';
```
RunDevelopment, line 205: You added roughly 100 LOC to a function already 400 LOC long. Could you move this out into a function at the top of this file?

RunDevelopment (approved):

PR #2596, opened 2024-02-19
Feature - Add MiniMap to chaiNNer

mrjschulte:
The minimap feature enables the user to quickly navigate around a large chaiNNer schematic and also allows for zooming/panning directly within the minimap to help ease schematic navigation.

It shows the entirety of the chaiNNer schematic with all of the nodes and also a highlighted "active window" area within the minimap to show the currently visible focus area.

The minimap is semi transparent to not obscure any of the content behind it.

Screenshot 2024-02-19 at 09 53 11

### DIFF
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -520,7 +520,8 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                     maskColor="rgb(11, 13, 18, 0.175)"
                     maskStrokeColor="rgb(255, 255, 255, 1.0"
                     maskStrokeWidth={10}
-                    nodeColor="rgb(255, 255, 255, 0.66)"
+                    nodeColor="rgb(255, 255, 255, 0.5)"
+                    zoomStep={3}
                 />
                 <Controls>
                     <ControlButton
--- a/src/renderer/components/ReactFlowBox.tsx
+++ b/src/renderer/components/ReactFlowBox.tsx
@@ -9,6 +9,7 @@ import ReactFlow, {
     Controls,
     Edge,
     EdgeTypes,
+    MiniMap,
     Node,
     NodeTypes,
     OnEdgesChange,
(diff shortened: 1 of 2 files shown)


src/renderer/components/ReactFlowBox.tsx
```diff
@@ -512,6 +513,15 @@ export const ReactFlowBox = memo(({ wrapperRef, nodeTypes, edgeTypes }: ReactFlo
                     size={1}
                     variant={BackgroundVariant.Dots}
                 />
+                <MiniMap
+                    pannable
+                    zoomable
+                    ariaLabel=""
+                    maskColor="rgb(11, 13, 18, 0.175)"
+                    maskStrokeColor="rgb(255, 255, 255, 1.0"
```
RunDevelopment, line 521: This isn't a valid color.

RunDevelopment (commented):
Nice work!

It works generally well, but I found 3 things I think we should improve:

1. The background is a bit too translucent IMO. The mini map isn't very high-contrast, so having nodes be quite visible through the background makes the mini map hard to read.
2. The zoom change when using the scroll wheel on the mini map is server times larger than in the editor (the node canvas). This is surprising and inconsistent. I think we should make using the scroll wheel in the mini map and in the editor behave the same in terms of zoom change.
3. Light mode isn't supported. It's usable, but it doesn't fit.
    image

src/renderer/components/ReactFlowBox.tsx
mrjschulte: Could you elaborate more on how this rgb triplet definition + opacity isn't a valid color?

mrjschulte: Part of the Minimap Ux, in my experience in other apps,  is the fact that you can quickly navigate around the entire canvas quickly with big strides, so I'm ok with having the scrollZoom factor increment be more than when you are directly scrolling on the canvas but I also see where you're coming from and that this is currently perhaps a little too much so I'll take it back by 50%.

mrjschulte:
Here are the latest versions with more low-contrast theming for both the "light" and "dark" theme options in chaiNNer.

Light:
Screenshot 2024-02-19 at 18 01 57

Dark:
Screenshot 2024-02-19 at 18 01 38

src/renderer/components/ReactFlowBox.tsx
joeyballentine: > Could you elaborate more on how this rgb triplet definition + opacity isn't a valid color?

It's missing a closing parenthesis

mrjschulte: Fixed!

joeyballentine:
Also, can the minimap on light mode be a lighter color than the nodes?

mrjschulte:
New updates, push comments speak for themselves. New Images below:

Light: 
Screenshot 2024-02-20 at 20 34 16

Dark:
Screenshot 2024-02-20 at 20 45 22

joeyballentine (approved):
Looks great and works great, thanks

PR #2602, opened 2024-02-22
Give the minimap a border radius

joeyballentine:
Makes it fit with the rest of the UI a bit more IMO, especially since its also floating in the viewport like the control buttons

image

image

### DIFF
--- a/src/renderer/global.scss
+++ b/src/renderer/global.scss
@@ -102,6 +102,13 @@ body {
     border-radius: 0 0 8px 8px;
 }
 
+.react-flow__minimap {
+    border-radius: 0.5rem;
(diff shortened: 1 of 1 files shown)


RunDevelopment (dismissed):
Just make the linter happy, then feel free to merge :)

src/renderer/global.scss
```diff
@@ -102,6 +102,13 @@ body {
     border-radius: 0 0 8px 8px;
 }
 
+.react-flow__minimap {
+    border-radius: 0.5rem;
+    svg {
```
joeyballentine, line 107:
```suggestion
    
    svg {
```

PR #2606, opened 2024-02-24
Create "Checkerboard" node

mrjschulte:
This node creates a checkerboard with user definable colors and size for the squares. 

Sometimes it can be useful to be able to quickly composite images over a checkerboard to evaluate the transparency/alpha channel association.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/create_images/create_checkerboard.py
+++ b/backend/src/packages/chaiNNer_standard/image/create_images/create_checkerboard.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import numpy as np
+
+import navi
+from nodes.impl.color.color import Color
+from nodes.properties.inputs import ColorInput, NumberInput
+from nodes.properties.outputs import ImageOutput
+
+from .. import create_images_group
+
+
+@create_images_group.register(
+    schema_id="chainner:image:create_checkerboard",
+    name="Create Checkerboard",
+    description="Create a Checkerboard of specified dimensions filled with the given colors for the squares.",
+    icon="MdBorderAll",
+    inputs=[
+        NumberInput("Width", minimum=1, unit="px", default=1024),
+        NumberInput("Height", minimum=1, unit="px", default=1024),
+        ColorInput(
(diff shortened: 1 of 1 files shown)


mrjschulte:
Screenshot 2024-02-23 at 21 26 21

backend/src/packages/chaiNNer_standard/image/create_images/create_checkerboard.py
```diff
@@ -0,0 +1,72 @@
...
+@create_images_group.register(
+    schema_id="chainner:image:create_checkerboard",
+    name="Create Checkerboard",
+    description="Create a Checkerboard of specified dimensions filled with the given colors for the squares.",
+    icon="MdBorderAll",
+    inputs=[
+        NumberInput("Width", minimum=1, unit="px", default=1024),
+        NumberInput("Height", minimum=1, unit="px", default=1024),
+        ColorInput(
+            "Color 1", channels=[3], default=Color.bgr((0.75, 0.75, 0.75))
```
joeyballentine, line 22: is there any reason we shouldn't allow grayscale or RGBA checkerboards?

mrjschulte: In my workflows it is just somewhat uncommon or bound to cause other "huh" issues down the road. The user can always make this Grayscale or RGB+A after the fact with another node.

joeyballentine: What if I want to make a cherboard with one transparent color and one solid color though?

mrjschulte: Probably a little uncommon in the designed-as use case, but fair enough, let me tweak this PR.

mrjschulte: Ok, I'm going to make the default channels 4 and allow the user to define transparency as needed. That should cover all of the use cases you called out. If the user wants a grayscale value they can just desaturate the triplet as needed.

joeyballentine (approved):
Thanks

PR #2610, opened 2024-02-25
Create "Resolutions" Node

mrjschulte:
This node provides chaiNNer with a selection of common, and at times industry specific, resolution presets to use as "Number" sources for further use in Chains to help drive other parameters and nodes.

Screenshot 2024-02-24 at 19 34 35

Screenshot 2024-02-24 at 19 34 48


backend/src/packages/chaiNNer_standard/utility/value/formats.py
```diff
@@ -0,0 +1,119 @@
...
+                ResList.SQ1024: "Square 1024x1024",
+                ResList.SQ2048: "Square 2048x2048",
+                ResList.SQ4096: "Square 4096x4096",
+                ResList.SQ8192: "Square 8192x8192",
+            },
+        ),
+    ],
+    outputs=[
+        NumberOutput("Width", output_type="1.."),
+        NumberOutput("Height", output_type="1.."),
```
RunDevelopment, line 72: We know that they are integers.

```suggestion
        NumberOutput("Width", output_type="int(1..)"),
        NumberOutput("Height", output_type="int(1..)"),
```

backend/src/packages/chaiNNer_standard/utility/value/formats.py
```diff
@@ -0,0 +1,119 @@
...
+    elif resolution_presets == ResList.SQ1024:
+        return 1024, 1024
+    elif resolution_presets == ResList.SQ2048:
+        return 2048, 2048
+    elif resolution_presets == ResList.SQ4096:
+        return 4096, 4096
+    elif resolution_presets == ResList.SQ8192:
+        return 8192, 8192
+    else:
+        return 1920, 1080
```
RunDevelopment, line 119: I think it would be better to define the resolutions in a dict and then to look them up. This will eliminate this long `elif` chain, and allows us to easily check that we didn't forget a format (`assert len(RESOLUTION) == len(ResList)`).

backend/src/packages/chaiNNer_standard/utility/value/formats.py
```diff
@@ -0,0 +1,119 @@
...
+    SQ1024 = "SQ1024"
+    SQ2048 = "SQ2048"
+    SQ4096 = "SQ4096"
+    SQ8192 = "SQ8192"
+
+
+@value_group.register(
+    schema_id="chainner:utility:formats",
+    name="Formats",
+    description="Industry formats.",
```
RunDevelopment, line 38: Let's explain the node a bit more.

```suggestion
    description="Returns the image dimensions of industry formats. E.g. HD 1080p and UHD 4K.",
```

RunDevelopment (commented):
I really like this idea for a node! It's simple and gets rid of magic constants in our users' chains.

I kind of hate how this node gets type information due to our execution of starting nodes, but it works, so why change it :)

backend/src/packages/chaiNNer_standard/utility/value/formats.py
```diff
@@ -0,0 +1,119 @@
...
+    SQ512 = "SQ512"
+    SQ1024 = "SQ1024"
+    SQ2048 = "SQ2048"
+    SQ4096 = "SQ4096"
+    SQ8192 = "SQ8192"
+
+
+@value_group.register(
+    schema_id="chainner:utility:formats",
+    name="Formats",
```
joeyballentine, line 37: I'm not quite sure the name "Formats" properly conveys what information this returns, but I also can't think of a better name.

RunDevelopment: Maybe "Resolutions"?

backend/src/packages/chaiNNer_standard/utility/value/formats.py
mrjschulte: Have an example node where this is already setup for me to look at? The elif will only grow as we add more resolutions in the future so I’d be interested in a more elegant approach. 

I’d also love to be able to do cascading menus in chaiNNer in an easier way for an even bigger “camera resolutions node” in the future which could have hundreds of presets broken down per camera.

backend/src/packages/chaiNNer_standard/utility/value/formats.py
joeyballentine: I think "Resolutions" would be good

mrjschulte: Ok, I'll repost a PR with the Resolutions name, and an attempt at the dict for the formats.

Do we want to add Pixel aspect ratio info in here as well? Or leave that for the user to deal with accordingly?

backend/src/packages/chaiNNer_standard/utility/value/formats.py
mrjschulte: I think from a chaiNNer node design/Ux paradigm having an easy means to do cascading menus would be a great addition. For now though I'll do my best attempt to learn some more Python fu and get this dict thing to work!

backend/src/packages/chaiNNer_standard/utility/value/formats.py
joeyballentine: You don't need to make a new PR, you can just update this one

mrjschulte: Done, new node with dict-fu in there. And of course it fails!

mrjschulte: Fixed, have a look and let me know.

backend/src/packages/chaiNNer_standard/utility/value/resolutions.py
```diff
@@ -0,0 +1,103 @@
...
+            },
+        ),
+    ],
+    outputs=[
+        NumberOutput("Width", output_type="int(1..)"),
+        NumberOutput("Height", output_type="int(1..)"),
+    ],
+)
+def resolutions_node(resolution_presets: ResList) -> tuple[int, int]:
+    assert len(RESOLUTIONS) == len(ResList)
```
RunDevelopment, line 101: Please put the assert right after the definition of `RESOLUTIONS`. I want the assert to fail when node is being imported, not when it's used. This will ensure that we can verify correctness in CI.

mrjschulte: Done

backend/src/packages/chaiNNer_standard/utility/value/formats.py
mrjschulte: Done, moved 99% of the presets to a dict.

backend/src/packages/chaiNNer_standard/utility/value/resolutions.py
```diff
@@ -0,0 +1,134 @@
...
+        NumberOutput("Height", output_type="int(1..)"),
+    ],
+)
+def resolutions_node(
+    resolution_presets: ResList,
+    width: float,
+    height: float,
+) -> tuple[int, int]:
+    if resolution_presets == ResList.CUSTOM:
+        return int(width), int(height)
```
joeyballentine, line 132:
```suggestion
    width: int,
    height: int,
) -> tuple[int, int]:
    if resolution_presets == ResList.CUSTOM:
        return width, height
```
You can just type these as ints, since with precision=0 they will be ints

backend/src/packages/chaiNNer_standard/utility/value/resolutions.py
```diff
@@ -0,0 +1,134 @@
...
+                ResList.CUSTOM: "Custom Resolution",
+            },
+        ),
+        if_enum_group(0, ResList.CUSTOM)(
+            NumberInput(
+                "Width",
+                minimum=1,
+                maximum=None,
+                default=1920,
+                precision=1,
```
joeyballentine, line 106:
```suggestion
                precision=0,
```
This will make these actually ints

backend/src/packages/chaiNNer_standard/utility/value/resolutions.py
```diff
@@ -0,0 +1,134 @@
...
+                controls_step=1,
+                unit="px",
+            ),
+            NumberInput(
+                "Height",
+                minimum=1,
+                maximum=None,
+                default=1080,
+                precision=1,
```
joeyballentine, line 115:
```suggestion
                precision=0,
```
ditto

RunDevelopment:
> I kind of hate how this node gets type information due to our execution of starting nodes, but it works, so why change it :)

Remember when I said that? Yeah, we need to change this now...

Since you added custom resolutions, the Resolutions node now has 2 number inputs which prevent it from being a starting node. So the node no longer shows the exact resolutions on its outputs. There are 2 ways around this:

1. Set `has_handle=False` for both number inputs. The condition for being a starting node is "no *possible* input handles," so preventing the number inputs from having any connections makes the node execute like before.
2. Add static type information to the outputs. Basically, we need one long `match` expression to encode all resolutions in the type system.

Which option you choose should depend on whether you want the "Width" and "Height" input to be connectable. If this isn't necessary, then go with option 1. If you think they should be connectable, then go with option 2 and I'll help you to get the output types right.

mrjschulte:
Option 1 it is!

joeyballentine (approved):

PR #2613, opened 2024-02-25
Support custom scales for PyTorch upscale

RunDevelopment:
This allows users to set a custom scale for the PyTorch models to upscale to any scale. This will hopefully make chainner easier to use for our users.

image

### DIFF
--- a/backend/src/nodes/impl/upscale/custom_scale.py
+++ b/backend/src/nodes/impl/upscale/custom_scale.py
@@ -1,0 +1,41 @@
+import math
+
+import numpy as np
+
+from nodes.impl.image_op import ImageOp
+from nodes.impl.resize import ResizeFilter, resize
+from nodes.utils.utils import get_h_w_c
+
+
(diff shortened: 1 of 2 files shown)


joeyballentine (approved):
As much as it saddens me that this is necessary (and defeats the point of having nodes), this will at least stop people from needing to ask why chaiNNer doesn't support custom scales like other UIs. So this is ultimately a good change.

RunDevelopment:
> As much as it saddens me that this is necessary (and defeats the point of having nodes)

I mean, it's a common-enough use case, so I honestly think we should have added this sooner. Having some duplicate functionality for convenience isn't a bad thing IMO. E.g. we also had 2 resize nodes until recently (and even Resize to Side) even though all of this functionality can be achieved with Resize (Resolution) and a few Math nodes (and a bit of pain).

joeyballentine:
True. I guess there's nothing wrong with adding a bit of convenience here and there :)

PR #2614, opened 2024-02-25
Allow mixed colors for Create Checkerboard

joeyballentine:
A followup to @mrjschulte's PR (#2606) that allows 1/3/4 channel colors to be used rather than just 4 channel colors.

### DIFF
--- a/backend/src/nodes/impl/color/color.py
+++ b/backend/src/nodes/impl/color/color.py
@@ -105,3 +105,26 @@ class Color:
             raise AssertionError(f"Colors with {len(values)} are not supported.")
 
         return {"kind": kind, "values": values}
+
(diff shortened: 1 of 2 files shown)


RunDevelopment (commented):
Not a fan of this color conversion code. Why not create a 1x1 image, use `as_target_channels`, and then go back to a color with `from_1x1_image`? This would allow us to use the usual color conversion functions instead of implementing new ones.

joeyballentine:
To me that seems like it's just adding unnecessary steps, but sure

RunDevelopment:
> it's just adding unnecessary steps

It kinda is. I mean, it's definitely more work for the CPU, but my goal was to reuse existing code.

PR #2625, opened 2024-02-28
Optimize node selector panel and use group hover

RunDevelopment:
Changes:
- Use `useMemo` for the JSX tree inside `RepresentativeNodeWrapper` and `RepresentativeNode`. While these components are already `memo`ed, `memo` is ineffective when one of the used contexts updates. Even if the updated context values do not lead to any changes in the DOM, React still has to re-create and diff the JSX tree of the component. `useMemo` not only caches the tree, it also allows React to immediately see that the cache tree didn't change (since they are the same object).
- I removed 3 instances of using `useState` to track whether a component is hovered, and replaced them with `_groupHover`. This makes the hover effects instant (because we don't need to re-render anything) and means that hover effects can't get stuck.

### DIFF
--- a/src/renderer/components/CustomEdge/CustomEdge.scss
+++ b/src/renderer/components/CustomEdge/CustomEdge.scss
@@ -21,7 +21,7 @@
     transition-timing-function: ease-in-out;
     stroke: var(--bg-700);
 
-    &.hovered {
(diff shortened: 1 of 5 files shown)


joeyballentine (approved):
Nice 👍

PR #2627, opened 2024-02-28
Optimize UI by only rendering what's visible

RunDevelopment:
This uses [the `react-intersection-observer` package](https://www.npmjs.com/package/react-intersection-observer) to implement an `IfVisibile` component. The idea is that replace elements that aren't visible with a placeholder to improve performance.

This has noticeably improved lag when:
1. opening chainner,
2. clearing a search query with few or no matches in the node selector panel,
3. collapsing/expanding the node selector panel,
4. opening the context selector pane,
5. clearing a search query with few or no matches in the context selector pane, and
6. opening the node documentation modal.

The `IfVisibile` is a bit hard to use in that it requires a `height` property. This needs to be the exact height of the content or else things will look funky. This is necessary to guarantee that the placeholder behaves exactly like the content. However, this restriction also allows it to further optimize layouting using the CSS `contain` property.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "react-hotkeys-hook": "^3.4.6",
         "react-i18next": "^12.0.0",
         "react-icons": "^4.10.1",
+        "react-intersection-observer": "^9.8.1",
         "react-markdown": "^8.0.3",
         "react-query": "^3.39.3",
         "react-time-ago": "^7.2.1",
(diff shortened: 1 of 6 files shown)


joeyballentine:
Played around with it and noticed that it does seem to cause some lag in the node selector sidebar if you scroll way too fast, but the other optimizations are worth that tradeoff

joeyballentine (approved):

PR #2628, opened 2024-02-29
Fix checkerboard node

RunDevelopment:
Looks like 2 people never actually tested the node.


joeyballentine:
In my defense, I tested it, just only ever with RGBA...

joeyballentine (approved):

PR #2635, opened 2024-02-29
Stretch type tags to fill available space

RunDevelopment:
I finally learned how to do this. It's finicky, but it works for strings and paths. So type tags will finally use the space available to them.

image
image


joeyballentine (approved):
Awesome

PR #2642, opened 2024-03-01
Update Electron Forge and remove EF patch

joeyballentine:
They've fixed the line we previously needed to patch, rendering the patch no longer necessary. I'm hoping they also maybe fixed something else related to macos releases...


RunDevelopment (approved):
I'm assuming this stuff works, because I have no idea what it does...

PR #2647, opened 2024-03-04
server_host.py refactor

joeyballentine:
This refactors server_host to use a new helper class that manages everything related to the executor process (on top of the already existing class that handles the subprocess itself)

This ultimately fixes the SSE reconnection issue by having this class manage both the process and the connection, allowing both to be restarted easily. (I tested this merged with the backend-killing branch locally, and confirmed this did fix the issue)

This should be merged before #2646

### DIFF
--- a/backend/src/server_process_helper.py
+++ b/backend/src/server_process_helper.py
@@ -1,0 +1,148 @@
+from __future__ import annotations
+
+import asyncio
+import os
+import subprocess
+import sys
+import threading
+
+import aiohttp
+from sanic import HTTPResponse, Request
+from sanic.log import logger
+
+from api import Package
+
+
+class ExecutorServerProcess:
+    def __init__(self, port: int, flags: list[str] | None = None):
+        self.process = None
+        self.stop_event = threading.Event()
+        self.finished_starting = False
+
+        self.port = port
+        self.flags = flags or []
+
+    def start_process(self):
+        server_file = os.path.join(os.path.dirname(__file__), "server.py")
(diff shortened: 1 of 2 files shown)


backend/src/server_process_helper.py
```diff
@@ -0,0 +1,148 @@
...
+    def _read_output(self):
+        if self.process is None or self.process.stdout is None:
+            return
+        for line in self.process.stdout:
+            if self.stop_event.is_set():
+                break
+            if not self.finished_starting:
+                if "Starting worker" in line.decode():
+                    self.finished_starting = True
+            print(line.decode().strip())
```
RunDevelopment, line 55: Why `print` and not the logger?

RunDevelopment (commented):
I think the general design of `ExecutorServerProcess` and `ExecutorServer` is good. There are a few implementation details I don't like (two-step init, too much API surface), but I would like to tackle those things myself after this is merged.

For now, I just think that we should rename some stuff. Calling classes and variables "server process" makes no sense to me, because everything here is the server. I think we should consistently call the host process/server "host" (like we already kinda do) and the worker process/server "worker" (suffixes are allowed, of course 😄).

backend/src/server_process_helper.py
joeyballentine: Because then we'd get double log prefixes (I think. I never actually tried it). So like `(time) [info]: (time) [info]: log`

RunDevelopment (approved):

PR #2648, opened 2024-03-04
Remove unnecessary `__init__` files and fix pylance auto-imports

RunDevelopment:
After the pain we had with spandrel, I finally figured out how to (1) remove the `__init__.py` files we had to make tests work and (2) how to make pylance auto-import with correct paths (no more `backend.src.nodes...`).


joeyballentine (approved):
> no more backend.src.nodes...

🥳🎉🍾

PR #2650, opened 2024-03-05
Avoid restarting backend if no deps installed

joeyballentine:
This PR moves the usage of `get_deps_to_install` to outside of the installing deps functions. This makes it so that we can choose to not restart the executor server if there are no dependencies installed, which should speed up startup times.

### DIFF
--- a/backend/src/server_host.py
+++ b/backend/src/server_host.py
@@ -218,23 +218,22 @@ async def import_packages(
             if dep.auto_update and is_installed:
                 to_install.append(dep)
 
-    if len(to_install) > 0:
-        try:
-            num_installed = await install_deps(to_install)
+    try:
+        num_installed = await install_deps(to_install)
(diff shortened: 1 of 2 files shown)


RunDevelopment:
Not a fan. You basically just added an invisible pre-condition to for `install_dependencies{,_sync}`. Just make those functions return how many dependencies changed (or something similar).

joeyballentine:
Replaced the other implementation with what you suggested

backend/src/server_host.py
```diff
@@ -219,16 +220,17 @@ async def install_deps(dependencies: list[api.Dependency]):
 
     if len(to_install) > 0:
```
RunDevelopment, line 221: We don't need that anymore.

RunDevelopment (approved):

PR #2656, opened 2024-03-08
Add key information to collapsed nodes

RunDevelopment:
image

This adds the key information of nodes next to the title of collapsed nodes. This should make collapsed nodes more useful as the functionality of the node can be seen more easily without needing to expand it.

As for the implementation: I used the term "key info" everywhere for now, but I don't really like it. It doesn't convey that this information is only used when the node is collapsed (although maybe we'll use it somewhere also too in future? idk).

### DIFF
--- a/backend/src/api/api.py
+++ b/backend/src/api/api.py
@@ -24,7 +24,7 @@ from .node_check import (
     check_naming_conventions,
     check_schema_types,
 )
-from .node_data import IteratorInputInfo, IteratorOutputInfo, NodeData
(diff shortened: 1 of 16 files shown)


mrjschulte:
This is great! Is there still a spot to allow for a "locked" node icon representation in that top-line on the node?

joeyballentine:
image
I fixed the text alignment

joeyballentine:
> This is great! Is there still a spot to allow for a "locked" node icon representation in that top-line on the node?

I'm sure we could squeeze it in somewhere

joeyballentine (approved):

PR #2659, opened 2024-03-09
Fix executor passthrough logging

joeyballentine:
The biggest thing I dislike about this is the amount of code that now has to run per-log. This could wind up being a minor performance impact if we have a lot of logging

### DIFF
--- a/backend/src/server_process_helper.py
+++ b/backend/src/server_process_helper.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import os
+import re
 import socket
 import subprocess
 import sys
@@ -25,6 +26,9 @@ def _find_free_port():
 def _port_in_use(port: int):
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
         return s.connect_ex(("127.0.0.1", port)) == 0
+
+
+SANIC_LOG_REGEX = r"\[.*\] \[\d*\] \[(\w*)\] (.*)"
 
 
 class _WorkerProcess:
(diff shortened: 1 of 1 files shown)


backend/src/server_process_helper.py
```diff
@@ -59,7 +59,20 @@ def _read_output(self):
         for line in self._process.stdout:
             if self._stop_event.is_set():
                 break
-            print(line.decode().strip())
+            stripped_line = line.decode().strip()
+            words = stripped_line.split()
+            log_level = words[4].lower()[1:-1]  # remove brackets
+            rest = " ".join(words[5:])
```
RunDevelopment, line 65: No. Please use a regex to do the parsing.

backend/src/server_process_helper.py
```diff
@@ -59,7 +59,20 @@ def _read_output(self):
...
+            if log_level == "debug":
+                logger.debug(rest)
+            elif log_level == "info":
+                logger.info(rest)
+            elif log_level == "warning":
+                logger.warning(rest)
+            elif log_level == "error":
+                logger.error(rest)
+            elif log_level == "critical":
+                logger.critical(rest)
```
RunDevelopment, line 75:
```
else:
    silentely_drop_line()
```

Please fall back to something to log the line.

backend/src/server_process_helper.py
```diff
@@ -59,7 +63,25 @@ def _read_output(self):
         for line in self._process.stdout:
             if self._stop_event.is_set():
                 break
-            print(line.decode().strip())
+            stripped_line = line.decode().strip()
+            match_obj = re.match(SANIC_LOG_REGEX, stripped_line)
```
RunDevelopment, line 67: You are saying you are concerned about performance, yet you are compiling a regex in every iteration of the loop. Curious.

Jokes aside, please don't do that. Compile the re outside the loop and use it inside.

backend/src/server_process_helper.py
```diff
@@ -27,6 +28,9 @@ def _port_in_use(port: int):
         return s.connect_ex(("127.0.0.1", port)) == 0
 
 
+SANIC_LOG_REGEX = r"\[.*\] \[\d*\] \[(\w*)\] (.*)"
```
RunDevelopment, line 31: You should anchor the pattern so it doesn't in the middle of the string.

> `\[.*\]`

`.` also matches `[` and `]`, so this may end up matching waaaay too much.

```suggestion
SANIC_LOG_REGEX = r"^\[[^\[\]]\] \[\d*\] \[(\w*)\] (.*)"
```

joeyballentine: Nice edit lol I was about to mention that the original didn't work right

RunDevelopment:
One last thing: I think it would be a good idea to give the logged message a prefix, so we know it's from the worker. Maybe something like `[sanic] [stuff] [INFO] Worker: <message>`. What do you think

joeyballentine:
I already added [Executor] to it

backend/src/server_process_helper.py
```diff
@@ -59,7 +63,25 @@ def _read_output(self):
         for line in self._process.stdout:
             if self._stop_event.is_set():
                 break
-            print(line.decode().strip())
+            stripped_line = line.decode().strip()
+            match_obj = re.match(SANIC_LOG_REGEX, stripped_line)
+            if match_obj is not None:
+                log_level, message = match_obj.groups()
+                message = f"[Executor] {message.strip()}"
```
joeyballentine, line 70: here

joeyballentine:
Gonna merge this since I need it for fixing #2646 and I don't think there's anything left to do. If you wanna change the message, just make a followup PR or reply to this and I'll do it.

PR #2664, opened 2024-03-10
Fixed vivid light blend mode implementation

BigBoyBarney:
I do not have access to photoshop, so I tested blend parity with https://www.photopea.com/ on a few images.
A double check with photoshop before merge would be great, just to make sure, particularly with regards to image input order (base - overlay / overlay - base)


mrjschulte (approved):
Confirmed as matching against PS 2024

joeyballentine (approved):
Thanks!

PR #2678, opened 2024-03-14
Fix glob negation

joeyballentine:
Discord link for context: https://discord.com/channels/930865462852591648/930875396277280788/1217655737178656798

Basically, the negate flag wasn't enough. I guess negating doesn't assume you want to match everything that doesnt match the negation by default, so to enable that i had to enable negateall.

Docs for negateall: https://facelessuser.github.io/wcmatch/glob/#negateall. No it wont break regular globs.

Specifically, this enables global pattern negation (which appears to be enabled on glob tester websites). For example, with this you can write `!**/*_ab.*` instead of `**/!(*_ab).*`


joeyballentine:
Merging this for now so it does not become stale

PR #2681, opened 2024-03-16
Make view image node image inputs optional

joeyballentine:
This is a draft because I'm not sure how to fix this error that results from this change:

image

### DIFF
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -158,7 +158,10 @@ class LargeImageOutput(ImageOutput):
             assume_normalized=assume_normalized,
         )
 
-    def get_broadcast_data(self, value: np.ndarray):
+    def get_broadcast_data(self, value: np.ndarray | None):
+        if value is None:
+            return None
+
         img = value
         h, w, c = get_h_w_c(img)
         image_size = max(h, w)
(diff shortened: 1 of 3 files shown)


RunDevelopment:
What's the motivation here?

joeyballentine:
https://discord.com/channels/930865462852591648/930865780663398451/1218596742681858120

RunDevelopment:
I see. Not a fan then.

You essentially made View Image a node with side effects that is allowed to do nothing. This doesn't break anything, but it messes with the way we handle side effects. The problem is that nodes with side effects are assumed to, well, have side effects. But you made View Image have side effects *conditionally*. This has consequences. 
E.g. the warning we show in the UI when a chain has no side effect nodes is effectual now. Having a single View Image node anyway in the chain will now suppress this because there is technically a node that is executed.

I see 3 possible solutions to address this problem:
1. Add a property that tells the frontend that a node with side effects can always safely be removed if it doesn't have inputs. The frontend already removes disabled and unused nodes, so we just extend this functionality to remove unused View Image nodes as well. (This would also be useful for Copy To Clipboard.)
2. Same as 1, but no property, just do it for all side-effect nodes. Basically, if a side-effect node has no inputs, remove it.
3. Blame the user. They could have just disabled or removed the node.

While option 3 is tempting, option 1 and 2 have their uses. What do you think we should go with?

joeyballentine:
I think I'm personally a fan of option 2. To me, it's an ok assumption to make that this can be applied automatically.

PR #2683, opened 2024-03-17
Server-side package installation

joeyballentine:
This PR implements installing packages through the backend, by having the host server kill the worker server, install the dependencies, then start the worker back up. Since it is now using our pip fork, it can also report download progress directly through pip.

This means the frontend no longer handles any pip dependency installation. The only remnant of this is the getFindLinks function used for manual/copy mode, which i will have the backend handle in a future PR.

The way it reports progress right now is a bit jank. I want to redo this in the future but i also wanted to make minimal changes at the moment so I will handle that once I get around to redoing the entire dependency manager page.

This works well enough for now.


backend/src/server_host.py
```diff
@@ -152,6 +157,102 @@ async def get_features(request: Request):
...
+    if package is None:
+        return json(
+            {"status": "error", "message": f"Package {package_to_uninstall} not found"},
+            status=404,
+        )
+
+    def update_progress(
+        message: str, progress: float, status_progress: float | None = None
+    ):
+        return AppContext.get(request.app).setup_queue.put_and_wait(
```
RunDevelopment, line 190: Any reason why you wait here? What does it matter to the backend whether the frontend read the status or not?

backend/src/server_host.py
```diff
@@ -152,6 +157,102 @@ async def get_features(request: Request):
...
+            },
+            timeout=1,
+        )
+
+    try:
+        await worker.stop()
+        await uninstall_dependencies(
+            deps_to_dep_info(package.dependencies), update_progress, logger
+        )
+        await worker.start()
```
RunDevelopment, line 207: This is dangerous. What if the uninstall fails for some reason? We'll have a backend without an active worker.

On that note: how do things like `/nodes` even behave when the worker isn't active? Will they wait forever until something starts the worker, or will they attempt to start the worker if no one is doing so/failed to do so already?

src/common/Backend.ts
```diff
@@ -237,6 +237,24 @@ export class Backend {
     features(): Promise<FeatureState[]> {
         return this.fetchJson('/features', 'GET');
     }
+
+    installPackage(pkg: Package): Promise<void> {
+        return this.fetchJson('/packages/install', 'POST', {
+            package: pkg.name,
```
RunDevelopment, line 243: Packages have IDs for reason... Please use them.

RunDevelopment (commented):
I only briefly looked at this, but did you remove the ability to install/update individual pypi packages? The API only deals with chainner packages.

joeyballentine:
> but did you remove the ability to install/update individual pypi packages?

We only ever installed things on a chaiNNer-package-level basis, so to simplify the process I made it go by chaiNNer package. 

Think of the install/uninstall process as installing or installing an extension rather than installing and uninstalling just the dependencies. Maybe in the future an extension would be more than just pypi packages. 

Do you have a usecase for being able to manage individual deps like that?

RunDevelopment:
> Do you have a usecase for being able to manage individual deps like that?

PyTorch. Spandrel updates are necessary, pytorch less so. So when we do a pytorch update, I typically stay with the older version purely because I don't want to wait for the download.

joeyballentine:
But that's never been possible within chaiNNer afaik, how were doing that?

RunDevelopment:
I happen to contribute to this project, so I get dep changes a little more frequently.

But you're right. I'm probably the only one with this problem, so let's not complicate things. I can still get the pip command and edit it if need be.

backend/src/server_host.py
joeyballentine: Nope. I just copy and pasted another line. I probably made this same mistake in other PRs as well actually

backend/src/server_host.py
joeyballentine: I think they'll just wait forever. I'm not sure having multiple things be able to start up the backend separately is a good idea though. i'll fix this however

backend/src/server_host.py
joeyballentine: Ok so for some reason it only works like this and not at all with put, and i can't explain why

PR #2684, opened 2024-03-18
Fix FLT_MAX ONNX -> NCNN error

joeyballentine:
Finally looked into this bug. Basically, there's code that checks if a default value is a string and does some stuff if it is. Since FLT_MAX isn't a number and is instead a string, it was triggering the code to handle strings, when it really should be handled like a number. So, I just added a case to the if statement to ignore FLT_MAX and -FLT_MAX

### DIFF
--- a/backend/src/nodes/impl/ncnn/model.py
+++ b/backend/src/nodes/impl/ncnn/model.py
@@ -254,7 +254,7 @@ class NcnnParamCollection:
         for v in self.param_dict.values():
             if v.value == v.default:
                 continue
-            if isinstance(v.default, str):
+            if isinstance(v.default, str) and "FLT_MAX" not in v.default:
                 pid = None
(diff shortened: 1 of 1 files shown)


joeyballentine:
Converting this to a draft until i can figure out whats happening with the resulting NCNN models crashing the backend

joeyballentine:
The model converts fine. The issue seems unrelated. Perhaps SPAN just can't be run in ncnn?

Kim2091:
@joeyballentine If you take a model converted to ONNX in new chaiNNer, and try to convert it to NCNN in an older version of chaiNNer this error happens no matter what arch it is.

joeyballentine:
@Kim2091 that's unrelated. The float max error is due to the clipping introduced via spandrel with the call API. The crash I believe is arch dependent. This PR merely addresses the float max problem

PR #2685, opened 2024-03-19
Improved error reporting

RunDevelopment:
This improves the error reporting in 2 ways:

1. Logged errors will no longer contain the stack trace of the `NodeExecutionError`. `NodeExecutionError` is a wrapper to add additional info to errors, its trace doesn't matter.
2. Send the full stack trace of an error to the frontend. Since the trace isn't usually interesting for users, it's tugged away inside a `<details>`.

Both changes together should make errors easier to diagnose and fix in the future. Stack traces are now shorter and don't spam logs anymore, and users can send us a full stack trace by simply clicking Copy To Clipboard on an error message.

image

---

Aside from those changes, I also refactored `AlertBoxProvider` to extract out the `AlertDialog` into its own component.

### DIFF
--- a/backend/src/events.py
+++ b/backend/src/events.py
@@ -44,6 +44,7 @@ class ExecutionErrorSource(TypedDict):
 class ExecutionErrorData(TypedDict):
     message: str
     exception: str
+    exceptionTrace: str
     source: ExecutionErrorSource | None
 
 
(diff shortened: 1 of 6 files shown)


joeyballentine (approved):

joeyballentine:
Awesome

PR #2697, opened 2024-03-21
Allow math expressions in number inputs

RunDevelopment:
I finally added a feature I wanted for a long time: using math expressions in number inputs. Sometimes I want to input a big number, or a number with many digits that can be easily computed, but I always to enter it digit by digit. Well, not anymore! All number inputs now support math expressions. E.g. `3+5`, `32 / 5`, `2^10`, `e ^ (pi i)`

Under the hood, math expressions are evaluated by [mathjs](https://www.npmjs.com/package/mathjs), so this isn't a simple `eval(string)`. It's safe.

image
image

image
image

image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "html-to-image": "^1.10.8",
         "i18next": "^22.0.5",
         "javascript-time-ago": "^2.5.9",
+        "mathjs": "^12.4.1",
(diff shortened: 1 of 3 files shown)


mrjschulte:
This is great!

joeyballentine (approved):
This is really cool. Awesome idea

PR #2702, opened 2024-03-21
Add Directory Go Into node

### LINKED ISSUE #606: Directory helper nodes
It would be nice if we had a good way of splitting and combining directories. Like being able to do os.path.join essentially, or splitting a directory into its different parts. 

Maybe this just means we need a way to convert a string to a directory, so that Text Append or Text Pattern would be able to be used for a directory

RunDevelopment:
Closes #2519
Closes #606

This adds the last node from #2519. As before I made a few changes:

- I named it "Directory Go Into" to be consistent with the "Directory Go Up" node.
- I added a few optional folder inputs, so users can join multiple folders without needing to know that `/` and `\` are usually used to denote this.
- More path validation: no invalid characters and absolute paths.

The output type of this node is kind of insane because Navi doesn't have early returns.

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/utility/directory/directory_go_into.py
+++ b/backend/src/packages/chaiNNer_standard/utility/directory/directory_go_into.py
@@ -1,0 +1,82 @@
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from nodes.groups import optional_list_group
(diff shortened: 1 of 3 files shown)


joeyballentine (approved):
LGTM. Damn the nesting on the Navi type is so gross

PR #2716, opened 2024-03-29
Move all fs calls to main process, invoke via IPC

joeyballentine:
This moves all the usage of FS out of the renderer.

One unfortunate thing I had to do was move a bit of code out of input-overrides, as that file imports from FS, so I moved the tiny bit that was used in the renderer out into its own file.

### DIFF
--- a/src/main/gui/main-window.ts
+++ b/src/main/gui/main-window.ts
@@ -1,5 +1,6 @@
 import { BrowserWindow, app, dialog, nativeTheme, powerSaveBlocker, shell } from 'electron';
 import EventSource from 'eventsource';
+import fs, { constants } from 'fs/promises';
 import { t } from 'i18next';
 import { BackendEventMap } from '../../common/Backend';
 import { Version } from '../../common/common-types';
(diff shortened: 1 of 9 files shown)


joeyballentine:
I finally figured out why Vite wasn't working for me when I tried switching to it: it doesn't support nodejs stuff being called from the browser. And technically it's bad practice and a security concern to have that enabled in webpack, I just never cared.

So, since it's technically correct and I want to try out Vite, we need to do this. (It doesn't mean we actually will switch to Vite for sure, I just haven't gotten the opportunity to really try it out since it won't compile)

src/renderer/components/settings/components.tsx
```diff
@@ -151,9 +155,11 @@ const CacheSetting = memo(({ setting, value, setValue }: SettingsProps<'cache'>)
                             locationPromise
                                 .then(async (cacheLocation) => {
                                     try {
-                                        await access(cacheLocation);
+                                        await ipcRenderer.invoke('fs-access', cacheLocation);
```
RunDevelopment, line 158: This relies on `access` throwing an error. I don't think errors are ICP compatible, so how does this work now?

src/main/gui/main-window.ts
```diff
@@ -195,6 +196,24 @@ const registerEventHandlerPreSetup = (
             })().catch(log.error);
         });
     }
+
+    // Handle filesystem
+    ipcMain.handle('fs-read-file', async (event, path) => fs.readFile(path, { encoding: 'utf8' }));
+    ipcMain.handle('fs-write-file', async (event, path, content) =>
+        fs.writeFile(path, content, { encoding: 'utf8' })
```
RunDevelopment, line 203: This doesn't make sense to me. If `content` can be a buffer, then how does that interact with `encoding: 'utf8'`? Will it ignore the encoding unless you pass in a string?

joeyballentine: Good point. I'll make the options passed through as well

src/renderer/components/settings/components.tsx
joeyballentine: Based on s a quick Google search it seems it handles errors fine now, but I'll try this out later to confirm

joeyballentine: can confirm it does properly pass through errors.

RunDevelopment (approved):

PR #2719, opened 2024-03-29
Send SSE event as one

RunDevelopment:
Changes:
- Only check `response` for None once.
- Send SSE event as one string instead of 2. This should make the IO a little more efficient.


joeyballentine:
Shouldn't this be done in server_host as well?

joeyballentine (approved):

PR #2720, opened 2024-03-30
Separate electron imports between main and renderer, use IPC to replace most renderer-side electron imports

joeyballentine:
electron provides (and recommends using) these segregated imports (electron/renderer, electron/main, electron/common) in order to ensure you aren't accidentally importing something in the wrong place.

I discovered a few places where we were importing main-process electron code in the renderer and have fixed it, along with modifying (almost) all the electron imports to be the helper ones. (I left one import the same because if I didn't then it thinks electron doesn't exist. Try it yourself, its really strange)

I'm opening this as a draft, because there's still somewhere that is importing main process code where it shouldn't, and it is causing chaiNNer to crash on startup.


joeyballentine:
So, I learned a couple things:
1) the electron webpack template is bad and you apparently need to externalize the modules for them to be imported properly. This took the most time to debug
2) the namespace imports _do_ prevent you from using them in the renderer, since you're supposed to only expose stuff via the preload script. This is unlike the main electron import that doesn't care. This means that i had to leave a few as regular electron imports in the meantime, until we swap over to exposing IPC via the preload script
3) Because of the above, i merged a separate branch into this one where i had replaced a lot of renderer-side imports with IPC calls. Theoretically these could be replaced by instead exposing the clipboard api through the preload script, but for now it was just easier to send that stuff through IPC and have IPC be the only thing still being imported the incorrect way. Then, we can move IPC over to the correct way, then replace some of the IPC with non-IPC in the future.

src/common/safeIpc.ts
```diff
@@ -1,13 +1,4 @@
...
-    BrowserWindow,
-    IpcMainEvent,
-    IpcMainInvokeEvent,
-    IpcRendererEvent,
-    MessagePortMain,
-    WebContents,
-    ipcMain as unsafeIpcMain,
-    ipcRenderer as unsafeIpcRenderer,
-} from 'electron';
+import { type FileFilter, type OpenDialogReturnValue } from 'electron'; // TODO: replace with electron/common
```
RunDevelopment, line 1: TODO

src/renderer/components/inputs/DirectoryInput.tsx
```diff
@@ -90,7 +93,9 @@ export const DirectoryInput = memo(
                     isDisabled={!displayDirectory}
                     onClick={() => {
                         if (displayDirectory) {
-                            clipboard.writeText(displayDirectory);
+                            ipcRenderer
+                                .invoke('clipboard-writeText', displayDirectory)
+                                .catch(log.error);
```
RunDevelopment, line 98: Why were we using electron clipboard in the first place? There's a [browser API for this](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText). Wouldn't that work too?

src/renderer/contexts/GlobalNodeState.tsx
```diff
@@ -28,7 +28,7 @@ import { log } from '../../common/log';
 import { getEffectivelyDisabledNodes } from '../../common/nodes/disabled';
 import { ChainLineage } from '../../common/nodes/lineage';
 import { TypeState } from '../../common/nodes/TypeState';
-import { ipcRenderer } from '../../common/safeIpc';
+
```
RunDevelopment, line 31: weird empty line
```suggestion
```

src/renderer/hooks/useIpcRendererListener.ts
```diff
@@ -1,6 +1,7 @@
-import { IpcRendererEvent } from 'electron';
+import { IpcRendererEvent } from 'electron'; // TODO: replace with electron/renderer
```
RunDevelopment, line 1: TODO

src/renderer/safeIpc.ts
```diff
@@ -0,0 +1,33 @@
+import { IpcRendererEvent, ipcRenderer as unsafeIpcRenderer } from 'electron'; // TODO: replace with electron/renderer
```
RunDevelopment, line 1: TODO

src/common/safeIpc.ts
joeyballentine: That's not a now todo, that's a "we can't do that til we do more stuff in the future" todo. As in, if I were to do that now this wouldn't compile.

src/renderer/components/inputs/DirectoryInput.tsx
joeyballentine: Probably yeah, I can check that

src/renderer/contexts/GlobalNodeState.tsx
joeyballentine: Sorry I thought I removed all these. Was from bad merge resolutions

src/renderer/hooks/useIpcRendererListener.ts
joeyballentine: Ditto above

src/renderer/safeIpc.ts
joeyballentine: Ditto above

src/common/safeIpc.ts
joeyballentine: I also explained this in point 2 of my comment. For reference:

> This means that i had to leave a few as regular electron imports in the meantime, until we swap over to exposing IPC via the preload script

src/renderer/components/inputs/DirectoryInput.tsx
joeyballentine: Ok, I replaced all the read/write text ones. Can't replace the others as the native clipboard doesn't support as many formats as electron's does.

joeyballentine:
Ok, back to a draft. i broke chain copy and paste from this

joeyballentine:
I think i've tested all the different ways to copy and paste, and everything looks good now.

RunDevelopment (approved):

PR #2724, opened 2024-03-31
Reset file inputs when file deleted

joeyballentine:
I'm not 100% sure why individual runs were getting stuck when files were deleted, but this PR fixes the issue regardless and I would consider this behavior preferred, personally.


RunDevelopment:
> I would consider this behavior preferred

What behavior?

RunDevelopment:
Okay, having reviewed the code, I don't like it. This PR is a workaround, not a fix. 

I also don't like the new behavior. It changes how file inputs and files are connected. Previously, file paths were just pointers, specifying the location of a file that most likely exists. If you delete the file, you'll get an error in the UI. Now file inputs are linked. A file input must now point to an *existing* file. This makes uses like, renaming a file to replace it with another impossible.

I think we should just fix the bug. 
You seem to have already spent some time and sanity on this, do you want me to take over?

joeyballentine:
Yes please

PR #2731, opened 2024-04-04
Remove splash screen and fix loading issue

joeyballentine:
The splash screen ended up being pretty easy to replace.

The splash screen getting stuck though was simply because the backend ready event was not being sent through webcontents for some reason, so the listener for it on the frontend wasn't triggering.


RunDevelopment:
Is seems like the renderer is trying to connect to the backend before realizing that it has to setup the backend on first start up.

image



The menu is active during loading. I think it would probably be best to completely hide the menu while the renderer is loading.

image

RunDevelopment:
Also, closing chainner doesn't close the backend. Steps to reproduce:

1. `npm run make-win-zip` to cut a fresh build.
2. Unzip and start chainner. Let it install everything.
3. Close chainner and try to delete the folder you unzipped chainner into.
4. See that you can't delete the folder, open task manager, and kill the Python process that takes up around 220MB (that's our backend).
5. Delete the folder for real this time.

RunDevelopment:
Auto updates also don't work correctly. It will show download progress, but then it fetches nodes and shows the chain editor as normal, while the dep installation silently fails in the background because the worker process is up.

RunDevelopment:
> That's not an issue with this PR. that's just a general issue right now. Feel free to try to fix it because I was going insane last night trying to myself

Damn. I really hoped this was related to this PR... Will investigate.

joeyballentine:
> while the dep installation silently fails in the background because the worker process is up.

I didn't change anything related to that part of the code, so if that is an issue (i haven't yet been able to reproduce) i don't think its related to this.

for me, the auto updating seems to work fine. do you have reproduction steps for this?

RunDevelopment:
> I didn't change anything related to that part of the code, so if that is an issue (i haven't yet been able to reproduce) i don't think its related to this.

This did work correctly before this PR, I tested that.

> do you have reproduction steps for this?

Manually install the old ONNX deps from before #2717. For some reason, auto updates don't close the worker, so the worker will prevent the ONNX auto update from succeeding.

joeyballentine:
Honestly I don't believe you that it's this PR that caused that, but I'll look into it anyway

joeyballentine:
See #2748

joeyballentine:
Since the issue is unrelated, is this good to merge now?

RunDevelopment:
> Is seems like the renderer is trying to connect to the backend before realizing that it has to setup the backend on first start up.

This is still an issue. We just can't see it with `npm start` or `npm run dev` because the renderer takes too long to show up. Make a portable zip of chainner, and this will happen for every fresh install.

backend/src/server_host.py
```diff
@@ -313,6 +314,11 @@ async def shutdown(request: Request):
     return json(success_response())
 
 
+@app.get("/status")
+async def status(request: Request):
+    return json({"ready": backend_ready})
```
RunDevelopment, line 319: Quick question: what does `ready: true` mean? As is: what state of the server does this represent?

joeyballentine: That means that the backend is done with all its initial setup, ie installing dependencies and starting the worker

joeyballentine:
fixed

joeyballentine:
nvm, its not ready. it hangs on subsequent opens... :(

joeyballentine:
fixed

src/renderer/main.tsx
```diff
@@ -61,9 +62,37 @@ export const Main = memo(() => {
...
+            setStatusProgress(progress.statusProgress > 0 ? progress.statusProgress : null);
+
+            if (progress.status) {
+                setStatus(progress.status);
+            }
+        });
+    }, []);
+
+    useEffect(() => {
+        ipcRenderer.send(isBackendReady ? 'enable-menu' : 'disable-menu');
```
RunDevelopment, line 82: I don't like the idea that the menu starts out enabled, then gets disabled shortly after, only to be re-enabled when the backend is ready.

Could we start with the menu disabled instead and skip the disabling from here?

src/main/gui/main-window.ts
```diff
@@ -399,7 +405,82 @@ export const createMainWindow = async (args: OpenArguments, settings: ChainnerSe
     });
 
     const progressController = new ProgressController();
-    addSplashScreen(progressController);
+
+    let progressFinished = false;
```
RunDevelopment, line 409: Could you please move all of this added code (so from `let progressFinished` to `progressController.addInterruptListener`) into its own function? `createMainWindow` is already quite complex.

src/renderer/main.tsx
```diff
@@ -61,9 +62,37 @@ export const Main = memo(() => {
         )
     );
 
+    const [status, setStatus] = useState(t('splash.loading', 'Loading...'));
```
RunDevelopment, line 65: I think we should replace all `splash.x` keys with `setup.x` (or similar). What do you think?

src/renderer/main.tsx
```diff
@@ -61,9 +62,37 @@ export const Main = memo(() => {
...
+    }, [isBackendReady]);
+
     if (connectionState === 'failed') return null;
 
-    if (connectionState === 'connecting' || !ready) {
+    if (
+        connectionState === 'connecting' ||
+        !ready ||
+        !isBackendReady ||
+        schemata.schemata.length === 0
```
RunDevelopment, line 94: > schemata.schemata.length === 0

This is weird. Is this to catch the case of a fresh install when the worker doesn't have any nodes for a few seconds?

src/renderer/main.tsx
joeyballentine: Sure

src/renderer/main.tsx
joeyballentine: There's some kind of race condition where the backend is ready but we haven't fetched the nodes yet. At least, there was. It's possible with my changes that this line is no longer necessary, but previously it was trying to get the first schema item and failing, so I added this

RunDevelopment: So the issue isn't with the frontend, but that the host server passes `/nodes` requests along to the worker before the backend is actually ready.

Let's leave this hack in for now. I already have an idea how to fix this.

RunDevelopment: lol

joeyballentine: oh whoops, i just removed it. in my latest testing it wasnt necessary anymore

RunDevelopment: It is still necessary, but don't worry about it. I'll fix it after this PR.

RunDevelopment (approved):

PR #2738, opened 2024-04-04
Fixed server getting stuck on deleted files

### LINKED ISSUE #2679: Deleting loaded images causes infinite loading
**Information:**

- Chainner version: Nightly builds2024-03-12
- OS: Windows 10

**Description**
https://drive.google.com/file/d/1Sa7PA5L7Hj7T57Y8xHOTTKCGB2JptRo-/view
Once loaded, if you delete an image, the chainner will start an infinite load.
The infinite loading continues even with a newly opened chain.
Even if you exit chainner, it will remain until the python process is manually terminated.

**Logs**
[renderer.zip](https://github.com/chaiNNer-org/chaiNNer/files/14603375/renderer.zip)

RunDevelopment:
Fixes #2679

The cause of this bug is absolutely stupid. The issue is that passing an exception as the second positional argument to `logger.debug/info/warn/error` will cause the messages to be formatted with it. Sanic essentially does `message.format(args)` behind the scenes, so the intended usage is `logger.error("My name is %s!", name)`. (We never use this because f-strings exist.) The problem is that formatting will fail if any positional argument goes unused. (Yes, it's great API design when something as important as logging can trivially fail...) For whatever reason, if you fail to format 3 times, the thread you log in will get stuck and never respond to anything again. 

The fix for now is to fix the formatting by not using it. The long term fix is to switch to a competent logger. 

---

Btw, I also check for other instances of positional arguments being used: there were none.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -46,25 +46,25 @@ def collect_input_information(
             if not enforced:
                 try:
                     value = node_input.enforce_(value)  # noqa
-                except Exception as e:
+                except Exception:
                     logger.error(
                         f"Error enforcing input {node_input.label} (id {node_input.id})",
-                        e,
+                        exc_info=True,
                     )
                     # We'll just try using the un-enforced value. Maybe it'll work.
 
             try:
                 input_dict[node_input.id] = node_input.get_error_value(value)
-            except Exception as e:
+            except Exception:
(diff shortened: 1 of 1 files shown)


joeyballentine (approved):
What the fuck

PR #2745, opened 2024-04-05
Fix orphaned python processes on exit

joeyballentine:


backend/src/server_process_helper.py
```diff
@@ -47,21 +47,28 @@ def __init__(self, flags: list[str]):
...
+        logger.info("Closing worker process...")
         self._stop_event.set()
-        self._process.terminate()
-        self._process.kill()
+        if hasattr(self, "_process"):
+            self._process.terminate()
+            self._process.kill()
+            del self._process
+        if hasattr(self, "t1"):
+            del self.t1
```
RunDevelopment, line 66: Don't do that... Please just set it to `None`, so pyright can do its thing.

backend/src/server_process_helper.py
```diff
@@ -47,21 +47,28 @@ def __init__(self, flags: list[str]):
         self._stop_event = threading.Event()
 
         # Create a separate thread to read and print the output of the subprocess
-        threading.Thread(
+        self.t1 = threading.Thread(
```
RunDevelopment, line 50: `t1` is a terrible name. Since you already named the thread "output reader", maybe we could call the variable `_reader_thread`.

backend/src/server_process_helper.py
```diff
@@ -47,21 +47,28 @@ def __init__(self, flags: list[str]):
...
+            self._process.kill()
+            del self._process
+        if hasattr(self, "t1"):
+            del self.t1
 
     def _read_output(self):
         if self._process.stdout is None:
             return
-        for line in self._process.stdout:
+        for line in iter(self._process.stdout):
```
RunDevelopment, line 71: Why?

src/main/backend/process.ts
```diff
@@ -147,11 +151,10 @@ export class OwnedBackendProcess implements BaseBackendProcess {
             return;
         }
 
+        await fetch(`${this.url}/shutdown`, { method: 'POST' });
```
RunDevelopment, line 154: Could you please go through a `Backend` instance instead? I don't want to have the low-level API fetch logic for `/shutdown` duplicated.
```suggestion
        await getBackend(this.url).shutdown();
```

src/main/backend/process.ts
```diff
@@ -147,11 +151,10 @@ export class OwnedBackendProcess implements BaseBackendProcess {
...
         if (this.process.kill()) {
             this.process = undefined;
             log.info('Successfully killed backend.');
-        } else {
-            throw new Error('Unable to the backend process. Kill returned false.');
         }
```
RunDevelopment, line 158: Why did you remove this?

src/main/gui/main-window.ts
```diff
@@ -281,10 +284,14 @@ const registerEventHandlerPostSetup = (
 
     const restartChainner = (): void => {
         if (backend.owned) {
-            backend.tryKill();
+            backend
+                .tryKill()
+                .then(() => {
+                    app.relaunch();
+                    app.exit();
+                })
```
RunDevelopment, line 292: I would use `finally` here in case `tryKill` ever throws. I know that `tryKill` isn't supposed to throw, but we don't have a way to enforce this in the type system, so we can't trust it.

```suggestion
                .finally(() => {
                    app.relaunch();
                    app.exit();
                })
```

backend/src/server_process_helper.py
joeyballentine: Set it to none where? I don't even know what was causing it to not be initialized in the first place (look at the failing CI commits)

backend/src/server_process_helper.py
joeyballentine: This was one of the 500 things I was trying, and I forgot to remove it

backend/src/server_process_helper.py
joeyballentine: Sorry for being dumb lol

src/main/backend/process.ts
joeyballentine: I think I did it assuming that it wouldn't always be able to be killed if it's sitting itself down now

RunDevelopment (approved):

PR #2747, opened 2024-04-05
Use None instead of del

RunDevelopment:
This resolves most of my review comments for #2745.


joeyballentine (approved):
Ooooooooh my deleting it is what caused that.... 🤦‍♂️

PR #2752, opened 2024-04-07
Improved node renaming UX

RunDevelopment:
Changes:
- Disable node renaming in docs. Just like collapsing nodes, it's not useful for docs and just gets in the way.
- Double-clicking the node header no longer toggles collapsing/expanding the node. 
- Double-clicking the icon or name now toggles renaming. The change is that I included the icon as well, which is necessary. If our users ever find a bug to enter something that is displayed as an empty name, they won't be able to rename the node without the icon also acting as a toggle.
- Changed design of text box to be closer to the regular header. So the text is now centered and larger.
- The text box will now fill all available space (and no more).
- Added input cleaning to normalize spaces.
- ESC now cancels renaming instead of confirming it. This can be useful sometimes, but I'm not sure whether we should keep this behavior.

image
image

### DIFF
--- a/src/renderer/components/NodeDocumentation/NodeExample.tsx
+++ b/src/renderer/components/NodeDocumentation/NodeExample.tsx
@@ -14,7 +14,7 @@ import {
 } from '../../../common/common-types';
 import { checkNodeValidity } from '../../../common/nodes/checkNodeValidity';
 import { TypeState } from '../../../common/nodes/TypeState';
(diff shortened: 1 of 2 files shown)


joeyballentine:
> ESC now cancels renaming instead of confirming it.

Oh, that's what I thought I had implemented for it anyway, so yeah I think we should keep it

joeyballentine (approved):

PR #2755, opened 2024-04-07
Don't restart worker unnecessarily when there are no deps to install

RunDevelopment:
The condition `if len(to_install) > 0:` was always true, because `to_install` contains the dependencies the worker requires and not the dependencies that pip will install. So I changed the condition to filter out deps that are already installed before. This means that the worker won't be started, stopped, and then started again anymore.

### DIFF
--- a/backend/src/server_host.py
+++ b/backend/src/server_host.py
@@ -18,6 +18,7 @@ import api
 from custom_types import UpdateProgressFn
 from dependencies.store import (
     DependencyInfo,
+    filter_necessary_to_install,
     install_dependencies,
     installed_packages,
     uninstall_dependencies,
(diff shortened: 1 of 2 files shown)


backend/src/server_host.py
```diff
@@ -383,18 +371,20 @@ async def install_deps(dependencies: list[api.Dependency]):
...
-            if config.close_after_start:
-                flags.append("--close-after-start")
+        restart_flags: list[str] = []
+        if config.error_on_failed_node:
+            restart_flags.append("--error-on-failed-node")
+        if config.close_after_start:
+            restart_flags.append("--close-after-start")
 
-            await worker.start(flags)
+        if len(restart_flags) > 0 or len(deps_to_install) > 0:
```
joeyballentine, line 382: why are the flags that we've passed in a condition that would make this restart? we only need to restart if we installed any deps. the flag stuff is just so that when we restart, we restart with the flags we need. we already have those flags under a normal start

RunDevelopment: > we already have those flags under a normal start

Nope. The first start is always with no flags.

joeyballentine: Then that's a bug isn't it?

joeyballentine: Actually, maybe that was intentional -- we probably don't want to do the error on failed node or close after start flags until the deps are installed properly, so that does make sense.

never mind then

joeyballentine (approved):

PR #2777, opened 2024-04-09
Attempt to fix macos build CI

joeyballentine:


joeyballentine:
So far, 2 runs worked. So, maybe fixed it, maybe didn't. Guess we'll find out

joeyballentine:
merging this without approval in hopes the macos cli gets fixed

PR #2781, opened 2024-04-10
Switch from Webpack to Vite

joeyballentine:
This was incredibly difficult to pull off and took 3 separate attempts (4 if you count me almost giving up during this attempt and basically starting over)

This required a bunch of prerequisite work to remove node-only code from the renderer and fixing a bunch of electron worst-practices. Not everything is currently perfect (we still need a bunch of polyfills for certain things and the way we do IPC is still technically not best-practice) but it is still better than it was before. This was necessary because unlike webpack, there was no option with Vite to just use the node libraries in the browser. This ultimately ended up being a good thing, because supposedly this all helps security.

The big win from switching to Vite though is the improved DX. Webpack + Babel is slow, Vite(esbuild) + SWC is really fast. Vite also has Hot Module Replacement which means the app doesn't have to reload everything when you make a change to a small component. Vite also allows hot-reloading the electron main process, so now we won't spend hours trying to figure out why our main process changes aren't working only to realize we never restarted the app.

There's still a few things left to complete on this before merging, but overall I'm super happy that we were able to pull this off as it will make chaiNNer development much better.


src/main/gui/main-window.ts
```diff
@@ -478,12 +498,13 @@ export const createMainWindow = async (args: OpenArguments, settings: ChainnerSe
...
-            nodeIntegration: true,
+            nodeIntegration: false,
             nodeIntegrationInWorker: true,
-            contextIsolation: false,
+            contextIsolation: true,
+            preload: path.join(__dirname, '/preload.js'),
         },
         icon: `${__dirname}/../public/icons/cross_platform/icon`,
-        show: false,
+        show: true,
```
RunDevelopment, line 507: Why did you change this?

vite.renderer.config.ts
```diff
@@ -0,0 +1,45 @@
...
+                buffer: 'buffer',
+                crypto: 'crypto-browserify',
+                stream: 'stream-browserify',
+                assert: 'assert',
+                http: 'stream-http',
+                https: 'https-browserify',
+                os: 'os-browserify',
+                url: 'url',
+                util: 'util',
+                path: '@chainner/node-path',
```
RunDevelopment, line 35: We need `path`, but do we need the rest?

src/main/gui/main-window.ts
```diff
@@ -478,12 +498,13 @@ export const createMainWindow = async (args: OpenArguments, settings: ChainnerSe
         roundedCorners: true,
         webPreferences: {
             webSecurity: false,
-            nodeIntegration: true,
+            nodeIntegration: false,
             nodeIntegrationInWorker: true,
```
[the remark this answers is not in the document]
joeyballentine: I'm not sure we ever actually used this tbh

src/main/gui/main-window.ts
joeyballentine: For debugging purposes when j first started this, and then I forgot to change it back. Oops

src/main/gui/main-window.ts
joeyballentine, line 502:
```suggestion
            nodeIntegrationInWorker: false,
```

src/main/gui/main-window.ts
joeyballentine, line 507:
```suggestion
        show: false,
```

RunDevelopment:
Found a bug. If you make a new chain and press Ctrl+N, you'll get the "Are you sure?" alert twice. This bug doesn't happen on main.

forge.config.js
```diff
@@ -40,6 +40,7 @@ const deletePycFiles = async (directory) => {
 /** @type {import("@electron-forge/shared-types").ForgeConfig} */
 const config = {
     packagerConfig: {
+        asar: true,
```
RunDevelopment, line 43: What's this?

Because of this option, there is a 200MB `app.asar` in all windows builds now.

RunDevelopment: Okay, `asar: true` itself isn't the problem, it just hides it. Setting `asar: false` and making a build reveals that vite copies the ***entire `node_modules` folder***.

image

The issue seems to be that vite doesn't inline our dependencies in `main/main.ts`

joeyballentine: Something somewhere told me to enable it. Hover over it in vscode and itll explain what it is

RunDevelopment: I looked it up. It's necessary for macos, so we have to do asar. But as I said, asar is not the problem. vite is.

joeyballentine: Is this what you want it to do? https://www.npmjs.com/package/vite-plugin-singlefile

joeyballentine: I can't find any other information on inlining node_modules. Are you sure that's something we want to do in the first place?

RunDevelopment: No. Firstly, this isn't about the renderer, so a single HTML file wouldn't be useful. Secondly, node_module in itself isn't a problem. 

The problem is that vite includes **all** our dependencies. It doesn't so any sort of tree shaking (or similar) to only include the dependencies we need. Almost everything inside the `node_modules` folder is dead weight. I mean, I only noticed this because our .zip distribution went from 100MB to 150MB. Same for the installer.

joeyballentine: I see. That definitely seems wrong, we're probably missing something like what Babel had with the exclude, maybe for the esbuild and/or SWC settings. Should be an easy fix if we just need to ignore it

joeyballentine: Apparently we can use https://www.npmjs.com/package/rollup-plugin-node-externals for it. I tried this, but it makes us unable to import the wasm file for rregex 🙃

RunDevelopment: That's not ideal.

joeyballentine: forgot to give some updates. i've tried several things but just keep running into further issues. will take some more time to investigate why and resolve (if even possible). this is the kind of issue that feels like it should be really simple to fix and yet seems basically impossible, which is really starting to piss me off and makes me want to just give up on this.

RunDevelopment: One idea: If you got everything except for the wasm file working, maybe we could inline the wasm file?

I'm thinking about using a plugin to replace our `import wasmUrl from "rregexp/lib/rregex.wasm?url"` with a data URL along the lines of `const wasmUrl = "data:blob;base64, ...";` Since we're inlining the wasm, it can't be missing, and since we're doing it ourselves, vite can't fuck it up.

joeyballentine: The problem isn't the wasm file anymore. I handled that. It's just the entire process of trying to get it to externalize the dependencies. I got it working to a certain point, but I got stuck when electron-log was complaining about being initialized twice, somehow

joeyballentine: I think it's safe to say this has defeated me yet again. I cannot for the life of me figure out how to do this properly

joeyballentine:
And finally, this is finished.

joeyballentine:
Almost there. Last thing is to figure out what broke npm start...

joeyballentine:
You're just gonna have to be ok with index.html being in the root

forge.config.js
```diff
@@ -40,6 +41,7 @@ const deletePycFiles = async (directory) => {
 /** @type {import("@electron-forge/shared-types").ForgeConfig} */
 const config = {
     packagerConfig: {
+        asar: false,
```
RunDevelopment, line 44: Don't we need this to be `true` for macOS?

joeyballentine: Do we? We weren't using this with webpack

RunDevelopment: Just checked, everything works with `asar: true`, so we can enable no problem.

RunDevelopment: > Do we?

I remember reading about this in the release notes of forge. I don't actually know though.

RunDevelopment: Okay, [reading the docs](https://www.electronjs.org/docs/latest/tutorial/asar-integrity), I think I remembered incorrectly. My bad.

RunDevelopment (approved):

Issue #2784, opened 2024-04-11
Load pytorch model


peamak:
With the latest version (0.23.0), I can't find the node to load a pytorch model… it's pretty useless without it… Everything was working fine in the previous version: did I miss something? Is the node somewhere else?

peamak:
Yes, it told me to update Pytorch indeed after I updated ChaiNNer, which I did.

Currently in the dependency Manager:
PyTorch 2.1.2

peamak:
OK, re-restarting it seems to work… my bad, sorry for the inconvenience.

joeyballentine:
All good. This actually is a bug. For some reason the nodes list isn't refreshing after installing the dependencies

PR #2787, opened 2024-04-12
Refresh nodes after changing dependencies

joeyballentine:
This fixes the issues pointed out in #2784


src/renderer/contexts/DependencyContext.tsx
```diff
@@ -692,6 +693,7 @@ export const DependencyProvider = memo(({ children }: React.PropsWithChildren<un
         supplier()
             .catch((error) => {
                 logOutput(String(error));
+                log.error(error);
```
joeyballentine, line 696: while i was at it, i noticed we weren't actually logging errors from this, only writing them to the terminal view component, so i fixed that as well

RunDevelopment (approved):

Issue #2734, opened 2024-04-04
Tab UI elements can be clicked with any mouse button


RunDevelopment:
**Description**
I found a small UI bug. Tab UI elements in nodes can be clicked with any mouse button (left, right, middle, back, etc.). This interferes with using the middle mouse button to pan the viewport of the chain editor.

**Expected Behavior**
Just like other button, tabs should react to left mouse button clicks.

---

Good example nodes to test this on are Resize, Crop, and Color From:

![image](https://github.com/chaiNNer-org/chaiNNer/assets/20878432/e5c0f2a5-b321-4eda-97ff-3cc3c05efdf2)

RaySit:
Hello, we are CS major students from Vanderbilt University. We did reproduce the bug, finding that this would also interfere with using the right mouse button to bring out the menu. We are trying to complete a pull request for our final project. Can you assign us to this issue? Thanks!

RunDevelopment:
Thank you for working on this @RaySit!

PR #2792, opened 2024-04-14
issue #2734 fixed

RaySit:
Changes to 'TabList' Component Behavior

Modification Overview:
This update modifies how tab changes are handled in the 'TabList' component to ensure that tab switching only occurs with left mouse clicks.

Key Changes:
1. Removal of Redundant 'onChange': The 'onChange' prop of the 'Tabs' component was found to be redundant because tab change handling is now customized. It has been replaced with a placeholder function to maintain the component structure without performing any operation.
2. Conditional Tab Switching: Introduced a 'handleTabClick' function that is triggered on the `onMouseDown` event. This function checks if the left mouse button ('event.button === 0') was clicked before changing the tab. This prevents unintended tab switches from right-clicks or middle-clicks.
3. Efficient Event Handling: By using the 'onMouseDown' event directly on each 'Tab', the tab-switching logic is more directly controlled and can be easily extended or modified to include additional conditions or behaviors.

Screenshot demonstrating right-clicking a tab:
The user node menu is brought out as normal, but the tab is not switched.
Screenshot 2024-04-13 at 21 35 07


src/renderer/components/inputs/elements/TabList.tsx
```diff
@@ -42,7 +43,7 @@ export const TabList = memo(({ value, onChange, reset, isDisabled, options }: Ta
             pt={1}
             size="sm"
             variant="unstyled"
-            onChange={handleChange}
+            onChange={() => {}} // Placeholder, no operation here
```
RunDevelopment, line 46: Please use [our `noop` function](https://github.com/chaiNNer-org/chaiNNer/blob/7fbc9972ad318149f017d64e1ffb2528aa3a0411/src/common/util.ts#L11). It's slightly more efficient to use the same function instead of creating a new one each time, because it allows React to (potentially) do less work.

RunDevelopment (changes requested):
Thank you for the work @RaySit!

Just one minor nit, and then this is good to go.

src/renderer/components/inputs/elements/TabList.tsx
joeyballentine: Does this even actually need to be set? Could the onChange prop just be removed altogether?

RaySit: @RunDevelopment Thank you for the notice! I have made the change to use noop. 

@joeyballentine I tried removing onChange and it seems like it functions correctly as well. Should I just go ahead and delete it?

RunDevelopment: Joey's right, we can remove this, so please go ahead. I thought this field was important, but I was wrong.

RaySit: The change has been made.

RunDevelopment (approved):
Thank you @RaySit!

joeyballentine (approved):
Thanks! (And thanks to your professor for having open source contribution as part of your final project 😄)

PR #2794, opened 2024-04-15
Add node to remove border/frame with constant color

RunDevelopment:
[Someone asked for this on discord](https://discord.com/channels/930865462852591648/1045361136196919366/1224787134834151506), and I thought this would be a useful feature.

image

This node has a few useful features:
- It automatically detects the frame/border color.
- It allows the user to select which non-frame/border section of the image they want.
- It allows the user to add a padding around the selected section.

The select options have the following effects:
image

The node is also quite useful for cropping images with a constant background color:

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_dimension/crop/crop_border.py
+++ b/backend/src/packages/chaiNNer_standard/image_dimension/crop/crop_border.py
@@ -5,12 +5,11 @@ from enum import Enum
 
 import numpy as np
 
-from nodes.impl.color.color import Color
 from nodes.properties.inputs import EnumInput, ImageInput, NumberInput, SliderInput
-from nodes.properties.outputs import ColorOutput, ImageOutput
+from nodes.properties.outputs import ImageOutput
 from nodes.utils.utils import Padding, Region, get_h_w_c
 
-from .. import compositing_group
+from .. import crop_group
 
 
 class SelectMode(Enum):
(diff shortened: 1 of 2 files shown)


joeyballentine:
Is the "crop to content" node still needed after this?

RunDevelopment:
Yes. While the 2 nodes are very similar, they define the "content of the image" differently. Crop to Content only looks at the alpha channel to determine whether a pixel is part of the content of the image. Crop Border looks at the average error of the pixel color to a background color.

This actually the only difference between the 2 nodes. Otherwise, Crop Border is a super set of Crop to Content. I even combined the 2 nodes into Crop to Content while working on this PR, but then discarded it, because the node became too complex. I think new users would have a hard time using this node when the equivalent filters in Gimp and other programs are so much simpler.

image

joeyballentine:
I feel like you don't even have to have separate tabs for them.... "transparent" is just a bg color

RunDevelopment:
> I feel like you don't even have to have separate tabs for them.... "transparent" is just a bg color

That would be a breaking change. 

Right now, Crop To content only considers the alpha channel. Suppose we set an opacity threshold of 50%. Then any pixels with alpha >50% will be considered content, regardless of their RGB values.

However, if we use "transparent" as the border, we'll have to also set its RGB values. So let's say we set the border color to transparent black (RGBA=0,0,0,0), then the absdiff to pixels of transparent white would be ΔRGBA=(1,1,1,0). Despite being fully transparent, this pixel would be considered "content" with a tolerance of 50% due to the large absdiff.

As I said, these 2 nodes define the "content of the image" differently. However, we could change how Crop Border defines it to be compatible with Crop to Content. But I can't think of a useful distance function for that.

And even if we find such a distance function, it would still be a breaking change, because the Crop Border node automatically determines the background/border color. So it even if we have the right distance function for transparent colors, there's no guarantee that the node will pick a transparent color as the background/border color.

joeyballentine (approved):
never mind then 👍

PR #2799, opened 2024-04-17
Install FFMPEG using the backend

### LINKED ISSUE #2271: Minimal bug report
Only few people will face it.(who cannot download it automatically but copy their own ffmpeg.exe to the location)😂😭

1' not checking whether ffprobe.exe exist or not when ffmpwg.exe exist.

2' not checking the version of ffprobe.exe which might raise 
_await timed_supplier_async error_

Bug reproduce
1.Choco install ffmpeg
2.copy ffmpeg.exe from choco installed location to chaineer/ffmpeg/
3.using video frame iterator node to make video.

Will reproduce bug 1.

3'. Copy ffprobe.exe to chaineer/ffmpeg/
4. using video frame iterator node to make video.

Will reproduce bug 2

😂😂😂🤷

RunDevelopment:
This PR moves the FFMPEG installation into the backend. FFMPEG will now be installed when a video node is used, instead of when chainner is first starting up after a fresh installation.

Since the backend now has to store files, I also added a `storage_dir` field to `NodeContext`, so nodes can now store files. This storage directory is set as a flag when starting the server. This allows the main electron process to set the storage dir when starting the backend. Since the main process doesn't start the backend in `npm run dev`, there's also some fallback logic in the backend to use chainner's app data dir.

Since `appdirs` is no longer used by our nodes, I moved it to `requirements.txt` to make it a dev dependency.

I also cleaned up the initialization of our `AppContext` classes a little.

Fixes #2271


joeyballentine (approved):
Code looks good and this works really well. 👍

PR #2805, opened 2024-04-19
Support passthrough/skipping nodes

### LINKED ISSUE #373: Node disable/passthrough

RunDevelopment:
Fixes #373.

I had to change a lot of stuff, because passthrough affects not only how chains are executed, but also the type state and visuals of the chain.

### The UI

I changed the disable toggle to be an icon that opens a popover. It now only displays the current state of the node (enabled, disabled, skipped) and has a tooltip explaining the current state.

image
image

When a node is skipped, all inputs that are *not* mapped to outputs will be faded out. This makes it easy to see that those inputs do not have an effect anymore. This is consistent with how disabling a node makes the whole node transparent.

image

### How inputs are mapped to outputs

To skip a node, we need to know how to rewire inputs to outputs. This is done is 2 ways:

1. By using simple rules to automatically determine the mapping.
1. By manually declaring the mapping in the backend.

This allows us to correct the automatic mapping if it doesn't work for a particular node (which should be very rare). The manual mapping is done with `Output().as_passthrough_of(input_id)`. 

I added a new `shape_as=input_id` property to `ImageOutput`. This new property has 3 functions: (1) it will set the output type to the shape of the given input image, (2) it will make this output a passthrough for the input, and (3) it will be used to check the shape of the returned image (not implemented yet). This means that we already have quite a few nodes that manually set the mapping, which is good.

The automatic method works almost exactly as proposed by @theflyingzamboni in #373. The only differences are that I handle a few more edge cases and that I always pick the first input (regardless of whether if there are other possible mappings). See the `PassthroughMap` class for details.

### How passthrough is implemented

I implemented passthrough as a frontend optimization. So the backend has no concept of passthrough except for manual mappings. It's implemented alongside the other optimizations (e.g. removing disabled nodes) and works by simply rewiring edges. 

From
image

To
image

The optimization for removing unused nodes without side effects will then take care of removing skipped nodes. Since we only remove one level of skipped nodes per optimization, I changed the optimizer to perform multiple passes. This will correctly remove chains of skipped nodes.


joeyballentine:
I wanna test this out before merging, but it'll be a few days before I can do that

joeyballentine (approved):
Works perfectly. Good job

Issue #2762, opened 2024-04-08
Switch node does not support multiple iterators with different lineage as inputs


RunDevelopment:
**Description**
It's impossible to connect 2 iterators with different lineage to the Switch node. Example:

![image](https://github.com/chaiNNer-org/chaiNNer/assets/20878432/b6e90f2e-975f-4d30-965b-b4d0a04e10ff)

Since the switch node is actually completely optimized away by the executor, connecting iterators of different lineage is valid since only the lineage of the selected iterator matters.

PR #2808, opened 2024-04-19
Add Skip existing files option for Save Image & lazy inputs

### LINKED ISSUE #2807: Skip existing images for Save Image
**Motivation**
Currently, Save Image will always overwrite existing images. This is intended, but not always what users want. We have gotten numerous requests to add an option to not overwrite existing images.

**Description**
Add an option to Save Image that allows users to skip existing images. So if there already exists a file that Save Image would overwrite, Save Image should do nothing.

This option should not only not overwrite existing images, it should also automatically optimize the execution to not compute the image that won't be saved. Not computing the image is important, because this will enable use cases such as #1770.

RunDevelopment:
Fixes #2807.

image

This PR adds lazy inputs and uses them to add a "Skip existing files" to Save Image. Lazy inputs work with the `Lazy` type which represent a lazy infallible computation. (While `Lazy` of course handles errors correctly, its users are supposed to assume errors don't happen/pass them up the stack.) This abstraction makes everything pretty easy to implement.

One bug that I haven't addressed yet is that lazy inputs mess with node timing. Since the time it takes to evaluate a lazy input currently counts towards both the node being lazily executed and the node evaluating the lazy execution, the time durations of Save Image nodes are wrong. 
Not sure if this is a blocker for this PR or whether this can be fixed later.

### DIFF
--- a/backend/src/api/__init__.py
+++ b/backend/src/api/__init__.py
@@ -1,6 +1,7 @@
 from .api import *
 from .group import *
 from .input import *
+from .lazy import *
 from .node_context import *
 from .node_data import *
 from .output import *
--- a/backend/src/api/input.py
+++ b/backend/src/api/input.py
@@ -66,13 +66,19 @@ class FormattedErrorValue(TypedDict):
     formatString: str
 
 
+class PendingErrorValue(TypedDict):
+    type: Literal["pending"]
+
+
 class UnknownErrorValue(TypedDict):
     type: Literal["unknown"]
     typeName: str
     typeModule: str
 
 
-ErrorValue = Union[LiteralErrorValue, FormattedErrorValue, UnknownErrorValue]
+ErrorValue = Union[
+    LiteralErrorValue, FormattedErrorValue, UnknownErrorValue, PendingErrorValue
+]
 
 T = TypeVar("T")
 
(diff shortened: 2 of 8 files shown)


joeyballentine:
So if I understand correctly, this works by not evaluating the path for each input until it's used, rather than before the node runs, so that if we don't need the input we don't evaluate it, but we evaluate everything else? 

If that's the case, would there be any downside to doing this all the time for every input? (Besides the major refactoring we'd have to do) Feels like we could get some free optimizations out of this, and even accomplish the switch node optimization through this

RunDevelopment:
That's exactly how it works.

> would there be any downside to doing this all the time for every input? Feels like we could get some free optimizations out of this, [...]

Aside from the obvious fact that we'd have to change every single node, this would also make simple nodes a lot more complex. Every node would have to deal with `Lazy`, which would make chainner nodes more difficult to implement for new contributors and plugin authors. Also, most node use all of their inputs, so there's no point in delaying the evaluation of their arguments. This would just make things slower overall, because lazy evaluation also has an overhead.

So not only would just make all nodes more complex, it would even make most chains slower.

> even accomplish the switch node optimization through this

Yes, and no. Yes, Switch could be implemented like this in its current form, but this would make #2762 impossible, so no. 

It's also plain more efficient to have Switch be a static optimization. Not only because the executor needs to plan for fewer nodes, but also because removing a Switch node may enable further chain optimizations. In general, any optimization that removes nodes/edges also improves other optimizations. Of course, we currently don't implement many optimizations, so I'm just speaking in the abstract right now.

joeyballentine (approved):

PR #2815, opened 2024-04-23
Manually set python encoding env var in subprocesses (to fix language issue)

joeyballentine:
Allegedly (according to [this](https://github.com/python/cpython/issues/105312#issuecomment-1919348455)), this is how you force a python subprocess to output utf-8 only. 

This should theoretically fix the remaining issues around people using non-utf8 windows languages. I have not tested this myself as my system language is english.

### DIFF
--- a/backend/src/dependencies/store.py
+++ b/backend/src/dependencies/store.py
@@ -21,6 +21,8 @@ UNINSTALLING_REGEX = re.compile(r"Uninstalling ([a-zA-Z0-9-_]+)-+")
 
 DEP_MAX_PROGRESS = 0.8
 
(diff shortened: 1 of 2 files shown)


joeyballentine:
I figured out a way to simulate it (just have the subprocess output "Я"), and this does not fix it. the good thing is, i should be able to figure out a solution now.

joeyballentine:
Seems that this does fix the issue: https://discord.com/channels/930865462852591648/1045361136196919366/1232400378608357470

The way I was testing this must have not been representative of the actual problem. So, this should be good to go and should fix the issue.

RunDevelopment (approved):

Issue #1525, opened 2023-01-31
AVIF support 


joeyballentine:
Someone asked about this, and it looks like its possible with a Pillow plugin https://pypi.org/project/pillow-avif-plugin/

joeyballentine:
Closed by #2820

PR #2820, opened 2024-04-26
#1525 add avif support

the-database:
Adds support for loading and saving images in the AVIF format, using `pillow-avif-plugin` as suggested in #1525. 

There are tons of potential options that could be added to the save node for AVIF (see options from [libavif](https://github.com/AOMediaCodec/libavif/blob/main/doc/avifenc.1.md) and options exposed by [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin/blob/aa7ac7f74c84ac409ad8668cd25f466334fd3753/src/pillow_avif/AvifImagePlugin.py#L119)). I kept things to a minimum and only exposed quality and subsampling. Notably, the lossless option is missing since `pillow-avif-plugin` doesn't expose it. I've read that AVIF lossless isn't that great at reducing file sizes anyway, but I'm not sure how true that is. 

Also, I made some guesses when working with `with_id` and `if_enum_group`, please let me know if those magic numbers need any corrections.

### DIFF
--- a/backend/src/nodes/impl/image_formats.py
+++ b/backend/src/nodes/impl/image_formats.py
@@ -68,6 +68,8 @@ def get_pil_formats():
         ".pnm",
         # TGA
         ".tga",
+        # AVIF,
+        ".avif",
     ]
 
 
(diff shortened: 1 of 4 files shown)


joeyballentine:
Thanks for this addition! Could you please format your code changes with ruff?

the-database:
> Thanks for this addition! Could you please format your code changes with ruff?

Done. I had to add `# noqa` to the import statements as ruff wanted to remove them for appearing unused, but they're needed to add AVIF support to pillow's `Image.save` and `Image.load`.

PR #2826, opened 2024-04-28
Use built-in nodejs UUID generator for image pasting

joeyballentine:
Fixes an issue i found with image copying and pasting images directly into chaiNNer.

I'm not really sure _why_ this was broken. It definitely had something to do with Vite, but I have yet to figure out what specifically. What i did discover though, is that UUID4 just uses the native crypto.randomUUID function under the hood, when its available (so not in a browser). For some reason, calling it directly works fine, but using the uuid package does not. It does however work fine in the renderer, which is even weirder.

Something important to note is that we can't replace the renderer's use of UUID4 with this, as it sits in common/utils, where crypto is off limits.

### DIFF
--- a/src/main/gui/main-window.ts
+++ b/src/main/gui/main-window.ts
@@ -9,9 +9,9 @@ import {
 } from 'electron/main';
 import EventSource from 'eventsource';
 import fs, { constants } from 'fs/promises';
+import crypto from 'node:crypto';
 import os from 'os';
 import path from 'path';
-import { v4 as uuid4 } from 'uuid';
 import { BackendEventMap } from '../../common/Backend';
 import { Version } from '../../common/common-types';
 import { log } from '../../common/log';
@@ -245,7 +245,7 @@ const registerEventHandlerPreSetup = (
     ipcMain.handle('clipboard-readImage-and-store', async () => {
         const clipboardData = clipboard.readImage();
         const imgData = clipboardData.toPNG();
-        const imgPath = path.join(os.tmpdir(), `chaiNNer-clipboard-${uuid4()}.png`);
+        const imgPath = path.join(os.tmpdir(), `chaiNNer-clipboard-${crypto.randomUUID()}.png`);
         await fs.writeFile(imgPath, imgData);
         return imgPath;
     });


RunDevelopment:
This isn't a fix for the underlying problem. We already use `uuid` in `src/common/util`, so this is a bomb just waiting to blow up in our faces again.

> What i did discover though, is that UUID4 just uses the native crypto.randomUUID function under the hood, when its available (so not in a browser).

Not exactly true. The uuid package has 2 separate builds: [one for nodejs](https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/src/native.js#L3) and [one for browsers](https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/src/native-browser.js#L2). (Their build system is super messed up, but it basically just replaces the code of `src/xyz.js` files with the contents of `src/xyz-browser.js`.) So it doesn't detect which one is available, the bundler makes that choice when it builds the app.

So vite is likely configured incorrectly to use the [browser](https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/package.json#L32) build of `uuid` over the [node](https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/package.json#L27) build.

joeyballentine:
Yeah I'm not really sure what to do about that to be honest. The crypto module exists and the function that it uses works on its own, so why it doesn't work when done via uuid4 i have no clue. And replacing crypto with a polyfill doesn't do anything either.

RunDevelopment:
This is the error we get:

```
Error occurred in handler for 'clipboard-readImage-and-store': Error: crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported
    at rng (C:\Users\micha\Git\chaiNNer\.vite\build\main.js:90887:13)
    at v4 (C:\Users\micha\Git\chaiNNer\.vite\build\main.js:90978:52)
    at C:\Users\micha\Git\chaiNNer\.vite\build\main.js:204600:78
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:89026)
    at WebContents.emit (node:events:513:28)
15:54:47.272 > Failed to read clipboard image and save to disk Error: Error invoking remote method 'clipboard-readImage-and-store': Error: crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported
```

`rng` [only throws this error in the browser build](https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/src/rng-browser.js#L19). So as I speculated above, the issue is that vite is using the browser build.

> The crypto module exists and the function that it uses works on its own, so why it doesn't work when done via uuid4 i have no clue.

It doesn't matter that the `crypto` module exists, because `uuid` it is not using it. The browser build of `uuid` doesn't import node modules.

> And replacing crypto with a polyfill doesn't do anything either.

I assume you tried to polyfill the `crypto` *module*. In that case, that doesn't help, because the browser build never imports the `crypto` module.

joeyballentine:
I see, so I should see if I can manually make it use the node one. Thanks

Issue #2654, opened 2024-03-07
Add more tiling options


Kim2091:
The main reason for this is just so we can more finely tune the tile size to prevent VRAM overflow. Having to choose between 1024 and 2048 is not good, it can go from 11 GB allocated to over 24, with no in-between

There are two possible ways to do this that I see:
- add more options to the dropdown
- allow the user to type in the number

joeyballentine:
We actually could just add a "custom" option to the dropdown super easily. We previously (a long time ago when this was first added) were restricted in not being able to do this, but now we definitely can. Good callout

joeyballentine:
Closed by #2654

PR #2828, opened 2024-04-28
#2654 add custom tiling option

the-database:
Adds a custom option to the tile size dropdown to allow entering any tile size.

### DIFF
--- a/backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
+++ b/backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
@@ -204,7 +204,7 @@ def upscale_image_node(
     img: np.ndarray,
     model: NcnnModelWrapper,
     tile_size: TileSize,
-    custom_tile_size: int,
+    custom_tile_size: int | None,
     separate_alpha: bool,
 ) -> np.ndarray:
     settings = get_settings(context)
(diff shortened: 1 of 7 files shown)


the-database:
Working on the pyright errors.

backend/src/packages/chaiNNer_ncnn/ncnn/processing/upscale_image.py
```diff
@@ -189,6 +204,7 @@ def upscale_image_node(
     img: np.ndarray,
     model: NcnnModelWrapper,
     tile_size: TileSize,
+    custom_tile_size: int,
```
joeyballentine, line 207: I would prefer for this to be an optional int, so that None can be passed instead of 0 for those other places that call the node, just to be a bit clearer

joeyballentine (commented):
Thanks for this! Just one minor nit

joeyballentine:
I think you just need to make it so the check is `TileSize(custom_tile_size) if tile_size == CUSTOM and custom_tile_size is not None else tile_size,`

the-database:
> I think you just need to make it so the check is `TileSize(custom_tile_size) if tile_size == CUSTOM and custom_tile_size is not None else tile_size,`

Done, thanks!

joeyballentine (approved):

joeyballentine:
Thanks again. I very much appreciate the contributions

PR #2830, opened 2024-04-30
Set conditions for vite module resolution

joeyballentine:
fixes the uuid issue


joeyballentine:
tbh this really doesnt need a review

PR #2845, opened 2024-05-12
Protect against huge image sizes

### LINKED ISSUE #2832: Protect against ludicrous image sizes
**Motivation**
sometimes when upscaling images in a batch, or resizing them, etc. its possible to crash the entire computer because i unwittingly ran an operation that would result in a 8192+ pixels wide/tall image, and processing it takes up 100% of both the cpu and gpu memory. somewhat related to #2671 

**Description**
when an operation will result in an image being more than 4k resolution at any point, pop up a confirmation window saying something like, "Warning: the current configuration will likely freeze your entire computer due to massive image sizes. do you wish to continue?"

**Alternatives**
nodes could visibly turn bright orange with caution stripes or something if they detect an overly large image, and hovering over them will show tooltip-esque text explaining the potential memory overuse. heres a very rough mockup:
image

another alternative, which deals with the root of the issue, would be to have a setting that hard-limits how much memory chainner is allowed to use at any given time- so i could set it to half my installed system memory, or half of my gpu memory, for example.
the following screenshot is from a minecraft launcher, and controls java memory limits, but communicates the same concept:
image

RunDevelopment:
Fixes #2832.

This adds the `assert_image_dimensions` util function I proposed to protect against huge image sizes at runtime.

I also implemented a type-based check, but I'm not satisfied with it yet. I'll try to improve the UX a bit more before requesting review.

### DIFF
--- a/backend/src/nodes/properties/outputs/numpy_outputs.py
+++ b/backend/src/nodes/properties/outputs/numpy_outputs.py
@@ -11,7 +11,7 @@ from api import BaseOutput, BroadcastData, InputId, OutputKind
 from ...impl.image_utils import normalize, to_uint8
 from ...impl.resize import ResizeFilter, resize
 from ...utils.format import format_image_with_channels
-from ...utils.utils import get_h_w_c, round_half_up
+from ...utils.utils import IMAGE_SIZE_LIMIT, get_h_w_c, round_half_up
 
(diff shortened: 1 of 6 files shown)


RunDevelopment:
Blocked by #2909.

PR #2847, opened 2024-05-14
Make type colors themeable

joeyballentine:
This PR adds support for changing the type colors via themes.

One important thing i realized when doing this is that we definitely need to not define themes via sass. The amount of dependency lists i've had to add `theme` to, even though its not used anywhere inside the thing being memo'd, is ridiculous. And the worst part is, I can't do anything about a couple of them. For example, collapsed handles won't change until the node state changes due to how its rendered. Since the colors aren't actually being modified by anything react can see, it has no reason to rerender unless i force it to by misusing dep arrays. I'm pretty sure my theme object solution would take care of this. I'm also open to other solutions.

Anyway, I only implemented different colors for the "ComfortUI" theme to sorta match their colors

image

It's more of a POC than anything else.


joeyballentine:
My solution was to move everything to a custom hook. It still ignores the exhaustive deps thing (idk how else to prevent doing that) but it at least abstracts that behavior out of all these files.

RunDevelopment (approved):
Nice!

PR #2848, opened 2024-05-14
Color Transfer: add Initial Reference Image parameter

Splendide-Imaginarius:
Among other things, this allows performing a consistent color transfer operation to multiple input images, or performing a color transfer to a large image based on the color difference in only a small portion of that image.

### DIFF
--- a/backend/src/nodes/impl/color_transfer/mean_std.py
+++ b/backend/src/nodes/impl/color_transfer/mean_std.py
@@ -89,6 +89,7 @@ def scale_array(
 def mean_std_transfer(
     img: np.ndarray,
     ref_img: np.ndarray,
+    init_img: np.ndarray,
     colorspace: TransferColorSpace,
     overflow_method: OverflowMethod,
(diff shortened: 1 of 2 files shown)


Splendide-Imaginarius:
(I see my formatting isn't to Ruff's liking, I'll fix that soon.)

RunDevelopment:
What problem does this solve?

Splendide-Imaginarius:
> What problem does this solve?

@RunDevelopment I have 3 input images, which are intended to be stacked together later (and which share a color scheme). I have one reference image, which corresponds to one of the input images. I want to do a color transfer from the reference image to its corresponding input image, and perform the same color transformation to the other two input images (such that the consistent color scheme in the 3 input images stays consistent in the 3 output images). This PR enables me to do that by using the same image as the initial reference image for all 3 color transfer operations.

Splendide-Imaginarius:
Ruff seems happy now; feel free to review.

PR #2852, opened 2024-05-16
Add per-node chain-finish cleanup functions + empty cuda cache only on chain finish

joeyballentine:
After doing some profiling, I noticed that clearing the cuda cache was taking significant time to do. And, this was happening after every upscale, which i'm pretty sure is incorrect and bad. We only need to release the memory when we're actually done with it (aka when the chain finishes).

So, I added a hook system to packages. Not sure if this is the best api for it, but packages can now define on_chain_finish callbacks to be run when the chain is finished, to do things like resource clearing. 

This will hopefully bring a little speed improvement to batch upscales

### DIFF
--- a/backend/src/api/node_context.py
+++ b/backend/src/api/node_context.py
@@ -1,7 +1,7 @@
 import time
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Literal
+from typing import Callable, Literal
 
 from .settings import SettingsParser
 
@@ -145,4 +145,12 @@ class NodeContext(Progress, ABC):
         The path of a directory where nodes can store files.
 
         This directory persists between node executions, and its contents are shared between different nodes.
+        """
+
+    @abstractmethod
+    def add_cleanup(self, fn: Callable) -> None:
+        """
+        Registers a function that will be called when the chain execution is finished.
+
+        The function will be called with no arguments.
         """
--- a/backend/src/api/node_context.py
+++ b/backend/src/api/node_context.py
@@ -151,6 +151,6 @@ class NodeContext(Progress, ABC):
     def add_cleanup(self, fn: Callable[[], None]) -> None:
         """
         Registers a function that will be called when the chain execution is finished.
-        
+
         Registering the same function (object) twice will only result in the function being called once.
         """
--- a/backend/src/api/node_context.py
+++ b/backend/src/api/node_context.py
@@ -148,9 +148,9 @@ class NodeContext(Progress, ABC):
         """
 
     @abstractmethod
-    def add_cleanup(self, fn: Callable) -> None:
+    def add_cleanup(self, fn: Callable[[], None]) -> None:
         """
         Registers a function that will be called when the chain execution is finished.
-
-        The function will be called with no arguments.
+        
+        Registering the same function (object) twice will only result in the function being called once.
         """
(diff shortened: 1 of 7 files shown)


RunDevelopment:
> This will hopefully bring a little speed improvement to batch upscales

"Hopefully"? Did... did you not measure whether your optimization actually made things faster?

And on that note: if you do an optimization, please also report how much faster this actually makes things. As a reviewer, I also have to weigh whether the speed-up is worth the tradeoff in code simplicity/quality/maintainability. I can of course measure things myself, but having you give a reference case/number is useful.

> Not sure if this is the best api for it, but packages can now define on_chain_finish callbacks to be run when the chain is finished, to do things like resource clearing.

I would suggest the following API: we have a `def add_cleanup(callback: Callable[[], None]) -> None` method to `NodeContext`. This will allow nodes to add cleanup functions that run when the chain is done. Cleanup callbacks are stored in a `set`, so nodes can decide to run a cleanup action once or once per node execution by doing:
```py
def do_clean_up(): ...

@register(...)
def my_node(context: NodeContext):
    # run once per node execution
    context.add_cleanup(lambda: do_clean_up())
    # run once per chain
    context.add_cleanup(do_clean_up)
``` 

This fixes the following issue with the current `set_on_chain_finish` API:

- `set_on_chain_finish` always runs cleanup actions even if no nodes from that package were executed.
- Cleanup actions were not allowed to assume the dependencies of the package.

joeyballentine:
> "Hopefully"? Did... did you not measure whether your optimization actually made things faster?

I did, after I made this PR, but that was with all of my changes together. I'll test this individually.

joeyballentine:
Turns out this is doing most of the heavy lifting for the optimizations I tested. a 2m20s processing time went down to 2m01s with just this change alone.

backend/src/api/node_context.py
```diff
@@ -146,3 +146,11 @@ def storage_dir(self) -> Path:
...
         This directory persists between node executions, and its contents are shared between different nodes.
         """
+
+    @abstractmethod
+    def add_cleanup(self, fn: Callable) -> None:
+        """
+        Registers a function that will be called when the chain execution is finished.
+
+        The function will be called with no arguments.
+        """
```
RunDevelopment, line 156: Why document what can be expressed with types?

Also, we should add the guarantee about registering the same function to the docs.

```suggestion
    def add_cleanup(self, fn: Callable[[], None]) -> None:
        """
        Registers a function that will be called when the chain execution is finished.
        
        Registering the same function (object) twice will only result in the function being called once.
        """
```

backend/src/process.py
```diff
@@ -342,6 +342,8 @@ def __init__(
         self.__settings = settings
         self._storage_dir = storage_dir
 
+        self.cleanup_fns: set[Callable] = set()
```
RunDevelopment, line 345: Me likes types.

```suggestion
        self.cleanup_fns: set[Callable[[], None]] = set()
```

backend/src/process.py
```diff
@@ -367,6 +369,9 @@ def settings(self) -> SettingsParser:
     def storage_dir(self) -> Path:
         return self._storage_dir
 
+    def add_cleanup(self, fn: Callable[..., Any]) -> None:
```
RunDevelopment, line 372:
```suggestion
    def add_cleanup(self, fn: Callable[[], None]) -> None:
```

RunDevelopment (commented):
Just some minor things.

RunDevelopment:
Probably needs an `from __future__ import annotations` somewhere.

RunDevelopment (approved):

PR #2853, opened 2024-05-17
Remove sleep in upscale

joeyballentine:
I'm assuming this has a reason to be here, otherwise I can't see why it's worth slightly slowing down every upscale.

If it doesn't, let's get rid of it. If it does, can we use a smaller number? Even though it's only a millisecond, that time can add up. A 5 minute video clip at 24fps would cost about 7 extra seconds in processing time because of this.

### DIFF
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -348,6 +348,10 @@ class _ExecutorNodeContext(NodeContext):
 
     @property
     def paused(self) -> bool:
+        # Python is single-threaded, so it's necessary for this thread to yield, so other threads can do some work.
+        # This is necessary because the thread for accepting the `/pause` endpoint would not be able to accept requests otherwise.
+        # This in turn would mean that `self.progress.paused` would never be set to True.
+        # For more information, see https://github.com/chaiNNer-org/chaiNNer/pull/2853
         time.sleep(0)
         return self.progress.paused
 
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -348,6 +348,7 @@ class _ExecutorNodeContext(NodeContext):
 
     @property
     def paused(self) -> bool:
+        time.sleep(0)
         return self.progress.paused
 
     def set_progress(self, progress: float) -> None:
--- a/backend/src/process.py
+++ b/backend/src/process.py
@@ -348,7 +348,6 @@ class _ExecutorNodeContext(NodeContext):
 
     @property
     def paused(self) -> bool:
-        time.sleep(0.001)
         return self.progress.paused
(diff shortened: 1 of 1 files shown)


RunDevelopment:
> I'm assuming this has a reason to be here

Because Python is a single-threaded piece of shit that I don't trust at all. Basically, I don't think Python does time division multi-threading, which means that one running thread with starve all other threads. Normally, blocking IO operations make threads yield, so other threads can get some CPU, but we are only doing computations. So we need some way for threads to yield and let other threads and let other threads do some works. In particular, the thread that accepts the `/pause` request would get starved if we don't sleep here, which makes pausing impossible.

That said, [`time.sleep(0)` is supposed to be make a thread yield](https://stackoverflow.com/questions/787803/how-does-a-threading-thread-yield-the-rest-of-its-quantum-in-python) without actually sleeping any time, but I think that didn't work when I tested it. I could be misremembering, though.

So you definitely can't remove the sleep, but you could try replacing it with `time.sleep(0)` and testing whether pausing still works.

> can we use a smaller number

Yes, but it probably won't do much. Windows' native sleep kernel API only supports millisecond precision. There is a high-precision sleep API, but I don't know if Python uses that. On top of that, Windows likes to sleep a little longer sometimes...

As I said above, you can try 0, though. Python has a special cause for `time.sleep(0)`, where it doesn't actually use the OS' sleep.

joeyballentine:
Ok, I'll try

joeyballentine:
Works perfectly fine. Tested by using a heavy model with 128 tile size. Was able to pause a few times while it was processing. I could actually hear my GPU stop processing while paused. Resuming eventually led to it finishing.

backend/src/process.py
```diff
@@ -348,7 +348,7 @@ def aborted(self) -> bool:
 
     @property
     def paused(self) -> bool:
-        time.sleep(0.001)
+        time.sleep(0)
```
RunDevelopment, line 351: I should have added a comment explaining this in the first place, so let's add it now.

```suggestion
        # Python is single-threaded, so it's necessary for this thread to yield, so other threads can do some work.
        # This is necessary because the thread for accepting the `/pause` endpoint would not be able to accept requests otherwise.
        # This in turn would mean that `self.progress.paused` would never be set to True.
        # For more information, see https://github.com/chaiNNer-org/chaiNNer/pull/2853
        time.sleep(0)
```

RunDevelopment (approved):
Let's add a comment explaining the `thread.sleep(0)`.

After that, feel free to merge this.

PR #2855, opened 2024-05-17
Add Regex Find node

RunDevelopment:
This adds the Regex Find node, a simple node that just executes a regex and returns the first match.

image

To make this node more powerful, I also added a text pattern option that allows users to use a text pattern to access the capturing groups of the match.

image

This should be enough to do some basic parsing for things like file names.

---

Other changes:
- I also improved `regexReplace` to return `never` for more combinations of invalid inputs.


joeyballentine (approved):

joeyballentine:
Nice!

PR #2858, opened 2024-05-17
Cache model bytes calculation

joeyballentine:


backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -55,15 +59,16 @@ def upscale(
...
-            element_size = 2 if use_fp16 else 4
-            model_bytes = sum(
-                p.numel() * element_size for p in model.model.parameters()
-            )
+            model_bytes = MODEL_BYTES_CACHE.get(model)
+            if model_bytes is None:
+                model_bytes = sum(p.numel() * 4 for p in model.model.parameters())
+                MODEL_BYTES_CACHE[model] = model_bytes
+            if options.use_fp16:
+                model_bytes = model_bytes // 2
```
RunDevelopment, line 67: I think that `if` needs to go inside the cuda block.

RunDevelopment (changes requested):

backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
joeyballentine: Oh yeah, true

RunDevelopment (approved):

PR #2860, opened 2024-05-17
Only copy non-writeable image when CPU upscaling

joeyballentine:
Copying takes time and memory. Why waste this time and memory if we are just going to immediately copy the image to the GPU?

Now the copy will only happen if the user is upscaling with CPU. On anything else, the image is simply moved to a tensor (sharing the same memory) after being made writeable, then copied to the GPU. The image is then set back to readonly.

This made a significant cut to processing time.

### DIFF
--- a/backend/src/nodes/impl/pytorch/auto_split.py
+++ b/backend/src/nodes/impl/pytorch/auto_split.py
@@ -52,8 +52,9 @@ def _into_tensor(
     img: np.ndarray, device: torch.device, dtype: torch.dtype
 ) -> torch.Tensor:
     img = np.ascontiguousarray(img)
+    writeable = img.flags.writeable
     try:
-        if not img.flags.writeable and device == torch.device("cpu"):
+        if not writeable and device == torch.device("cpu"):
             img = np.copy(img)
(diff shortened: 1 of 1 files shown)


backend/src/nodes/impl/pytorch/auto_split.py
```diff
@@ -72,9 +72,13 @@ def upscale(img: np.ndarray, _: object):
...
             # convert to tensor
             img = np.ascontiguousarray(img)
-            if not img.flags.writeable:
+            if not img.flags.writeable and device == torch.device("cpu"):
                 img = np.copy(img)
+            else:
+                # since we are going to copy the image to the GPU, we can skip the copy here
+                img.flags.writeable = True
             input_tensor = torch.from_numpy(img).to(device, dtype)
+            img.flags.writeable = False
```
RunDevelopment, line 81: This is quite complex, so please move this into a function.

It's also not 100% correct. If `from_numpy` or `to` throws, the image might be incorrectly flagged as writeable.

RunDevelopment (approved):

Issue #148, opened 2022-05-06
Add support for branching


RunDevelopment:
We could do branching by using a special value representing a not-taken branch:

The main problem is that the current execution engine takes all possible paths. This is good and what we want it to do, but it makes branching (where we only want it to take some paths) difficult. So how about we use a special value, let's call it D, to represent a branch that is not taken?
To create a dead branch, a node simply has to output D in any of its outputs. Nodes would then handle D inputs as follows: if any of the required inputs is D, output D, and if an optional input is D, set the input to its default value. This is enough to implement branching (assuming no cycles).

This system would transparently implement branching. The D-input check would happen before actually running nodes, so our current nodes will just work in this system.

That being said, this only how to do branching. I don't have any good ideas as for how we could do the condition expressions. 

For more information, see [here](https://discord.com/channels/930865462852591648/930865780663398451/972239321732091996).

joeyballentine:
The condition expressions, i assume, would just be something like 

```
inputs:
-[number input]
-[dropdown(==, <=, >=, <, >, !=)]
-[number input]
outputs:
[true]-
[false]-
```

PR #2866, opened 2024-05-18
Add Compare & Conditional node

### LINKED ISSUE #2671: logical flow control / image size output
**Motivation**
it would be nice to be able to have switch nodes automatically switch for complicated tasks based on logic. the other day i was upscaling a large number of images with no way to know the original resolution of every single one, but some were already decent resolution and didnt need an upscale, just a sharpening. It would be beneficial to have this so i could set up a thing where if the image size is already bigger than a certain amount, it would not upscale it but route it to a sharpening part of the graph instead.

**Description**
this suggestion is kindof 3 different but related things:

1. the choice of which output on a switch node should have an input that works based on an int number (1 = A, 2 = B etc)
2. there should be a math comparison node that works like an IF statement depending on the comparison (if input A == input B then C else D, and perhaps the comparison could be configured by the user to be ==, >= , <=, etc.)
3. load image node / load images node should have an output of width / height size in pixels

**Alternatives**
I haven't thought of any alternative suggestions that would be beneficial on a broader scale; however one alternative for my specific use case would be to just detect in upscaling nodes if the resulting image would be a ludicrous size like 8192 pixels wide and skip the upscale in that case.

However, i think having numerical logic / automatic flow control would be more applicable for more uses.

RunDevelopment:
This PR adds 2 new nodes: Compare and Conditional.

*Compare* is a simple node that takes 2 numbers and returns either ==, !=, <, <=, >, >=. The result of this node will be a boolean. This is currently the only node that can produce boolean values that can be connected other nodes.

*Condition* is the node equivalent of `return a if cond else b`. The node only evaluates the input that is actually returned, so we don't perform unnecessary computation. This is done with lazy inputs added in #2808. Pass through is supported by passing through the "If true" value.

Other changes:
- Added `BoolOutput` for returning boolean values.
- Added `has_handle: bool` parameter to `BoolInput`. This currently switches between the old enum/dropdown based implementation and a new generic-input based implemented. We should improve this in the future, but it works for now.
- Added type tags for the `bool` type.

Fixes #2671

Not that this does **not** solve the issue #148. #148 describes a more general branching implementation that is more powerful than the one implemented here.

image
image
image


joeyballentine (approved):
Awesome. Genius implementation

joeyballentine:
That's what i get for using github's merge resolver instead of doing it locally...

RunDevelopment:
No worries. I fixed it.

joeyballentine (approved):

PR #2870, opened 2024-05-18
Fix custom tile size type

joeyballentine:
Since our CI no longer fails for invalid types, this got missed. Easy fix though.


joeyballentine:
Man, it's my fault for this. In my review I asked him to make this accept None instead of passing 0 in, not knowing that the value actually isn't none for hidden inputs. oh well

RunDevelopment (approved):

PR #2872, opened 2024-05-18
Give threshold softness extra precision

joeyballentine:
This was typed as a float yet was actually an int. I assumed it was meant to have greater precision.


RunDevelopment:
Nope. It was meant to be int. I had it be a float during development, but figured that having 11 levels (0-10) was enough.

RunDevelopment (approved):

joeyballentine:
For documentary purposes, the title of this PR should have been changed before merging. That's my bad.

PR #2878, opened 2024-05-19
Add "order" dropdown for both spritesheet nodes

### LINKED ISSUE #2851: Having a different mode for rxc vs cxr would be helpful on image sequences that start with coloumn number and row as second and vise versa
**Motivation**
Need the feature to merge images that start with column number and second the row number and vise versa

**Description**
I have a Image sequence with the syntax "text1<ColoumnNumberStartsWith000><RowsNumerStartsWith000>_text2.tga"
If I use the "standard" the Merged Spritesheet Image looks not correct. I think It should use the first number as column index and second number for put it horizontally right to it.

**Alternatives**
Best case would be some kind of expression function to describe witch number or "part" of a file name would be stand for "column" and which for "rows"

joeyballentine:
closes #2851

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/batch_processing/split_spritesheet.py
+++ b/backend/src/packages/chaiNNer_standard/image/batch_processing/split_spritesheet.py
@@ -34,7 +34,7 @@ class OrderEnum(Enum):
             "The number of columns to split the image into. The width of the image must be a multiple of this number."
         ),
         EnumInput(OrderEnum, label="Order", default=OrderEnum.ROW_X_COLUMN).with_docs(
-            "The order in which the images are combined."
+            "The order in which the images are separated."
         ),
     ],
     outputs=[
--- a/backend/src/packages/chaiNNer_standard/image/batch_processing/split_spritesheet.py
+++ b/backend/src/packages/chaiNNer_standard/image/batch_processing/split_spritesheet.py
@@ -75,22 +75,16 @@ def split_spritesheet_node(
         if order == OrderEnum.ROW_MAJOR:
             row = index // columns
             col = index % columns
-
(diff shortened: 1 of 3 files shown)


RunDevelopment:
The "row x column" and "column x row" names are confusing. The infix "x" typically denotes multiplication which most people think of as commutative. By this interpretation, these two options describe the same thing (`a*b = b*a`).

The terms you are looking for are [row major and column major](https://en.wikipedia.org/wiki/Row-_and_column-major_order). Unfortunately, these terms aren't really used outside CS and math, so we need to explain them. I would probably give a short description of both and then use code blocks to visually show the difference like this:

```
Row major:
→ 0 1 2
  3 4 5
  6 7 8

Column major:
↓ 0 3 6
  1 4 7
  2 5 8
```

Also, the description should be a hint (so with a little (i)) IMO. We can't expect people to know this, so we should make the explanation easy to access.

backend/src/packages/chaiNNer_standard/image/batch_processing/split_spritesheet.py
```diff
@@ -60,13 +72,24 @@ def split_spritesheet_node(
...
+            row = index % rows
 
-        sprite = sprite_sheet[
-            row * individual_h : (row + 1) * individual_h,
-            col * individual_w : (col + 1) * individual_w,
-        ]
+            sprite = sprite_sheet[
+                row * individual_h : (row + 1) * individual_h,
+                col * individual_w : (col + 1) * individual_w,
+            ]
```
RunDevelopment, line 90: Those `sprite = sprite_sheet[...]` are identical in both cases, so they can be moved outside the if/else.

RunDevelopment (approved):

PR #2879, opened 2024-05-19
Add a "simple" mode for saving video

joeyballentine:
People often have confusion as to why the quality (crf) slider is what it is, or what the preset option does. If you don't know FFMPEG, then those things make no sense. Those users probably also don't know anything about encoders or formats.

So, I made a "simple" mode. It gives only a few options for formats, and uses a quality slider from 0-100 that sets both the CRF and preset values. This should make the node much more easily usable and intuitive for non-superusers 

Open questions:
1) Should AVI be an option for lossless video?
2) Should we use HEVC instead of H.265 in the parens?

image


backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -323,16 +353,50 @@ def save_video_node(
...
+            SimpleVideoFormat.MP4_H265: VideoFormat.MP4,
+            SimpleVideoFormat.WEBM: VideoFormat.WEBM,
+            SimpleVideoFormat.GIF: VideoFormat.GIF,
+        }[simple_video_format]
+
+        encoder = {
+            SimpleVideoFormat.MP4_H264: VideoEncoder.H264,
+            SimpleVideoFormat.MP4_H265: VideoEncoder.H265,
+            SimpleVideoFormat.WEBM: VideoEncoder.VP9,
+        }[simple_video_format]
```
RunDevelopment, line 381: This will throw an error for GIF.

backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -323,16 +353,50 @@ def save_video_node(
     _: None,
     save_dir: Path,
     video_name: str,
+    simplicity: Simplicity,
     container: VideoFormat,
     encoder: VideoEncoder,
     video_preset: VideoPreset,
     crf: int,
     advanced: bool,
     additional_parameters: str | None,
```
RunDevelopment, line 362: You still use `additional_parameters` in simple mode. See the "Append additional parameters" section below.

backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -323,16 +353,50 @@ def save_video_node(
...
+    simple_video_format: SimpleVideoFormat,
+    quality: int,
     fps: float,
     audio: Any,
     audio_settings: AudioSettings,
 ) -> Collector[np.ndarray, None]:
+    if simplicity == Simplicity.SIMPLE:
```
RunDevelopment, line 369: I feel like this entire block should be a function.

RunDevelopment (changes requested):

RunDevelopment:
> Should AVI be an option for lossless video?

Not sure. If it's easy to add, then why not. I just think that it should have "lossless" in the name of the option if you add it.

> Should we use HEVC instead of H.265 in the parens?

I think both are used, so maybe both?

joeyballentine:
I decided not to add AVI for now. We can add that if anyone wants it. I also added a AVC and HEVC to the h264 and h265 labels, respectively. Also, I made it hide the quality slider for GIF (like advanced mode does)

backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -354,27 +438,28 @@ def save_video_node(
...
-                for nop in non_overridable_params:
-                    if not key.startswith(nop):
-                        output_params[key] = value
-                    else:
-                        raise ValueError(f"Duplicate parameter: -{parameter}")
-            else:
-                global_params.append(f"-{parameter}")
+    if simplicity == Simplicity.ADVANCED:
+        # Append additional parameters
+        if advanced and additional_parameters is not None:
```
RunDevelopment, line 444: Could you combine this into a single if statement?

backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -98,6 +98,62 @@ class AudioSettings(Enum):
...
+
+def get_simple_format(
+    simple_video_format: SimpleVideoFormat, quality: int
+) -> tuple[VideoFormat, VideoEncoder, VideoPreset, int]:
+    container = None
+    encoder = None
+    video_preset = None
+    crf = None
+
+    container = {
```
RunDevelopment, line 121: Those declarations shouldn't be necessary.

```suggestion
    container = {
```

backend/src/packages/chaiNNer_standard/image/video_frames/save_video.py
```diff
@@ -98,6 +98,62 @@ class AudioSettings(Enum):
...
+
+    if simple_video_format != SimpleVideoFormat.GIF:
+        encoder = {
+            SimpleVideoFormat.MP4_H264: VideoEncoder.H264,
+            SimpleVideoFormat.MP4_H265: VideoEncoder.H265,
+            SimpleVideoFormat.WEBM: VideoEncoder.VP9,
+        }[simple_video_format]
+    else:
+        # Encoder will be ignored, it just needs to be a value in the case of GIF
+        encoder = VideoEncoder.H264
```
RunDevelopment, line 136: This feels like the logic is backwards. I think of GIF as the special case, but this looks like GIF is related to the general case instead. I would turn that around like this:

```suggestion
    if simple_video_format == SimpleVideoFormat.GIF:
        # Encoder will be ignored, it just needs to be a value in the case of GIF
        encoder = VideoEncoder.H264
    else:
        encoder = {
            SimpleVideoFormat.MP4_H264: VideoEncoder.H264,
            SimpleVideoFormat.MP4_H265: VideoEncoder.H265,
            SimpleVideoFormat.WEBM: VideoEncoder.VP9,
        }[simple_video_format]
```

joeyballentine: ended up just adding a case to the dict for gif

RunDevelopment:
I think we need a label for the additional params text box. Otherwise gtg

RunDevelopment (approved):

PR #2880, opened 2024-05-19
Allow using a connection for custom tile size

joeyballentine:
Not sure why this was disabled in the first place


backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -184,7 +184,7 @@ def estimate():
                 max=None,
                 default=TILE_SIZE_256,
                 unit="px",
-                has_handle=False,
+                has_handle=True,
```
RunDevelopment, line 187: Number inputs have handles by default.

```suggestion
```

RunDevelopment (changes requested):
The other upscale nodes have the same problem.

RunDevelopment (approved):

PR #2884, opened 2024-05-19
Fix for writeable flag error

joeyballentine:
Some numpy arrays cannot be set to writeable -- this seems to include arrays that come from rust/chainner-ext. These would cause errors from attempting to make them writeable.

So, I just wrapped it with a try/except, and made it copy if it cannot be set to writeable. Easy fix, and we still get the optimization from it if it is able to be set to writeable


RunDevelopment:
Will the `finally` not cause the same issue?

joeyballentine:
It doesn't. I tested this.

joeyballentine:
Gonna merge this so it's fixed in the nightly

PR #2886, opened 2024-05-19
Use "put_nowait" for event queue

joeyballentine:
Why wait and waste time when not waiting works?

Based on skimming a trace, this actually does appear to make the SSE requests no longer blocking, I think. I could be entirely wrong though, since put_nowait theoretically doesnt do anything different with an infinite size queue


RunDevelopment:
Could you make the functions in the executor that no longer need to be async normal function again?

RunDevelopment (approved):

PR #2897, opened 2024-05-22
Add "Split Color" node

joeyballentine:
This node takes in a color and either outputs its colors as numbers or as grayscale colors.

This is **NOT** complete yet. I have something wrong with the type that I cannot figure out, which prevents the Color option from actually working.

image

Issue:
image


backend/src/packages/chaiNNer_standard/utility/color/split_color.py
```diff
@@ -0,0 +1,102 @@
...
+    schema_id="chainner:utility:split_color",
+    name="Split Color",
+    description="Split a color into its RGBA values, either as numbers or grayscale colors.",
+    icon="MdColorLens",
+    inputs=[
+        ColorInput(),
+        EnumInput(SplitColorMode, label="Mode", default=SplitColorMode.NUMBER),
+    ],
+    outputs=[
+        NumberOutput("Red", output_type=output_navi_type),
```
RunDevelopment, line 38: This doesn't work, because `NumberOutput` "casts" `output_type` to a number. You need to make a new output class that doesn't do this.

backend/src/packages/chaiNNer_standard/utility/color/split_color.py
```diff
@@ -0,0 +1,102 @@
...
+output_navi_type = """match Input1 {
+    SplitColorMode::Number => int(0..255),
+    SplitColorMode::Color => Color { channels: 1 },
+    _ => never }
+"""
+
+
+@color_group.register(
+    schema_id="chainner:utility:split_color",
+    name="Split Color",
```
RunDevelopment, line 30: IMO this node should be called "Separate Color" to be consistent with the Separate RGBA node.

backend/src/packages/chaiNNer_standard/utility/color/split_color.py
```diff
@@ -0,0 +1,102 @@
...
+        NumberOutput("Alpha", output_type=output_navi_type),
+    ],
+)
+def split_color_node(
+    color: Color,
+    mode: SplitColorMode,
+) -> tuple[int, int, int, int] | tuple[Color, Color, Color, Color]:
+    length = len(color.value)
+    if length == 1:
+        if mode == SplitColorMode.NUMBER:
```
RunDevelopment, line 50: This implementation is overly complex. You essentially need to do 2 things: split the color into RGBA and output the channels. By doing these things after another instead of integrated into each other, the code becomes simpler. So like this:

```py
# get rgba
if len == 1:
    r, g, b, a = ...
elif ...:
    ...

# output
if mode == number:
    return to_num(r), to_num(g), to_num(b), to_num(a)
elif ...:
    ...
```

backend/src/packages/chaiNNer_standard/utility/color/split_color.py
```diff
@@ -0,0 +1,102 @@
...
+
+from .. import color_group
+
+
+def _norm(n: float) -> int:
+    return int(max(min(n * 255, 255), 0))
+
+
+class SplitColorMode(Enum):
+    NUMBER = 0
```
RunDevelopment, line 17: Number is too generic of a name.

backend/src/packages/chaiNNer_standard/utility/color/split_color.py
```diff
@@ -0,0 +1,102 @@
...
+
+from nodes.impl.color.color import Color
+from nodes.properties.inputs import ColorInput, EnumInput
+from nodes.properties.outputs import NumberOutput
+
+from .. import color_group
+
+
+def _norm(n: float) -> int:
+    return int(max(min(n * 255, 255), 0))
```
RunDevelopment, line 13: Please round (with `round_half_up`).

RunDevelopment (changes requested):

backend/src/packages/chaiNNer_standard/utility/color/split_color.py
joeyballentine: Oh right, duh 🤣 🤦‍♂️

joeyballentine:
image

backend/src/packages/chaiNNer_standard/utility/color/separate_color.py
```diff
@@ -0,0 +1,93 @@
...
+) -> (
+    tuple[int, int, int, int]
+    | tuple[float, float, float, float]
+    | tuple[Color, Color, Color, Color]
+):
+    length = len(color.value)
+    if length == 1:
+        r = color.value[0]
+        g = 0
+        b = 0
```
RunDevelopment, line 71: Why do you put the grayscale level into the red channel?

backend/src/packages/chaiNNer_standard/utility/color/separate_color.py
```diff
@@ -0,0 +1,93 @@
...
+        a = color.value[3]
+    else:
+        raise AssertionError("Invalid number of color channels")
+
+    if mode == SeparateColorMode.UINT8:
+        return _to_num(r), _to_num(g), _to_num(b), _to_num(a)
+    elif mode == SeparateColorMode.PERCENT:
+        return r, g, b, a
+    elif mode == SeparateColorMode.COLOR:
+        return Color((r,)), Color((g,)), Color((b,)), Color((a,))
```
RunDevelopment, line 91: The `Color` constructor is supposed to be private. I just couldn't do this, because Python doesn't support it. Use `Color.gray` instead.

backend/src/packages/chaiNNer_standard/utility/color/separate_color.py
```diff
@@ -0,0 +1,93 @@
...
+    UINT8 = 0
+    PERCENT = 1
+    COLOR = 2
+
+
+LABELS = {
+    SeparateColorMode.UINT8: "Uint8",
+    SeparateColorMode.PERCENT: "Percent",
+    SeparateColorMode.COLOR: "Color",
+}
```
RunDevelopment, line 27: Uint8 and Percent aren't good names. People that aren't programmers don't know what a uint8 is, and a percentage is typically a number between 0 and 100.

I would suggest labelling uint8 as "8 Bit". Not sure what to do about percentage tbh.

joeyballentine: I could make percent an int between 0 and 100 instead?

RunDevelopment: Idk tbh. Bot 0-1 and 0-100 would be useful.

joeyballentine: I think 0-100 would be better. It can always just be divided by 100 after all

RunDevelopment (approved):

PR #2899, opened 2024-05-22
Disable reloading for main process

RunDevelopment:
As I said on Discord, the hot reloading of the main process gets in the way of my development workflow, because it restarts chainner completely and loses the current chain/state. This is a huge problem for me, because I frequently change files in `src/common` (e.g. type system stuff) that cause these reloads. So not having it is better.


joeyballentine:
Since I personally prefer having this, could you make it so that there's s separate command for having this enabled instead of just removing it completely?

RunDevelopment:
Done. You can now have hot reloading with `npm run dev:reload`.

joeyballentine (approved):

PR #2903, opened 2024-05-22
Select best-matching node based on search query

RunDevelopment:
This improves the UX of the context node selector by setting the selected index to the node that matches the search query the best. This should make it a lot easier to find common nodes such as Text, Number, and Color.

image image image image image image


joeyballentine (approved):
Awesome

PR #2904, opened 2024-05-22
Fix Color type color not shading properly

joeyballentine:


RunDevelopment:
The real problem is that we parse those hex colors incorrectly. See `colorTools.ts` -> `hexToRgb`. You could probably just use `parseRgbHex` from `colorUtil` to fix this.

Also, we could probably clean up or color utility...

joeyballentine:
Yeah probably, but this fixes the present issue. Redoing how this works can be a separate pr

RunDevelopment:
This PR isn't a fix but a workaround. Once the real bug is fixed, there's no point in making this change.

joeyballentine:
I don't really care about fixing the actual problem behind this right now. This works. Why not merge it?

RunDevelopment:
Because you're just accumulating tech debt, when the actual fix is pretty simple. Let me make a PR for it.

RunDevelopment:
Actually, let's go. I'm gonna just refactor our color stuff after this.

RunDevelopment (approved):

RunDevelopment:
Also, what issue was this even fixing? You never mentioned.

joeyballentine:
This was for edges being invisible when you clicked on them

PR #2915, opened 2024-05-28
Add similarity check to ONNX model conversions

### LINKED ISSUE #2585: Perform a similarity check + inference test after converting to onnx
The main con of this would be that then onnx inference would be required for conversion. Maybe it could just be an optional thing if onnx is installed.

Why is this necessary? According to musl:

> [Check] If the converted onnx is relatively close to the outputs of pytorch inference. This is not strictly necessary, but the [documentation recommends it](https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html) (np.testing.assert_allclose). It obviously won't change the model, but I found some situations where the converted onnx gave like 20% difference between pytorch and onnx, which should not be acceptable. So if this is not tested with assert_allclose, the outputs could be bad while still exporting just fine.

RunDevelopment:
Closes #2585.

This adds a new check to ensure that the ONNX models of PyTorch models produce the same outputs. This new check is options and off by default, because it requires an ONNX runtime.

I also cleaned up some other code along the way.

To the reviewer: please test this. I don't really do ONNX conversions, so I just tested it on a few models and it seemed to work.

### DIFF
--- a/backend/src/packages/chaiNNer_onnx/onnx/utility/convert_to_ncnn.py
+++ b/backend/src/packages/chaiNNer_onnx/onnx/utility/convert_to_ncnn.py
@@ -11,8 +11,6 @@ from nodes.properties.outputs import NcnnModelOutput, TextOutput
 
 from .. import utility_group
 
-FP_MODE_32 = 0
-
 
 @utility_group.register(
     schema_id="chainner:onnx:convert_to_ncnn",
--- a/backend/src/packages/chaiNNer_pytorch/pytorch/utility/convert_to_ncnn.py
+++ b/backend/src/packages/chaiNNer_pytorch/pytorch/utility/convert_to_ncnn.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from sanic.log import logger
 from spandrel import ImageModelDescriptor
 from spandrel.architectures.DAT import DAT
 from spandrel.architectures.HAT import HAT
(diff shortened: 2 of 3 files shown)


joeyballentine:
Errors when running in fp16 mode:

image

Full stack trace:

```Error

An error occurred in a Convert To ONNX node:

[ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Unexpected input data type. Actual: (tensor(float)) , expected: (tensor(float16))

Input values:
• PyTorch Model: Value of type 'spandrel.__helpers.model_descriptor.ImageModelDescriptor'
• Data Type: fp16
• Opset: 17
• Verify: Yes

Stack Trace:
Traceback (most recent call last):
  File "C:\Users\joeyj\Documents\Git\chaiNNer\backend\src\process.py", line 174, in run_node
    raw_output = node.run(context, *enforced_inputs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\joeyj\Documents\Git\chaiNNer\backend\src\packages\chaiNNer_pytorch\pytorch\utility\convert_to_onnx.py", line 113, in convert_to_onnx_node
    verify_models(model, onnx_model_bytes, fp16)
  File "C:\Users\joeyj\Documents\Git\chaiNNer\backend\src\packages\chaiNNer_pytorch\pytorch\utility\convert_to_onnx.py", line 132, in verify_models
    onnx_out = session.run(
               ^^^^^^^^^^^^
  File "C:\Users\joeyj\AppData\Local\Programs\Python\Python311\Lib\site-packages\onnxruntime\capi\onnxruntime_inference_collection.py", line 220, in run
    return self._sess.run(output_names, input_feed, run_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Unexpected input data type. Actual: (tensor(float)) , expected: (tensor(float16))
```

RunDevelopment:
Okay, I think it should work now.

joeyballentine:
image
looks like it works properly when there's a mismatch

joeyballentine (approved):

PR #2919, opened 2024-05-29
Add support for ONNX type and tag broadcasting

### LINKED ISSUE #1370: Add broadcasting to ONNX models
Right now, we can't provide any information about images output from ONNX models. This is in part due to our self-imposed limitation of not requiring the onnx module for the OnnxModel class. I also ran into complications and failure points when trying to get the required info from the graph, so this may be more complicated to attempt than I'd originally thought. May be possible to do it by limiting the set of usable ONNX models (i.e. how we eventually prevented NCNN models with scales less than 1 from loading).

RunDevelopment:
Closes #1370

This adds support for type and tag broadcasting for all ONNX models. While we did have type broadcasting for ONNX models before, those broadcasts did not contain much information. Now they contain the model's input/output channels and scale, just like for PyTorch and NCNN models. This finally allows us to compute output types for ONNX's Upscale Image node.

A model's input/output channels and scale(s) are detected using `onnx.shape_inference` [[1]](https://onnx.ai/onnx/api/shape_inference.html) [[2]](https://github.com/onnx/onnx/blob/main/docs/ShapeInference.md). This allows us to statically determine the shape of the output tensor for a given input tensor shape without actually running the model (=no inference session). While this is roughly 10-100x faster than creating an `ort.InferenceSession`, it's still not free. Loading generic ONNX models will now take roughly 3-4x longer. This might sound bad, but it's not. E.g. an ESRGAN ONNX model previously took 0.11s and now it takes 0.38s. A lot slower, but not a problem.

image
image

---

Changes:
- Extract out loading ONNX models into its own file. This ensures that the type definitions for ONNX models don't depend on the ONNX package.
- Store ONNX model information in new `OnnxInfo` class.
- Removed `is_rembg_model` function, because it wasn't necessary.
- Refactored `create_inference_session`.

### DIFF
--- a/backend/src/nodes/impl/onnx/load.py
+++ b/backend/src/nodes/impl/onnx/load.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+import time
+
+import onnx
+import onnx.inliner
+import re2
+from sanic.log import logger
+
+from .model import OnnxGeneric, OnnxInfo, OnnxModel, OnnxRemBg
+from .utils import (
+    get_opset,
+    get_tensor_fp_datatype,
+    image_to_image_shape_inference,
+)
+
+re2_options = re2.Options()
+re2_options.dot_nl = True
+re2_options.encoding = re2.Options.Encoding.LATIN1
+
+U2NET_STANDARD = re2.compile(b"1959.+1960.+1961.+1962.+1963.+1964.+1965", re2_options)
+U2NET_CLOTH = re2.compile(
+    b"output.+d1.+Concat_1876.+Concat_1896.+Concat_1916.+Concat_1936.+Concat_1956",
+    re2_options,
+)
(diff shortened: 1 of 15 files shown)


backend/src/nodes/impl/onnx/load.py
```diff
@@ -0,0 +1,80 @@
...
+
+            info.scale_width = get_scale(i_w, o_w)
+            info.scale_height = get_scale(i_h, o_h)
+
+            info.input_channels = i_c
+            info.output_channels = o_c
+        except Exception:
+            pass
+
+        logger.info(f"Model info took {time.time() - start:.2f}s")
```
joeyballentine, line 79:
```suggestion
        logger.debug(f"Model info took {time.time() - start:.2f}s")
```

joeyballentine (approved):

PR #2929, opened 2024-06-02
Switch to using total VRAM instead of free VRAM to estimate tile size

joeyballentine:
We recently switched to not clearing out the cuda cache after each upscale and rather at the end of the chain. This means that pytorch now keeps VRAM usage somewhat high. This is fine though, because this is actually PyTorch's intended behavior to improve performance. Even though it is capturing this VRAM and it looks like its in use, PyTorch is able to reuse that memory for allocating tensors. In fact [users should not be using clear cache at all, apparently](https://discuss.pytorch.org/t/about-torch-cuda-empty-cache/34232).

However, since it _looks_ like this vram is in use and unable to be used, we need to switch to using total system VRAM for estimations rather than free VRAM, since otherwise we might be estimating incorrectly. 

Since the total amount is likely have some usage already, i lowered the budget calculation a bit to compensate.

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
+++ b/backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
@@ -68,10 +68,10 @@ def upscale(
                 if options.use_fp16:
                     model_bytes = model_bytes // 2
                 mem_info: tuple[int, int] = torch.cuda.mem_get_info(device)  # type: ignore
-                free, _total = mem_info
+                _free, total = mem_info
                 if options.budget_limit > 0:
-                    free = min(options.budget_limit * 1024**3, free)
-                budget = int(free * 0.8)
+                    total = min(options.budget_limit * 1024**3, total)
+                budget = int(total * 0.75)
 
                 return MaxTileSize(
                     estimate_tile_size(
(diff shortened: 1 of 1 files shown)


backend/src/packages/chaiNNer_pytorch/pytorch/processing/upscale_image.py
```diff
@@ -68,10 +68,13 @@ def estimate():
                 if options.use_fp16:
                     model_bytes = model_bytes // 2
                 mem_info: tuple[int, int] = torch.cuda.mem_get_info(device)  # type: ignore
-                free, _total = mem_info
+                _free, total = mem_info
+                # only use 75% of the total memory
+                total = int(total * 0.75)
```
RunDevelopment, line 73: I feel like this is too conservative. Even with chainner, chroma, steam, and VSCode running, my GPU uses less than 10% (1GB) of my total VRAM. Plus, because of the 80% down there, we estimate with only 60% of total VRAM.

Honestly, I would drop the first `int(total * 0.75)` and change the 80% below to 75%. That should be conservative enough. If it consistently estimates too high for some users, they can manually set a memory budget.

joeyballentine: In my testing this was equivalent to what we had before, so maybe this needs to be re-thought a bit

Kim2091: IMO: it's better to be too conservative with it than deal with the constant VRAM issues

joeyballentine: I would agree with that. If its not using enough VRAM, set a large tile size or set it to max. Auto should cause the fewest issues (imo)

RunDevelopment: Alright. Then let's be more conservative for now. If we find that it's too much, we can raise it later.

RunDevelopment (approved):

PR #2930, opened 2024-06-02
Add a setting for clearing CUDA cache after each inference

joeyballentine:
While I'd prefer for this to just not be an option, I admit it might be useful as a debugging tool for users or to potentially prevent issues. If users are fine with the slowdown and it stops an issue from happening, then it's probably useful to have this.

image

### DIFF
--- a/backend/src/packages/chaiNNer_pytorch/pytorch/processing/align_image_to_reference.py
+++ b/backend/src/packages/chaiNNer_pytorch/pytorch/processing/align_image_to_reference.py
@@ -221,14 +221,19 @@ def align_image_to_reference_node(
     blur_strength: float,
 ) -> np.ndarray:
     context.add_cleanup(safe_cuda_cache_empty)
+    exec_options = get_settings(context)
     multiplier = precision.value / 1000
-    return align_images(
-        context,
-        target_img,
-        source_img,
-        precision,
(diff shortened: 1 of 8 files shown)


backend/src/packages/chaiNNer_pytorch/pytorch/processing/align_image_to_reference.py
```diff
@@ -221,14 +221,19 @@ def align_image_to_reference_node(
...
+            source_img,
+            precision,
+            multiplier=multiplier,
+            alignment_passes=alignment_passes,
+            blur_strength=blur_strength,
+            ensemble=True,
+        )
+    finally:
+        if exec_options.force_cache_wipe:
+            safe_cuda_cache_empty()
```
RunDevelopment, line 239: Maybe instead of doing this everywhere, we could make `context.add_cleanup(fn, after="node")` a thing. Then we could write all of this as just
```py
context.add_cleanup(safe_cuda_cache_empty, after="node" if exec_options.force_cache_wipe else "chain")
```

joeyballentine: That's a better idea for sure

RunDevelopment (approved):

Issue #2933, opened 2024-06-03
Crop to a absolute size (with anchor)


simolg:
**Motivation**
I have many images with multiple resolutions (therefore multiple aspect ratios),
I would like a function that cuts automatically these images to a fixed size.

**Description**
I don't know the initial aspect ratio of each images I can't calculate the every edges;
therefore the crop node will have to cut "automatically" to a "fixed" size
keeping each image centered (or other types of anchors)

generalizing:
- if the image are wider you will have to cut the left and right edges
- if the image are taller you will have to cut the upper and lower edges
- if the image is totally different you must cut all edges

on the contrary:
- if the canvas needs to be larger, borders must be added

**Alternatives**
paintdot.net has a similar function (canvas size), which is very useful,
let's call it an "advanced automatic crop" ;)

joeyballentine:
You can already achieve this technically, with math nodes and "edges" crop mode. Simply take the width and height of the image, subtract the target width and height, and then divide by 2. You will be left the amount you need to take off of each side to get a "center crop"

simolg:
I tried it this way ... but things get complicated because the edges could be horizontal or vertical, a lot of calculations and comparisons (unnecessary)

In addition to this problem, if the number is not even we will have a difference of 1 pixel
example 831px, I want 800px: (831-800)/2 = 15.5 (crop 16px per side), at the end of the work I find an image of 799 pixels :/

I don't see a simple solution with the current nodes.

joeyballentine:
I don't understand the horizontal/vertical thing, if you're just doing a center crop it doesn't matter because it's even on either side.

As for the uneven different, you'd need to either do a mod (modulo) operation to crop or pad to an even number, or just round the resulting values from the math. A dedicated node or crop mode for center cropping would have the same issue

joeyballentine:
The difference of course is instead of a dedicated mode making that decision for you, you can decide how you want to do it.

simolg:
Thanks joeyballentine for the tips, it was very helpful

my mistake was that I wanted to do more things at once:
resize an image and change the canvas size,
_all without changing the aspect ratio of the source and without having black bars._

if the image is "landscape" ( aspect_ratio_source > aspect_ratio_destination )
the left and right edges will be cut

if the image is "portrait" ( aspect_ratio_source < aspect_ratio_dest_destination )
the top and bottom edges will be cut

I attach a chaiNNer file with the result
[...]

PR #2940, opened 2024-06-04
Add UI for anchor/position dropdowns

RunDevelopment:
This PR adds a new preferred style for dropdowns that describe anchors or positions. The Text as Image node is that only node using this new style right now, but I plan to use it for #2933/#777 as well. We might even use it for Blend Images, but that needs some more work.

image image

Other changes:
- Rename position options in Text As Image and Blend Images. The labels are now shorter.
- Add `useValidDropDownValue` hook to ensure valid dropdown values. This removes some code duplication.


joeyballentine (approved):
This is really cool, good job

PR #2952, opened 2024-06-15
Add pip cache to backend type check workflow

joeyballentine:
Noticed this one was being a bit slow, so i checked and it wasn't caching pip. oops


joeyballentine:
In hindsight, i now realize that not only was this not necessary since we were caching chainner_pip, it also didn't fix the issue. I just completely forgot we were caching chainner_pip instead and assumed this would fix it.

It is not fixed, and i'm not sure what the issue is either.

PR #2958, opened 2024-06-23
Switch to using Axios for requests

### LINKED ISSUE #2957: cli fetch timeout error, reversion of PR-1853
**Description**
When using the CLI on MacOS I get the 5 minute timeout mentioned in [PR-1853 ](https://github.com/chaiNNer-org/chaiNNer/pull/1853)

It seems these changes were recently reverted.

https://github.com/chaiNNer-org/chaiNNer/commit/270ab746b47d99ecb7001f85e9763782972c6195#diff-247357a19207f9fd53e150436a901b3a7005aa8614004edce4968d7d0dbd00b8L158

joeyballentine:
Closes #2957 

IMO, the fact that the browser and nodejs implementations of fetch differ, and there is no way to natively specify timeout, is really stupid. So, I decided to just switch to using Axios instead of trying to use native fetch. Axios is able to work in both the browser and nodejs, and it defaults to having no timeout. So, we should expect this issue to never pop up again with this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@fontsource/open-sans": "^4.5.10",
         "@octokit/rest": "^19.0.13",
         "@react-nano/use-event-source": "^0.12.0",
+        "axios": "^1.7.2",
         "bezier-js": "^6.1.4",
         "chakra-ui-markdown-renderer": "^4.1.0",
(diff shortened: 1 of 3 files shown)


joeyballentine:
Merging this so it's fixed in the nightly

PR #2960, opened 2024-06-25
Add length info for sequences

joeyballentine:
This reuses the existing "length type" string that was previously implemented but not used. This then adds the rest of the functionality for these types to be usable. Right now, they are only used for adding type tags for sequence length.

They are not used for checking connectability. I plan on doing that in a separate PR.

I also plan on figuring out a way to do things like get folder file amounts and video frame length ahead of time, also in a separate PR.

Also, I can't seem to figure out if its possible to get the input type on merge_spritesheet to be exact. It seems to always want to give it int >= 1 for some reason.

image

Also, I'm not sure if this is necessarily the best way to represent length (using a number inside brackets) since I just borrowed it from programming, but i wasn't sure what else to do

### DIFF
--- a/backend/src/api/node_data.py
+++ b/backend/src/api/node_data.py
@@ -9,7 +9,15 @@ import navi
 from .group import NestedIdGroup
 from .input import BaseInput
 from .output import BaseOutput
-from .types import FeatureId, InputId, NodeKind, OutputId, RunFn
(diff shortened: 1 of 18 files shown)


joeyballentine:
I'm going to assume that since you PR'd to it, it's good now.

PR #2966, opened 2024-06-30
Add Padding input to PyTorch Upscale node

RunDevelopment:
I recently upscaled a few tileable normal maps, and using Pad + Crop to add padding was really annoying. It's doable, but requires a bunch of nodes if you want to do it for arbitrary model scales, and even that isn't enough with Custom Scale. So I added a "Padding" option that automatically adds 16px of padding to the image depending on what the user selects.

Note that the padding size isn't configurable. The Padding option is supposed to be a simple way to get commonly-used functionality that is *good enough*. It's just like how you can't choose an interpolation method. This simplicity is by design. If you need more control, you can implement it yourself with chaiNNer's existing nodes.

This should also make upscaling easier for beginners, as a very simple Upscaling chain now provides a lot of features. In particular, simple upscaling chains are now more powerful than CupScale. With this, I would consider the Upscale Image node to be feature-complete. 

This
image

is equivalent to this (assuming Custom Scale isn't used):
image


joeyballentine (approved):
Pretty good solution, nice work 👍

PR #2967, opened 2024-06-30
Fix canceling regression

joeyballentine:


src/common/Backend.ts
```diff
@@ -182,6 +182,9 @@ export class Backend {
                 if (error.response?.data) {
                     return error.response.data as T;
                 }
+                if (axios.isCancel(error)) {
+                    return {} as T;
```
RunDevelopment, line 186: What even is this?

joeyballentine: As is, the function has to return T. What would you recommend doing instead?

RunDevelopment: Are.... Are you joking? The function has to return a `T`, so you just cast some random value to `T`?

This PR is supposed to fix a bug, not add another one that's harder to detect.

joeyballentine: Ok, yeah

src/common/Backend.ts
```diff
@@ -133,6 +133,31 @@ export class ServerError extends Error {
...
+        this.config = config;
+    }
+
+    static isCancel(cancel: unknown): cancel is CancelError {
+        if (typeof cancel !== 'object' || cancel === null) {
+            return false;
+        }
+        const obj = cancel as Record<string, unknown>;
+        return typeof obj.message === 'string' && obj.type === 'cancel';
+    }
```
RunDevelopment, line 155: Why are you doing all that when `x instanceof CancelError` exists?

joeyballentine: I was just doing what we did for server error

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -478,7 +479,10 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                 });
             }
         } catch (err: unknown) {
-            if (!(err instanceof DOMException && err.name === 'AbortError')) {
+            if (
+                !(err instanceof DOMException && err.name === 'AbortError') &&
```
RunDevelopment, line 483: Is this even needed anymore? I don't think so, but just to be sure, I think it would be good if `fetchJson` automatically converted these `DOMException`s to our `CancelError`, so we only have to deal with our own errors.

When moving this to `Backend.ts`, be careful to check whether `DOMException` is defined, because it's a web-only class. So the condition needs to be something like `if (typeof DOMException !== "undefined" && err instanceof DOMException && err.name === 'AbortError')`.

joeyballentine: You know, I think we actually don't need that part at all, like I think the axios cancel completely replaces this.

At least, I cant get that condition to trigger now

src/common/Backend.ts
```diff
@@ -133,6 +134,27 @@ export class ServerError extends Error {
         return new ServerError(json.message, json.description, json.status);
     }
 }
+export class CancelError extends Error {
+    type = 'cancel';
```
RunDevelopment, line 138: This is unused.

src/common/Backend.ts
```diff
@@ -133,6 +134,27 @@ export class ServerError extends Error {
...
+
+    constructor(message: string | undefined, config?: AxiosRequestConfig) {
+        super();
+        this.message = message ?? 'The request was cancelled';
+        this.config = config;
+    }
+
+    static isCancel(cancel: unknown): cancel is CancelError {
+        return cancel instanceof CancelError;
+    }
```
RunDevelopment, line 152: Please remove this method. `ServerError` only has `isJson` to detect whether some JSON sent by the server is a server error. Obviously, we don't need to do something like that for cancellation.

src/renderer/contexts/ExecutionContext.tsx
```diff
@@ -478,7 +479,7 @@ export const ExecutionProvider = memo(({ children }: React.PropsWithChildren<{}>
                 });
             }
         } catch (err: unknown) {
-            if (!(err instanceof DOMException && err.name === 'AbortError')) {
+            if (!CancelError.isCancel(err)) {
```
RunDevelopment, line 482:
```suggestion
            if (!(err instanceof CancelError)) {
```

RunDevelopment (commented):
Last nits

RunDevelopment (approved):

PR #2968, opened 2024-06-30
PR changes

RunDevelopment:
PR changes for #2960

Changes:
- Added `struct Sequence { length: uint }` Navi type to hold the type information of a sequence. Everything now uses this type instead of just being a number. This has the advantage that it will be trivial to add more information in the future.
- Improved type tags for sequences.
- Added sequence length types for Load Images and Load Video to account for Use Limit.
- Major rework for how input/output types are ordered. Types for (iter) inputs/outputs just work.
- Iter input/output infos now have IDs. This just makes it easier to refer to them, because output types might depend on sequence lengths in the future. E.g. a Count node that returns the length of the input sequence. Since nodes can only have at most 1 input/output sequence rn, I just hard-coded the ID as 0 in the backend for now.

### DIFF
--- a/backend/src/api/node_data.py
+++ b/backend/src/api/node_data.py
@@ -9,7 +9,15 @@ import navi
 from .group import NestedIdGroup
 from .input import BaseInput
 from .output import BaseOutput
-from .types import FeatureId, InputId, NodeKind, OutputId, RunFn
+from .types import (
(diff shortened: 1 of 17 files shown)


joeyballentine (approved):
This is perfect, thank you 🙏

Issue #2950, opened 2024-06-11
Add additional save options for TIFF image format


Wetzer2:
**Motivation**
Some programs recommend TIFF format as import format for textures. If I create a TIFF with chaiNNer this wouldn't recognized as valid.
So I guess it depends on the export options

**Description**
If there actual one of the following file options not the "default" option of tiff in the chaiNNer save node, please add this as select-able file format option:
* color mode: RGB, 8bits per channel
* Compression: LZW (preferred) or ZIP
* Pixel Order: Interleaved (RGBRGB)
* Byte Order: IBM PC
* Layer Compression: Discard Layers and Save a Copy

**Alternative**
There's no alternative to chaiNNer!
THX guys for this great tool. :)

RunDevelopment:
Could you please tell us where you got these file format options from or what they mean? Especially "Byte Order: IBM PC" makes no sense to me...

PR #2970, opened 2024-06-30
Add support for 3 TIFF compression modes

RunDevelopment:
Progress towards #2950.

This PR adds support for None, LZW, and ZIP compression for 8 and 16 bit images. This is the best that OpenCV would let me do.

image

---

So OpenCV's TIFF compression is completely fucked. Not only did they *forget* to define the enum constants for TIFF compression modes, the compression modes themselves don't really work. E.g. it says that it support a JPEG compression mode, but this mode always errors no matter what I do. For 32 bit, OpenCV only supports a single compression mode. Not even NONE is an option.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image/io/save_image.py
+++ b/backend/src/packages/chaiNNer_standard/image/io/save_image.py
@@ -92,6 +92,17 @@ class TiffColorDepth(Enum):
     F32 = "f32"
 
 
+class TiffCompression(Enum):
+    NONE = 1
+    LZW = 5
+    ZIP = 8
+
+    @property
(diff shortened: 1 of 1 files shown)


joeyballentine (approved):
Maybe in the future we can use the image crate's tiff encoding. I'm guessing it's probably better than opencv 😆

RunDevelopment:
> Maybe in the future we can use the image crate's tiff encoding. I'm guessing it's probably better than opencv 😆

I just looked, and it looks like [it has even fewer features](https://docs.rs/tiff/0.9.1/tiff/encoder/struct.TiffEncoder.html) :(

PR #2971, opened 2024-07-01
Add broadcasts for sequences

joeyballentine:
Wanted to make this a draft to get some early feedback. There's a few things that still have to be done, mainly fixing the issue where it shows the _previous_ broadcast value, which I have been unable to find even a shred of a lead at what's causing it.

Note: I decided to store the lengths in the outputNarrowing map under the "length" key. This is obviously not an ideal solution, but the alternative was duplicating all the same kind of code for storing it in a new map, and i felt it was similar enough to just throw it in there, at least for a proof of concept first.

# TODO:
- [x] Expand autorun capability to all the generators
- [x] Remove generators from the cache after sending the broadcast (to avoid them resuming from the previous index)
- [x] Fix the stale broadcast issue
- [x] Probably more

image

### DIFF
--- a/backend/src/server.py
+++ b/backend/src/server.py
@@ -28,7 +28,7 @@ from api import (
     NodeId,
 )
 from chain.cache import OutputCache
-from chain.chain import Chain, FunctionNode
+from chain.chain import Chain, FunctionNode, GeneratorNode
 from chain.json import JsonNode, parse_json
 from chain.optimize import optimize
 from dependencies.store import installed_packages
@@ -287,7 +287,22 @@ async def run_individual(request: Request):
         node_id = full_data["id"]
         ctx.cache.pop(node_id, None)
 
-        node = FunctionNode(node_id, full_data["schemaId"])
+        schema_data = api.registry.nodes.get(full_data["schemaId"])
+
+        if schema_data is None:
+            raise ValueError(
+                f"Invalid node {full_data['schemaId']} attempted to run individually"
+            )
+
+        node_data, _ = schema_data
+        if node_data.kind == "generator":
+            node = GeneratorNode(node_id, full_data["schemaId"])
+        elif node_data.kind == "regularNode":
+            node = FunctionNode(node_id, full_data["schemaId"])
+        else:
+            raise ValueError(
+                f"Invalid node kind {node_data.kind} attempted to run individually"
+            )
         chain = Chain()
         chain.add_node(node)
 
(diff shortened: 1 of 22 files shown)


backend/src/server.py
```diff
@@ -207,6 +207,16 @@ async def run(request: Request):
         chain = parse_json(full_data["data"])
         optimize(chain)
 
+        # Remove all Generator values from the cache for each new run
+        # Otherwise, their state will cause them to resume from where they left off
```
RunDevelopment, line 211: This makes no sense IMO. Why do you need to clear the cache? Are you telling me that iterators start iterating their first item when we run them in `/run/individual`? Because if that's the case, the problem isn't that their stale iterator is cached, but that they started iterating.

joeyballentine: I guess I never actually fully confirmed it, but it appeared to be staring from the last index whenever I stopped and restarted the chain. So let's say I had a load images node which got to 600/1000 and I stoppee the chain. When I pressed play again, it would go from 0-600 really quickly then continue going slower at 600.

So, I did this and confirmed it removed the nodes from the cache in case that was the issue. I never actually got to confirm it because my kill requests would end up restarting the executor backend which would reset it anyway.

joeyballentine:
Whats the status on this? Is there anything that needs to be actively addressed?

RunDevelopment:
Same as before, no? 

- Caching is either bugged or otherwise broken for iterators.
- You don't know what the underlying issue is and just nuked the cache.

So please fix the underlying issue, because if we are running one iteration, that'd be a huge problem because it'll run nodes with side effects.

joeyballentine:
> because if we are running one iteration, that'll be a huge problem because it'll run nodes with side effects.

That's not at all what's happening. The problem I mentioned is if you run the chain, stop it, then start again. It is not running an iteration when it auto runs. It's just running the node function, which returns the generator object. it seemed that keeping the same object in the cache treats it the same way as pausing, which is why I needed to dump the cache. 

I'm not sure how you got that we're running an iteration from anything I said

RunDevelopment:
Oh. Well, this PR is about broadcasts, so I assumed that change had something to do with it... I assumed that you discovered this bug because of broadcasts (`/run/individual`) and then talked about stopping since it happens there too. But I guess it has nothing to do with broadcasts, and is only a fix for stopping iterators? Is that it?

joeyballentine:
Correct

RunDevelopment:
Alright, then it should be good.

RunDevelopment (approved):

PR #2976, opened 2024-07-02
Use Sequence lengths when determining validity for connections

joeyballentine:
image

### DIFF
--- a/src/renderer/helpers/canConnect.ts
+++ b/src/renderer/helpers/canConnect.ts
@@ -85,6 +85,37 @@ export const canConnect = (
         );
     }
 
+    const outputSequenceType = sourceFn.outputSequence.get(sourceOutputId);
+    if (outputSequenceType !== undefined) {
+        const schema = targetSchema;
+        const input = schema.inputs.find((i) => i.id === targetInputId)!;
+
+        const inputSequenceType = targetFn.inputSequence.get(targetInputId);
+        if (inputSequenceType) {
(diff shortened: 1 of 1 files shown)


RunDevelopment:
This check is trivial to defeat in 2 ways:

1. By having a regular node in between
     image
1. By having the connection be valid at first, and then changing the sequence length of either node to make it invalid.
    image

joeyballentine:
making sequence length types get passed down to things you have connected sounds like a separate effort / separate PR to me

The other part sounds like I just have to do this same kind of check in the other place we're checking validity? But I thought it used the same check for both? I'll have to do some digging

joeyballentine:
Yeah I'll be honest here -- I just tried to do the second suggestion and I cannot figure it out. I'm confused on how the current input errors even work, and how i could do something similar for sequence mismatches. There doesn't seem to be any input-to-output matching going on and yet somehow it knows the types are incompatible. I don't get it

joeyballentine:
TBH, both of these things feel like they should be separate efforts. This just adds the validation for connectivity (since there's 0 validation right now), and in separate PRs we can do what's necessary for getting these other things working

joeyballentine:
Those two concerns are now addressed in #2996 and #2997

RunDevelopment:
Blocked by #2996.

This PR doesn't work correctly until #2996, so I won't merge this PR until I can test it.

PR #2977, opened 2024-07-02
Remove assertNever assertion for Axios requests

joeyballentine:
There appears to be an error type that axios does not have a type for: Network errors. Because of this, the error value is not really never and therefore we cannot assertNever there.

This does not fix the initial issue though, as the error will still throw.

Should it throw? I'm not really sure what even causes the network error (i did get it myself once when working on a different PR), but it seems to have to do with the backend restarting from code changes, which regular users will not have to deal with.

### DIFF
--- a/src/common/Backend.ts
+++ b/src/common/Backend.ts
@@ -17,7 +17,6 @@ import {
     SchemaId,
     Version,
 } from './common-types';
-import { assertNever } from './util';
 
 export interface BackendSuccessResponse {
     type: 'success';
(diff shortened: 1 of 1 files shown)


RunDevelopment:
I looked into this a little and the problem is that the type guard for `isCancel` is wrong: https://github.com/axios/axios/issues/5153. There is a PR for a fix (https://github.com/axios/axios/pull/5595), but the maintainer of the package seems to not be very active.

So I think the real fix is to not use the bugged `isCancel` for the time being and instead check for `instanceof axios.CanceledError` directly.

PR #2990, opened 2024-07-10
Add logic operation node

### LINKED ISSUE #2982: Confusing and very limited conditional nodes capability
Apologies if none of this made too much sense, I'm no expert on this. Feel free to correct or ask for clarification, I'll try my best.

**Motivation**
While trying to make an automation to check if the image resolution is above my specified threshold that switches the upscaling multiplier to either `1` or `4` - I realized how limiting the conditional node is. There is no easy way to create even slightly advanced logic conditions.  

image

(All of this to create an XOR)  
The documentation is very not helpful, all it gave was what it expects from input and what it spits out at output.  
  

**Description**
There needs to be an easier way and give finer control for more advanced conditional scenarios  
  
I'm thinking of:
- Adding logic operators as their own nodes. (AND, ANDOR, OR, XOR, NOT, etc)
- Even better if the Conditional node can handle an "else if" condition that expects a new logic rule.  

**Alternatives**
The only workaround is the screenshot above. Even then that only accepts 2 values, inevitably it will need even more nodes to accept more values.
image

RunDevelopment:
Fixes #2982

This adds a node to perform boolean operations. Supported operations are: and, or, xor, not. These operations are [functionally complete](https://en.wikipedia.org/wiki/Functional_completeness), but we may want to add more in the future if needed.

Just like for the Math node, I added suggestions, so users can quickly create specific operations. Since I wanted to support text, I also had to slightly adjust how search queries are matched.

image

### DIFF
--- a/backend/src/packages/chaiNNer_standard/utility/math/logic_operation.py
+++ b/backend/src/packages/chaiNNer_standard/utility/math/logic_operation.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from enum import Enum
+
+from api import Lazy, SpecialSuggestion
+from nodes.groups import if_enum_group
+from nodes.properties.inputs import BoolInput, EnumInput
+from nodes.properties.outputs import BoolOutput
+
+from .. import math_group
+
+
+class LogicOperation(Enum):
+    AND = "and"
+    OR = "or"
+    XOR = "xor"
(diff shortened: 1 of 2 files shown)


joeyballentine (approved):

joeyballentine:
Awesome

PR #2998, opened 2024-07-16
fix: Fix SwinIR 4X being referenced with wrong name

PereViader:
SwinIR is not properly named.

Add the underscore so the Automatic111 api properly picks it ups.

image

As a workaround for now and without needing to rebuild the windows app, I have updated Automatic111
image


joeyballentine (approved):

joeyballentine:
Thanks!

PR #3028, opened 2024-09-23
Clarify "Load Model" SafeTensors documentation

JeremyRand:
The "Load Model" node description inaccurately implied that SafeTensors wasn't supported. This was presumably because the documentation in question was split between two files, and only one was updated.

This PR consolidates the supported file type documentation in one file, so that this bug is less likely to happen again. This also avoids making the description tooltip unwieldily verbose, as would have happened if I had instead edited the description in-place.

Also note in the documentation that SafeTensors should be used for better security, since it's not really reasonable to expect users to read the Spandrel GitHub README to find that.


joeyballentine (approved):
Great change, thank you for this

PR #3030, opened 2024-09-23
Save Model: support TorchScript

JeremyRand:
Various software requires models in TorchScript format. As one example, this can be used to convert PyTorch models to ncnn models via PNNX, without using ONNX as an intermediary.


JeremyRand:
I'm not quite sure what the Pyright warning is about; that line is based on PyTorch and ncnn docs, and it seems to work fine. Suggestions on how to make Pyright happy would be welcome.

joeyballentine:
If there's an assertion you can make about the type of `trace` to make Pyright know it's an instance of some class, then that would be one way. Otherwise, if you're sure this works and it's complaining for no reason, just throw an ignore comment on it so it stops complaining

PR #3055, opened 2024-12-05
Improve "Add Noise" node to support subtle noise generation (#3048)

### LINKED ISSUE #3048: [Bug] Add Noise node is way too powerful
**Information:**

- Chainner version: Alpha 0.24.1

**Description**
When you want to add camera sensor noise to an image, you need fine control over the Gaussian noise.

ChaiNNer's Add Noise node is extremely heavy-handed. Even at the lowest value, 1, it adds waaaay too much noise to the image:

image

It seems like it might be throwing an integer into an algorithm that should be using a low-value float. The amount of noise it adds at the lowest setting is extreme!

**The solution:** Make the control even finer, by turning the INTEGER slider into a FLOAT slider.

I can perform a very slow workaround for this problem (it takes around 1 second to process this per image):

- Get Dimensions node on the image I want noise on.
- Create Noise node, with those dimensions as input. Scale: 1, Method: Simplex, Brightness: 2.0 (this controls how much noise; 2.0 adds a perceptible yet very subtle amount which makes images look realistic and not AI generated).
- Blend Images node, base input: Original image. overlay input: The noise. Blend mode: Subtract.

It would be much better if the Add Noise node was fixed to use a working scale parameter, because Add Noise runs in a few milliseconds compared to 1 second for my workaround. :D

Arcitec:
Implements fine-grained control over the noise amount, by switching to a float-based noise multiplier instead.

(The previous implementation was only able to generate noise in increments of +2.55 on the 0-255 pixel luminance scale.)

---

Detailed description:

- The node's previous noise implementation was not granular enough, causing huge jumps between the noise amounts generated by different slider values. The problem was the division factor (an integer scale of 0-100). Think about it in terms of the 0-255 range of normal uint8 image channels. The previous algorithm divided the input slider by 100, and the lowest input value was 1, meaning that the lowest possible noise amount "step" was `0.01`. Now let's look at what happens when we generate such noise: `0.01 * 255 = 2.55`. This means the smallest noise amount step was causing roughly 3 brightness steps for each pixel, which made the entire node very heavy-handed and not very useful. It was most noticeable when generating monochromatic noise on RGB images, where the equal impact on every RGB channel made the heavy noise very invasive (because ChaiNNer adds the same monochromatic noise texture equally to all 3 channels then).

- To fix the node in a backwards-compatible way, the slider has been changed to a float with a single decimal, ranging from `0.0` to `100.0`. By adding a single decimal to the slider, we thereby make the noise amount go in steps of `0.1 / 100 = 0.001` instead, which represents `0.001 * 255 = 0.255` on the regular uint8 scale. So to generate noise that adds +1 on the uint8 scale, you would now just need to go roughly +0.4 on the float slider.

This improvement finally makes the noise node granular enough to be useful for subtle, artistic noise.

---

Closes #3048

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_filter/noise/add_noise.py
+++ b/backend/src/packages/chaiNNer_standard/image_filter/noise/add_noise.py
@@ -45,7 +45,15 @@ class NoiseType(Enum):
                 NoiseColor.GRAY: "Monochrome",
             },
         ),
-        SliderInput("Amount", min=0, max=100, default=50),
+        SliderInput(
+            "Amount",
+            min=0.0,
+            max=100.0,
+            default=50.0,
+            precision=1,
+            step=0.1,
+            slider_step=0.1,
+        ),
         seed_group(SeedInput()),
     ],
     outputs=[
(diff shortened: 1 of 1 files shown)


Arcitec:
@FNsi Since you expressed interest on the ticket, you may be interested in downloading this patch. Just letting you know. :)

Arcitec:
@joeyballentine @RunDevelopment Thank you both so incredibly much for creating ChaiNNer. It's a stunningly beautiful piece of software, with great backend and frontend engineering. Truly a work of art.

I love the way you re-implement algorithms and neural networks in clever ways (spandrel, your genius noise generator code, and so on...). Seriously, thank you both so much.

This was the only issue I had with the application. The heavy-handed noise generator has now been fixed, and it's a small and uncontroversial and fully backwards-compatible change, so hopefully you find the time to merge it at your leisure sometime. :)

FNsi:
Haven't checked my account days , nice job man, I'd try it tomorrow!

O-J1:
Just tried it and can confirm it works.

joeyballentine (approved):
Thanks for this

Arcitec:
@joeyballentine Thank you too, man, for everything you did creating this amazing software. I hope you have a great new year! :)

PR #3045, opened 2024-10-28
simply change __init__.py, add onnxruntime with dml support.

FNsi:
switch Non-NVIDIA onnx version from CPU to directml version to support whatever GPU in windows,
Screenshot 2024-10-28 155759
 which could handle CPU inference too.
Screenshot 2024-10-28 155346

PR #3072, opened 2025-01-04
Allow onnx DML backend 

FNsi:
An accident caused last pull request closed....it's my fault.
This is only the 'control  V' from the closed one.

### DIFF
--- a/backend/src/packages/chaiNNer_onnx/__init__.py
+++ b/backend/src/packages/chaiNNer_onnx/__init__.py
@@ -82,3 +82,6 @@ onnx_category = package.add_category(
     color="#63B3ED",
     install_hint=inst_hint,
 )
+
+
+logger.debug(f"Loaded package {package.name}")
--- a/backend/src/packages/chaiNNer_onnx/__init__.py
+++ b/backend/src/packages/chaiNNer_onnx/__init__.py
@@ -42,7 +42,7 @@ def get_onnx_runtime():
                 pypi_name="onnxruntime",
                 version="1.17.1",
                 size_estimate=6 * MB,          
-        }
+            )
 
 package = add_package(
     __file__,
--- a/backend/src/packages/chaiNNer_onnx/__init__.py
+++ b/backend/src/packages/chaiNNer_onnx/__init__.py
@@ -33,7 +33,7 @@ def get_onnx_runtime():
             display_name="ONNX Runtime (DirectMl)",
             pypi_name="onnxruntime-directml",
             version="1.17.1",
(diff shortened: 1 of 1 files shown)


FNsi:
See detail #3045

joeyballentine (approved):
Thanks

FNsi:
> Thanks

😁 

Should I say you're welcome?
But I'm only a User.

So ,
Thank you
Haha

Kim2091:
@FNsi I'm on the latest nightly build which includes this. DML is not showing up as an option with all default dependencies installed.

Manually installing `onnxruntime-directml==1.17.1` in chaiNNer's python environment does not allow it to show in the settings menu either. TensorRT, CUDA, and CPU are the only options.

Uninstalling other ONNX related libraries and only installing directml allows it to appear, but of course it won't function like that.

FNsi:
> Uninstalling other ONNX related libraries and only installing directml allows it to appear, but of course it won't function like that.


Sorry for the inconvenience, that could happened by chance...

Even only the onnxruntime-directml installed, the showed option still might be "azure and cpu"....

Maybe there's a installing sequence need to follow, but only needed is fix for the search....

Current using Chainner-pip uninstall onnxruntime then install onnxruntime-directml can solve this problem...

Sorry again for inconvenience.

FNsi:
Quick ugly fix in my mind, change default provider to DML.

def get_providers():
    providers = cast(List[str], ort.get_available_providers())

    default = providers[0]
    if "CUDAExecutionProvider" in providers:
        default = "CUDAExecutionProvider"
    elif "DmlExecutionProvider" in providers:
        default = "DmlExecutionProvider"

    return providers, default


def get_provider_label(identifier: str) -> str:
    label = identifier.replace("ExecutionProvider", "")
    if label.lower() == "tensorrt":
        label = "TensorRT"
    return label


execution_providers, default_provider = get_providers()

Kim2091:
@FNsi Thank you for the suggestion. I'm unable to compile chaiNNer right now, but I did try the fix in the final install, but to no avail.

Kim2091:
This may be an issue with ORT itself. I printed the available providers with the default chaiNNer ONNX packages + the ONNX DML package:
>>> print(providers)
['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']

FNsi:
> print(providers)
> ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']

Also I found out people in other repo compliant about that 😅😅, I think you are right.

PR #3089, opened 2025-03-05
Add a node to get the Network type description

tumuyan:
image


joeyballentine:
It would probably better to just turn the model scale node into a more complete "get model info" node that returns the scale, architecture (what you're returning here, it's not a description), etc

PR #3090, opened 2025-03-06
Change "model scale" node to "get model info" node

tumuyan:
Output additional model information, similar to #3089 
The original "get model scale" node has been modified and renamed, but since the `schema_id` unchanged and the scale is set as output0, it will not affect any existing *.chn files contain the "get model scale" node.

image


tumuyan:
update

image

joeyballentine (dismissed):
Looks good to me

joeyballentine:
Oops, I guess there's some issues

joeyballentine (approved):

PR #3091, opened 2025-03-10
Update "Align Image to Reference" node to v3.0.0

pifroggi:
This brings the node up to date to match the [Vapoursynth version](https://github.com/pifroggi/vs_align/tree/main).

Full changelog:
* Overhaul to how the alignment works, which makes it easier to get good results. Previously it could be hard to finetune the options. Sometimes higher precision was better, sometimes multiple passes were better, sometimes more blur, sometimes a combination of all. The new version should now consistently improve with a higher precision level and no longer require much time to find the right combination.
* `Precision` levels now produce different results. There are now 4 levels instead of 5. Alignment quality is better than before for a given precision level, but also slower.
* Added new optional `Mask` input. It takes a black & white image that can be used to exclude areas in white from warping, like a watermark or text that is only on one image. Masked areas will instead be warped like the surroundings.
* Added new `Wide Search` toggle and removed now obsolete `Alignment Passes` slider. This enables a larger search radius at the cost of some speed. When enabled completely different crops like 4:3 and 16:9, shearing, and rotations up to 45° can be aligned. It works by doing a fast homography alignment with the help of [XFeat](https://github.com/verlab/accelerated_features) first to move the corners of the image roughly into the right position before proceeding with the main alignment step. This is both better and faster than the old approach. 
* Added new `Low Quality Input` toggle and removed now obsolete `Blur Strength` slider.
This enables better handling for low-quality input images. When enabled general shapes are prioritized over high-frequency details like noise, grain, or compression artifacts by averaging the warping across a small area. This works better than the old approach and does not require finetuning.
* Added fp16 support.

### DIFF
--- a/backend/src/nodes/impl/pytorch/rife/IFNet_HDv3_v4_14_align.py
+++ b/backend/src/nodes/impl/pytorch/rife/IFNet_HDv3_v4_14_align.py
@@ -1,23 +1,33 @@
-# type: ignore
 # Original Rife Frame Interpolation by hzwer
 # https://github.com/megvii-research/ECCV2022-RIFE
 # https://github.com/hzwer/Practical-RIFE
 
-# Modifications to use Rife for Image Alignment by tepete/pifroggi ('Enhance Everything!' Discord Server)
+# Modifications to use Rife for Image Alignment by pifroggi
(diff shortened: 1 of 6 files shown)


joeyballentine (approved):
Awesome, thanks. Heard this works really well

PR #3103, opened 2025-05-08
Update torch & torchvision to support RTX 50 series

LlamaEnjoyer:
Torch updated to 2.7.0 and vision to 0.22.0.

### DIFF
--- a/backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
+++ b/backend/src/nodes/impl/pytorch/convert_to_onnx_impl.py
@@ -57,8 +57,8 @@ def convert_to_onnx_impl(
     with BytesIO() as f:
         torch.onnx.export(
             m,
-            dummy_input,
-            f,
+            (dummy_input,),
+            f,  # type: ignore[arg-type]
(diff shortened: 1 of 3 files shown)


joeyballentine (approved):

joeyballentine:
Looks like this introduced some type issues. Are you able to handle fixing these?

LlamaEnjoyer:
I'll give it a try tomorrow

LlamaEnjoyer:
@joeyballentine I updated pyright to most recent version, since 1.1.338 is ancient. I'm hoping this will resolve the problem.

joeyballentine:
It did not. You're gonna have to fix those type issues (or if you don't know how, I can do it for you if i end up having some time)

LlamaEnjoyer:
Allright, one last attempt. The classic "works for me" moment. If that doesn't do the trick then I rest my case and pass the torch you :)

geomax22:
Not sure how this works but when is a compiled version of chaiNNer with  RTX 50 support likely to be available (a nightly build?)
Thanks

PR #3222, opened 2025-11-18
Add IsGrayscale node to detect if images are grayscale

Velorexe:
This PR provides a new node to detect if an image is grayscale or not. Most images from the web are not correctly labeled as such (i.e. limiting the amount of channels), therefore the `Image Statistics` node doesn't always come in clutch. When up scaling or editing a directory of images, it can happen that some images require different models to be loaded in based on if the image is grayscale or not.

This node provides such functionality by comparing the channels to a tolerance level. Due to noise or compression, it can happen that a seemingly grayscale image has some differing values within the channels, hence a tolerance input seemed as the right way to handle this bit of (unfortunate) logic.

This PR is always open for feedback and changes.

### DIFF
--- a/backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/is_grayscale.py
+++ b/backend/src/packages/chaiNNer_standard/image_utility/miscellaneous/is_grayscale.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+import numpy as np
+
+from nodes.properties.inputs import ImageInput, SliderInput
+from nodes.properties.outputs import BoolOutput
+
+from .. import miscellaneous_group
+
+
+@miscellaneous_group.register(
+    schema_id="chainner:image:is_grayscale",
+    name="Is Grayscale",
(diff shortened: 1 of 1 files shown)


Velorexe:
Spitballing some ideas for potential improvements here:
* Add an extra `NumberOutput` with the difference between channels, perhaps the tolerance handling could be in a different node to keep this one as small as possible/not copy already existing number comparison logic?
*  Use OpenCV instead of Numpy to do the comparison, as it *should* give better performance on larger images. Not confirmed, can check.

joeyballentine:
Whoops sorry didn't meant to press close there, my bad.

0-255 is fine in my opinion, but I think we do tend to use percentages for these things as 0-100 is more intuitive for casual users than 0-255. But I don't think it matters much, really.

joeyballentine (changes requested):

joeyballentine (changes requested):

joeyballentine (approved):

PR #3224, opened 2025-11-20
Added translations and translation loaders for nodes

Velorexe:
This PR aims to add translations to the nodes in the backend by overwriting parts of the original node schema in the `server.py`. By adding an `xml` translation file next to nodes, they can be picked up by the `translation_loader.py`. They don't *strictly* have to be next to the node itself, but it's recommended, else it can become quite a mess fast.

The reason I picked XML is because it's easy to write schemas for it that can be enforced in an editor. Besides that, it also makes it easier to assign properties on object level, instead of having to make it a separate property that has to live on the object itself, like with JSON.

As this introduces big changes, I'll let it remain in a draft for now. I'd like to add tests for it in the future as well, so this isn't in a merge-able state as of yet.

### DIFF
--- a/backend/utility/translation_linter.py
+++ b/backend/utility/translation_linter.py
@@ -1,0 +1,128 @@
+import ast
+import sys
+import xml.etree.ElementTree as ET
+from pathlib import Path
+
+
+class TranslationLinter(ast.NodeVisitor):
+    def __init__(self, file_path: Path):
+        self.file_path = file_path
+        self.violations: list[tuple[int, str]] = []
+
+    def _get_decorator_arg(self, decorator: ast.Call, arg_name: str):
+        """Extract a keyword argument value from a decorator call."""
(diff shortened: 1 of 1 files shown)


joeyballentine:
In general I think I like this. The only thing would be that I can see it potentially becoming an issue of maintaining it when inputs change. But I can't think of anything that would help with that and we already have systems in place for keeping input ids the same when we swap them around, so I think it should be fine.

@RunDevelopment if you don't mind, I'd love to get your opinion on this.

Velorexe:
Apparently you can't write custom ruff plugins as of yet, my idea was to write a pylint check that'd do it for me. Runtime sounds like it's a step too "late" in my opinion, though it would come up when testing the project, it wouldn't show up that well when a pipeline check on linting is run.

I'll draft a custom AST script for now, so there's utility to check the content beforehand.

PR #3309, opened 2026-07-04
distance_transform: Add more algorithms and cut-off parameter

CrossVR:
This PR depends on the https://github.com/chaiNNer-org/chaiNNer-rs/pull/36 PR in chaiNNer-rs.

The following features are added for distance fields:
- Added the EDTAA3 and EDTAA4 algorithms
- Added cut-off parameter to Distance Transform node to adjust the pivot point of the signed-distance field
- Added dedicated distance-to-alpha node that increases the contrast of the contour edge to return to an alpha mask
- Added optional RGBA image input support
  - For Binary the RGB channels are simply passed through
  - For ESDF and EDTAA the RGB channels are alpha-filled using the results of the distance transform

These new features give the user more options for upscaling alpha masks using signed distance fields.


joeyballentine (changes requested):
Since this changes an existing node, it requires a migration.

PR #3315, opened 2026-07-26
Add wrapping modes to Average Color Fix node

ignamiranda:
Adds a **Padding** dropdown (None / Reflect / Wrap / Replicate) to the Average Color Fix node, matching the pad-and-crop pattern already used by the PyTorch upscale node.

The reference image gets a fixed 16px border (the diff is computed at ref resolution), and the input gets proportionally larger padding so both images have the same border size at the ref level after downscaling. The border is cropped from the output, so output dimensions are unchanged and existing workflows are unaffected (default is None).


joeyballentine:
I'm not 100% sure that the direction i want to go with chaiNNer is adding things like this to every node where it could be useful. the same thing can be done with individual nodes, adding padding and then cropping it off after. that's only 2 nodes in the chain, or a few more if you wanna make it dynamic with math nodes, and that can apply to ANY node and not just the ones that have this built in

